PR #323, opened 2025-10-08
Fix e2e tests to properly fail CI when they have errors

bokelley:
## Problem
E2E tests were failing in CI (42 errors + 10 failures) but CI was still passing due to `continue-on-error: true`.

## Root Causes
1. **CI configuration**: `continue-on-error: true` on e2e test step allowed failures to pass
2. **Test summary**: e2e-tests not included in dependency list, so failures were ignored
3. **Schema validation tests**: Test data used wrong field names (`formats` vs `format_ids`) and missing required fields
4. **Port configuration**: e2e conftest.py had incorrect port defaults that didn't match docker-compose.yml

### CI Configuration (.github/workflows/test.yml)
- Remove `continue-on-error: true` from e2e test step
- Add `e2e-tests` to test-summary job dependencies
- Update test-summary to check e2e-tests result and fail CI if they fail
- Remove misleading comment about tests being "not yet stable"

### Schema Validation Tests (tests/e2e/test_comprehensive_adcp_compliance.py)
- Replace `formats` field with `format_ids` per AdCP spec
- Add `adcp_version` to all test responses (required field)
- Add `property_tags` to all products (required by oneOf constraint in schema)
- Fix test_invalid_format_type expectation (schema allows custom format IDs)
- **All 14 schema validation tests now pass** ✅

### Port Configuration (tests/e2e/conftest.py)
- Fix MCP port default: 8126 → 8092 (matches docker-compose)
- Fix A2A port default: 8091 → 8094 (matches docker-compose)
- Fix Admin port default: 8087 → 8093 (matches docker-compose)
- Fix Postgres port default: 5518 → 5435 (matches docker-compose)
- CI will override these with env vars (8080, 8091, etc.) so this doesn't affect CI

## Impact
E2E tests will now properly fail CI when they have errors instead of being silently ignored.

## Testing
- ✅ All 14 schema validation tests pass locally
- ✅ Pre-commit hooks pass
- ⏳ Waiting for CI to run e2e tests with proper error reporting

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -315,7 +315,6 @@ jobs:
         export PATH="$HOME/.cargo/bin:$PATH"
         # Let pytest manage Docker Compose via conftest.py
         uv run pytest tests/e2e/ -v --tb=short
-      continue-on-error: true  # E2E tests not yet stable - see docs/testing/e2e-test-stabilization-plan.md
 
     - name: Cleanup Docker services
       if: always()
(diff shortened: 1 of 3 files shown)


bokelley:
## CI Failure Analysis

The e2e tests were timing out waiting for the MCP server to become healthy. 

**Root Cause**: 60-second timeout was too short for CI environment where Docker builds and service startup are slower.

**Fix**: 
- Increased timeout from 60s → 120s
- Added detailed logging to show progress every 10s
- Added docker log dump on failure for easier debugging
- Show elapsed time when services become ready

This should either fix the issue (if it was just slow startup) or give us better logs to diagnose if there's an actual crash.

bokelley:
## Second CI Failure Analysis & Fix

The improved logging revealed the actual issue! 🎉

**Root Cause**: The MCP server was crashing on startup with a database constraint violation:
```
duplicate key value violates unique constraint "tenants_pkey"
Key (tenant_id)=(default) already exists
```

**Problem**: PostgreSQL volumes were persisting between test runs in CI. When tests run multiple times, the database initialization tries to insert the default tenant again, which already exists from the previous run.

**Fix** (commit 03ad718):
- Run `docker-compose down -v` before starting services
- The `-v` flag removes volumes, ensuring a clean database for each test run
- This is the standard pattern for e2e tests that need isolated environments

**Diagnosis was successful thanks to the logging improvements** that dumped the Docker logs when health checks failed!

PR #753, opened 2025-11-13
feat: Add pricing to delivery

BaiyuScope3:
# Add Pricing Fields to Delivery Reports

Adds `pricing_model`, `rate`, and `currency` fields to `by_package` delivery reports.

**Changes:**
- Added pricing fields to `PackageDelivery` schema
- Uses real per-package metrics from adapters (not equal division)
- Mock adapter now returns `by_package` breakdown
- Pricing info fetched from `package_config["pricing_info"]`

**Impact:** Delivery webhooks/API now include pricing per package with accurate per-package metrics.

### DIFF
--- a/src/core/tools/media_buy_delivery.py
+++ b/src/core/tools/media_buy_delivery.py
@@ -230,7 +230,8 @@ def _get_media_buy_delivery_impl(
             total_impressions_from_adapter = 0
             
             if not any([testing_ctx.dry_run, testing_ctx.mock_time, testing_ctx.jump_to_event, testing_ctx.test_session_id]):
-                # Call adapter to get REAL per-package delivery metrics
+                # Call adapter to get per-package delivery metrics
+                # Note: Mock adapter returns simulated data, GAM adapter returns real data from Reporting API
                 try:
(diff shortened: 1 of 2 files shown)


src/core/tools/media_buy_delivery.py
```diff
@@ -242,40 +287,63 @@ def _get_media_buy_delivery_impl(
...
+                        # Use real metrics from adapter
+                        pkg_metrics = adapter_package_metrics[package_id]
+                        package_spend = pkg_metrics["spend"]
+                        package_impressions = pkg_metrics["impressions"]
+                        package_clicks = pkg_metrics.get("clicks")
+                    else:
+                        # Fallback: divide equally if adapter didn't return this package
+                        package_spend = spend / len(packages) if packages else spend
+                        package_impressions = impressions / len(packages) if packages else impressions
+                        package_clicks = None
```
BaiyuScope3: QQ: Will all ad server adaptors have per package metrics? If not, should we fallback on evenly divide or just don't providing these
Currently the transaction generation will need per package metrics

PR #901, opened 2025-12-30
fix: src.core.format_spec_cache undefined

roofeel-fm:
https://github.com/adcontextprotocol/salesagent/issues/900

### DIFF
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -230,7 +230,7 @@ def _extract_creative_url_and_dimensions(
 
 
 def _get_format_spec_sync(agent_url: str, format_id: str) -> Any | None:
-    """Get format specification synchronously using asyncio event loop.
+    """Get format specification synchronously using asyncio.run().
     
     This helper function wraps the async registry.get_format() call to make it
     usable in synchronous contexts. The registry uses in-memory cache (30min TTL)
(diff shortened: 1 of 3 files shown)


roofeel-fm:
I have read the IPR Policy

bokelley (commented):
## Review

Thanks for fixing this bug - the `format_spec_cache` module doesn't exist so this would cause runtime failures in the auto-approval flow.

### The fix is correct

The core approach is sound: extract a helper function to unify the two places that need synchronous access to the async `registry.get_format()` call.

### Suggestions

**1. Consider using `asyncio.run()` instead of manual loop management**

The current pattern:
```python
loop = asyncio.new_event_loop()
asyncio.set_event_loop(loop)
try:
    return loop.run_until_complete(registry.get_format(agent_url, format_id))
finally:
    loop.close()
```

Could be simplified to:
```python
return asyncio.run(registry.get_format(agent_url, format_id))
```

`asyncio.run()` (Python 3.7+) handles loop creation, cleanup, and avoids issues with `set_event_loop()` polluting global state. The manual pattern can cause issues if called from certain contexts.

**2. Test uses MagicMock for format spec**

In `test_validate_creatives_missing_required_fields`, the test creates a `MagicMock()` for the format spec. This pattern can hide interface mismatches. Consider using a real `FormatSpec` object or `SimpleNamespace` with explicit attributes to catch any field access bugs.

**3. Minor: test location**

The test for `_get_format_spec_sync` is a unit test but lives in `tests/integration_v2/`. Not blocking, just a nit.

---

Overall this looks good to merge - it fixes a real bug and the tests cover the key paths. The suggestions above are minor improvements that could be addressed in a follow-up if needed.

roofeel-fm:
Thanks for the thorough review and the clear suggestions! 

Summary of changes:
- Use asyncio.run for simpler handling of asynchronous operations.
- Use SimpleNamespace instead of MagicMock.
- Encapsulated helper function logic testing to dedicated unit test modules

PR #943, opened 2026-01-08
fix: ignore url not configured when use mock adapters

roofeel-fm:
- Moved agent URL retrieval after the auto-verification block 
- This allows the function to return early with auto-verified results for mock/dev environments
- Only production tenants with real adapters need the agent URL for actual adagents.json verification
- Prevents "Agent URL not configured" errors when using mock adapters in testing scenarios


roofeel-fm:
https://github.com/adcontextprotocol/salesagent/commits/main/
image
https://github.com/adcontextprotocol/salesagent/actions/runs/20865067316/job/59954042406
image

Hi @EmmaLouise2018  The test suite on the main branch is failing with the same error, so it doesn’t appear to be caused by this PR.

PR #945, opened 2026-01-09
feat: update to adcp 2.18.0 with new assets field support

BaiyuScope3:
## Summary

Updates the sales agent to support AdCP-client 2.18.0's new `assets` field while maintaining backward compatibility with older clients that use the deprecated `assets_required` field.
Per `AdCP protocol 2.6.0` spec.

## Motivation

AdCP-client 2.18.0 introduces a new `assets` field in the `Format` schema that includes both required and optional assets (with an explicit `required: boolean` flag). The previous `assets_required` field only contained required assets and is now deprecated.

This change ensures:
- **Old buyer agents** (adcp protocol < 2.6.0) can still use `assets_required` 
- **New buyer agents** (adcp protocol >= 2.6.0) can use the new `assets` field with optional asset support

### Core Changes
- **`pyproject.toml`**: Upgraded `adcp client python` from 2.17.0 to 2.18.0
- **`src/core/tools/creative_formats.py`**: 
  - Added `_ensure_backward_compatible_format()` function that ensures both fields are populated
  - Uses adcp utilities (`has_assets`, `uses_deprecated_assets_field`, `get_format_assets`, etc.)
  - If format has only `assets` → populates `assets_required` from required assets
  - If format has only `assets_required` → populates `assets` using `normalize_assets_required()`
- **`src/core/tools/media_buy_create.py`**: Updated to use `get_format_assets()` utility for consistent asset handling

### Test Updates
- Updated `tests/helpers/adcp_factories.py` to use new `Assets` type
- Updated integration and unit tests to use `assets` field instead of deprecated `assets_required`

## Backward Compatibility

| Creative Agent Returns | Sales Agent Returns | Old Buyer Works | New Buyer Works |
|----------------------|---------------------|-----------------|-----------------|
| Only `assets` (new) | Both fields | ✅ | ✅ |
| Only `assets_required` (old) | Both fields | ✅ | ✅ |
| Both fields | Both fields (unchanged) | ✅ | ✅ |

## Testing

- [x] Unit tests pass
- [x] Verified backward compatibility function with all scenarios
- [x] Tested against local creative agent (2.18.0) - confirms both fields returned

### Test sales agent MCP tools
- Old version compatible check
image

- New version check
image

### DIFF
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -187,7 +187,8 @@ def _extract_creative_url_and_dimensions(
     url = None
     if creative_data.get("assets") and format_spec and has_assets(format_spec):
         # Use format spec to find the correct asset_id for image/video/url assets
-        # Only check individual assets (not repeatable groups) which have asset_type/asset_id
+        # Note: We only support individual assets here (not repeatable groups).
+        # This matches previous behavior - repeatable groups were never supported in this function.
         for asset_spec in get_individual_assets(format_spec):
             # Type guard: get_individual_assets only returns Assets, not Assets1 (repeatable groups)
             if not isinstance(asset_spec, Assets):
@@ -206,7 +207,8 @@ def _extract_creative_url_and_dimensions(
     height = None
     if creative_data.get("assets") and format_spec and has_assets(format_spec):
         # Use format spec to find the correct asset_id for image/video assets
-        # Only check individual assets (not repeatable groups) which have asset_type/asset_id
+        # Note: We only support individual assets here (not repeatable groups).
+        # This matches previous behavior - repeatable groups were never supported in this function.
         for asset_spec in get_individual_assets(format_spec):
             # Type guard: get_individual_assets only returns Assets, not Assets1 (repeatable groups)
             if not isinstance(asset_spec, Assets):


src/core/tools/creative_formats.py
```diff
@@ -20,6 +34,52 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _ensure_backward_compatible_format(f: FormatT) -> FormatT:
```
BaiyuScope3: This ensures our sales agent always return both `assets` (new) and `assets_required` (deprecated) in the format, so buyer agents using older ADCP versions remain compatible.

src/core/tools/media_buy_create.py
```diff
@@ -179,15 +181,21 @@ def _extract_creative_url_and_dimensions(
...
         - Dimensions extracted from asset types: image, video
         - Type validation: width/height must be int or coercible to int
+        - Uses adcp.utils.get_individual_assets() for backward compatibility with assets_required
     """
     # Extract URL from assets using format specification
     url = None
-    if creative_data.get("assets") and format_spec and format_spec.assets_required:
+    if creative_data.get("assets") and format_spec and has_assets(format_spec):
         # Use format spec to find the correct asset_id for image/video/url assets
-        for asset_req in format_spec.assets_required:
```
BaiyuScope3: Previous logic was not handling the `repeatable group assets`, so let's not add extra logic to handle it in this PR too
We can address the repeatable group assets support in a seprate task

uv.lock
```diff
@@ -4490,14 +4490,14 @@ wheels = [
 
 [[package]]
 name = "werkzeug"
-version = "3.1.4"
+version = "3.1.5"
```
BaiyuScope3: vulerability fix , it is not related to adcp protocol 2.6 updates
image

Issue #1019, opened 2026-02-09
'FormatId' object is not subscriptable


jamiedruce:
We are just testing the media buy end-to-end, and can get all the way through to execution of a media buy, when we get stuck at the following error:

`'MEDIA_BUY_CREATION_ERROR', "Failed to create media buy: 'FormatId' object is not subscriptable"`

This is just a vanilla salesagent, latest version as of this writing.

The products "format_ids" from the db is: `[{"id": "display_image", "agent_url": "<https://creative.adcontextprotocol.org>"}, {"id": "display_html", "agent_url": "<https://creative.adcontextprotocol.org>"}, {"id": "display_js", "agent_url": "<https://creative.adcontextprotocol.org>"}]`

And the `packages.0.format_ids` within the call to `create_media_buy` is exactly the same.

We've tried re-creating the product from scratch with no luck. Just wondering if we need to re-setup with a blank DB incase the issue is from a legacy migration.

Any thoughts here would be much appreciated, as we are currently unsure even what side of the fence the problem is; the request from the buyer agent, or our handling of it in the seller agent.

PR #1051, opened 2026-02-18
refactor: eliminate model_dump antipatterns and migrate to adcp library base classes

KonstantinMirin:
## Summary

This builds on #1044, which established the typed model boundaries principle: Pydantic models stay intact across internal layers, and serialization (`model_dump()`) only happens at system boundaries (protocol I/O, database, API responses).

This PR continues that work in three areas:

1. **Migrated ~20 schema types to adcp library base classes** — eliminating our parallel type definitions in favor of inheriting from the spec
2. **Removed remaining `model_dump()` calls from internal code paths** — functions now pass typed models through instead of cracking them open into dicts
3. **Migrated all DateTime columns to timezone-aware TIMESTAMPTZ** — aligning with the library's `AwareDatetime` type

We hit one architecture limitation we couldn't resolve in this PR: the response serialization pipeline for `get_products` has three divergent paths (MCP, A2A explicit, A2A natural language) that each serialize differently. Unifying them requires a `serialize_response()` pipeline that's designed but not yet implemented (tracked separately).

### Library base class migrations

These models previously duplicated fields from the adcp spec. Now they extend the library types with minimal overrides (`model_config` for extra-field handling, plus any fields the code generator missed):

`PricingModel`, `CreativePolicy`, `QuerySummary`, `SyncCreativeResult`, `SyncCreativesRequest`, `ListCreativesResponse`, `Signal`, `GetSignalsRequest`, `GetSignalsResponse`, `GetMediaBuyDeliveryResponse`, `ActivateSignalRequest`, `SyncCreativesResponse`, `Placement`, `GetProductsRequest`, `SignalDeployment`, `SignalPricing`, `ReportingPeriod`

Each migration follows the same pattern: remove duplicated field definitions, inherit from the library type, keep only local overrides.

### model_dump() elimination

Every case where a model was serialized to a dict just to pass it to another internal function:

- **`_create_media_buy_impl`** — replaced 14 untyped parameters with a single `CreateMediaBuyRequest` model
- **`_sync_creatives_impl`** — accepts `Creative` models and `SyncCreativesRequest` instead of dicts. Extracted from a 2354-line monolith into a `creatives/` package with 7 focused submodules
- **`_list_creatives_impl`** — accepts typed models, uses `model_validate`
- **`PolicyCheckService`** — accepts `Product` and `BrandManifest` models instead of dicts
- **`ranking_agent`** — accepts `list[Product]` instead of `list[dict]`
- **`apply_testing_hooks`** — returns a typed `TestingHooksResult` dataclass instead of mutating a data dict
- **Adapter targeting** — 4 adapters now use typed `Targeting` model attributes instead of `targeting_dict["geo"]["countries"]` dict key access
- **Validation functions** — `validate_geo_overlap`, `validate_targeting` etc. accept `Targeting` model directly
- **Database writes** — `JSONType` and the SQLAlchemy engine `json_serializer` now accept Pydantic models directly, so code no longer needs to call `model_dump()` before storing to JSONB columns
- **Webhook delivery** — delivery scheduler sets webhook metadata fields directly on the response model (they exist on the library type) instead of serializing to dict and mutating

Also removed: `hasattr()` defensive checks on typed models, `hasattr(x, 'root')` RootModel unwrapping, `to_context_object()` no-op wrappings, `serialize_for_json()` helper, and `dict[str, Any]` context fields replaced with typed `ContextObject`.

### A2A serialization centralization

A2A skill handlers previously did ad-hoc serialization — some returned dicts, some returned models, some called `model_dump()` conditionally. Now:

- Handlers return typed response models
- `_serialize_for_a2a()` handles model→dict at the protocol boundary
- `_handle_explicit_skill` dispatch applies serialization consistently
- Request parameters are validated via `model_validate()` with typed request models

V2 compatibility fields (`is_fixed`, `rate`, `floor`) are now gated on `adcp_version` — clients on v3+ get clean responses without legacy shims.

### TIMESTAMPTZ migration

All `DateTime` columns migrated to timezone-aware `TIMESTAMPTZ` via a reversible Alembic migration. This aligns with the adcp library's `AwareDatetime` type for `ReportingPeriod.start`/`end` and ensures consistent timezone handling across the stack. Includes a rollback test that verifies the migration can be reversed cleanly.

### Remaining limitation

The `get_products` response has three serialization paths (MCP handler, A2A explicit handler, A2A natural language handler) that each apply v2 compatibility transforms differently. Unifying them into a single `serialize_response()` pipeline is designed but blocked on the current A2A server architecture. This is tracked as a follow-up task.

## Why tests changed

**78 test files changed** — here's why:

**Tests that now use valid data (bulk of changes):**
With `extra='forbid'` enforced on buyer-facing request models, tests that previously passed arbitrary extra fields now fail validation. These tests were updated to use valid AdCP data — not to make tests pass, but because the old test data was technically invalid and only worked because validation was too permissive.

**Tests updated for typed APIs:**
Functions that changed from `dict` parameters to typed models required test call sites to pass models instead of raw dicts. For example, tests calling `_create_media_buy_impl` now construct a `CreateMediaBuyRequest` instead of passing 14 keyword arguments.

**Tests updated for TIMESTAMPTZ:**
Date strings like `"2025-01-01"` changed to `"2025-01-01T00:00:00Z"` where the library now requires timezone-aware datetimes.

**New test files:**
- `test_build_creative_data.py` — tests for extracted `_build_creative_data` helper
- `test_extract_url_from_assets.py` — tests for extracted `_extract_url_from_assets` helper
- `test_format_resolver.py` — tests for format resolution logic
- `test_policy_typed_models.py` — tests for `PolicyCheckService` typed model API
- `test_v2_compat_version_gating.py` — tests that v2 compat fields are only added for pre-3.0 clients
- `test_timestamptz_migration.py` — integration test for migration up/down
- `test_pydantic_schema_alignment.py` — validates our models match the AdCP JSON schemas
- `test_schema_validation_modes.py` — expanded to cover `extra='forbid'` vs `extra='ignore'` behavior

**No test logic was weakened.** One earlier commit (`b88284a6`) explicitly restored test assertions that had been accidentally weakened during the refactoring.

## Test results

| Suite | Result |
|-------|--------|
| Unit | 1972 passed, 31 skipped |
| Integration (Legacy) | 532 passed, 36 skipped, 18 deselected, 1 xfailed |
| Integration V2 | 184 passed |
| E2E | 71 passed, 13 skipped |
| Smoke | 14 passed |
| Formatting | Pass |
| Linting | Pass |
| Mypy | Pass |

### DIFF
--- a/src/core/schemas.py
+++ b/src/core/schemas.py
@@ -2942,7 +2942,7 @@ class UpdateMediaBuyRequest(LibraryUpdateMediaBuyRequest1):
             unwrapped = []
             for pkg in values["packages"]:
                 if isinstance(pkg, RootModel):
-                    unwrapped.append(pkg.root.model_dump())
+                    unwrapped.append(pkg.root.model_dump(mode="json"))
                 else:
                     unwrapped.append(pkg)
             values["packages"] = unwrapped
--- a/src/core/tools/creatives/_assets.py
+++ b/src/core/tools/creatives/_assets.py
@@ -64,7 +64,7 @@ def _build_creative_data(
         Data dict for storing in the creative's data field.
     """
     if context is not None and not isinstance(context, dict):
-        context = context.model_dump()
+        context = context.model_dump(mode="json")
 
     data: dict[str, Any] = {
         "url": url,


src/core/tools/products.py
```diff
@@ -825,17 +809,16 @@ async def get_products(
     response = await _get_products_impl(req, ctx)
 
     # Return ToolResult with human-readable text and structured data
-    # Apply v2.x backward-compat fields to pricing_options for older clients
     response_dict = response.model_dump()
```
ChrisHuie: looks like this guy got missed with `mode="json"`

KonstantinMirin:
This PR (#1051) doesn't fix #1019 — the `fmt["agent_url"]` dict subscript on that line is still present on this branch. #1053 does address it directly.

That said, #1053 can be merged independently — it's a two-line fix in `media_buy_create.py` and doesn't conflict with this PR. They touch the same file but different sections. No need to wait for this larger release.

PR #1053, opened 2026-02-18
fix: handle FormatId objects in format validation during media buy creation

### LINKED ISSUE #1019: 'FormatId' object is not subscriptable

kmoegling-scope3:
## Summary
- Fixes `'FormatId' object is not subscriptable` error during media buy creation
- The `format_ids` from a product can be either raw dicts (from database JSONB) or Pydantic `FormatId` objects (after deserialization). The code used dictionary subscript access (`fmt["agent_url"]`) which crashes when `fmt` is a `FormatId` object.
- Adds `isinstance()` check to handle both cases, matching the defensive pattern already used later in the same function (line ~2650)

## Root Cause
In `src/core/tools/media_buy_create.py`, the format validation loop at line 2572 accessed `format_ids` entries with `fmt["agent_url"]` and `fmt["id"]`. When Pydantic deserializes these into `FormatId` model instances, subscript access fails because Pydantic models don't support `[]` by default.

Fixes #1019

## Test plan
- [x] All 1965 unit tests pass (`uv run pytest tests/unit/ -x`)
- [x] Ruff formatting and linting pass
- [ ] Verify with end-to-end media buy creation using format_ids

### DIFF
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -2570,10 +2570,8 @@ async def _create_media_buy_impl(
                 product_format_keys: set[tuple[str | None, str]] = set()
                 if pkg_product.format_ids:
                     for fmt in pkg_product.format_ids:
-                        # pkg_product.format_ids are dicts from database JSONB (type annotation says FormatId but runtime is dict)
-                        agent_url = fmt["agent_url"]  # type: ignore[index]
-                        normalized_url = str(agent_url).rstrip("/") if agent_url else None
-                        product_format_keys.add((normalized_url, fmt["id"]))  # type: ignore[index]
+                        normalized_url = str(fmt.agent_url).rstrip("/") if fmt.agent_url else None
+                        product_format_keys.add((normalized_url, fmt.id))
 
                 # Build set of requested format keys for comparison
                 requested_format_keys: set[tuple[str | None, str]] = set()


src/core/tools/media_buy_create.py
```diff
@@ -2570,10 +2570,8 @@ def unwrap_po(po: Any) -> Any:
                 product_format_keys: set[tuple[str | None, str]] = set()
                 if pkg_product.format_ids:
                     for fmt in pkg_product.format_ids:
-                        # pkg_product.format_ids are dicts from database JSONB (type annotation says FormatId but runtime is dict)
-                        agent_url = fmt["agent_url"]  # type: ignore[index]
-                        normalized_url = str(agent_url).rstrip("/") if agent_url else None
-                        product_format_keys.add((normalized_url, fmt["id"]))  # type: ignore[index]
+                        normalized_url = str(fmt.agent_url).rstrip("/") if fmt.agent_url else None
```
BaiyuScope3: let's be preventive and handle both Dict and Pydantic types

KonstantinMirin:
@kmoegling-scope3 I an happy to mere this as soon as you do the IPR Agreement. It's a clean fix, let's merge it.

PR #1054, opened 2026-02-19
fix: resolve property_ids/property_tags authorization in property discovery

lclaudon:
## Summary

- Fixes property discovery failing when publishers use `property_ids` or `property_tags` authorization types in their adagents.json (properties defined at top-level, agents reference them by ID/tag)
- Fixes multi-domain adagents.json files incorrectly linking all properties to a single publisher domain

## Problem

When publishers like Prisma Media (capital.fr, geo.fr, etc.) use `authorization_type: "property_ids"` in their adagents.json, the property discovery service returned 0 properties because:

1. `get_all_properties()` only reads inline `agent.properties` arrays — it doesn't resolve `property_ids` references to top-level properties
2. The unrestricted agent fallback doesn't trigger because the agent HAS a `property_ids` field
3. Even if properties were found, a multi-domain adagents.json (one file listing properties for capital.fr, geo.fr, cotemaison.fr, etc.) would link ALL properties to whichever publisher domain was being synced

## Fix

**Property resolution** (`property_discovery_service.py`): Added `agent_url` parameter. When provided, uses `get_properties_by_agent()` from adcp library which handles all authorization types (property_ids, property_tags, inline, publisher_properties). Backward compatible — without `agent_url`, old code path is preserved.

**Domain filtering** (`property_discovery_service.py`): After resolving properties, filters them to only keep those whose domain identifier matches the publisher domain being synced. Mobile apps (no domain identifier) are kept for all publishers.

**Caller updates** (`publisher_partners.py`): Both production and dev/mock sync paths now pass the `agent_url` (already computed in those functions) to the discovery service.

## Test plan

- [x] `property_ids` authorization resolves correctly with agent_url (Prisma Media pattern)
- [x] Multi-domain adagents.json links each property to correct publisher only
- [x] `property_tags` authorization resolves correctly
- [x] `publisher_properties` cross-domain selectors are filtered out
- [x] Backward compatibility: without agent_url, old code path still works
- [x] Sync wrapper passes agent_url through
- [x] All 1971 tests pass, `make quality` clean

### DIFF
--- a/src/services/property_discovery_service.py
+++ b/src/services/property_discovery_service.py
@@ -17,7 +17,7 @@ from adcp import (
     get_all_properties,
     get_all_tags,
 )
-from adcp.adagents import get_properties_by_agent
+from adcp.adagents import get_properties_by_agent, normalize_url
 from sqlalchemy import select
 
 from src.core.database.database_session import get_db_session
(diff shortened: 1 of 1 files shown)


lclaudon:
recheck

lclaudon:
I have read the IPR Policy

PR #1063, opened 2026-02-23
feat: implement get_media_buys tool with delivery snapshots

bokelley:
## Summary

- Adds \`get_media_buys\` MCP/A2A tool returning media buys with status, creative approval state, and optional delivery snapshots
- Fixes \`ADCP_AUTH_TEST_MODE\` not blocking Google OAuth redirects in admin UI login routes
- Fixes GAM client using deprecated/disabled API version \`v202411\` — now uses \`GAM_API_VERSION\` constant
- Upgrades Flask to 3.1.3 (GHSA-68rp-wp8r-4726) and Werkzeug to 3.1.6 (GHSA-29vq-49wr-vm6x)

## Changes

**\`get_media_buys\` tool** (\`src/core/tools/media_buy_list.py\`)
- Returns media buys for the authenticated principal with computed status (\`active\`, \`pending_activation\`, \`completed\`)
- Supports filtering by \`media_buy_ids\`, \`buyer_refs\`, or \`status_filter\` — status filter applies consistently across all three paths
- Optional \`include_snapshot=true\` fetches near-real-time delivery data from the adapter
- Raises \`ToolError\` for unsupported \`account_id\` parameter rather than silently ignoring it

**GAM adapter** (\`src/adapters/google_ad_manager.py\`)
- \`get_packages_snapshot\` adds tenant_id filter to prevent cross-tenant data leakage
- Moves ORM attribute access inside the session to prevent \`DetachedInstanceError\`

**Schemas** (\`src/core/schemas.py\`)
- \`GetMediaBuysPackage\`, \`GetMediaBuysMediaBuy\`, \`GetMediaBuysRequest\`, \`GetMediaBuysResponse\`
- \`GetMediaBuysMediaBuy.model_dump()\` override ensures nested packages serialize correctly
- \`Snapshot\`, \`SnapshotUnavailableReason\`, \`DeliveryStatus\`, \`CreativeApproval\`

**Adapters**
- \`AdapterCapabilities.supports_realtime_reporting\` flag replaces \`hasattr()\` check
- GAM and mock adapters both declare \`supports_realtime_reporting=True\`

## Test plan

- [ ] Unit tests: \`uv run pytest tests/unit/test_get_media_buys.py -v\` (23 tests)
- [ ] Full unit suite: \`uv run pytest tests/unit/ -x -q\` (2004 passing)
- [ ] Tool verified against running server: returns active media buys, respects status filter, handles empty results
- [ ] Security audit: Flask 3.1.3 / Werkzeug 3.1.6 patch CVEs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -351,6 +351,9 @@ jobs:
         # Don't set ADCP_TESTING=true - let E2E tests manage their own Docker stack
         ADCP_SALES_PORT: 8080
         A2A_PORT: 8091
+        # Set explicit ports to avoid dynamic allocation race conditions
+        POSTGRES_PORT: 5435
+        ADMIN_UI_PORT: 8093
         # Enable Docker BuildKit for faster builds with caching
         DOCKER_BUILDKIT: 1
         COMPOSE_DOCKER_CLI_BUILD: 1
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -24,8 +24,9 @@ def find_free_port(start_port: int = 10000, end_port: int = 60000) -> int:
     """Find an available port in the given range."""
     for port in range(start_port, end_port):
         with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 0)
             try:
-                s.bind(("127.0.0.1", port))
+                s.bind(("0.0.0.0", port))  # Match Docker's binding interface
                 return port
             except OSError:
                 continue


bokelley:
> Hey Brian — nice addition, the `_impl()` pattern and status computation look solid, and the auth test mode fix is a genuine catch.
> 
> A few questions before I approve:
> 
> **1. Merge conflict heads-up** #1062 touches some of the same files — GAM API version centralization, Flask/Werkzeug bumps, and the `buying_mode` schema validation test. Happy to rebase whichever lands second. Just want to coordinate so we don't step on each other.
> 
> **2. `hasattr` check for snapshot support (`media_buy_list.py`)** `get_packages_snapshot` is defined on `BaseAdServer` (raises `NotImplementedError`), so `hasattr(adapter, "get_packages_snapshot")` will always be `True` for every adapter — including ones that haven't overridden it. The `NotImplementedError` then gets caught by the broad `except Exception` and the user silently gets no snapshot with no indication why. Would a capability flag or checking for the override work better here?
> 
> **3. Nested serialization (`schemas.py`)** `GetMediaBuysResponse` has a few layers of nesting (`GetMediaBuysMediaBuy` → `GetMediaBuysPackage` → `Snapshot`/`CreativeApproval`). We've been bitten by the `model_dump()` not recursing into nested models before — could we add a roundtrip serialization test to make sure the A2A path gets the right shape back?
> 
> **4. `status_filter` + `media_buy_ids` interaction (`media_buy_list.py`)** When `media_buy_ids` is provided, the `status_filter` parameter doesn't get applied. Is that intentional? If so, a comment would help — otherwise it might surprise callers who pass both.

good feedback! think I addressed them. Either way on #1, happy to merge after you

PR #1066, opened 2026-02-24
refactor: FastAPI migration — unify MCP + A2A + Admin into single process

### LINKED ISSUE #1050: Architecture Proposal: FastAPI as Unified Application Framework
Inspired by this suggestion: #1048 and my own learnings from the refactoring attempts.

## Motivation

The project serves three transports — MCP (AI agent tools), A2A (agent-to-agent JSON-RPC), and Admin UI (Flask). The business logic is shared via `_impl()` functions, but everything between "request arrives" and "call `_impl()`" is reimplemented per transport:

- **MCP**: FastMCP handles routing, validation, serialization, errors, dependency injection. Handlers are clean.
- **A2A**: No framework. 500+ lines of hand-rolled routing, auth, validation, serialization, error handling in `adcp_a2a_server.py`. Handlers mix framework concerns with business logic.
- **Admin UI**: Flask handles its own routes. Separate process.

This causes real problems:

1. **Handlers do framework work.** Every A2A handler does its own auth extraction, parameter validation, error formatting, and response serialization. Adding a new skill means copying 30 lines of boilerplate.

2. **Errors are dicts, not exceptions.** Six A2A handlers return `{"success": False, "message": "..."}` dicts instead of throwing typed exceptions. The dispatch can't distinguish error responses from success responses without `isinstance(result, dict)`.

3. **Three serialization paths for the same data.** `get_products` has three divergent serialization flows (MCP, A2A explicit, A2A natural language) that produce slightly different output for the same business logic call.

4. **Version compat is tangled into handlers.** V2 backward-compatibility (adding `is_fixed`, `rate`, `price_guidance.floor` to pricing options for pre-3.0 clients) is applied differently in each path — some unconditionally, some gated on version, some not at all.

5. **No REST API.** AdCP promises REST support. Adding it today means building a fourth path with a fourth reimplementation of auth/validation/serialization.

The root cause: there is no shared application framework. FastMCP is a framework for MCP. A2A has no framework. Each transport reinvents the same infrastructure.

### Deployment

```
nginx (:8000)
  /mcp/*    ->  adcp-server (:8080)   [FastMCP, mcp.run()]
  /a2a/*    ->  adcp-server (:8091)   [Starlette via A2A SDK]
  /admin/*  ->  admin-ui (:5000)      [Flask]
```

Two Python processes for MCP and A2A. Three frameworks.

### Framework Capabilities by Transport

| Capability              | MCP (FastMCP)  | A2A               | Admin (Flask)  |
|-------------------------|----------------|--------------------|----------------|
| Routing                 | @mcp.tool      | Hand-rolled        | @app.route     |
| Request validation      | TypeAdapter    | Manual per-handler | Manual         |
| Response serialization  | to_jsonable_py | model_dump() everywhere | jsonify    |
| Error handling          | ToolError exc  | Dict returns + catch-all | abort()   |
| Auth/DI                 | Context inject | Manual per-handler | decorators     |
| Middleware              | Middleware API | None               | before_request |

### A2A Request Flow (current)

```
on_message_send(params)
  |
  |-- Parse text/data parts (manual, 50 lines)
  |-- Extract auth token (manual)
  |-- Route: keyword match for NL, skill name for explicit (manual, 80 lines)
  |
  |-- _handle_explicit_skill(skill_name, parameters, auth_token)
  |     |-- Map skill_name to handler (dict lookup)
  |     |-- Call handler
  |     |-- _serialize_for_a2a(result) -- adds protocol fields
  |     |-- Build Task/Artifact/DataPart (manual, 60 lines)
  |
  |-- Individual handlers (15 of them):
        |-- Auth extraction (repeated in every handler)
        |-- Parameter validation (repeated, returns error dicts)
        |-- Call core _raw() function
        |-- Serialization (inconsistent per handler)
        |-- Return dict or model (inconsistent)
```

### MCP Request Flow (current)

```
FastMCP receives JSON-RPC call
  |-- Validates args against function signature (automatic)
  |-- Injects Context (automatic)
  |-- Calls @mcp.tool function
  |     |-- Build request model
  |     |-- Call _impl()
  |     |-- response.model_dump()      <-- handler does serialization (wrong)
  |     |-- add_v2_compat()            <-- handler does version compat (wrong)
  |     |-- Return ToolResult(dict)
  |-- FastMCP sends response (automatic)
```

Even in MCP where the framework is good, handlers reach into the serialization layer.

### Use FastAPI as the unified application framework.

FastAPI is built on Starlette (already a dependency). It provides routing, Pydantic validation, automatic serialization, dependency injection, exception handlers, middleware, OpenAPI docs, and streaming — everything we're hand-rolling or missing.

The A2A SDK already ships `A2AFastAPIApplication` — a FastAPI-native variant of the Starlette integration. FastMCP's `http_app()` returns a Starlette app that mounts onto FastAPI.

### Target Architecture

```
FastAPI (:8080)
  |
  |-- Middleware pipeline (shared):
  |     |-- CORS
  |     |-- Auth extraction
  |     |-- Request logging
  |     |-- Version compat (post-response transform)
  |
  |-- /mcp/*     ->  FastMCP mounted as Starlette sub-app
  |-- /a2a/*     ->  A2A SDK mounted via A2AFastAPIApplication
  |-- /api/v1/*  ->  REST routes (FastAPI native)
  |
  |-- Exception handlers (shared):
  |     |-- AdCPValidationError -> 400 + AdCP error format
  |     |-- AuthenticationError -> 401
  |     |-- ToolError -> mapped to transport-specific error
  |
  |-- Dependency injection (shared):
        |-- get_auth_context() -> AuthContext
        |-- get_adcp_version() -> str | None
```

### Handler Pattern (target)

Handlers become pure async functions. No auth. No validation. No serialization. No error dicts.

```python
# Before (A2A handler, current):
async def _handle_get_products_skill(self, parameters: dict, auth_token: str | None) -> Any:
    try:
        tool_context: ToolContext | MinimalContext
        if auth_token:
            tool_context = self._create_tool_context_from_a2a(auth_token=auth_token, tool_name="get_products")
        else:
            tool_context = MinimalContext.from_request_context()

        brand_manifest = parameters.get("brand_manifest")
        if isinstance(brand_manifest, str):
            brand_manifest = {"url": brand_manifest}
        elif brand_manifest is not None and not isinstance(brand_manifest, dict):
            raise ServerError(InvalidParamsError(...))

        brief = parameters.get("brief", "")
        if not brief and not brand_manifest:
            raise ServerError(InvalidParamsError(...))

        if isinstance(tool_context, ToolContext):
            mcp_ctx = self._tool_context_to_mcp_context(tool_context)
        else:
            mcp_ctx = cast(ToolContext, tool_context)

        response = await core_get_products_tool(...)

        adcp_version = parameters.get("adcp_version")
        if needs_v2_compat(adcp_version):
            if isinstance(response, dict):
                response_data = response
            else:
                response_data = response.model_dump(mode="json")
            if "products" in response_data:
                response_data["products"] = add_v2_compat_to_products(response_data["products"])
            return response_data
        return response

    except Exception as e:
        raise ServerError(InternalError(message=f"Unable to retrieve products: {str(e)}"))


# After (pure handler):
async def get_products(req: GetProductsRequest, ctx: AuthContext) -> GetProductsResponse:
    return await _get_products_impl(req, ctx)
```

Auth, validation, serialization, version compat, error handling — all handled by the framework.

### Version Compat as Middleware

```python
# Version compat is a response transform, not handler logic.
# Registered per-tool (only tools that need it).

_version_transforms: dict[str, Callable] = {}

def register_version_transform(tool_name: str):
    def decorator(fn):
        _version_transforms[tool_name] = fn
        return fn
    return decorator

@register_version_transform("get_products")
def get_products_v2_compat(response_data: dict, adcp_version: str | None) -> dict:
    if needs_v2_compat(adcp_version) and "products" in response_data:
        response_data["products"] = add_v2_compat_to_products(response_data["products"])
    return response_data
```

The middleware applies the registered transform after serialization, before sending the response. Same transform runs regardless of transport (MCP, A2A, REST).

### Error Handling as Exceptions

```python
# Before: dict returns mixed with model returns
if missing_params:
    return {
        "success": False,
        "message": f"Missing required AdCP parameters: {missing_params}",
        "errors": [{"code": "validation_error", "message": f"Missing: {missing_params}"}],
    }

# After: typed exceptions caught by framework
if missing_params:
    raise AdCPValidationError(
        message=f"Missing required AdCP parameters: {missing_params}",
        code="validation_error",
    )

# Framework exception handler (registered once):
@app.exception_handler(AdCPValidationError)
async def handle_adcp_validation(request, exc):
    return JSONResponse(status_code=400, content={
        "success": False,
        "message": str(exc),
        "errors": [{"code": exc.code, "message": str(exc)}],
    })
```

### A2A Integration

The A2A SDK's `on_message_send` still handles the A2A JSON-RPC protocol (parsing message parts, building Task/Artifact responses). But skill dispatch calls the same handlers as REST:

```python
async def on_message_send(self, params, context):
    # A2A protocol parsing (SDK concern)
    skill_invocations = parse_skill_invocations(params.message)

    for invocation in skill_invocations:
        # Call the shared handler (same as REST endpoint)
        response = await dispatch_skill(invocation.skill, invocation.parameters, auth_context)
        # Version compat applied by shared pipeline
        # Serialization applied by shared pipeline

    # A2A envelope (SDK concern)
    return build_task_with_artifacts(results)
```

The `dispatch_skill` function uses the same routing, validation, and serialization as REST. A2A's `on_message_send` is just a protocol translator.

### MCP Integration

FastMCP tools can delegate to the shared handlers:

```python
@mcp.tool
async def get_products(brand_manifest: BrandManifest | None = None, brief: str = "", ..., ctx: Context) -> GetProductsResponse:
    req = create_get_products_request(brief=brief, brand_manifest=brand_manifest, ...)
    return await _get_products_impl(req, ctx)
    # FastMCP serializes the response via to_jsonable_python()
    # Version compat applied by FastMCP middleware (on_call_tool hook)
```

FastMCP's `Middleware.on_call_tool` hook applies the version compat transform after the tool returns.

## Key Properties

1. **One handler, three transports.** Business logic is written once. MCP, A2A, and REST are transport adapters.
2. **Framework handles boilerplate.** Auth, validation, serialization, error handling are FastAPI's job.
3. **Errors are exceptions.** Handlers never return error dicts. The framework catches exceptions and formats error responses per transport.
4. **Version compat is a registered post-response transform.** Not handler logic. Applied uniformly regardless of transport.
5. **REST API for free.** Adding a REST endpoint is `@app.post("/api/v1/endpoint")` pointing at the same handler.

### New dependency

- `fastapi` (built on Starlette, which is already installed)

### Structural changes

- Create a shared handler layer (plain async functions with typed params)
- Move auth to FastAPI dependency injection
- Move validation errors from dict returns to exceptions
- Create FastAPI exception handlers for AdCP error types
- Mount FastMCP and A2A SDK on the FastAPI app
- Version compat as middleware/post-response hook

### Migration path

This can be done incrementally, one handler at a time:
1. Add FastAPI, create the main app, mount existing MCP and A2A sub-apps (no behavior change)
2. Extract one handler (e.g., get_products) to the shared pattern
3. Wire it through all three transports
4. Repeat for remaining handlers
5. Delete the hand-rolled A2A dispatch code as handlers migrate

### What stays the same

- `_impl()` functions (business logic core) — unchanged
- Database layer — unchanged
- Adapter layer (GAM, Mock) — unchanged
- Admin UI (Flask) — separate concern, can stay as-is or migrate later
- A2A SDK's `RequestHandler.on_message_send` — still handles A2A protocol, just delegates to shared handlers
- FastMCP's `@mcp.tool` — still registers MCP tools, just delegates to shared handlers

## Open Questions

1. **Single process or keep separate?** Currently MCP and A2A are in separate processes. FastAPI can host both, but merging processes changes the deployment topology. -- I think it is better to keep a single one

2. **Admin UI migration?** Flask admin UI is a separate concern (HTML templates, OAuth, etc.). Could stay as Flask mounted on FastAPI via WSGIMiddleware, or migrate per #1048 suggestion

3. **FastMCP middleware vs FastAPI middleware for MCP tools?** FastMCP has its own middleware pipeline (`Middleware.on_call_tool`). Version compat for MCP could use either FastMCP middleware or be applied in the tool function. FastMCP middleware only runs through the MCP protocol path — it doesn't run when calling `ToolManager.call_tool` directly. -- So I'd say better to use it as library and unify the middleware and handling

4. **A2A natural language path?** NL queries currently use keyword matching to route to handlers. This is independent of the framework question — it's application-level routing that stays in `on_message_send`. The fix is simpler: NL routes through the same explicit handler (pricing queries already do this).

5. **Gradual vs big-bang migration?** The incremental path (one handler at a time) is safer but means a period where both patterns coexist. -- I'd just "shoot" :-)

KonstantinMirin:
Closes #1050

## Summary

Unify the three separate processes (MCP, A2A, Flask Admin) into a single FastAPI process with clean separation of concerns — and fix the test suite from 138 warnings / flaky failures to 3,490 passing tests, 0 warnings, 51% line coverage.

## The core problem

Every transport (MCP, A2A, Admin) reimplemented auth extraction, tenant resolution, error formatting, and serialization. Every `_impl()` function reached into framework concerns — reading tenant from ContextVars, calling `set_current_tenant()`, catching transport-specific `ToolError`. This made it impossible to add a transport without duplicating all of that, and impossible to change any of it without touching every transport.

## What changed: data flow unification

The key idea is that each concern is handled **exactly once, in one place**.

### Before: every layer does everything

Each handler and `_impl()` function mixes auth, tenant resolution, error handling, serialization, and version compat. Adding a new tool means copying all of it.

```mermaid
flowchart TD
    subgraph MCP["MCP Handler"]
        M1[Extract auth token]
        M2[Call _impl]
        M3[set_current_tenant]
        M4[get_current_tenant from ContextVar]
        M5[get_adapter via ContextVar]
        M6[Catch ToolError]
        M7[response.model_dump]
        M8[add_v2_compat]
        M1 --> M2 --> M3 --> M4 --> M5 --> M6 --> M7 --> M8
    end

    subgraph A2A["A2A Handler"]
        A1[Extract auth token]
        A2[Validate parameters manually]
        A3[Call _impl]
        A4[set_current_tenant]
        A5[get_current_tenant from ContextVar]
        A6[get_adapter via ContextVar]
        A7[Return error dict]
        A8[Serialize for A2A]
        A1 --> A2 --> A3 --> A4 --> A5 --> A6 --> A7 --> A8
    end

    style M1 fill:#f9c74f
    style A1 fill:#f9c74f
    style M3 fill:#f4845f
    style M4 fill:#f4845f
    style A4 fill:#f4845f
    style A5 fill:#f4845f
    style M6 fill:#c1121f
    style A7 fill:#c1121f
    style M7 fill:#90be6d
    style M8 fill:#90be6d
    style A8 fill:#90be6d
    style A2 fill:#577590
```

🟡 Auth  🔴 Tenant ContextVar  🔵 Validation  🟢 Serialization — all duplicated across transports and mixed into `_impl()`

### After: each concern lives in one layer

```mermaid
flowchart TD
    subgraph boundary["Transport Boundary (MCP / A2A / REST)"]
        direction TB
        B1["Auth extraction → principal_id"]
        B2["Tenant resolution → LazyTenantContext"]
        B3["Build ResolvedIdentity"]
        B1 --> B2 --> B3
    end

    subgraph impl["Business Logic (_impl)"]
        direction TB
        I1["Receive (request, identity)"]
        I2["identity.tenant_id — no DB query"]
        I3["identity.tenant.ad_server — lazy DB load"]
        I4["get_adapter(tenant=tenant) — explicit"]
        I5["Raise AdCPError — transport-agnostic"]
        I6["Return typed response model"]
        I1 --> I2 --> I3 --> I4 --> I5 --> I6
    end

    subgraph translate["Error Boundary"]
        direction TB
        E1["AdCPError → ToolError (MCP)"]
        E2["AdCPError → ServerError (A2A)"]
        E3["AdCPError → JSON 4xx (REST)"]
    end

    subgraph serial["Response"]
        direction TB
        S1["Framework serializes model"]
        S2["Version compat transform"]
    end

    boundary --> impl --> translate --> serial

    style B1 fill:#f9c74f
    style B2 fill:#f9c74f
    style B3 fill:#f9c74f
    style I1 fill:#4895ef
    style I2 fill:#4895ef
    style I3 fill:#4895ef
    style I4 fill:#4895ef
    style I5 fill:#4895ef
    style I6 fill:#4895ef
    style E1 fill:#c1121f,color:#fff
    style E2 fill:#c1121f,color:#fff
    style E3 fill:#c1121f,color:#fff
    style S1 fill:#90be6d
    style S2 fill:#90be6d
```

🟡 Auth + tenant — resolved once at boundary  🔵 Business logic — pure functions, no global state  🔴 Errors — translated at the edge  🟢 Serialization — framework handles it

### Data flow through the layers

```mermaid
sequenceDiagram
    participant T as Transport<br/>(MCP / A2A / REST)
    participant B as Boundary<br/>(resolve_identity)
    participant I as _impl()<br/>(business logic)
    participant L as LazyTenantContext
    participant DB as Database

    T->>B: incoming request + auth token
    B->>B: extract principal_id from token
    B->>B: resolve tenant_id from header/subdomain
    B->>B: create LazyTenantContext(tenant_id)
    Note over B: No DB query yet
    B->>I: _impl(request, ResolvedIdentity)

    I->>L: identity.tenant.tenant_id
    L-->>I: "test_tenant" (immediate, no DB)

    I->>L: identity.tenant.ad_server
    L->>DB: get_tenant_by_id("test_tenant")
    DB-->>L: {tenant_id, name, ad_server, ...}
    Note over L: Cached — subsequent<br/>accesses don't hit DB
    L-->>I: "gam"

    I->>I: get_adapter(tenant=tenant)
    I->>I: business logic...
    I-->>T: GetProductsResponse (typed model)

    alt Success
        T->>T: framework serializes response
    else AdCPError raised
        T->>T: translate to transport-specific error
    end
```

### Lazy tenant loading detail

~85% of call paths only need `tenant_id`, which is available without a DB query. Full tenant data loads transparently on first access to any other field.

```mermaid
stateDiagram-v2
    [*] --> Unloaded: LazyTenantContext(tenant_id)

    state Unloaded {
        [*] --> Available
        Available: tenant_id ✅ (immediate)
        Available: __contains__ ✅ (no DB)
        Available: __bool__ ✅ (always True)
    }

    Unloaded --> Loaded: Access .name, .ad_server,<br/>.approval_mode, etc.

    state Loaded {
        [*] --> Resolved
        Resolved: All 22 fields available
        Resolved: Backed by TenantContext model
        Resolved: set_current_tenant() called once
        Resolved: Cached — no repeat queries
    }

    note right of Unloaded: ~85% of _impl calls<br/>stay here
    note right of Loaded: Only when _impl needs<br/>full tenant config
```

## Concrete changes

**1. Unified process** — Single FastAPI app replaces 3 subprocesses
- FastMCP v2 → v3.0.2, mounted at `/mcp` via `http_app()`
- A2A mounted via `A2AFastAPIApplication.add_routes_to_app()`
- Flask Admin mounted via `WSGIMiddleware`
- nginx configs → single upstream, Docker Compose → single service

**2. Identity resolution** — Auth + tenant resolved once at boundary
- `ResolvedIdentity` — frozen Pydantic model carrying resolved auth + tenant context
- `TenantContext` / `LazyTenantContext` — typed model replacing raw dicts, defers DB load until needed (~85% of calls never hit DB)
- `get_adapter()` takes explicit `tenant` parameter (no ContextVar reads)
- All 12 `_impl` modules use `identity.tenant` exclusively
- Single pure ASGI `UnifiedAuthMiddleware` replaces 3 separate auth middlewares
- Each transport delegates to `resolve_identity()` via its own mechanism: FastMCP Middleware (MCP), `_resolve_a2a_identity` (A2A), FastAPI `Depends` (REST)

**3. Error boundary** — Transport-agnostic errors in business logic
- `AdCPError` hierarchy (`Validation`, `Authentication`, `Authorization`, `Adapter`) replaces `ToolError` in 42+ sites across `_impl` functions
- Error boundary middleware translates at each transport edge
- `_impl` never sees `ToolError`, `ServerError`, or HTTP status codes

**4. Structural enforcement** — 77 tests verify architecture invariants
- `_impl` functions never touch ContextVars
- `resolve_identity()` internals are pure (no `set_current_tenant()` side effects)
- Token extraction is consistent across all transports
- Auth infrastructure uses shared constants and immutable types

**5. REST API** — Full feature parity, for free

Because all business logic lives in shared `_impl` functions, adding a REST transport was a single 183-line file (`src/routes/api_v1.py`) with zero changes to business logic. It has complete feature parity with MCP and A2A — same 10 tools, same auth, same validation, same errors — just JSON-RPC over HTTP instead of MCP/A2A protocols. Swagger UI auto-generated at `/docs`, OpenAPI spec at `/openapi.json`. This layer is effectively maintenance-free: when an `_impl` changes, all three transports pick it up automatically.

### Benefits

- **Simplicity**: Adding a new transport (REST took 1 file) doesn't require reimplementing auth/tenant/errors
- **Security**: Auth is resolved once at the boundary, not scattered across handlers. Debug endpoints gated behind `ADCP_TESTING`
- **Stability**: No ContextVar race conditions, no global state mutations in business logic, lazy tenant loading prevents unnecessary DB queries

## Test suite health

The test suite was in rough shape — 138 warnings, flaky failures from resource leaks, unawaited coroutines, stale mocks, and tests that didn't actually test what they claimed. This PR fixes all of that:

| Metric | Before | After |
|--------|--------|-------|
| Passing tests | ~2,474 | **3,490** |
| Warnings | 138 | **0** |
| Failures/flaky | Several | **0** |
| Unit suite time | 54s | **17s** |
| Full suite parallelism | Sequential | **5 suites in parallel (tox)** |
| Coverage reporting | None | **51% line + branch** |

### What was fixed

- **Resource leaks**: Unclosed async generators, HTTPServer sockets not released, asyncio ResourceWarnings from dependent libraries
- **Stale mocks**: Tests patching deleted functions, xfail markers on passing tests, MagicMock objects where ResolvedIdentity was needed
- **Missing coverage**: 29 new unit tests for `update_performance_index` and `get_adcp_capabilities`, 8 new e2e tests for discovery endpoints and multi-tenant isolation
- **Broken context propagation**: `list_authorized_properties` MCP wrapper silently dropped the `context` parameter
- **Slow tests**: `time.sleep` in retry/backoff tests replaced with mocks (54s → 17s)

### Test infrastructure

- **tox + tox-uv** for parallel test orchestration with combined coverage
- **`./run_all_tests.sh`** — one command: starts Docker, runs `tox -p` (5 suites in parallel), combines coverage, tears down Docker
- **`make test-stack-up` / `make test-stack-down`** — Docker lifecycle separated for fast iteration
- **Coverage**: `htmlcov/index.html` (HTML) + `coverage.json` (machine-readable), per-env isolation prevents SQLite race conditions
- **Auto-pruning**: `test-results/` keeps only last 10 runs
- **JSON reports**: Per-suite test results in `test-results/<timestamp>/`

## Test plan

- [x] 3,490 tests pass (2,688 unit + 543 integration + 174 integration_v2 + 81 e2e + 4 ui)
- [x] 0 warnings, 0 failures
- [x] 51% line coverage (branch coverage enabled)
- [x] `make quality` passes (ruff format + lint + mypy + unit tests)
- [x] `tox -e unit` verified independently
- [x] Full `./run_all_tests.sh` run verified end-to-end
- [x] 77 structural tests enforce architecture invariants (no ContextVar in _impl, pure resolver, consistent auth)
- [ ] `docker compose up` — unified stack serves all routes
- [ ] Admin UI accessible at `/admin/`
- [ ] Swagger UI at `/docs` lists all 10 REST endpoints

## Nginx configuration simplification

Since all three services (MCP, A2A, Admin) now run in a single FastAPI process, the nginx configs were simplified from ~40 location blocks to catch-all patterns. Each config now uses a single `location /` with WebSocket-safe headers (via `map $http_upgrade $connection_upgrade`) instead of repeating identical proxy blocks per path. Total reduction: 957 → 472 lines (50%) across the three configs (development, single-tenant, multi-tenant).

## Post-review hardening (29 commits)

The initial review surfaced gaps between the claimed architecture and the actual implementation — auth was described as "resolved once at the boundary" but each transport still had its own extraction logic, resolver internals still mutated ContextVars, and A2A had duplicated tenant detection with the wrong strategy order. 29 follow-up commits close these gaps.

**Auth unification completed**
- Replaced 3 separate auth middlewares with a single pure ASGI `UnifiedAuthMiddleware`
- MCP: auth centralized via FastMCP Middleware → `resolve_identity()`
- A2A: identity resolved once at transport boundary and threaded through all handlers; inline tenant detection replaced by delegation to `resolve_identity()`
- REST: converted to FastAPI `Depends` with `Annotated` types
- Eliminated ContextVar dual-write from A2A auth path

**ContextVar side effects eliminated from resolver internals**
- `_detect_tenant()` made pure — 6 `set_current_tenant()` calls removed
- `get_principal_from_token()` returns `(principal_id, tenant_dict)` tuple instead of mutating ContextVar
- `resolve_identity()` no longer falls back to `get_current_tenant()`
- `set_current_tenant()` now called only at transport boundaries, never inside the resolver

**A2A handler correctness**
- Fixed duplicate tenant detection that used the wrong strategy order
- Eliminated double header parsing in A2A middleware
- `CallContextBuilder` wired to thread SDK `ServerCallContext` to handler helpers
- Task IDs use UUID; removed debug log statements; stub methods raise `ServerError`

**Infrastructure hardening**
- `AuthContext.headers` wrapped in `MappingProxyType` (immutable after creation)
- `AUTH_CONTEXT_STATE_KEY` shared constant replaces string literal duplication
- Token extraction priority reconciled between middleware and `resolve_identity()`
- Test paths ported to `pathlib` for cross-platform portability
- Stale comments and docs updated to reflect post-migration architecture

### DIFF
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -229,7 +229,7 @@ def _validate_creatives_before_adapter_call(packages: list[MediaPackage], tenant
         tenant_id: Tenant ID for database lookup
 
     Raises:
-        ToolError: If any creative is missing required fields (URL, dimensions)
+        AdCPValidationError: If any creative is missing required fields (URL, dimensions)
     """
     from sqlalchemy import select
 
@@ -951,7 +951,7 @@ def _validate_pricing_model_selection(
         }
 
     Raises:
-        ToolError: If pricing_model validation fails
+        AdCPValidationError: If pricing_model validation fails
     """
     from decimal import Decimal
 
@@ -1102,7 +1102,7 @@ async def _validate_and_convert_format_ids(
         List of validated FormatId dicts with {agent_url, id}
 
     Raises:
-        ToolError: If any format_id is invalid, unregistered, or doesn't exist
+        AdCPValidationError: If any format_id is invalid, unregistered, or doesn't exist
     """
     from src.core.creative_agent_registry import CreativeAgentRegistry
 
--- a/src/core/helpers/creative_helpers.py
+++ b/src/core/helpers/creative_helpers.py
@@ -519,7 +519,7 @@ def process_and_upload_package_creatives(
         - uploaded_ids_by_product: Mapping of product_id -> uploaded creative IDs
 
     Raises:
-        ToolError: If creative upload fails for any package (CREATIVES_UPLOAD_FAILED)
+        AdCPAdapterError: If creative upload fails for any package (CREATIVES_UPLOAD_FAILED)
 
     Example:
         >>> packages = [PackageRequest(product_id="p1", creatives=[creative1, creative2])]
@@ -585,7 +585,6 @@ def process_and_upload_package_creatives(
         except Exception as e:
             error_msg = f"Failed to upload creatives for package with product_id {product_id}: {str(e)}"
             logger.error(error_msg)
-            # Re-raise as ToolError for consistent error handling
             raise AdCPAdapterError(error_msg, details={"error_code": "CREATIVES_UPLOAD_FAILED"}) from e
 
     return updated_packages, uploaded_by_product
--- a/src/core/tools/media_buy_list.py
+++ b/src/core/tools/media_buy_list.py
@@ -19,6 +19,7 @@ from fastmcp.tools.tool import ToolResult
 from pydantic import RootModel, ValidationError
 from sqlalchemy import select
 
+from src.core.exceptions import AdCPAuthenticationError, AdCPValidationError
 from src.core.resolved_identity import ResolvedIdentity
 from src.core.tool_context import ToolContext
 
@@ -82,10 +83,10 @@ def _get_media_buys_impl(
     from src.core.helpers.context_helpers import ensure_tenant_context
 
     if identity is None:
-        raise ToolError("Identity is required")
+        raise AdCPAuthenticationError("Identity is required")
 
     if req.account_id is not None:
-        raise ToolError("account_id filtering is not yet supported")
+        raise AdCPValidationError("account_id filtering is not yet supported")
 
     testing_ctx = identity.testing_context
     ensure_tenant_context(identity)
--- a/src/core/tools/products.py
+++ b/src/core/tools/products.py
@@ -280,7 +280,6 @@ async def _get_products_impl(
     if policy_result and policy_result.status == PolicyStatus.BLOCKED:
         # Always block if policy says blocked
         logger.warning(f"Brief blocked by policy: {policy_result.reason}")
-        # Raise ToolError to properly signal failure to client
         raise AdCPAuthorizationError(
             policy_result.reason or "Blocked by policy", details={"error_code": "POLICY_VIOLATION"}
         )
@@ -781,7 +780,6 @@ async def get_products(
     except ValidationError as e:
         raise AdCPValidationError(format_validation_error(e, context="get_products request")) from e
     except ValueError as e:
-        # Convert ValueError from helper to ToolError with clear message
         raise AdCPValidationError(f"Invalid get_products request: {e}") from e
 
     # Read identity pre-resolved by MCPAuthMiddleware
--- a/tests/integration/test_gam_pricing_models_integration.py
+++ b/tests/integration/test_gam_pricing_models_integration.py
@@ -27,7 +27,6 @@ from src.core.database.models import (
     PropertyTag,
     Tenant,
 )
-from src.core.exceptions import AdCPAdapterError
 from src.core.resolved_identity import ResolvedIdentity
 from src.core.schemas import CreateMediaBuyRequest
 from src.core.testing_hooks import AdCPTestContext
--- a/tests/unit/test_get_media_buys.py
+++ b/tests/unit/test_get_media_buys.py
@@ -381,12 +381,46 @@ class TestGetMediaBuysImpl:
         assert pkg_response.snapshot_unavailable_reason == SnapshotUnavailableReason.SNAPSHOT_UNSUPPORTED
 
     def test_identity_required(self):
-        """identity=None raises ToolError."""
+        """identity=None raises AdCPAuthenticationError, not MCP ToolError.
+
+        _impl functions are shared across MCP, A2A, and REST transports.
+        They must raise domain exceptions (AdCPError hierarchy), never
+        transport-specific types like ToolError.
+        """
+        from src.core.exceptions import AdCPAuthenticationError
+
+        req = self._make_request()
+        with pytest.raises(AdCPAuthenticationError, match="Identity is required"):
+            _get_media_buys_impl(req, None)
+
+    def test_account_id_filtering_raises_domain_exception(self):
+        """account_id filtering raises AdCPValidationError, not MCP ToolError.
+
+        Same principle: _impl must use domain exceptions for all transports.
+        """
+        from src.core.exceptions import AdCPValidationError
+
+        req = self._make_request(account_id="some_account")
+        with pytest.raises(AdCPValidationError, match="account_id filtering is not yet supported"):
+            _get_media_buys_impl(req, make_identity())
+
+    def test_impl_does_not_raise_tool_error(self):
+        """_impl must never raise ToolError — it's an MCP transport concern.
+
+        Regression guard: if someone re-imports ToolError into _impl,
+        this test catches it.
+        """
         from fastmcp.exceptions import ToolError
 
         req = self._make_request()
-        with pytest.raises(ToolError, match="Identity is required"):
+        # identity=None should raise, but NOT ToolError
+        with pytest.raises(Exception) as exc_info:
             _get_media_buys_impl(req, None)
+        assert not isinstance(exc_info.value, ToolError), (
+            f"_get_media_buys_impl raised ToolError({exc_info.value}) — "
+            f"_impl functions must raise domain exceptions (AdCPError hierarchy), "
+            f"not MCP transport types"
+        )
 
 
 class TestGetMediaBuysResponseStructure:


src/routes/api_v1.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    from src.core.auth_context import AuthContext
+    from src.core.resolved_identity import ResolvedIdentity, resolve_identity
+
+    auth_ctx: AuthContext = getattr(request.state, "auth_context", AuthContext.unauthenticated())
+
+    if not auth_ctx.auth_token:
+        return None, None
+
+    principal_id = get_principal_from_token(auth_ctx.auth_token)
```
ChrisHuie: this makes 2 DB queries (Principal SELECT + Tenant SELECT) to validate the token without a tenant_id.

src/core/tools/media_buy_create.py
```diff
@@ -2249,7 +2261,7 @@ def unwrap_po(po: Any) -> Any:
         # Lazy import to avoid circular dependency with main.py
         from src.core.main import get_product_catalog
```
ChrisHuie: imports old get_product_catalog => `from src.core.tools.products import get_product_catalog`

Issue #1067, opened 2026-02-25
bug: new product server error


githubMarcin:
Hi

I’ve found two issues (not sure if they are specific to my setup) related to the new product features.

I’m using Docker Desktop on Windows 11 to run the Prebid Sales Agent.

With version 3.11, I’m unable to save a new product. The system requires selecting a pricing model, but even after choosing one, the product cannot be saved.

With the current version, I get a 500 server error when trying to access the New Product page.

Screenshot attached.
https://drive.google.com/file/d/1V6a_AHsDTjpaTjIIh0QjHkm9RUg8p4Dg/view?usp=sharing
https://drive.google.com/file/d/1Cr6WRX83Vyf7hET5UiqL_ENh1v3hUjBK/view?usp=sharing
https://drive.google.com/file/d/1x_51Q7qSZmFqKkmLGL9VOyahYRThnIwN/view?usp=sharing

Thank you

numaras:
On version 3.11, are you using GAM or other AdServer? 
If you go to console in developer options, do you get 

A quick debug points to a broken function:
/src/blueprints/products.py selects templates/add_product_gam.html or add_product.html depending on which adserver is configured. The update pricing fields function in add_product.html is with a lot of "\", completely breaking the syntax:

```
window[\`updatePricingFields\${index}\`] = function() {
        const pricingModel = document.querySelector(\`select[name="pricing_model_\${index}"]\`).value;
        const rateGroup = document.getElementById(\`rate-group-\${index}\`);
        const floorGroup = document.getElementById(\`floor-group-\${index}\`);
        const priceGuidance = document.getElementById(\`price-guidance-\${index}\`);

        // Determine if fixed or auction from pricing model selection
        const isAuction = pricingModel.includes('_auction') || pricingModel === 'vcpm' || pricingModel === 'cpc';

        if (isAuction) {
            // Auction mode (cpm_auction or cpc_auction)
            rateGroup.style.display = 'none';
            floorGroup.style.display = 'block';
            priceGuidance.style.display = 'block';
            document.querySelector(\`input[name="rate_\${index}"]\`).required = false;
            document.querySelector(\`input[name="floor_\${index}"]\`).required = true;
        } else {
            // Fixed rate mode
            rateGroup.style.display = 'block';
            floorGroup.style.display = 'none';
            priceGuidance.style.display = 'none';
            document.querySelector(\`input[name="rate_\${index}"]\`).required = true;
[...]

Issue #1043, opened 2026-02-13
chore: Upgrade adcp library to 3.5.0


KonstantinMirin:
## Summary

Upgrade the adcp Python library from 3.2.0 to 3.5.0 to gain access to new AdCP v3 types and enable 9 feature issues currently waiting on library support.

### New request/response types
- `SyncAccountsRequest/Response`, `ListAccountsRequest/Response`, `Account` — multi-advertiser account management
- `GetMediaBuyArtifactsRequest/Response`, `Artifact` — media buy artifact retrieval
- `LogEventRequest/Response`, `SyncEventSourcesRequest/Response` — conversion tracking
- `GetCreativeDeliveryRequest/Response` — creative-level delivery metrics
- `DeliveryForecast`, `ForecastMethod`, `ForecastPoint`, `ForecastRange` — delivery forecasting

### New targeting and pricing types
- `DaypartTarget` — daypart targeting
- `PricingModel.cpa`, `PricingModel.time` — CPA and time-based pricing

### Improved core types
- `PaginationRequest` (cursor, max_results) and `PaginationResponse` (cursor, has_more, total_count) — cursor-based pagination
- `BrandManifestReference` (`RootModel[BrandManifest | AnyUrl]`) — unified brand manifest handling
- `GetProductsRequest` now includes all 10 spec fields (product_selectors, pagination, account_id, etc.)
- `Creative` restructured with `variants` and `variant_count` for multi-variant creative support
- `Property` updated with `identifier` and `type` fields

## Migration scope

The upgrade requires updating models that extend library types to align with the new field structures:

| Area | Scope | Est. test updates |
|------|-------|-------------------|
| Creative model | Align with variants-based structure | ~30 tests |
| Property model | Align with identifier/type fields | ~4 tests |
| Pagination | Adopt cursor-based PaginationRequest/PaginationResponse, update import paths | ~10 tests |
| GetProductsRequest | Extend library directly (zero local fields needed) | ~5 tests |
| GetMediaBuyDeliveryRequest | Remove redundant account_id override | ~2 tests |
| Import paths | Pagination moved from `list_creatives_request/response` to `core/pagination_request/pagination_response` | ~5 tests |

### Confirmed migration pattern

GetProductsRequest migration has been validated — all 10 fields inherited from library with zero local re-declarations:

```python
class GetProductsRequest(LibraryGetProductsRequest):
    model_config = ConfigDict(extra=get_pydantic_extra_mode())
```

[...]

Issue #1041, opened 2026-02-13
Bug: update_media_buy broken when manual approval enabled


estofko-scope3:
When `manual_approval_required=true`, all `update_media_buy` operations fail silently. The function returns early with empty `affected_packages` and no changes are persisted. This blocks all updates: creative assignments, budget changes, pause/resume, targeting, and date changes.

**Reproduction Steps**
1. Configure tenant with manual_approval_required=true
2. Create a media buy (works, enters approval workflow)
3. Call update_media_buy with any changes (creative_ids, budget, paused, etc.)
4. Expected: Changes applied or queued for approval
5. Actual: Returns empty affected_packages, nothing happens

KonstantinMirin:
**Update:** Further tracing revealed that `update_media_buy` also lacks the protocol-level `status` wrapper that `create_media_buy` uses (via `CreateMediaBuyResult`). Even after fixing the approval gate, the A2A path would still report the task as `completed` instead of `submitted`. This is part of a broader architectural issue where TaskStatus (protocol envelope concern) is mixed into domain responses rather than being set by the protocol layer. We'll fix update_media_buy to match the existing pattern for now; proper separation is tracked separately.

ChrisHuie:
Re-triaged at HEAD a55db82d.

- Core bug (manual-approval updates accepted but never applied) is live. Root cause traced: the update path never sets `status="pending_approval"` and the admin approval handlers ignore `mapping.action`, so the stored update is never replayed; the only executor is the create path (`src/admin/blueprints/workflows.py:199`, `src/admin/blueprints/operations.py:368`, `src/core/tools/media_buy_update.py:376-410`).
- Removing the "blocked by #1069" status. #1069 (Post-2.0/Low) targets the auto-create path and states the manual-approval defer-to-approval pattern is already correct; #1037 fixed the same class on the create path without it.
- Secondary pause/resume "workflow step stuck in in_progress" bug is fixed (`src/core/tools/media_buy_update.py:551`, regression test present).
- Original fix (populate `affected_packages` on the success shape) is superseded by the pinned spec's `UpdateMediaBuySubmitted` envelope.
- Rescoped to full end-to-end for the manual-approval update path and reparented under #1531. Create-response / delivery envelopes stay in #1305.

Body rewritten to match.

PR #1071, opened 2026-03-01
feat: AdCP v3.6 upgrade — schema migration, auth hardening, repository pattern, multi-tenant isolation

KonstantinMirin:
## Summary

- Upgrade to adcp library 3.6.0 with full schema migration (Product, Creative, Delivery, MediaBuy)
- Unify auth infrastructure across MCP/A2A/REST into single `resolve_identity()` path (from PR #1066)
- Introduce `MediaBuyRepository` + Unit of Work pattern with tenant-scoped queries
- Fix 16 cross-tenant query leaks discovered during repository migration
- Add 8 AST-scanning structural guards enforcing architecture invariants on every `make quality`
- Grow test suite from ~2,400 to 3,231 unit tests (3,218 passing, 10 skipped, 3 xfailed)

## Issues addressed

| Issue | Title | Status |
|-------|-------|--------|
| #1050 | Architecture Proposal: FastAPI as Unified Application Framework | Core infrastructure merged (via #1066) |
| #1066 | refactor: FastAPI migration — unify MCP + A2A + Admin into single process | Merged into this branch |
| #1067 | bug: new product server error | **Fixed** |
| #1041 | Bug: update_media_buy broken when manual approval enabled | **Fixed** |
| #997 | feat: Implement property_list filtering in get_products | **Implemented** |
| #1043 | chore: Upgrade adcp library to 3.5.0 | **Superseded** — upgraded directly to 3.6.0 |
| #1005 | AdCP v3 Gap Analysis - Tracking Issue | Partial progress (schema alignment, property_list, MediaChannel) |

### 1. AdCP v3.6 Schema Migration

All Pydantic models now extend `adcp` library 3.6.0 base types via inheritance (Critical Pattern #1). No field duplication — our models only declare local extensions and overrides.

- **Product**: `channels` aligned to `MediaChannel` enum, `property_targeting_allowed` added, `delivery_measurement` mapped
- **Creative**: `CreativeAsset` replaces legacy asset model, `CreativeGroup` schema added
- **GetProductsRequest**: `property_list` filtering, `buying_mode` enum, `catalog` support
- **UpdateMediaBuyRequest**: extends library `UpdateMediaBuyRequest1` (oneOf variant)
- **GetMediaBuyDeliveryRequest**: new fields from updated spec (`reporting_dimensions`, `account`)
- 30 new v3.6 Product field contract tests (`test_adcp_contract.py`)
- Schema alignment test validates all request models against live AdCP JSON schemas, with `KNOWN_SCHEMA_LIBRARY_MISMATCHES` allowlist for fields the library hasn't shipped yet (`fields`, `refine`, `idempotency_key`)

### 2. Auth Infrastructure Hardening (from #1066)

Three separate auth implementations (MCP middleware, A2A inline, REST decorator) replaced with single `resolve_identity()` called once at each transport boundary.

- `ResolvedIdentity` (immutable Pydantic model) replaces raw dicts and ContextVar reads in `_impl` functions
- `UnifiedAuthMiddleware` (pure ASGI) replaces 3 transport-specific middlewares
- MCP uses `FastMCP Middleware`, A2A uses `_resolve_a2a_identity()`, REST uses `FastAPI Depends`
- ContextVar dual-writes eliminated — tenant context set once at boundary, not scattered across layers
- `AdCPTestContext` extracted from headers at boundary, threaded through `ResolvedIdentity`
- `AUTH_CONTEXT_STATE_KEY` shared constant replaces 4 hardcoded string literals

#### MediaBuyRepository + Unit of Work

`MediaBuyRepository` encapsulates all `MediaBuy` and `MediaPackage` data access behind tenant-scoped methods. Every query includes `tenant_id` in the WHERE clause automatically — callers cannot bypass isolation.

```python
class MediaBuyRepository:
    def __init__(self, session: Session, tenant_id: str):
        self._session = session
        self._tenant_id = tenant_id

    def get_by_id(self, media_buy_id: str) -> MediaBuy | None:
        return self._session.scalars(
            select(MediaBuy).where(
                MediaBuy.tenant_id == self._tenant_id,
                MediaBuy.media_buy_id == media_buy_id,
            )
        ).first()
```

`MediaBuyUoW` wraps session lifecycle (begin/commit/rollback) and exposes the repository. `_impl` functions receive the UoW via dependency injection instead of calling `get_db_session()` inline.

Write methods (`create_from_request`, `update_status`, `update_fields`, `create_package`, `create_packages_bulk`) add objects to the session but never commit — the UoW handles commit/rollback at the boundary.

`create_from_request()` accepts the Pydantic request model directly and serializes `raw_request` internally — this eliminated 4 `model_dump()` calls from `_create_media_buy_impl`, keeping serialization at the repository boundary where it belongs.

#### Cross-tenant query leaks fixed

After composite PK migration made `principal_id` and `creative_id` no longer globally unique, 16 queries across 8 files were missing `tenant_id` filters:

| File | Queries Fixed |
|------|--------------|
| `auth_utils.py` | 1 — Principal lookup by principal_id |
| `media_buy_update.py` | 2 — MediaBuy queries in update flow |
| `gam/orders.py` | 1 — Creative query in `create_line_items` |
| `creatives.py` (admin) | 8 — Creative, CreativeReview, MediaBuy, Product, CreativeAssignment |
| `media_buy_delivery.py` | 1 — PricingOption query |
| `queries.py` | 3 — `get_creative_reviews`, `get_creative_with_latest_review` |

Each fix adds `tenant_id` to the WHERE clause. The `test_architecture_no_raw_media_package_select.py` guard prevents new `select(MediaPackage)` calls outside the repository (MediaPackage has no `tenant_id` column — isolation comes through the parent MediaBuy FK).

#### Structural guard: repository pattern

`test_architecture_repository_pattern.py` enforces:
- No `get_db_session()` calls in `_impl` functions (data access belongs in the repository layer)
- No inline `session.add()` in test bodies (use factories or fixtures)

### 4. Business Rule Implementations

- `buyer_ref` uniqueness validation in `create_media_buy` — rejects duplicate buyer refs within a tenant
- Empty update validation in `update_media_buy` — rejects requests with no actionable fields
- Creative format vs product `format_ids` validation — rejects creative formats not supported by the product
- `dry_run=True` now skips both adapter calls and DB writes (was only skipping adapter)
- `delete_missing=True` archive logic in `_sync_creatives_impl`

### 5. Structural Guards (8 total)

AST-scanning tests that run on every `make quality` and fail the build on new violations:

| Guard | Enforces | Allowlist |
|-------|----------|-----------|
| No ToolError in _impl | `_impl` raises `AdCPError`, never `ToolError` | 0 |
| Transport-agnostic _impl | `_impl` has zero imports from fastmcp/a2a/starlette/fastapi | 0 |
| ResolvedIdentity in _impl | `_impl` accepts `ResolvedIdentity`, not `Context` | 0 |
| Schema inheritance | Schemas extend adcp library base types | 27 intentional overrides |
| Boundary completeness | MCP/A2A wrappers forward all `_impl` parameters | 3 |
| Query type safety | DB queries use types matching column definitions | 1 |
| No model_dump in _impl | `_impl` returns model objects, never calls `.model_dump()` | 25 (down from 29) |
| Repository pattern | No `get_db_session()` in `_impl` | 0 |

Plus two domain-specific guards:
- **No raw `select(MediaPackage)`** outside repository — tenant isolation through FK join
- **Obligation coverage** — 733 uncovered test obligations tracked in allowlist (only shrinks)

### 6. Entity Test Suites

Three entity test suites with complete obligation mapping from AdCP spec scenarios:

| Entity | Tests | Confirmed | Unspecified | xfail |
|--------|-------|-----------|-------------|-------|
| MediaBuy | 130 | 78 | 52 | 5 |
| Creative | 147 | — | 83 | 3 |
| Delivery | 59 | 30 | 24 | 0 |
| **Total** | **336** | | | **8** |

Each test is mapped to a specific obligation ID from the AdCP spec BDD scenarios. `UNSPECIFIED` tests document implementation-defined behavior not mandated by the spec.

## Test stubs to develop with new protocol functionality

The following xfail markers and UNSPECIFIED stubs represent planned work that will become real tests as new AdCP features are implemented:

**xfail — blocked on implementation:**

| Stub | Entity | Blocked on |
|------|--------|-----------|
| `test_creative_group_schema_creation` | Creative | `create_creative_group` tool not implemented yet |
| `test_creative_group_listing` | Creative | `list_creative_groups` / `get_creative_groups` tool not yet |
| `test_adapt_creative_request_validation` | Creative | `adapt_creative` tool not implemented yet |
| `test_execute_approved_calls_adapter` | MediaBuy | Integration: adapter status update not flowing back |
| `test_creative_assignments_with_weights` | MediaBuy | Integration: weighted assignment not wired |
| `test_invalid_placement_ids_rejected` | MediaBuy | Integration: placement validation in adapter |
| `test_snapshot_populated_when_requested` | MediaBuy | Integration: delivery snapshot population |
| `test_creative_approvals_populated` | MediaBuy | Integration: creative approval field population |

**UNSPECIFIED stubs — implementation-defined behavior (167 total):**

These document behavior our code implements but the AdCP spec doesn't mandate. They serve as regression guards — if a future spec version defines this behavior, we update the stub to `CONFIRMED` and verify alignment:

- **MediaBuy (52)**: Budget helpers, spend cap enforcement, package-level budget aggregation, result wrapper patterns, approval workflow edge cases
- **Creative (83)**: Asset serialization, format validation, review workflow states, sync conflict resolution, approval state transitions
- **Delivery (24+7)**: Date range defaults, ID resolution strategy, status filter precedence, webhook trigger criteria, circuit breaker behavior, MCP transport details

**Schema-library mismatches (spec fields not yet in adcp library):**

| Schema | Missing Fields |
|--------|---------------|
| `get-products-request.json` | `fields`, `refine` |
| `update-media-buy-request.json` | `idempotency_key` |
| `get-media-buy-delivery-request.json` | `account`, `reporting_dimensions` |

## Test plan

- [x] `make quality` — 3,218 passed, 10 skipped, 3 xfailed, 0 failed
- [x] `./run_all_tests.sh` — all 5 suites (unit, integration, integration_v2, e2e, ui) green
- [x] All 8 structural guards passing
- [x] Schema alignment test validates against live AdCP spec
- [x] Obligation coverage guard reconciled (733 allowlist entries)
- [ ] Manual smoke test: create media buy via MCP and A2A transports
- [ ] Manual smoke test: admin UI product creation and media buy approval flow

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -508,14 +508,14 @@ wheels = [
 
 [[package]]
 name = "authlib"
-version = "1.6.6"
+version = "1.6.9"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "cryptography" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/bb/9b/b1661026ff24bc641b76b78c5222d614776b0c085bcfdac9bd15a1cb4b35/authlib-1.6.6.tar.gz", hash = "sha256:45770e8e056d0f283451d9996fbb59b70d45722b45d854d58f32878d0a40c38e", size = 164894, upload-time = "2025-12-12T08:01:41.464Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/af/98/00d3dd826d46959ad8e32af2dbb2398868fd9fd0683c26e56d0789bd0e68/authlib-1.6.9.tar.gz", hash = "sha256:d8f2421e7e5980cc1ddb4e32d3f5fa659cfaf60d8eaf3281ebed192e4ab74f04", size = 165134, upload-time = "2026-03-02T07:44:01.998Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/54/51/321e821856452f7386c4e9df866f196720b1ad0c5ea1623ea7399969ae3b/authlib-1.6.6-py2.py3-none-any.whl", hash = "sha256:7d9e9bc535c13974313a87f53e8430eb6ea3d1cf6ae4f6efcd793f2e949143fd", size = 244005, upload-time = "2025-12-12T08:01:40.209Z" },
+    { url = "https://files.pythonhosted.org/packages/53/23/b65f568ed0c22f1efacb744d2db1a33c8068f384b8c9b482b52ebdbc3ef6/authlib-1.6.9-py2.py3-none-any.whl", hash = "sha256:f08b4c14e08f0861dc18a32357b33fbcfd2ea86cfe3fe149484b4d764c4a0ac3", size = 244197, upload-time = "2026-03-02T07:44:00.307Z" },
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


bokelley:
## Note: `Error.recovery` field (RC1)

The AdCP RC1 spec adds a `recovery` field to the `Error` schema with three values:
- `transient` — retry after delay (rate limits, temporary unavailability)
- `correctable` — fix the request and retry (validation errors, invalid IDs)
- `terminal` — don't retry (permission denied, resource permanently gone)

This PR introduces `AdCPError` replacing `ToolError` in `_impl` functions, which is the right foundation. The next step is to have `AdCPError` populate the `recovery` hint so buyer agents can self-heal:

```python
raise AdCPError(
    code="BUDGET_EXHAUSTED",
    message="Media buy budget fully spent",
    recovery="terminal",
)
```

Standard error codes from RC1: `BUDGET_EXHAUSTED`, `CONFLICT`, `PRODUCT_NOT_FOUND`, `PRODUCT_UNAVAILABLE`, `RATE_LIMITED`, `PROPOSAL_EXPIRED`.

This could be a follow-up PR once this one merges — the `AdCPError` base class just needs the `recovery` field added.

KonstantinMirin:
Brilliant idea — this is exactly the kind of self-healing behavior buyer agents need.

Already tracked in #1072 (depends on this PR's `AdCPError` base class). The recovery hints + standard error codes will be a clean follow-up once this merges.

src/core/tools/media_buy_create.py
```diff
@@ -3605,30 +3726,33 @@ async def create_media_buy(
...
+            buyer_campaign_ref=buyer_campaign_ref,
+            ext=ext,
         )
     except ValidationError as e:
         raise AdCPValidationError(format_validation_error(e, context="request")) from e
 
     # Read identity and context_id pre-resolved by MCPAuthMiddleware
     identity = (await ctx.get_state("identity")) if isinstance(ctx, Context) else None
     _ctx_id = (await ctx.get_state("context_id")) if isinstance(ctx, Context) else None
+    # FIXME(salesagent-v0kb): boundary-completeness — push_notification_config not passed to _impl
```
ChrisHuie: Would address this now because MCP is not working the same way as A2A by not forwarding the `push_notification_config` right now .


The A2A wrapper (create_media_buy_raw) at line ~3823 forwards push_notification_config correctly.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1355,13 +1361,13 @@ async def _handle_explicit_skill(
...
 
-        Aligned with adcp v1.2.1 spec - brand_manifest must be a dict.
+        Aligned with adcp spec - brand must be a BrandReference dict.
 
         NOTE: Authentication is OPTIONAL for this endpoint. Access depends on tenant's
         brand_manifest_policy setting (public/require_brand/require_auth).
         """
         try:
-            # Normalize brand_manifest: URL string → dict (adcp v1.2.1)
+            # Normalize brand_manifest: URL string → dict (adcp v1.2.1 compat)
```
ChrisHuie: Interesting comment. Looks like the `brand_manifest` variable goes unused after
 normalization. 🤷‍♂️

src/a2a_server/adcp_a2a_server.py
ChrisHuie: The A2A explicit `get_products` skill handler doesn't forward `property_list`. Both `get_products()` and `get_products_raw()` in products.py accept and forward property_list, but this handler doesn't extract or pass `parameters.get("property_list")`

src/core/tools/products.py
```diff
@@ -159,23 +244,14 @@ async def _get_products_impl(
...
-        if isinstance(brand_manifest_unwrapped, BrandManifest):
-            if brand_manifest_unwrapped.name:
-                offering = brand_manifest_unwrapped.name
-            elif brand_manifest_unwrapped.url:
-                offering = f"Brand at {brand_manifest_unwrapped.url}"
-        else:
-            # AnyUrl — use the URL string
-            offering = f"Brand at {brand_manifest_unwrapped}"
+    if req.brand:
+        brand_dict: dict[str, Any] = req.brand if isinstance(req.brand, dict) else {}
```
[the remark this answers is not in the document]
KonstantinMirin: Correct — `isinstance(req.brand, dict)` is always `False` after Pydantic coercion. Filed as salesagent-7bzt. Will fix to use `req.brand.domain` directly and remove the stale comment.

KonstantinMirin:
Re: missing `get_current_tenant` import in `auth_utils.py:103` —

Good catch — but it turns out this function is actually dead code. Nobody imports `get_principal_object` from `auth_utils.py` — every caller uses the one in `auth.py` (which has the import and works correctly). That's why the missing import never caused a runtime error.

Removed the dead copy entirely in c2f21945 and updated the tenant-isolation guard to scan `auth.py` instead.

Your comment also surfaced a deeper issue: `get_principal_object` shouldn't be called inside `_impl` at all. The Principal should be resolved at the transport boundary and included in `ResolvedIdentity` (same lazy-loading pattern we use for tenant). Filed #1088 for that refactor.

PR #1087, opened 2026-03-05
feat: fix hardcoded URLs for release

ChrisHuie:
Upgrade authlib which had vulneraabilities in the prior release so was blocking with a new update to the package. Also, fixed some hardcoded urls that were messing up things :)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -508,14 +508,14 @@ wheels = [
 
 [[package]]
 name = "authlib"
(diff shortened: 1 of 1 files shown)


KonstantinMirin (approved):
Looks good to merge.

I also updated authlib in one of the PRs, I think it closes the pending ticket we had from a few weeks ago on vulnerability existing in dependent packages

PR #1103, opened 2026-03-10
fix: remove unauthenticated /init-api-key endpoint and harden control-plane auth

### LINKED ISSUE #1098: security: Unauthenticated control-plane first-claim takeover (/init-api-key) CRITICAL
F-01 Unauthenticated control-plane first-claim takeover (/init-api-key)

### Severity
Critical

### Status
Confirmed

### Description
The control-plane bootstrap endpoint `POST /admin/api/v1/tenant-management/init-api-key` can be called without authentication. The same module applies API-key enforcement to health and tenant-management operations, but the bootstrap route omits that gate and relies only on whether the key already exists. This creates a trust-boundary failure at first initialization: any unauthenticated caller who reaches an uninitialized deployment can claim the management key and become the effective control-plane owner.

Evidence:

request:
`curl -sS -i -X POST http://localhost:8000/admin/api/v1/tenant-management/init-api-key`
response/output:
`HTTP/1.1 201 CREATED` with JSON body containing a newly issued `api_key`.

Code:
- salesagent/src/admin/tenant_management_api.py:550–586
```
@tenant_management_api.route("/init-api-key", methods=["POST"])
def initialize_api_key():
    stmt = select(TenantManagementConfig).filter_by(config_key="tenant_management_api_key")
    existing_config = db_session.scalars(stmt).first()
    if existing_config:
        return jsonify({"error": "API key already initialized"}), 409

    api_key = f"sk-{secrets.token_urlsafe(32)}"
    ...
    return jsonify({"message": "Tenant management API key initialized", "api_key": api_key, ...}), 201
```

### Impact
Unauthenticated first caller can claim privileged control-plane credentials.

### Steps to reproduce
1) Issue an unauthenticated POST request to the bootstrap endpoint.
2) Confirm whether key material is returned.

```
curl -sS -i -X POST \
  http://localhost:8000/admin/api/v1/tenant-management/init-api-key
```

### Suggested fix
Require a bootstrap credential for initialization and enforce it at the route level, or move initialization to an out-of-band provisioning path (startup/CLI only) that is never publicly exposed. Keep the endpoint returning `401/403` for anonymous requests and add regression tests that assert unauthenticated initialization fails.

### Affected files
- `salesagent-main/src/app.py:271-275`
- `salesagent-main/src/admin/tenant_management_api.py:31-56`
- `salesagent-main/src/admin/tenant_management_api.py:550-586`

KonstantinMirin:
## Summary

Closes #1098. Removes the unauthenticated `POST /api/v1/tenant-management/init-api-key` endpoint that allowed first-caller takeover of the tenant management API key.

- **Remove `/init-api-key` endpoint** — eliminates the unauthenticated first-claim race condition
- **Add environment variable support** — `TENANT_MANAGEMENT_API_KEY` and `SYNC_API_KEY` env vars (falls back to database lookup for backward compatibility)
- **Fix timing-attack-vulnerable key comparison** — both `require_tenant_management_api_key` decorators now use `hmac.compare_digest()` instead of `!=`
- **Remove key values from log messages** — no longer logs partial key values on auth failures or key generation
- **Document in `.env.template`** — new section with generation instructions

### Key provisioning (replaces the removed endpoint)

Operators generate keys before deployment:
```bash
python -c "import secrets; print(f'sk-{secrets.token_urlsafe(32)}')"
```

Then set in `.env.secrets` or deployment secrets:
```
TENANT_MANAGEMENT_API_KEY=sk-your-key-here
SYNC_API_KEY=sk-your-key-here
```

### Backward compatibility

Both keys still check the database (`superadmin_config` table) as a fallback when env vars are not set. Existing deployments that initialized keys via the old endpoint or CLI script will continue to work — the env var simply takes priority.

## Test plan

- [x] `./run_all_tests.sh` — all 5 suites pass (unit, integration, integration_v2, e2e, ui)
- [x] Integration test verifies the `/init-api-key` route returns 404
- [x] Existing tenant management API tests pass with database-backed key auth

### DIFF
--- a/tests/integration/test_tenant_management_api_integration.py
+++ b/tests/integration/test_tenant_management_api_integration.py
@@ -392,16 +392,19 @@ class TestSyncApiAuth:
         return app
 
     def test_env_var_auth_succeeds(self, sync_app, monkeypatch):
-        """SYNC_API_KEY env var → auth succeeds."""
+        """SYNC_API_KEY env var → auth succeeds (hits real route)."""
         monkeypatch.setenv("SYNC_API_KEY", "sync-test-key")
 
         with sync_app.test_client() as client:
             resp = client.get(
-                "/api/v1/sync/status",
+                "/api/v1/sync/stats",
                 headers={"X-API-Key": "sync-test-key"},
             )
-            # 200 or 404 (route may not exist for GET) — but NOT 401/503
-            assert resp.status_code not in (401, 503)
+            # Auth passed — we get past the decorator. Route may fail on DB
+            # but must NOT return 401 (missing/invalid key) or 503 (unconfigured).
+            assert resp.status_code not in (401, 403, 503), (
+                f"Auth should have succeeded but got {resp.status_code}: {resp.json}"
+            )
 
     def test_missing_header_returns_401(self, sync_app, monkeypatch):
         """Request without X-API-Key header returns 401."""
--- a/.claude/skills/.agent-db.env
+++ b/.claude/skills/.agent-db.env
@@ -1,6 +1,6 @@
-export DATABASE_URL="postgresql://adcp_user:secure_password_change_me@localhost:50003/adcp_test"
+export DATABASE_URL="postgresql://adcp_user:secure_password_change_me@localhost:50002/adcp_test"
 export AGENT_PG_CONTAINER="agent-pg-skills"
-export AGENT_PG_PORT=50003
+export AGENT_PG_PORT=50002
 export ADCP_TESTING=true
 export ENCRYPTION_KEY="PEg0SNGQyvzi4Nft-ForSzK8AGXyhRtql1MgoUsfUHk="  # TEST ONLY — never use in production
 export GEMINI_API_KEY="test_key"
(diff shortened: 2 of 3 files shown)


ChrisHuie:
src/admin/sync_api.py:36-73 and src/admin/tenant_management_api.py:33-71

DRY opportunity. Consider extracting shared auth logic

tests/integration/test_tenant_management_api_integration.py
```diff
@@ -309,5 +303,135 @@ def test_soft_delete_tenant(self, client, mock_api_key_auth, test_tenant):
...
+        """SYNC_API_KEY env var → auth succeeds."""
+        monkeypatch.setenv("SYNC_API_KEY", "sync-test-key")
+
+        with sync_app.test_client() as client:
+            resp = client.get(
+                "/api/v1/sync/status",
+                headers={"X-API-Key": "sync-test-key"},
+            )
+            # 200 or 404 (route may not exist for GET) — but NOT 401/503
+            assert resp.status_code not in (401, 503)
```
ChrisHuie: This route `GET /api/v1/sync/status` doesn't exist , the registered route is `GET /api/v1/sync/status/<sync_id>`

PR #1117, opened 2026-03-12
fix: normalize admin UI to canonical /admin routes

### LINKED ISSUE #1105: Unable to login since 1.5.0 update
The SSO redirects back to `/auth/oidc/callback` to a 404

> The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.

Nothing else was changed in the update our side. Has there been a breaking change in this flow that we need to adjust for?

numarasSigmaSoftware:
## Background

This regression traces back to the FastAPI unification in commit `7d2b1d9e` (`refactor: FastAPI migration — unify MCP + A2A + Admin into single process (#1066)`).

That change moved the admin UI from a standalone Flask service behind proxy-managed root paths into a Flask app mounted inside the unified FastAPI process. After that migration, `/admin/...` became the only consistently correct external admin path, but parts of the UI and some helper services still generated legacy root-style admin URLs such as `/tenant/...`, `/login`, and `/auth/oidc/callback`.

This PR does not revert that architecture change. Instead, it aligns the app with the unified deployment model by standardizing all admin URL generation on canonical `/admin/...` paths and updating tests to validate that contract.

## Summary

Normalize the unified admin UI to a single canonical external path model.

This change makes `/admin/...` the only supported public admin URL space in the unified FastAPI deployment. It removes env-based prefix guessing, removes HTML response rewriting, updates admin link generation to use request-aware prefixes, and updates OIDC callback generation to `/admin/auth/oidc/callback`.

This PR does not add compatibility redirects. Legacy root-style admin entry URLs such as `/tenant/...`, `/login`, `/test/...`, and `/auth/...` are intentionally no longer treated as supported public admin paths in the unified app.

### Routing
- mount the Flask admin app only at `/admin` in `src/app.py`
- remove duplicate-prefix admin mounts for `/tenant`, `/auth`, `/login`, `/logout`, `/signup`, `/test`, and related root aliases
- update admin landing redirect to `/admin/login`

### Prefix handling
- stop deriving `script_name` from `PRODUCTION`
- derive template prefixing from `request.script_root` / `SCRIPT_NAME` in `src/admin/app.py`
- remove the after-request HTML rewriting hook that tried to inject `/admin`

### Canonical callback generation
- change OIDC redirect URI generation to `/admin/auth/oidc/callback` in `src/services/auth_config_service.py`

### Admin URL generation cleanup
Updated audited templates and backend URL producers so they emit canonical admin-aware URLs instead of unsupported root-style admin paths.

Key areas updated:
- dashboard and settings pages
- setup checklist pages
- policy pages
- product inventory and order/browser pages
- workflow/media-buy pages
- setup checklist service URLs
- business activity service URLs
- Slack notification URLs
- tenant bootstrap `admin_ui_url`
- tenant landing/admin entry URL generation

### Core routing / app behavior
- `src/app.py`
- `src/admin/app.py`
- `src/admin/blueprints/tenants.py`
- `src/services/auth_config_service.py`

### Backend URL generators
- `src/services/setup_checklist_service.py`
- `src/admin/services/business_activity_service.py`
- `src/services/slack_notifier.py`
- `src/landing/landing_page.py`
- `src/admin/tenant_management_api.py`

### Templates updated
- `templates/tenant_dashboard.html`
- `templates/tenant_settings.html`
- `templates/policy_settings.html`
- `templates/policy_settings_comprehensive.html`
- `templates/policy_rules.html`
- `templates/policy_review.html`
- `templates/setup_checklist.html`
- `templates/components/setup_checklist_widget.html`
- `templates/product_inventory_config.html`
- `templates/media_buy_approval.html`
- `templates/workflows.html`
- `templates/orders_browser.html`
- `templates/inventory_browser.html`
- `templates/inventory_unified.html`
- `templates/gam_line_item_viewer.html`
- `templates/components/inventory_picker.html`
- `templates/edit_product_mock.html`
- `templates/signup_complete.html`

### Unified-app / canonical path coverage
- `tests/unit/test_fastapi_app_regression.py`
- `tests/ui/test_comprehensive_pages.py`
- `tests/integration/test_tenant_settings_comprehensive.py`
- `tests/e2e/test_landing_pages.py`

### Direct Flask app coverage
- `src/admin/tests/integration/test_admin_app.py`

### URL producer / notification coverage
- `tests/integration/test_notification_urls_exist.py`
- `tests/integration/test_setup_checklist_service.py`
- `tests/unit/test_slack_notification_urls.py`

## Verification

Completed:
- `python3 -m py_compile` on all touched Python files

Not completed in this environment:
- `make quality`
- `pytest`
- `tox -e integration`

I could not run the full suites here because the environment does not have the required runtime/test dependencies installed.

## Breaking change

This PR intentionally changes the public admin URL contract for the unified deployment.

Supported public admin URLs:
- `/admin/...`

No longer supported as canonical public admin URLs:
- `/tenant/...`
- `/auth/...`
- `/login`
- `/logout`
- `/signup`
- `/test/...`

Existing bookmarks and IdP callback configurations using root-style admin URLs will need to be updated.

## Issues

Closes #1105

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -30,14 +30,13 @@ def _install_admin_mounts() -> None:
     remaining paths.
     """
 
+    from a2wsgi import WSGIMiddleware
     from starlette.routing import Mount
 
     filtered_routes = []
     for route in app.router.routes:
-        if isinstance(route, Mount) and route.app.__class__.__name__ == "WSGIMiddleware" and route.path in {
-            "/admin",
-            "",
-        }:
+        # Remove any prior compatibility mounts so we can re-add them at the end.
+        if isinstance(route, Mount) and isinstance(route.app, WSGIMiddleware) and route.path in {"/admin", ""}:
             continue
         filtered_routes.append(route)
 


KonstantinMirin:
Hi Nicolas, thanks for the thorough investigation and the detailed PR. You've correctly identified the root cause — the FastAPI migration in #1066 (my commit) broke the routing.

I wanted to share some context on why it broke and propose a simpler fix that avoids dropping the legacy URLs.

**How it worked before #1066 (nginx):**
```
location /auth {
    proxy_pass http://localhost:8001/auth;
}
```
Nginx forwarded the full path to Flask. A request to `/auth/oidc/callback` arrived at Flask as `PATH_INFO=/auth/oidc/callback` — matched the blueprint route, worked fine. Same for `/login`, `/tenant/...`, etc.

**Why it broke after #1066 (Starlette mount):**
```python
app.mount("/auth", admin_wsgi)
```
Starlette's `app.mount()` **strips the mount prefix** before forwarding to the WSGI app. So `/auth/oidc/callback` arrives at Flask as `PATH_INFO=/oidc/callback` — Flask has no route for that, 404. The same prefix-stripping issue affects every other root-level mount (`/login`, `/logout`, `/tenant`, `/api`, etc.).

**Proposed fix — 2 lines, no breaking changes:**
```python
app.mount("/admin", admin_wsgi)  # Canonical prefix
app.mount("/", admin_wsgi)       # Fallback — Flask gets the full path, just like nginx did
```
The `/` fallback mount only catches requests that don't match any FastAPI route or more specific mount. Flask receives the full path unchanged — exactly like the pre-#1066 nginx setup. This restores `/login`, `/tenant/...`, `/auth/oidc/callback`, and everything else without dropping support for any existing URLs.

No template changes, no service URL changes, no IdP reconfiguration needed. Existing bookmarks and callback configurations keep working.

I fully agree we need the regression tests you've added (the `TestAdminCanonicalMount` and `TestCanonicalOidcCallback` classes are great), but I don't think we need to drop the root-level URLs that were working before my migration broke them. The goal should be to restore the pre-#1066 behavior, not change the URL contract.

If you agree with this approach, it would be great if you could update the PR to use the fallback mount and remove the test that expects a 404 on `/login` (`test_root_login_path_is_not_exposed_by_unified_app`) — that URL should keep working.

numarasSigmaSoftware:
Hi Konstantin,

Considering a quick fix post release, I agree we could take a simpler route. However, for the next release I think we should consider a cleanup of those legacy routes. This will help the ongoing support (as the agents will have less things to worry and less test to run) and will create routes that are logical. There are ways we can do this and still support old endpoints for a while.

With the quick fix, we will be supporting two admin contracts. I'll proceed with the quick fix to get users back online for the moment, but let's think of this for the next release.

numarasSigmaSoftware:
Done

I added a small follow-up on top of the fallback-mount fix.

The root Flask fallback was catching a few FastAPI-owned paths too aggressively:
- unit tests that register temporary FastAPI routes after importing `src.app`
- `/a2a/`, which was falling through to Flask and returning 404

To fix that, I:
- moved the admin mount installation to startup so the fallback stays last in the route table
- added an explicit `/a2a/ -> /a2a` FastAPI redirect

That preserves the compatibility approach while keeping FastAPI-owned routes behaving correctly.

PR #1139, opened 2026-03-17
fix: require authenticated principal for task management tools

### LINKED ISSUE #1132: security: Currency Integrity Drift on Update (Implicit USD)
### Severity
High

### Status
Confirmed

### Description
The `update_media_buy` path accepts top-level `budget` as a float without requiring an explicit currency and then implicitly assigns `USD`. That inferred value is persisted together with the new budget into `media_buys`, which can silently overwrite an existing non-USD campaign currency during a budget-only update. The wrapper layer also still defaults omitted currency to `USD` when constructing the `Budget` object. The trust boundary failure: untyped external budget input is allowed to mutate the canonical financial currency field, not just the numeric amount.

Evidence:
request:
1) Create a temporary mock-backed tenant and seed a media buy with `currency='EUR'`
2) Call MCP `update_media_buy` with `budget: 500.0` only
3) Query `media_buys` for the updated row
4) Remove the temporary validation rows
response/output:
- MCP returned success
- database row showed updated budget and `currency='USD'` even though no currency change was requested

Code:
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1062-1098`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1288-1303`

```
if req.budget is not None:
    if isinstance(req.budget, int | float):
        total_budget = float(req.budget)
        budget_currency = "USD"
    ...
    uow.media_buys.update_fields(
        req.media_buy_id,
        budget=total_budget,
        currency=budget_currency,
    )
```

### Impact
- Silent currency mutation can flip campaign policy context (for example, EUR to USD) without operator intent.
- Currency-specific spend and limit checks can execute against the wrong `CurrencyLimit` bucket.
- Billing, reconciliation, and approvals can diverge from contractual currency intent.
- Most impactful scenario: a buyer updates only numeric budget, currency silently flips to USD, and subsequent package updates are validated under looser USD thresholds, enabling materially higher spend than intended before detection.

### Steps to reproduce
1) Prepare a media buy with non-USD currency in the database (for example `EUR`).
2) Initialize MCP and call `update_media_buy` with a float budget only (no currency field).
3) Re-check the row and observe currency drift.

### Suggested fix
Preserve existing `media_buys.currency` for numeric budget-only updates and reject implicit currency changes. Require explicit currency for cross-currency updates and validate allowed transitions before persistence. Add regression tests for float-only budget updates on non-USD records to assert currency immutability unless explicitly requested.

### Affected files
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1062-1098`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1288-1303`

numarasSigmaSoftware:
Closes #1132 (F-03)

**Summary**

list_tasks, get_task, and complete_task checked identity.tenant but not identity.principal_id. Because the localhost tenant fallback resolves a tenant even for requests with no auth token, the existing guard was silently bypassed — an unauthenticated caller could read task state and advance or complete workflow steps on behalf of any tenant.

**Changes**

src/core/tools/task_management.py

Added identity.is_authenticated check immediately after the existing tenant check in all three functions, preserving the existing error order: tenant resolution fails first, then authentication, then DB access.
Unauthenticated callers now receive AdCPAuthenticationError before any DB access occurs.
Scope is intentionally limited to these three task-mutating operations. The localhost tenant fallback for read-only tools such as get_products is unchanged.
tests/unit/test_task_management_auth.py (new)

6 rejection cases: unauthenticated callers (no principal_id, no identity) across all three tools.
3 regression cases confirming authenticated callers pass through normally.

**Note on architecture compliance**

While reviewing the auth fix I noticed a pre-existing violation worth flagging: task_management.py imports Context from fastmcp.server.context at module level and accepts it as a parameter in all three functions. This is the pattern the transport-boundary guards are designed to prevent — business logic functions should accept ResolvedIdentity, not Context.

The structural guard in test_transport_agnostic_impl.py does not catch this because it only scans functions whose names end in _impl. These three functions are named list_tasks, get_task, and complete_task, so the AST sweep never inspects them.

This is not introduced by this PR. I'm not fixing it here to keep the scope of the auth change narrow, but I think it warrants a follow-up issue so it doesn't get lost. The fix would be:

Rename to _list_tasks_impl, _get_task_impl, _complete_task_impl
Remove the context: Context | None parameter from each
Move identity resolution from context into the MCP/A2A wrapper layer, consistent with every other tool in src/core/tools/
Happy to open that issue separately if useful.

### DIFF
--- a/tests/integration/test_creative_agent_live.py
+++ b/tests/integration/test_creative_agent_live.py
@@ -13,6 +13,7 @@ enable_live_mode fixture so they call the real creative agent.
 
 import pytest
 
+from src.core import creative_agent_registry as creative_agent_registry_module
 from src.core.creative_agent_registry import CreativeAgent, CreativeAgentRegistry
 
 # The live creative agent URL
@@ -31,10 +32,21 @@ def enable_live_mode(monkeypatch):
     monkeypatch.setenv("ADCP_TESTING", "false")
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def registry():
-    """Fresh registry instance for each test."""
-    return CreativeAgentRegistry()
+    """Shared registry instance so live tests reuse the format cache.
+
+    A per-test registry defeats caching and can hammer the public creative
+    agent hard enough to trigger rate limiting in CI.
+    """
+    previous_registry = creative_agent_registry_module._registry
+    shared_registry = CreativeAgentRegistry()
+    shared_registry._format_cache.clear()
+    creative_agent_registry_module._registry = shared_registry
+    try:
+        yield shared_registry
+    finally:
+        creative_agent_registry_module._registry = previous_registry
 
 
 class TestCreativeAgentLiveConnection:


KonstantinMirin:
I think this a great catch regarding the architecture compliance - I suggest adding them ti the structural guard and allowlist and file a follow-up refactoring issue.

Issue #1133, opened 2026-03-16
security: Minimum Spend Policy Bypass via Update Path


sigma-antonnash:
### Severity
High

### Status
Confirmed

### Description
Create flow enforces minimum spend constraints per currency and pricing option, but update flow still does not re-apply equivalent minimum-spend validation before accepting budget reductions. In update, package budget changes are checked for structural validity and max-daily limits, then forwarded to the adapter even if the new value falls below configured minimums. The trust boundary failure: policy enforcement exists at admission time only and is missing on mutable state transitions.

Evidence:
request:
1) Create a temporary mock-backed tenant whose currency policy sets `min_package_budget=100.00`
2) Create or seed a package at a compliant value
3) Call MCP `update_media_buy` with a package budget of `1.0`
4) Query `media_packages` after the update
5) Remove the temporary validation rows
response/output:
- MCP returned success instead of a minimum-spend validation error
- `media_packages.package_config` was rewritten with the below-minimum budget
- `media_packages.budget` remained at the previous compliant value on the validation adapter path

Code:
- `salesagent-main-latest/src/core/tools/media_buy_create.py:1713-1797`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:369-400`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:488-514`

```
# create path
if currency_limit.min_package_budget:
    ...
    if package_min_spend and package_budget < package_min_spend:
        raise ValueError(error_msg)

# update path
if currency_limit.max_daily_package_spend and req.packages:
    ...  # only max daily validation

result = adapter.update_media_buy(
    action="update_package_budget",
    budget=int(budget_amount),
    ...
)

```
### Impact
- Buyers can pass create-time minimum thresholds, then lower budgets below contractual floors via update.
- Premium inventory commitments can be undercut post-approval, causing direct revenue leakage.
- Forecasting and operational allocation can remain based on compliant initial budgets while live spend commitment drops.
- Most impactful scenario: bulk premium campaigns are approved at minimum-compliant budgets, then rapidly reduced to near-zero through updates, creating large-scale under-commitment and delayed reconciliation disputes.

### Steps to reproduce
1) Create a media buy/package at or above configured minimum spend.
2) Issue `update_media_buy` package budget updates below the minimum threshold.
3) Confirm update succeeds and persisted/package budget reflects below-minimum value.

### Suggested fix
[...]

PR #1140, opened 2026-03-17
fix: enforce update budget guardrails and preserve currency in media …

### LINKED ISSUE #1130: security: Financial guardrail bypass via extreme budget updates
### Severity
High

### Status
Confirmed

### Description
The top-level budget update path accepts numeric budget values and enforces only that the amount is positive. The code then writes the value directly into `media_buys.budget` with no upper-bound control at this layer. The nearby currency-limit logic is still tied to package-level daily checks and does not cap extreme top-level campaign budgets. On the current local deployment, the default tenant is wired for `google_ad_manager` but does not have working GAM credentials, so live validation used a disposable mock-backed tenant in the same deployment. The trust boundary failure is unchanged: externally supplied budget values cross into persistent financial state without a max guardrail.

Evidence:
request:
1) Create a temporary mock-backed tenant and media buy for validation
2) Call MCP `update_media_buy` with an extreme top-level budget value (for example `999999999`)
3) Query `media_buys` for the updated row
4) Remove the temporary validation rows
response/output:
- MCP returned success
- database row persisted the extreme budget value with no upper-bound rejection
EVIDENCE_BLOCK_END

Code:
- `salesagent-main-latest/src/core/tools/media_buy_update.py:369-400`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1062-1098`

I```
f req.budget is not None:
    if isinstance(req.budget, int | float):
        total_budget = float(req.budget)
        budget_currency = "USD"
    ...
    if total_budget <= 0:
        return UpdateMediaBuyError(...)
    ...
    uow.media_buys.update_fields(
        req.media_buy_id,
        budget=total_budget,
        currency=budget_currency,
    )
```

### Impact
- Extreme budget values can be committed without top-level ceiling checks.
- Financial reporting and budget-control integrity can be materially distorted.

### Steps to reproduce
1) Authenticate for a media buy update path that executes immediately.
2) Submit a very large top-level budget and verify persisted value.

### Suggested fix
Introduce explicit top-level budget ceilings in the update path and enforce them before persistence, using the same policy source used for financial limits to keep create/update parity. Add tests for over-limit rejection and in-range acceptance.

### Affected files
- `salesagent-main-latest/src/core/tools/media_buy_update.py:369-400`
- `salesagent-main-latest/src/core/tools/media_buy_update.py:1062-1098`

numarasSigmaSoftware:
### Summary
Addresses #1130, #1132, and #1133.

This PR tightens financial validation in update_media_buy by:

rejecting campaign budget updates above a hard ceiling preserving the existing media buy currency for float-only budget updates instead of silently defaulting to USD enforcing minimum package spend on package budget updates, matching create-time validation These changes make update-time behavior safer and consistent with existing financial policy enforcement.

### Changes

src/core/tools/media_buy_update.py

Added MAX_CAMPAIGN_BUDGET as a configurable guardrail, defaulting to 10,000,000 via MAX_CAMPAIGN_BUDGET_USD. Campaign-level budget updates now return budget_ceiling_exceeded when the requested budget exceeds that ceiling. Float-only campaign budget updates now preserve the existing media buy currency from the database instead of defaulting to USD. Float-only package budget updates now preserve the existing media buy currency from the database instead of defaulting to USD. Package-level budget updates now enforce CurrencyLimit.min_package_budget and return budget_below_minimum when the requested budget is too low. tests/unit/test_budget_guardrails.py

Added focused unit coverage for:
budget ceiling enforcement
env override for MAX_CAMPAIGN_BUDGET_USD
currency preservation for float budgets
minimum package spend enforcement on updates
tests/unit/test_media_buy.py

Updated mocks to reflect the additional CurrencyLimit lookup introduced by the new minimum-spend validation path. tests/unit/test_update_media_buy_behavioral.py

Updated behavioral tests to reflect the new validation flow. Made currency-limit mock setup less brittle where update flows now perform additional reads. Added min_package_budget to the test currency-limit fixture so update-path validation matches production behavior. tests/unit/test_architecture_no_model_dump_in_impl.py

Updated allowlist line numbers after the new constant block shifted media_buy_update.py. Why
Before this change:

campaign budget updates had no explicit upper bound float-only budget updates could silently default to USD instead of preserving the existing media buy currency package budget updates did not enforce min_package_budget, even though create-time validation already did This PR closes those gaps and aligns update-time validation with the financial rules already expected elsewhere in the system.

### Testing

UV_CACHE_DIR=/tmp/uv-cache uv run pytest \
  tests/unit/test_update_media_buy_behavioral.py \
  tests/unit/test_media_buy.py \
  tests/unit/test_architecture_no_model_dump_in_impl.py \
  tests/unit/test_budget_guardrails.py \
  -v --tb=short
Result

203 passed
If you want the strictest GitHub wording, you can also change the first line to:

Closes #1130, #1132, and #1133.


KonstantinMirin (changes requested):
Nicholas, thanks for the rework — the `CurrencyLimitRepository` and shared validators are a genuine improvement over the first iteration, and we appreciate the effort.

We've documented the architecture patterns in [PR #1145](https://github.com/prebid/salesagent/pull/1145) (`docs/development/patterns-reference.md`). The codebase has a lot of legacy code that doesn't follow these patterns, so it's easy to pattern-match against the wrong thing — the doc maps each pattern to its canonical file and identifies what's tracked debt vs. target architecture. We'd welcome your input on that document too — if anything is unclear or you think something is missing, please comment on that PR.

Items to address before we can merge:

**1. Restore the FIXME on line 141.** The diff removes `# FIXME(salesagent-9f2): raw session usages below should migrate to repository methods` but the deprecated `session = uow.session` call and 16 raw session calls are still there. FIXMEs are contracts that debt is known and tracked — they can only be removed when the underlying code is fixed. See [Section 3](https://github.com/prebid/salesagent/blob/docs/architecture-patterns-reference/docs/development/patterns-reference.md#3-structural-guards-and-allowlists).

**2. Tests must use the test harness.** `test_budget_guardrails.py` rebuilds a 15-line MagicMock scaffold in each test function. The project has a test harness at `tests/harness/` specifically to avoid this — see [Section 4](https://github.com/prebid/salesagent/blob/docs/architecture-patterns-reference/docs/development/patterns-reference.md#4-writing-tests--the-test-harness). Read `tests/harness/_base.py` (BaseTestEnv) and `tests/harness/delivery_poll_unit.py` (DeliveryPollEnv) for the pattern. For `_update_media_buy_impl`, either create a `MediaBuyUpdateEnv` in `tests/harness/` or at minimum reuse the `standard_mocks` fixture from `tests/unit/test_update_media_buy_behavioral.py`. We don't want to multiply the scaffolding debt that already exists in older test files.

**3. Error message `.replace()` chains are fragile.** The create path calls the shared validator and then does `.replace("The same minimum applies to updates as to creation.", "for products in this package")`. If the shared message text changes, these replacements silently become no-ops. Add a context parameter to the validator instead.

PR #1141, opened 2026-03-17
fix: harden login redirect validation and test-auth gate (F-06, F-02)

### LINKED ISSUE #1127: security: Test auth/default credential privileged session issuance
### Severity
Critical

### Description
The UI flow allows disabling tenant Setup Mode and then displays a Production Mode message that says only SSO authentication is allowed and test credentials are disabled. The same UI flow’s “Need Help” guidance instructs operators to start the system, log in with test credentials (`test_super_admin@example.com` / `test123`), configure SSO, and then disable Setup Mode so only SSO remains. Runtime validation showed a mismatch in that model: even after Setup Mode is disabled and Production Mode is shown, test login can still succeed when global `ADCP_AUTH_TEST_MODE=true` is enabled at deployment level.

On the current local deployment, `default.auth_setup_mode=false` in the database, but the running container still has `ADCP_AUTH_TEST_MODE=true`. The effective runtime path is `/admin/test/auth`; the legacy `/test/auth` path is not reachable on this deployment. The vulnerability exists because authentication control is split across two scopes with different semantics. Tenant-level `auth_setup_mode` drives the Users & Access “Production Mode” UI state, while global `ADCP_AUTH_TEST_MODE` still enables the effective `/admin/test/auth` endpoint regardless of that tenant setting. The trust boundary fails at operator intent versus effective runtime policy, because the UI primarily reflects tenant setup state and not the global override, which can mislead operators into believing test authentication is fully disabled.

Evidence:
request:
1) Confirm `default.auth_setup_mode=false` and `ADCP_AUTH_TEST_MODE=true`
2) `POST /test/auth` with `email=test_super_admin@example.com`, `password=test123`, `tenant_id=default`
3) `POST /admin/test/auth` with `email=test_super_admin@example.com`, `password=test123`, `tenant_id=default`
4) `GET /admin/tenant/default` using returned cookie
response/output:
1) `/admin/test/auth` returned `HTTP/1.1 302 Found` with `Set-Cookie: session=...`
2) follow-up `GET /admin/tenant/default` returned `HTTP/1.1 200 OK`

Code:
- `salesagent-main-latest/src/app.py:271-275`
- `salesagent-main-latest/src/admin/blueprints/auth.py:805-878`

```
@auth_bp.route("/test/auth", methods=["POST"])
def test_auth():
    ...
    env_test_mode = os.environ.get("ADCP_AUTH_TEST_MODE", "").lower() == "true"
    tenant_setup_mode = False
    ...
    if not env_test_mode and not tenant_setup_mode:
        abort(404)

    test_users = {
        os.environ.get("TEST_SUPER_ADMIN_EMAIL", "test_super_admin@example.com"): {
            "password": os.environ.get("TEST_SUPER_ADMIN_PASSWORD", "test123"),
            "role": "super_admin",
        },
        ...
    }

    if is_super_admin(email) and password == "test123":
        session["is_super_admin"] = True
        session["authenticated"] = True
        ...
```

### Impact
Reachable test-auth flow enables immediate privileged session issuance via known credentials.

### Steps to reproduce
1) Disable tenant Setup Mode for the target tenant.
2) Leave `ADCP_AUTH_TEST_MODE=true` in the deployment.
3) Submit default test credentials to `/admin/test/auth`.
4) Reuse the session cookie to request a protected admin page.

### Suggested fix
Disable test authentication in production mode with a hard startup guard, remove static default passwords, and require one-time bootstrap credentials with forced rotation in development workflows. Add a regression test that confirms the effective mounted endpoint (`/admin/test/auth`) is unreachable in the production profile.

### Affected files
- `salesagent-main-latest/src/app.py:271-275`
- `salesagent-main-latest/src/admin/blueprints/auth.py:805-917`

### LINKED ISSUE #1131: security: Open Redirect via Login `next` Parameter
### Severity
Medium

### Status
Confirmed

### Description
The login flow stores a user-supplied `next` parameter into session state and later redirects to that value after authentication without validating destination safety. In the unified deployment, the effective live exploit chain uses mounted admin aliases: `GET /admin/login?next=...` to prime the session and `POST /admin/test/auth` to trigger the post-auth redirect. Because absolute external URLs, scheme-relative URLs, and encoded variants are accepted and replayed through `redirect(next_url)`, an attacker can prime a victim session and force a post-auth redirect to an attacker-controlled domain. The trust boundary fails where untrusted navigation input is reused as a privileged redirect sink.

Evidence:

request:
1) `GET /admin/login?next=https://evil.example.com/phish`
2) `POST /admin/test/auth` with the same cookie
3) Repeat with `next=//evil.example.com/phish` and an encoded absolute URL

response/output:
- authentication response returned `HTTP/1.1 302 Found` with `Location: https://evil.example.com/phish`
- scheme-relative and encoded absolute URL variants were also reflected into `Location`
- root `/login` did not reproduce the same admin redirect flow on this deployment

Code:
- `salesagent-main-latest/src/app.py:271-275`
- `salesagent-main-latest/src/admin/blueprints/auth.py:223-226`
- `salesagent-main-latest/src/admin/blueprints/auth.py:596-599`
- `salesagent-main-latest/src/admin/blueprints/auth.py:697-699`
- `salesagent-main-latest/src/admin/blueprints/auth.py:759-761`
- `salesagent-main-latest/src/admin/blueprints/auth.py:875-883`
- `salesagent-main-latest/src/admin/blueprints/auth.py:905-913`

```
next_url = request.args.get("next")
if next_url:
    session["login_next_url"] = next_url
```

### Impact
- Enables phishing-style redirect chains after successful authentication.
- Increases risk of session abuse through trusted-domain to attacker-domain transitions.

### Steps to reproduce
1) Prime the session using an external `next` URL.
2) Authenticate with that session and inspect the redirect target.

### Suggested fix
Validate redirect targets at both ingestion and sink: permit only relative internal paths or strict same-origin allowlist destinations, and fall back to a safe internal page when validation fails. Add tests for absolute URLs, scheme-relative URLs, and encoded bypass variants.

### Affected files
- `salesagent-main-latest/src/app.py:271-275`
- `salesagent-main-latest/src/admin/blueprints/auth.py:223-226`
- `salesagent-main-latest/src/admin/blueprints/auth.py:596-599`
- `salesagent-main-latest/src/admin/blueprints/auth.py:697-699`
- `salesagent-main-latest/src/admin/blueprints/auth.py:759-761`
- `salesagent-main-latest/src/admin/blueprints/auth.py:875-883`
- `salesagent-main-latest/src/admin/blueprints/auth.py:905-913`

numarasSigmaSoftware:
Closes #1127 Closes #1131

Summary
Two security findings from the audit addressed in this PR:

F-06 (Medium): Open redirect via login next parameter — user-supplied redirect targets were stored and replayed without validation, allowing a post-authentication redirect to an attacker-controlled domain. F-02 (Critical): Test auth bypass — ADCP_AUTH_TEST_MODE=true alone was sufficient to enable /test/auth regardless of a tenant's auth_setup_mode setting, allowing privileged session issuance after an operator had explicitly disabled Setup Mode via the Admin UI. F-06 — Open redirect fix
Added _safe_redirect() to src/admin/blueprints/auth.py:

def _safe_redirect(url: str | None, fallback: str) -> str:
    """Return url only if it is a safe local path, otherwise fallback."""
    decoded = unquote(url).strip()
    if parts.scheme or parts.netloc or decoded.startswith(("//", "\\\\")) ...:
        logger.warning("[SECURITY] Rejected unsafe redirect URL: %r", url)
        return fallback
    return decoded
Applied at both the ingestion point and every redirect sink:

/login and /tenant/<id>/login now call _safe_redirect() before storing login_next_url in session — absolute URLs, scheme-relative URLs, and encoded bypass variants are all rejected at entry. All four post-authentication redirect sinks (super-admin callback, single-tenant callback, multi-tenant tenant selector callback, and test_auth) now call _safe_redirect() again before the final redirect() call — defence in depth at the sink. F-02 — Test auth gate fix
Gate change in test_auth():

# Before — vulnerable: env var alone was sufficient if not env_test_mode and not tenant_setup_mode:
    abort(404)
# After — both required
if not env_test_mode or not tenant_setup_mode:
    abort(404)
A tenant operator disabling Setup Mode via the Admin UI now immediately blocks test auth for that tenant, regardless of the deployment-level ADCP_AUTH_TEST_MODE flag. The env var is now a necessary but not sufficient condition.

Hardcoded super-admin bypass removed:

# Before — bypassed test_users table entirely
if is_super_admin(email) and password == "test123":
    ...
# After — all credentials go through test_users, no special cases if email in test_users and test_users[email]["password"] == password:
    ...
Production hard-block added via is_admin_production() — PRODUCTION=true or ENVIRONMENT=production causes an immediate 404 before any credential check, regardless of all other flags.

Files changed
File	Change
src/admin/blueprints/auth.py	_safe_redirect(), validated ingestion + sinks, F-02 gate, removed bypass src/admin/utils/helpers.py	Added is_admin_production() src/admin/utils/__init__.py	Exported is_admin_production tests/unit/test_auth_setup_mode.py	Updated TestTestAuthEndpointLogic for AND gate tests/unit/test_test_auth_production_guard.py	Full gate coverage + TestSuperAdminCredentialPath Test results
25 new/updated tests passing. Pre-existing webhook DNS failures in test_webhook_delivery.py are unrelated (sandbox DNS restriction).


numarasSigmaSoftware:
Both points addressed — tests now exercise actual endpoints.

1. Gate tests converted to Flask test client

TestTestAuthEndpointLogic in test_auth_setup_mode.py and TestProductionHardBlock in test_test_auth_production_guard.py are replaced with Flask test client tests that POST to the actual /test/auth endpoint and assert the HTTP status code. Each test creates a real Flask app, mocks only get_db_session to return a tenant with the appropriate auth_setup_mode, and sets the relevant env vars. If someone changes the gate logic in auth.py these tests will now catch it.

2. Open redirect regression tests

Added TestOpenRedirectRejection in test_test_auth_production_guard.py with three tests:

test_external_next_url_not_stored_at_login — GET /login?next=https://evil.example.com confirms _safe_redirect() at ingestion rejects the external URL so login_next_url is never stored in session
test_external_next_url_does_not_redirect_to_attacker_domain — reproduces the original attack: prime session with evil next → POST /test/auth with valid credentials → assert Location header does not contain evil.example.com
test_session_injected_next_url_rejected_at_auth_sink — defense in depth: directly injects login_next_url = "https://evil.example.com" into the session and confirms the /test/auth sink still rejects it via its own _safe_redirect() call independently of the ingestion check

numarasSigmaSoftware:
@KonstantinMirin thanks for the review! Test is now fixed and issue #1149 is open so we don't lose track of it

PR #1167, opened 2026-03-28
fix: raise on anomalous empty format responses instead of silent return []

### LINKED ISSUE #1136: fix: skip format validation when creative agent returns no formats
## Summary

When creating or editing a product with display or video formats, the admin UI rejects all format IDs with:
> Invalid format IDs: video_standard, video_vast, video_standard, video_vast, ... These formats do not exist in the creative agent.
> Invalid format IDs: display_image, display_html, display_js, display_image, display_html, display_js, ... These formats do not exist in the creative agent.

## Root cause

The product form validates submitted format IDs against the creative agent registry (`registry.list_all_formats()`). When the creative agent at `https://creative.adcontextprotocol.org` is unreachable **but does not raise `ADCPConnectionError` or `ADCPTimeoutError`** — e.g., it returns an empty result, times out internally, or the SDK catches the error — the registry returns 0 formats. The validation builds an empty `valid_format_ids` set, so **every** submitted format ID is flagged as invalid.

The existing graceful degradation only catches `ADCPConnectionError` and `ADCPTimeoutError`. An empty response from the registry (0 formats, no exception) falls through to the validation path, which then rejects everything.

**Affected code:** `src/admin/blueprints/products.py` — both `add_product()` (~line 730) and `edit_product()` (~line 1400).

## Fix

- When `valid_format_ids` is empty (no formats returned from any agent), skip validation and save formats without checking — same graceful degradation as the connection error path. Hard validation still happens downstream at media buy creation time.
- Extracted `_parse_format_entries()` helper to DRY up the 4 identical "parse format entries without validation" blocks.

KonstantinMirin:
## Summary

Fixes #1136 — but **not** with the approach suggested in the issue.

The issue suggested changing the condition in `products.py` from `if result.errors and not result.formats` to `if not result.formats` (skip validation whenever no formats are returned). This would regress the intentional design from PR #1137 (commit c10cfed3), which **deliberately distinguishes three cases**:

| `formats` | `errors` | Meaning | Behavior |
|---|---|---|---|
| non-empty | any | Agent has data | Validate against them |
| empty | non-empty | Agent down (errors explain why) | Graceful degradation — save without validation |
| empty | empty | Agent up, genuinely 0 formats | Reject (format IDs can't be valid) |

The real bug is upstream in `creative_agent_registry.py`: five code paths silently return `[]` on anomalous/failed responses instead of raising. This makes failures **indistinguishable from "agent up, no formats"** — `list_all_formats_with_errors` records no error, and `products.py` correctly rejects all format IDs based on the (misleading) data it receives.

### The fix

Convert 5 silent `return []` to raises in `creative_agent_registry.py`, and propagate proper AdCPError subclasses throughout the format lookup chain:

| Location | Was | Now |
|---|---|---|
| `_fetch_formats_from_agent`: `completed` + `data=None` | `return []` | `raise AdCPAdapterError` |
| `_fetch_formats_from_agent`: `submitted` status | `return []` | `raise AdCPAdapterError` |
| `_fetch_formats_from_agent`: unexpected status | `return []` | `raise AdCPAdapterError` |
| `_fetch_formats_raw_mcp`: no `result` key in response | `return []` | `raise AdCPAdapterError` |
| `_parse_mcp_tool_result`: no text content items | `return []` | `raise AdCPAdapterError` |
| `format_resolver.get_format`: format not found | `raise ValueError` | `raise AdCPNotFoundError` |
| `signals_agent_registry`: 3 anomalous response paths | `raise ValueError` | `raise AdCPAdapterError` |
| `signals_agent_registry`: 4 transport error re-raises | `raise RuntimeError` | `raise AdCPAdapterError` |

**Error semantics:**
- `AdCPAdapterError` (502, transient) — agent misbehaved, returned garbage, or transport failed
- `AdCPNotFoundError` (404, terminal) — format genuinely not found after searching all sources

**Catch sites updated to handle specific AdCPError subclasses:**
- `get_format_by_id` (`_base.py`): catches `(ValueError, AdCPNotFoundError)` → returns `None`. Lets `AdCPAdapterError` propagate (broken agent ≠ "not found").
- GAM `orders.py`: catches `(ValueError, AdCPNotFoundError, AdCPAdapterError)` → wraps in contextual error message for line item creation failures.

These raises are caught by `list_all_formats_with_errors`'s existing `except Exception` handler (line 664), which records them as errors. The existing `result.errors and not result.formats` condition in `products.py` then correctly triggers graceful degradation.

**Cache poisoning is also fixed**: previously, silent `[]` returns were cached for 1 hour (line 658), meaning a single transient failure would cause all format validation to reject for an hour. Now, raises propagate before reaching the cache write.

The **only** path that still returns `[]` is `completed` with valid `data` containing `formats: []` — a legitimate response from a custom creative agent with no formats configured.

### Also included

- **`cryptography` 46.0.5 → 46.0.6**: Fixes CVE-2026-34073 (name constraints not applied to wildcard DNS SANs during verification)
- **`.duplication-baseline`** tests count restored from 84 → 111: PR #1143's merge accidentally reverted PR #1146's bump due to merge conflict resolution picking the wrong side

## Test plan

- [x] 6 new regression tests in `test_silent_empty_format_bug.py` (all failed before fix, all pass after)
- [x] 5 new regression tests in `test_signals_agent_silent_empty_bug.py`
- [x] 4 existing tests updated in `test_creative_agent_fallback.py` (now assert `AdCPAdapterError` instead of `[]`)
- [x] Existing tests updated in `test_creative_agent_registry.py`, `test_signals_agent_registry.py`, `test_format_resolver.py`, `test_creative_agent_live.py`
- [x] Full suite (test-results/300326_1646/): unit 4116, integration 1773, e2e 82, admin 4, bdd 822 — **all 0 failures** (6797 total)

### DIFF
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -1,0 +1,318 @@
+# Test Architecture
+
+This file is the authoritative guide to writing tests in this project.
+**Agents must read this before writing any test code.**
+
+## The Harness System (Use This)
+
+The test harness (`tests/harness/`) is the central testing abstraction. It manages mocks,
+identity, database sessions, and multi-transport dispatch. **All new tests must use it.**
+
+### How it works
+
+```python
+from tests.harness import DeliveryPollEnv
+
+with DeliveryPollEnv(tenant_id="t1", principal_id="p1") as env:
+    # env auto-patches external dependencies, creates identity, binds DB session to factories
+    tenant = TenantFactory(tenant_id="t1")
+    principal = PrincipalFactory(tenant=tenant, principal_id="p1")
+    buy = MediaBuyFactory(tenant=tenant, principal=principal)
+
+    env.set_adapter_response(buy.media_buy_id, impressions=5000)
+    result = env.call_impl(media_buy_ids=[buy.media_buy_id])
+
+    assert result.deliveries[0].impressions == 5000
+```
+
+### Environment hierarchy
+
+| Class | Mode | Domain | File |
+|-------|------|--------|------|
+| `BaseTestEnv` | Unit (mocked DB) | Base class | `tests/harness/_base.py` |
+| `IntegrationEnv` | Integration (real DB) | Base class | `tests/harness/_base.py` |
+| `DeliveryPollEnv` | Integration | Delivery metrics | `tests/harness/delivery_poll.py` |
+| `DeliveryPollEnvUnit` | Unit | Delivery metrics | `tests/harness/delivery_poll_unit.py` |
+| `WebhookEnv` | Integration | Webhook delivery | `tests/harness/delivery_webhook.py` |
+| `CircuitBreakerEnv` | Integration | Circuit breaker | `tests/harness/delivery_circuit_breaker.py` |
+| `CreativeSyncEnv` | Integration | Creative sync | `tests/harness/creative_sync.py` |
+| `CreativeFormatsEnv` | Integration | Format discovery | `tests/harness/creative_formats.py` |
+| `CreativeListEnv` | Integration | Creative listing | `tests/harness/creative_list.py` |
+| `ProductEnv` | Integration | Product catalog | `tests/harness/product.py` |
+| `ProductEnvUnit` | Unit | Product catalog | `tests/harness/product_unit.py` |
+| `MediaBuyUpdateEnv` | Unit | Media buy updates | `tests/harness/media_buy_update.py` |
+
+### Key capabilities
+
+- **`EXTERNAL_PATCHES`**: Dict of `{name: patch_target}` — auto-started as `unittest.mock.patch` on `__enter__`
+- **`ASYNC_PATCHES`**: Set of names that need `AsyncMock` instead of `MagicMock`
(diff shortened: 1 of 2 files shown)


ChrisHuie:
A couple gaps now for `RuntimeError` escaping


Gap site 1 — schemas/_base.py:740-743 (get_format_by_id):                                                                                                           
  try:                                                                                                                                                              
      return get_format(format_id, tenant_id=tenant_id)                                                                                                               
  except ValueError:                                                                                                                                                
      return None  # RuntimeError escapes here                                                                                                                      
                                                                                                                                                                      
  Gap site 2 — gam/managers/orders.py:504 (GAM order creation):
  try:                                                                                                                                                                
      format_obj = get_format(format_id_str, agent_url=agent_url, ...)                                                                                              
  except ValueError as e:                                                                                                                                             
      error_msg = f"Format lookup failed for '{format_display}': {e}"                                                                                                 
      raise ValueError(error_msg)  # RuntimeError escapes here

Issue #1169, opened 2026-03-28
refactor: migrate remaining raw select(AdapterConfig) to AdapterConfigRepository


KonstantinMirin:
## Summary

After the AdapterConfigRepository is introduced in the GAM service account fix PR, there will still be ~20 raw `select(AdapterConfig)` calls in the codebase that should be migrated to use the repository.

Related to #1158 (same pattern for `select(Tenant)` in auth.py).

### Admin blueprints (13 raw selects)
- `src/admin/blueprints/gam.py` — 5 calls (OAuth storage, connection test, SA upload, network verify)
- `src/admin/blueprints/adapters.py` — 2 calls (save/test adapter config)
- `src/admin/blueprints/auth.py` — 1 call (OAuth callback get-or-create)
- `src/admin/blueprints/inventory.py` — 1 call (custom targeting key sync)
- `src/admin/sync_api.py` — 3 calls (sync trigger, status, order discovery)
- `src/admin/tenant_management_api.py` — 2 calls (read + update)

### GCP service account service (3 raw selects)
- `src/services/gcp_service_account_service.py` — 3 calls (read SA, check existence, delete SA)

### Core business logic (1 raw select)
- `src/core/tools/media_buy_create.py:1713` — reads AXE keys + custom_targeting_keys

### GAM adapter internals (7 raw selects)
- `src/adapters/google_ad_manager.py:610` — custom_targeting_keys
- `src/adapters/gam/managers/targeting.py` — 3 calls (AXE keys + custom_targeting_keys)
- `src/adapters/gam/managers/orders.py:399` — naming templates
- `src/adapters/gam/managers/workflow.py:160` — naming templates
- `src/adapters/gam_reporting_api.py` — 2 calls

### Relationship-based access (tenant.adapter_config)
- Not raw selects but also scattered — `src/admin/utils/helpers.py`, `settings.py`, `publisher_partners.py`, `principals.py`, `tenants.py`, `core.py`, `tenant_status.py`, `setup_checklist_service.py`

## Approach

Each file group can be a separate PR. The structural guard (`test_architecture_repository_pattern.py`) tracks the allowlist — entries should shrink with each migration.

## Acceptance criteria

- [ ] All `select(AdapterConfig)` outside `AdapterConfigRepository` removed
- [ ] Repository allowlist entries for AdapterConfig removed
- [ ] `make quality` passes

Issue #1158, opened 2026-03-26
refactor: migrate raw select(Tenant) in auth blueprint to TenantRepository


numarasSigmaSoftware:
`src/admin/blueprints/auth.py` contains multiple raw `select(Tenant)` calls directly in the blueprint (lines ~193, ~212, ~245, ~293, ~767, ~879, ~1038). This violates the repository pattern enforced by `test_architecture_repository_pattern.py` and tracked in the pre-existing allowlist.

As the allowlist baseline ratchets down, these violations will need to be resolved.

**Action:** Create a `TenantRepository` (or extend an existing one) with lookup methods (`get_by_subdomain`, `get_by_virtual_host`, `get_by_id`, `get_default`). Replace all raw `select(Tenant)` calls in `auth.py` with repository calls. Remove the entries from the allowlist.

PR #1171, opened 2026-03-29
refactor: AdapterConfigRepository + GAM service account auth consolidation

### LINKED ISSUE #1163: fix: get_adapter() drops service account credentials for GAM adapter
## Summary

`get_adapter()` in `src/core/helpers/adapter_helpers.py` does not include `service_account_json` in the config dict when building the GAM adapter configuration. This means tenants using service account authentication can sync inventory but cannot execute media buys.

## Root Cause

`get_adapter()` (line 110-119) builds the GAM config inline and only sets `refresh_token`:

```python
elif adapter_type == "google_ad_manager":
    adapter_config["network_code"] = config_row.gam_network_code or ""
    adapter_config["refresh_token"] = config_row.gam_refresh_token or ""  # ← empty string for SA tenants
    adapter_config["trafficker_id"] = config_row.gam_trafficker_id or ""
```

`service_account_json` is never added to the dict. When `GoogleAdManager.__init__` (line 139) checks auth credentials, `refresh_token` is `""` (falsy), `service_account_json` is `None` (missing from dict), and `service_account_key_file` is `None` — so all three are falsy and it raises the error.

Meanwhile, `build_gam_config_from_adapter()` in `src/adapters/gam/__init__.py` correctly handles both auth methods — but `get_adapter()` doesn't use it. The inventory sync path (`background_sync_service.py:193`) also correctly handles service account auth with its own inline logic.

## Impact

- Tenants using `gam_auth_method == "service_account"` cannot create or execute media buys
- Inventory sync works fine (different code path)
- Tenants using OAuth (`refresh_token`) are unaffected

## Reproduction

1. Configure a tenant with GAM service account auth (`gam_auth_method = "service_account"`, `gam_service_account_json` populated)
2. Sync inventory — succeeds
3. Create and execute a media buy — fails with:
   > GAM config requires either 'service_account_key_file', 'service_account_json', or 'refresh_token'

KonstantinMirin:
## Summary

Closes #1163. References #1158, #1169.

Introduces `AdapterConfigRepository` to centralize all `AdapterConfig` database access. Fixes service account authentication broken in 7+ locations, eliminates adapter-to-DB circular dependencies, and shrinks the no-raw-select structural guard allowlist by 6 entries.

### What was broken
`get_adapter()` only set `refresh_token` in the GAM config dict, ignoring `gam_auth_method` and `gam_service_account_json`. Service-account tenants could sync inventory but could NOT create/update/list media buys.

### What changed
- **New `AdapterConfigRepository`** with `has_gam_credentials()` (single source of truth), `get_gam_config()`, `get_gam_targeting_config()`, `get_gam_naming_templates()`
- **Migrated 12 raw `select(AdapterConfig)` calls** to repository (get_adapter, 4 validation gates, 3 GAM services, 3 background services)
- **Eliminated adapter→DB circular dependency** — `GoogleAdManager.__init__` pre-loads config once, passes to managers via constructor/method params
- **Structural guard allowlist shrank by 6 entries**

## Test plan
- [x] `./run_all_tests.sh` — 6,805 tests pass (unit:4105 integration:1792 e2e:82 admin:4 bdd:822)
- [x] 21 new tests (14 repo integration + 5 auth harness + 2 config injection)
- [x] Allowlist shrank, not grew


src/adapters/google_ad_manager.py
```diff
@@ -134,6 +134,23 @@ def __init__(
...
+            from src.core.database.database_session import get_db_session
+            from src.core.database.repositories.adapter_config import AdapterConfigRepository
+
+            with get_db_session() as session:
+                repo = AdapterConfigRepository(session, self.tenant_id)
+                self._targeting_config = repo.get_gam_targeting_config()
+                self._order_name_template, self._line_item_name_template = repo.get_gam_naming_templates()
+        except Exception as e:
+            logger.warning(f"Could not pre-load adapter config for tenant {self.tenant_id}: {e}")
+
```
ChrisHuie: Don't think we want this because it could hijack reporting of other more specific errors like `ProgrammingError` or `TypeError` 

This problem is also here => `google_ad_manager.py:622-634` and `workflow.py:156-167` for Gemini key loading

templates/inventory_unified.html
```diff
@@ -994,95 +957,6 @@ <h5 class="mb-0">
         });
 }
```
ChrisHuie: this partially replaced the custom `renderAdUnitTree()` which had `nameSpan.onclick = selectNode(unit.id, e)`. Doesn't provide click-to-select though.

`inventory_browser.html` solved this with `attachNodeClickHandlers()` post-render

PR #1173, opened 2026-03-29
fix: prevent stale signup_flow from bypassing tenant access lookup

### LINKED ISSUE #1057: fix: Stale signup_flow session flag bypasses tenant access lookup
Users with a stale `signup_flow` cookie are redirected to signup even when they have access to existing tenants, causing duplicate tenant creation.

Two bugs:
1. `signup_flow` bypasses tenant lookup - In `google_callback()`, the code checks `signup_flow` BEFORE calling `get_user_tenant_access()`. If the flag is set, tenant lookup never runs.
2. Session not saved on callback - The callback doesn't call `save_session()` on redirects, so clearing `signup_flow` after signup doesn't persist. The stale flag remains in the browser indefinitely.

From the logs:
```
[SESSION_DEBUG] NO Set-Cookie on /auth/google/callback (session.modified=True, keys=['signup_flow', ...])
```

Fix:

In src/admin/blueprints/auth.py (google_callback):
1. Call `get_user_tenant_access()` BEFORE checking `signup_flow`
2. Only redirect to signup if `signup_flow=True` AND `total_access == 0`
3. Clear `signup_flow` if user has existing tenant access
4. Add explicit `save_session()` calls on all redirect responses

This was introduced by #993 - the fix to preserve `signup_flow` across session.clear() now preserves STALE flags from previous visits, and the callback doesn't properly clear them because save_session() isn't called on redirects.

ChrisHuie:
fix for https://github.com/prebid/salesagent/issues/1057

Move get_user_tenant_access() before the signup_flow check in google_callback() so returning users with a stale session flag are not incorrectly redirected to onboarding when they already have tenant access. The flag is now cleared with an info log when the user has existing tenants, and only redirects to onboarding when total_access is genuinely zero.


KonstantinMirin (commented):
## Code Review (5-agent automated review)

**Verdict: APPROVE** — fix is correct, well-tested, architecturally sound. 3 non-blocking suggestions below.

### The fix

Moving `get_user_tenant_access()` before the `signup_flow` check is the right approach — authorization check runs before session-flag routing, and the stale flag becomes a hint instead of an override. 

### Suggestion: use existing `ui_client` fixture instead of manual setup

The test builds its own Flask app with `_create_test_app_with_mocked_oauth()` and manages DB cleanup manually with `get_db_session()` + `session.delete()`. There's already a `ui_client` fixture in `tests/admin/conftest.py` that handles app creation, test mode, and cleanup:

```python
# tests/admin/conftest.py
@pytest.fixture
def ui_client(ui_test_mode):
    """Provide Flask client configured for UI testing."""
    ...
    client = app.test_client()
    yield client
```

And `integration_db` (already imported there) handles DB lifecycle with transaction rollback — no manual cleanup needed. Using these would:
1. Eliminate the manual `get_db_session()` + `session.delete()` cleanup blocks
2. Eliminate the manual `ALL_FACTORIES` session binding loop
3. Follow the established pattern in `tests/admin/test_comprehensive_pages.py`

The `UserFactory` addition is clean and follows all conventions.

### Minor notes

- The mock target `"src.admin.blueprints.auth.get_user_tenant_access"` works because the import is inline in `google_callback()`. Worth a comment so future refactors don't accidentally break it by hoisting to module-level.
- The `ALL_FACTORIES` session binding in the test mutates class-level state — safe today (single-threaded) but would break under parallel execution. Another reason to use `integration_db` which handles this via `IntegrationEnv`.

KonstantinMirin (commented):
Great work on this, Chris! The fix is clean and the root cause analysis is spot-on — the ordering issue with `signup_flow` bypassing tenant access was a subtle bug.

A few suggestions that might help:

### Existing test infrastructure you could leverage

There's a `ui_client` fixture in `tests/admin/conftest.py` (from #1146) that handles Flask app creation, test mode auth, and cleanup:

```python
@pytest.fixture
def ui_client(ui_test_mode):
    """Provide Flask client configured for UI testing."""
    ...
```

And `integration_db` handles the DB lifecycle — factory session binding and cleanup via transaction rollback, so you wouldn't need the manual `get_db_session()` + `session.delete()` try/finally blocks.

There's also `AdminAccountEnv` in `tests/harness/admin_accounts.py` (from #1170) which wraps Flask test_client for admin scenarios with automatic tenant/session management. Might be worth looking at for patterns.

Not a blocker at all — the manual approach works and the tests are solid. Just pointing these out in case they save some boilerplate.

### Minor notes

- The mock target `"src.admin.blueprints.auth.get_user_tenant_access"` relies on the import being inline in `google_callback()` — maybe worth a small comment so future refactors don't accidentally break the mock by hoisting the import to module level.
- `UserFactory` is clean and follows all the established patterns. Nice addition.

ChrisHuie:
Feedback addressed

jrosendahl-opt:
@ChrisHuie Please resolve merge conflicts

KonstantinMirin (commented):
## Review — PR #1173

**Overview** — The core fix is right: moving `get_user_tenant_access()` ahead of the `signup_flow` check makes the session flag a hint instead of an override, and the bug-fix branch is genuinely pinned by a real-DB integration test (`test_stale_signup_flow_with_existing_tenants_bypasses_onboarding` creates a real tenant+user via factories and does not mock the lookup — a regression would fail it). Two duplication defects introduced by the diff need resolving before merge, one of which the maintainer already raised.

### Should fix

**Duplicated `factory_session` fixture — ↩ raised in review of 2026-04-03, still open.** The new fixture in `tests/integration/test_self_service_signup.py` (the `ALL_FACTORIES` bind/unbind loop) re-implements logic that already exists two ways: the reusable helper `_bind_factories_to_session()` in `tests/utils/database_helpers.py` (importable from anywhere), and an identically-named `factory_session` fixture in `tests/admin/conftest.py` (same assert-None precondition, same `SASession(bind=get_engine())`, same `finally` unbind). The 2026-04-03 review asked to drop this hand-rolled binding ("Eliminate the manual `ALL_FACTORIES` session binding loop"); the thread was marked "Feedback addressed" (2026-04-13) but the current push still carries the loop. A bug in the bind/teardown protocol now has to be fixed in every copy. Rebuild the fixture on `_bind_factories_to_session()`, or migrate the two tests to `integration_db`/`ui_client` per that suggestion and delete the fixture.

**Signup-flag clear copied to a third site.** The pair
```python
session.pop("signup_flow", None)
session.pop("signup_step", None)
```
now lives at three places: the super-admin branch (`auth.py:552-553`, 24 lines above in the *same* function), the new stale-flag branch (`auth.py:576-577`, this diff), and post-provisioning (`public.py:285-286`). Same logical operation, no parameter variation — add a fourth signup key later and one of the three sites gets missed. Extract `clear_signup_session(session)` (shared, since `public.py` needs it too) and route the new site plus the siblings through it.

### Nice to have

- `tests/factories/core.py` docstring now lists `UserFactory`, but that class is defined in `tests/factories/user.py`, not `core.py`. The new test imports it from the package, so the docstring line serves no purpose and points readers at the wrong module — drop `, UserFactory` from the `core.py` header.
- The two OAuth-callback tests resolve `get_user_tenant_access` opposite ways (one mocks it → `total_access=0`, the other uses real factory rows). Both defensible, but a one-line comment on the stale-flag test explaining why it deliberately hits the real lookup would make the pair read as a unit.
- No BDD scenario covers the signup/onboarding routing. Coverage here is real and this is an Admin-UI OAuth view with no AdCP transport surface, so the four-transport bar does not apply — but a hand-authored `BR-ADMIN-*` scenario for "returning user with stale signup_flow lands on the tenant selector, not onboarding" would be more durable grounding if you want it.

- Resolved since the last round: the mock target moved to the source module (`patch("src.admin.domain_access.get_user_tenant_access")`), which is robust to the inline import inside `google_callback()` — this addresses the earlier "mock relies on inline import" note without needing a comment. The onboarding-path test's `assert_called_once_with("newuser@example.com")` also catches the exact ordering regression this PR fixes.
- Not re-raising: the class-level `ALL_FACTORIES` mutation's parallel-safety worry — the fixture guards nesting with an assert and xdist isolates per process; it folds into the fixture-duplication fix above anyway. The merge-conflict request from @jrosendahl-opt appears satisfied (head is a merge of origin/main).

src/admin/blueprints/auth.py
```diff
@@ -561,19 +561,26 @@ def google_callback():
...
-        # Get all accessible tenants
         tenant_access = get_user_tenant_access(email)
 
+        # Only redirect to onboarding if user genuinely has no tenants
+        if session.get("signup_flow"):
+            if tenant_access["total_access"] == 0:
+                flash(f"Welcome {user.get('name', email)}!", "success")
+                return redirect(url_for("public.signup_onboarding"))
+            # Stale flag — user already has tenant access, clear and continue
+            session.pop("signup_flow", None)
```
KonstantinMirin, line 576: This `session.pop("signup_flow", None)` + `session.pop("signup_step", None)` pair is now the third verbatim copy of the signup-flag clear — one sits 24 lines up in this same function (super-admin branch), and a third is in `public.py`. Add a fourth signup key later and one site gets missed. Extract a shared `clear_signup_session(session)` helper and route this site plus both siblings through it.

tests/integration/test_self_service_signup.py
```diff
@@ -448,3 +511,32 @@ def client():
     app = create_app({"TESTING": True, "SECRET_KEY": "test_key"})
     with app.test_client() as test_client:
         yield test_client
+
+
+@pytest.fixture
+def factory_session(integration_db):
```
KonstantinMirin, line 517: This `factory_session` fixture re-inlines the `ALL_FACTORIES` bind/unbind loop that already exists as the importable helper `_bind_factories_to_session()` in `tests/utils/database_helpers.py` and as an identically-named fixture in `tests/admin/conftest.py`. This is the boilerplate the maintainer asked to drop on 2026-04-03. Build the fixture on `_bind_factories_to_session()`, or migrate the two tests to `integration_db`/`ui_client` and delete it.

tests/factories/core.py
```diff
@@ -1,6 +1,6 @@
 """Factory_boy factories for core tenant-related models.
 
-Factories: TenantFactory, CurrencyLimitFactory, PropertyTagFactory, PublisherPartnerFactory
+Factories: TenantFactory, CurrencyLimitFactory, PropertyTagFactory, PublisherPartnerFactory, UserFactory
```
KonstantinMirin, line 3: `UserFactory` is defined in `tests/factories/user.py`, not this module — the new test imports it from the package, so listing it here is inaccurate and serves no purpose. Drop `, UserFactory` from the header.

PR #1175, opened 2026-03-30
feat: universal request normalization for AdCP backward compatibility

### LINKED ISSUE #1151: create_media_buy rejects brand_manifest due to strict FastMCP tool schema
## Bug

When a buyer agent sends `brand_manifest` (v2.5 field name) to `create_media_buy`, the call fails with:

```
Failed to submit media buy to publisher: 1 validation error for call[create_media_buy]
brand_manifest.    Unexpected keyword argument [type=unexpected_keyword_argument, input_value='https://rspb.co.uk', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/unexpected_keyword_argument
```

## Root Cause

The `create_media_buy` function in `src/core/tools/media_buy_create.py` is registered as a FastMCP tool via:

```python
mcp.tool()(with_error_logging(create_media_buy))
```

FastMCP introspects the function signature and auto-generates a **strict** Pydantic model from the typed parameters. Since `brand_manifest` is not in the function signature (only `brand` is), FastMCP rejects it as an unexpected keyword argument — **before the salesagent code ever runs**.

## Spec Compliance

The ADCP v3 `create-media-buy-request.json` schema explicitly sets `"additionalProperties": true` at the root level:
https://adcontextprotocol.org/schemas/v3/media-buy/create-media-buy-request.json

This means unknown fields MUST be accepted per the spec. The current behavior is non-compliant.

Additionally, the v2.5 schema uses `brand_manifest` while v3 uses `brand`. Buyer agents sending v2.5-style requests will always hit this error.

## Suggested Fix

Either:

1. **Add `**kwargs` or an explicit `brand_manifest` param** to the `create_media_buy` function signature to accept additional fields
2. **Normalize `brand_manifest` → `brand`** in the function body for backward compatibility with v2.5 callers
3. **Configure FastMCP** to allow additional properties on the generated tool schema (if FastMCP supports this)

Option 2 is probably the most robust since it handles both the spec compliance issue and the v2/v3 field name migration.

## Impact

Any buyer agent sending `brand_manifest` (the v2.5 field name) cannot create media buys against this salesagent. Confirmed failing with campaign `campaign_c3923c43-415b-4628-a7d1-87c982a026fd`.

KonstantinMirin:
## Summary

Adds a universal request normalization layer and real end-to-end transport testing infrastructure. What started as fixing `brand_manifest` rejection (#1151) evolved into a comprehensive overhaul of how the seller agent handles schema evolution, validates input, and tests across transports.

Closes #1151, #1180. Supersedes #1161.

## The Problem

Three problems, one root cause — the AdCP spec evolves faster than implementations can track:

1. **MCP hard-rejected deprecated fields** (`brand_manifest` → `ValidationError`) while A2A and REST silently dropped them. Publishers blocked.
2. **FastMCP's TypeAdapter validates before our Pydantic models** — a second validation gate we don't control, creating inconsistent behavior across transports.
3. **Test infrastructure bypassed real transport paths** — `call_mcp()` called tool functions directly with a mock Context, never exercising middleware, TypeAdapter, or auth. Bugs were invisible.

### 1. Universal Request Normalization

Single `normalize_request_params()` function shared by all three transports. Translates deprecated AdCP fields:

| Deprecated | Current | Scope | Translation |
|---|---|---|---|
| `brand_manifest` | `brand` | `get_products`, `create_media_buy` | Extract domain from URL → `{domain: hostname}` |
| `campaign_ref` | `buyer_campaign_ref` | `create_media_buy` only | Rename |
| `account_id` | `account` | All tools | Wrap: `"x"` → `{account_id: "x"}` |
| `optimization_goal` | `optimization_goals` | Package-level | Scalar → array |
| `catalog` | `catalogs` | Package-level | Scalar → array |
| `promoted_offerings` | `catalogs` | All tools | v2.5 rename |

Integrated via:
- **MCP**: `RequestCompatMiddleware` (FastMCP `on_call_tool` — official API, no monkey-patching)
- **A2A**: Called in `_handle_explicit_skill()` dispatcher
- **REST**: `RestCompatMiddleware` (Starlette middleware normalizes JSON body)

### 2. Schema-Aware Deep-Strip Middleware (Production Forward-Compat)

The middleware has a three-stage pipeline, **environment-gated**:

| Stage | Dev Mode | Production Mode |
|---|---|---|
| 1. Translate deprecated fields | ✅ Always | ✅ Always |
| 2. Strip unknown top-level fields | ❌ Fail loudly (schema drift detection) | ✅ Strip with WARNING |
| 3. Deep-strip + retry on TypeAdapter rejection | ❌ Fail loudly | ✅ Recursive strip → retry |

**Stage 3: Deep-strip** is the key innovation. When TypeAdapter rejects arguments (e.g., buyer sends a future field inside a nested object with `additionalProperties: false`), the middleware:

1. Gets the tool's full JSON Schema via `tool.parameters`
2. Recursively walks the value alongside the schema tree
3. At each object level with `additionalProperties: false`, strips unknown properties
4. Handles `anyOf`/`oneOf` unions by scoring variants on declared-property matches
5. Handles `allOf` by merging properties from all members
6. Retries with cleaned arguments — TypeAdapter accepts, Pydantic models validate

This is **schema-independent from AdCP** — strips against what our tool actually accepts, not what the external spec says. If stripping didn't change anything (e.g., type mismatch, not unknown fields), the original error propagates — no infinite retry loop.

**Architecture principle**: TypeAdapter is the gate we bypass. Everything after it (Pydantic model validation, `_impl` business logic) is intentional validation whose errors must reach the buyer. The middleware never swallows business logic errors.

### 3. Real End-to-End Transport Testing

**Before**: `call_mcp()` called tool functions directly with a mock Context. No middleware, no TypeAdapter, no auth chain. Bugs invisible.

**After**: Every transport exercises the real pipeline:

| Transport | Path | Auth |
|---|---|---|
| **MCP** | `Client(mcp)` in-memory → middleware chain → TypeAdapter → tool fn | Real: `get_http_headers` → token → DB lookup |
| **A2A** | `AdCPRequestHandler.on_message_send()` → skill routing → normalization → `model_validate` → handler | Real: `_resolve_a2a_identity` (single mock point) |
| **REST** | Starlette `TestClient` → middleware → FastAPI route → Pydantic body | Real: dependency injection auth |
| **impl** | Direct `_impl()` call | Identity object |

### 4. Forward-Compatibility Test Suite (100+ tests)

Comprehensive test coverage for the deep-strip mechanism, derived using systematic three-pass methodology:

**`test_deep_strip.py` (56 tests)**:
- P1-P8: Postcondition scenarios for each stripping behavior (top-level, nested, arrays, unions, refs, open schemas)
- INV-1: Known fields never removed
- INV-2: Stripping is idempotent
- INV-3: Data preservation — known field VALUES are bit-identical after strip (strings, unicode, ints, floats, bools, None, empty strings)
- Adversarial: mixed `additionalProperties` in `anyOf`, `allOf` handling, `oneOf` discriminated unions, 10-level nesting

**`test_forward_compat_acceptance.py` (36 tests)**:
- MCP pipeline: 9 payload shapes accepted in production (current spec, future fields, v2.5 legacy, multi-level extras)
- Dev mode: 3 tests verify unknown fields rejected loudly
- Deep-strip retry E2E: nested extras → TypeAdapter rejects → strip → retry → succeeds
- Data preservation E2E: intercept `_impl` to verify tool receives buyer's data intact after strip
- Error propagation: 5 adversarial tests — business logic errors, AdCPError, RuntimeError all propagate (never retried)
- Middleware adversarial: schema lookup failure → graceful degradation, concurrent calls → no interference, retry error propagates (not stale original)

**`test_mcp_compat_middleware.py` (10 tests)**: `_should_retry` distinguishes TypeAdapter from business logic errors

**`test_mcp_client_dispatch.py` (3 tests)**: Client(mcp) pipeline baseline

### 5. Review Fixes

- **`get_media_buys` account_id bug**: A2A handler bypassed `model_validate()`, using `parameters.get("account_id")` which returned None after normalization. Fixed: handler uses `model_validate()`, `GetMediaBuysRequest` gains `account: AccountReference | None` field. MCP/raw wrappers updated to `account: dict | None`.
- **Dead `except PermissionError`**: Removed from A2A skill dispatch — `AdCPAuthorizationError` (subclass of `AdCPError`) already caught above.
- **`tests/harness/` in tox**: Added to unit env so harness meta-tests and dispatch tests run in CI.
- **Defensive `json.dumps`**: `error.details` serialization wrapped in try/except to prevent crashes on non-serializable values.
- **ToolError branch dropped**: `_should_retry` only catches `pydantic.ValidationError` with `title.startswith("call[")`. TypeAdapter never raises ToolError.
- **Redundant schema alignment test deleted**: Tested spec against library (circular — library is generated from spec). Broke whenever spec site updated ahead of library release.

### 6. Transport-Agnostic Error Assertions

Added `assert_rejected(result, field=..., reason=...)` helper that checks observable behavior: "was this request rejected because field X had problem Y?" Works across all transports without knowing which layer caught the error.

### 7. Additional Fixes

- **BDD empty-step guard**: Extended to catch empty `@given`/`@when` steps (previously only `@then`)
- **CreativeAssetFactory**: Pydantic schema factory for valid `CreativeAsset` objects
- **ToolError details preservation**: Error `details` serialized as JSON 4th arg in ToolError, parsed back by unwrapper
- **MCP tool signature fixes**: `sync_accounts`, `list_authorized_properties` migrated from `req:` to individual typed params

## Known Limitations

- **5 BDD tests xfailed** (#1184): `supported_billing` and `account_approval_mode` are stored on identity by test harness but don't exist in DB. Real MCP auth resolves identity from DB → policies missing → billing/approval not enforced. Needs DB migration.
- **Nested forward-compat is MCP-only**: adcp library types (`BrandReference`, `AccountReference`) use `extra='forbid'` hardcoded. Deep-strip handles this on MCP. A2A/REST rely on production `extra='ignore'` for top-level fields, but nested library types always reject unknowns. Upstream fix: adcp library could use `extra='ignore'` on these types.

## Test Results

```
unit:        4222 passed, 0 failed
harness:      109 passed, 0 failed (deep-strip, forward-compat, dispatch)
admin:          10 passed, 0 failed
e2e:            93 passed, 0 failed
integration: 1848 passed, 0 failed
bdd:         1068 passed, 0 failed, 5 xfailed (#1184)
```

## Follow-up Issues

- #1181: Empty BDD Given step bodies from #1170
- #1182: MCP tool signature gaps (remaining tools)
- #1183: Agent-facing tool descriptions (0/150 params have descriptions)
- #1184: Move billing/approval policy from identity to tenant DB config

### DIFF
--- a/tests/harness/test_forward_compat_acceptance.py
+++ b/tests/harness/test_forward_compat_acceptance.py
@@ -442,6 +442,174 @@ class TestDeepStripRetryE2E:
 
 
 # ---------------------------------------------------------------------------
+# Data preservation: known values survive deep-strip + retry unchanged
+# ---------------------------------------------------------------------------
+
+
+class TestDataPreservationE2E:
+    """Verify that after deep-strip retry, the tool function receives exactly
+    the buyer's data for all known fields — no truncation, coercion, or loss.
+
+    The test intercepts the _impl call to capture what arguments the tool
+    function actually receives after the middleware pipeline.
+    """
+
+    def test_brand_domain_preserved_after_strip(self):
+        """Buyer sends brand with extra field. After strip + retry, _impl
+        receives brand.domain exactly as sent.
+        """
+        from fastmcp import Client
+
+        from src.core.main import mcp
+
+        captured_req = {}
+
+        async def _call():
+            patches = _get_products_patches()
+
+            # Intercept _get_products_impl to capture the request
+            original_impl = None
+
+            async def capturing_impl(req, identity=None):
+                captured_req["brand"] = req.brand
+                captured_req["brief"] = req.brief
+                # Call original to produce a valid response
+                return await original_impl(req, identity)
+
+            with patch.dict(os.environ, {"ENVIRONMENT": "production"}):
+                for p in patches:
+                    p.start()
+                # Capture the impl
+                import src.core.tools.products as products_mod
+
+                original_impl = products_mod._get_products_impl
+                with patch.object(products_mod, "_get_products_impl", side_effect=capturing_impl):
+                    try:
+                        async with Client(mcp) as client:
+                            result = await client.call_tool(
+                                "get_products",
+                                {
+                                    "brief": "holiday campaign — Q4",
+                                    "brand": {
+                                        "domain": "my-brand.com",
+                                        "future_field": "should be stripped",
+                                    },
+                                },
+                                raise_on_error=False,
+                            )
+                            assert not result.is_error, f"Should succeed: {result.content}"
+                    finally:
+                        for p in patches:
+                            p.stop()
+
+            # Verify the _impl received exactly what the buyer sent for known fields
+            assert captured_req["brief"] == "holiday campaign — Q4"
+            assert captured_req["brand"] is not None
+            assert captured_req["brand"].domain == "my-brand.com"
+
+        asyncio.run(_call())
+
+    def test_context_session_id_preserved_after_strip(self):
+        """Buyer sends context with extra field. After strip, session_id preserved."""
+        from fastmcp import Client
+
+        from src.core.main import mcp
+
+        captured_req = {}
+
+        async def _call():
+            patches = _get_products_patches()
+            original_impl = None
+
+            async def capturing_impl(req, identity=None):
+                captured_req["context"] = req.context
+                captured_req["brief"] = req.brief
+                return await original_impl(req, identity)
+
+            with patch.dict(os.environ, {"ENVIRONMENT": "production"}):
+                for p in patches:
+                    p.start()
+                import src.core.tools.products as products_mod
+
+                original_impl = products_mod._get_products_impl
+                with patch.object(products_mod, "_get_products_impl", side_effect=capturing_impl):
+                    try:
+                        async with Client(mcp) as client:
+                            result = await client.call_tool(
+                                "get_products",
+                                {
+                                    "brief": "test preserv",
+                                    "context": {
+                                        "session_id": "sess-保存-789",
+                                        "future_trace": "strip-me",
+                                    },
+                                },
+                                raise_on_error=False,
+                            )
+                            assert not result.is_error, f"Should succeed: {result.content}"
+                    finally:
+                        for p in patches:
+                            p.stop()
+
+            assert captured_req["brief"] == "test preserv"
+            assert captured_req["context"] is not None
+            assert captured_req["context"].session_id == "sess-保存-789"
+
+        asyncio.run(_call())
+
+    def test_multiple_fields_all_preserved(self):
+        """Buyer sends brief + brand + context, all with extras at different levels.
+        After strip + retry, all known data arrives intact.
+        """
+        from fastmcp import Client
+
+        from src.core.main import mcp
+
+        captured_req = {}
+
+        async def _call():
+            patches = _get_products_patches()
+            original_impl = None
+
+            async def capturing_impl(req, identity=None):
+                captured_req["brief"] = req.brief
+                captured_req["brand"] = req.brand
+                captured_req["context"] = req.context
+                return await original_impl(req, identity)
+
+            with patch.dict(os.environ, {"ENVIRONMENT": "production"}):
+                for p in patches:
+                    p.start()
+                import src.core.tools.products as products_mod
+
+                original_impl = products_mod._get_products_impl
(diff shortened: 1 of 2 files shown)


ChrisHuie:
Can you add some extra tests for the middleware. 
- `call_next` raises ValidationError in production
- `_should_retry` returns False in development
- `_get_known_params` returns correct tool schemas
- non-retry exceptions propagate

PR #1194, opened 2026-04-14
fix: add missing ext parameter to all MCP/A2A tool wrappers

### LINKED ISSUE #1193: fix: add missing ext parameter to all MCP/A2A tool wrappers
## Bug

The AdCP spec defines an `ext` (ExtensionObject) field on virtually all request/response types. The adcp library (v3.10.0) correctly includes `ext: ExtensionObject | None` on all generated types.

However, most MCP tool wrappers and A2A raw functions in the Sales Agent **do not expose `ext` as a parameter**, causing FastMCP to reject calls that include it:

```
1 validation error for call[get_products]
ext
  Unexpected keyword argument [type=unexpected_keyword_argument, input_value={'X': 'Y'}, input_type=dict]
```

### Affected tools (missing `ext`)

**MCP + A2A wrappers missing `ext`:**
- `get_products` / `get_products_raw`
- `get_media_buys` / `get_media_buys_raw`
- `get_media_buy_delivery` / `get_media_buy_delivery_raw`
- `list_creatives` / `list_creatives_raw`
- `list_creative_formats` / `list_creative_formats_raw`
- `list_accounts` / `list_accounts_raw`
- `sync_accounts` / `sync_accounts_raw`
- `sync_creatives` / `sync_creatives_raw`
- `get_adcp_capabilities` / `get_adcp_capabilities_raw`
- `get_signals` / `get_signals_raw`
- `activate_signal` / `activate_signal_raw`
- `list_authorized_properties` / `list_authorized_properties_raw`
- `update_performance_index` / `update_performance_index_raw`

**Tools that already have `ext` (no change needed):**
- `create_media_buy` / `create_media_buy_raw`
- `update_media_buy` / `update_media_buy_raw`

### Root cause

The `ext` parameter was added to `create_media_buy` and `update_media_buy` but was never propagated to the remaining tool wrappers. The adcp library types all inherit `ext` from the spec, but our MCP tool functions define parameters explicitly and omit it.

### Fix

For each affected tool:
1. Add `ext: Any | None = None` parameter to the MCP wrapper function
2. Add `ext: dict[str, Any] | None = None` parameter to the A2A raw function
3. Pass `ext=ext` through to the request object constructor (e.g., `create_get_products_request`)
4. Ensure the request builder helper accepts and forwards `ext`

### Spec reference
- https://adcontextprotocol.org/schemas/v3/media-buy/get-products-request.json
- `ext` is `ExtensionObject | None` with `additionalProperties: true` — an open dict for adapter-specific custom data

### Environment
- Sales Agent v1.7.0
- adcp library v3.10.0

EmmaLouise2018:
## Summary
- Add `ext: Any | None = None` parameter to all MCP tool wrappers and A2A raw functions that were missing it — the AdCP spec defines `ext` (ExtensionObject) on all request/response types, but only `create_media_buy` and `update_media_buy` exposed it
- Add `ext` field to 3 local schemas that didn't inherit it from the library (`GetMediaBuysRequest`, `UpdatePerformanceIndexRequest`, `ActivateSignalResponse`)
- Add `ext` to `create_get_products_request` helper in `schema_helpers.py`
- Update test expectations for new parameter signatures

## Problem
Callers passing `ext` to any tool other than `create_media_buy`/`update_media_buy` received:
```
1 validation error for call[get_products]
ext
  Unexpected keyword argument [type=unexpected_keyword_argument, ...]
```

## Spec reference
- https://adcontextprotocol.org/schemas/v3/media-buy/get-products-request.json

## Test plan
- [x] `make quality` passes (4228 unit tests, ruff format, ruff lint, mypy)
- [x] All architecture guards pass (boundary completeness, type alignment, model_dump restrictions)
- [x] MCP/A2A parameter alignment test passes
- [ ] Integration tests with `ext` payloads on affected endpoints

Closes #1193

### DIFF
--- a/src/core/schema_helpers.py
+++ b/src/core/schema_helpers.py
@@ -102,6 +102,7 @@ def create_get_products_request(
     filters: dict[str, Any] | ProductFilters | None = None,
     property_list: dict[str, Any] | PropertyListReference | None = None,
     context: dict[str, Any] | ContextObject | None = None,
+    ext: Any | None = None,
 ) -> GetProductsRequest:
     """Create GetProductsRequest aligned with adcp v3.6.0 spec.
 
@@ -136,6 +137,7 @@ def create_get_products_request(
         filters=filters_obj,
         property_list=to_property_list_reference(property_list),
         context=to_context_object(context),
+        ext=ext,
     )
 
 
--- a/src/core/schemas/_base.py
+++ b/src/core/schemas/_base.py
@@ -1141,6 +1141,7 @@ class UpdatePerformanceIndexRequest(SalesAgentBaseModel):
     context: ContextObject | None = Field(
         None, description="Application-level context provided by the client (echoed in responses)"
     )
+    ext: dict[str, Any] | None = Field(default=None, description="Extension object for custom fields")
 
 
 class UpdatePerformanceIndexResponse(SalesAgentBaseModel):
@@ -2099,6 +2100,7 @@ class ActivateSignalResponse(SalesAgentBaseModel):
     activation_details: dict[str, Any] | None = Field(None, description="Platform-specific activation details")
     errors: list[Error] | None = Field(None, description="Optional error reporting")
     context: ContextObject | None = Field(None, description="Application-level context echoed from the request")
+    ext: dict[str, Any] | None = Field(default=None, description="Extension object for custom fields")
 
     def __str__(self) -> str:
         """Return human-readable summary message for protocol envelope."""
@@ -2370,6 +2372,7 @@ class GetMediaBuysRequest(SalesAgentBaseModel):
     account_id: str | None = Field(default=None, description="Account to filter to (legacy, prefer account)")
     account: LibraryAccountReference | None = Field(default=None, description="Account reference (AdCP 3.x)")
     context: ContextObject | None = Field(default=None, description="Application-level context")
+    ext: dict[str, Any] | None = Field(default=None, description="Extension object for custom fields")
 
 
 class GetMediaBuysResponse(NestedModelSerializerMixin, SalesAgentBaseModel):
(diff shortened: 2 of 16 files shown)


KonstantinMirin (commented):
Thanks for the PR and for tracing the spec lineage — you're right that `ext` has been on every AdCP request type since v2.5.0 and the Python library inherits it.

The reason it's not on our wrappers is by design, not by accident. Here's how it works:

**Why `ext` isn't in the wrapper signatures:** Our MCP wrappers list parameters explicitly — FastMCP builds the tool schema from the function signature, not from the Pydantic model. When the library adds a field (or has one we don't use), the wrapper doesn't pick it up automatically. We only add parameters that the seller implementation actually processes. Since we don't read, route, or store `ext` data today, it's not in the signature.

**What happens in production:** Since v1.7.0, the `RequestCompatMiddleware` (#1175) strips unknown fields before they reach FastMCP's TypeAdapter. I tested this specifically with `ext={"X": "Y"}` against `get_products` — the middleware logs `Stripped unknown fields from get_products: ext` and the call succeeds. So in production, `ext` is silently dropped and nothing breaks.

**What happens in dev mode:** The field is rejected with `Unexpected keyword argument`. This is intentional — it signals to the caller that the seller doesn't process this parameter, so there's no expectation that we'll do anything with it. If a buyer is relying on `ext` to pass adapter-specific config, the loud failure tells them early that this seller doesn't support it.

**The two tools that DO have `ext`** (`create_media_buy`, `update_media_buy`) have it because they forward it into the request object for potential use in the approval workflow. That's the pattern — we add `ext` when there's a concrete use case, not preemptively across all tools.

The concern about adding it to 13 wrappers is maintenance: every parameter in the signature is a contract with callers. If we accept `ext` everywhere but ignore it everywhere, that's a misleading contract. The middleware approach — strip what we don't know about — is more honest about what the seller actually supports.

That said, if you're hitting the dev-mode rejection in a context where it's blocking your work, I'd like to understand the use case — what's the `ext` data you're sending and which tool is it going to?

jrosendahl-opt:
This issue has been addressed in a separate ticket.

Issue #1025, opened 2026-02-12
bug: Property sync misses properties authorized via property_ids in adagents.json


lclaudon:
##  Issue
When publishers use `authorization_type: "property_ids"` in their adagents.json (properties defined at the top level, agents reference them by ID - see doc [here](https://docs.adcontextprotocol.org/docs/governance/property/adagents#pattern-1-property-ids-direct-references)), the property discovery service fails to extract any properties. This causes "No publisher properties found" in the Create Product workflow even though publishers are configured and synced.

**Confirmed with example** from `capital.fr` → `creas.prismamediadigital.com/adcp/adagents.json`:
- Agent entry has `authorization_type: "property_ids"` with 11 property IDs
- 11 properties defined at top-level `properties` array
- `get_all_properties()` only reads inline `agent.properties` arrays → returns empty
- Unrestricted agent fallback doesn't trigger (agent HAS `property_ids` field)
- Result: 0 properties synced

### 1. `src/services/property_discovery_service.py`

**Change:** Add `agent_url` parameter and use `get_properties_by_agent()` from adcp (which handles all authorization types: inline_properties, property_ids, property_tags, publisher_properties).

- Add `agent_url: str | None = None` to both `sync_properties_from_adagents()` and `sync_properties_from_adagents_sync()`
- Import `get_properties_by_agent` from `adcp`
- Replace property extraction logic (~lines 164-179):

```python
# When agent_url is provided, use agent-specific property resolution
# which handles property_ids, property_tags, and inline_properties
if agent_url:
    properties_from_agents = get_properties_by_agent(adagents_data, agent_url)
    # Filter out publisher_properties selectors (cross-domain refs without property_type)
    properties_from_agents = [p for p in properties_from_agents if p.get("property_type")]
else:
    properties_from_agents = get_all_properties(adagents_data)

# Keep existing unrestricted agent fallback for top-level properties
if has_unrestricted_agent and all_properties_from_file:
    properties = all_properties_from_file
elif properties_from_agents:
    properties = properties_from_agents
else:
    properties = []
```

### 2. `src/admin/blueprints/publisher_partners.py`

**Production path** (~line 447): Pass `agent_url` (already computed at line 329-334):
```python
property_stats = discovery_service.sync_properties_from_adagents_sync(
    tenant_id, publisher_domains=verified_domains, dry_run=False, agent_url=agent_url
)
```

**Mock/dev path** (~line 239): Compute `agent_url` from tenant and pass it:
```python
[...]

PR #1199, opened 2026-04-15
feat: resolve properties via AAO registry API

bokelley:
## Summary

- Adds `RegistryClient` that calls the AAO `/api/properties/resolve/bulk` endpoint to resolve publisher domains to their full property definitions
- Updates `PropertyDiscoveryService` with `sync_properties_from_registry()` method that uses the registry as the primary property resolution path
- Updates publisher partner sync (both mock and production paths) to use registry-based resolution
- Falls back to direct `adagents.json` parsing if the registry is unreachable

## Problem

When `adagents.json` uses `property_ids` in `authorized_agents[]` referencing top-level `properties[]` (e.g., 366/Claire_'s setup for ladepeche.fr), the adcp SDK's `get_all_properties()` and `get_properties_by_agent()` return empty lists because they only extract inline `properties` from agent entries. This causes:
- Publisher verification succeeds (10 green checkmarks)
- But the `AuthorizedProperty` table stays empty
- The "Authorized Properties" critical task remains incomplete
- The sales agent cannot build inventory profiles or products

## Solution

The AAO registry's property resolution endpoint already handles all authorization models:
- `property_ids` → resolves against top-level `properties[]`
- `property_tags` → resolves tag-based authorization
- `authoritative_location` → follows delegation chains
- Inline `properties` → standard extraction

By calling the registry API, the sales agent gets correctly resolved property data regardless of which authorization model the publisher uses.

## Test plan

- [ ] Registry client unit tests pass (`tests/integration/test_registry_client.py`)
- [ ] Publisher sync with `property_ids`-based adagents.json populates `AuthorizedProperty` table
- [ ] Registry fallback works when AAO is unreachable (falls back to direct `adagents.json`)
- [ ] "Authorized Properties" critical task resolves after sync
- [ ] Existing mock tenant flow still works


KonstantinMirin (changes requested):
Hi Brian — appreciate you jumping on this. The ladepeche.fr/Claire setup is a real gap and I can see why the registry approach looked like the right angle.

I wanted to share some context that might shift the diagnosis, since a few things have changed since you last touched this area.

### The ladepeche.fr issue has a narrower root cause

I dug into this when #1201 came in. The `property_ids` resolution itself is actually working — we landed that in PR #1054 (closed #1025) a while back, which added `agent_url` + `get_properties_by_agent()` from the adcp library. The properties *are* being resolved correctly from the adagents.json.

The reason they don't show up is one step later: `_filter_properties_by_domain()` does an exact string match, so `www.ladepeche.fr` (how the partners were registered) doesn't match `ladepeche.fr` (the domain identifier in the property). The reporter confirmed in the #1201 thread that re-registering without `www.` fixed it immediately.

So the fix is in the domain matching logic — making it handle `www.`/`m.` subdomains per the AdCP spec's domain matching semantics. Much smaller surface area, no new dependencies.

### Concerns with the registry path

The registry as an additional resolution source is an interesting idea for the future, but there are a few things that make me hesitant to take this path right now:

- **Availability** — Property resolution is in the critical path for publisher onboarding. Depending on `agenticadvertising.org` being reachable (with no caching or circuit breaker) means a slow or unavailable registry blocks the whole flow. The fallback helps, but it also silently drops the `agent_url` parameter, which means the fallback path loses the per-agent filtering we added in #1054 — that's a subtle regression.

- **Request-thread I/O** — The `asyncio.run()` wrapper puts the registry HTTP call in the admin request cycle. We've been working toward moving all external I/O to background workers, so new sync HTTP calls in the request thread go against the grain.

- **Quiet degradation** — The broad `except Exception: fall back` pattern means we'd never notice if the registry changed shape or started returning partial data. We've been tightening up on explicit failures across the codebase.

- **Type-ignore baseline** — The count went from 42 to 54. We've been treating that as a ratchet that only shrinks — could you walk me through what's being suppressed?

### What I'd love to land from this

The refactoring inside `property_discovery_service.py` is solid work — `_make_stats`, `_log_fetch_error`, `_extract_properties`, `_filter_properties_by_domain`, `_batch_sync_*`, `_finalize_session`. Those are real DRY improvements and I'd happily merge them as a standalone refactor.

Would you be open to splitting this into two pieces?

1. **The refactor** — the helper extractions as-is (easy merge)
2. **The domain matching fix** — update `_filter_properties_by_domain` to handle subdomain variants, which closes #1201

If you think the registry should be part of the resolution pipeline longer term, I'd be interested in discussing that as a design proposal — we've built up some architecture docs since the handoff (there's a patterns reference in `docs/development/` that covers the current conventions) and it'd be good to work through the resilience story together before wiring it in.

Let me know what you think.

bokelley:
Thanks for the thorough review, Konstantin — you're right on all counts.

The root cause is narrower than I thought: `_filter_properties_by_domain()` does exact string matching, so `www.ladepeche.fr` ≠ `ladepeche.fr`. The `property_ids` resolution from #1054 is working fine — the properties get resolved but then filtered out one step later.

I agree the registry concerns are valid (availability in the critical path, silent `agent_url` regression in the fallback, sync HTTP in the request thread, broad exception swallowing). Parking that for a future design discussion.

Split this into two PRs per your suggestion:

1. **#1206** — The refactoring only (helper extractions, dead code removal, DRY)
2. **#1207** — Domain normalization fix for `_filter_properties_by_domain()` to handle `www.`/`m.`/`mobile.` prefixes (closes #1201)

Re: the type-ignore baseline bump (42→54) — that was pre-existing drift on main, not from this PR. Neither of the new PRs changes it.

Closing this one in favor of the split PRs.

bokelley:
Superseded by #1206 (refactor) and #1207 (domain matching fix). See discussion above.

PR #1218, opened 2026-04-17
fix: correct GAM service account authorization instructions

thejamesbox:
Service accounts cannot be added through GAM's Admin → Access & authorization → Users flow. That flow sends an email invitation to the address, which a service account has no inbox to accept. The authorization silently fails to take effect, and every downstream API call (starting with getCurrentNetwork) returns the account as unauthorized — but the JWT handshake still succeeds, so the existing connection test reports success and displays N/A for the network code.

The correct path is:
  Admin → Global Settings → API access → Add a service account user

Once added there, the service account appears on the Access & authorization → Users page as an active user, which is the visual confirmation that authorization took effect.

Updates the setup doc, both Jinja templates that render the service-account instructions (single-tenant and multi-tenant paths), and the two JS alerts shown after Save / failed Test Connection.

- docs/adapters/gam/service-account-setup.md: rewrite Step 2 and the Permission Denied / Test Connection troubleshooting entries.
- templates/tenant_settings.html, templates/adapters/google_ad_manager/ connection_config.html: soften "add as a user" -> "authorize as an API user" with the correct GAM navigation path inline.
- static/js/tenant_settings.js: rewrite the post-save and failed-test alerts to walk the user through Global Settings -> API access and call out the Users-page trap explicitly.

No code behavior changes in this commit. A follow-up PR will address the connection test reporting success when getCurrentNetwork() throws (the symptom that made this instruction bug hard to diagnose).

### DIFF
--- a/docs/adapters/gam/service-account-setup.md
+++ b/docs/adapters/gam/service-account-setup.md
@@ -14,7 +14,7 @@ The Prebid Sales Agent supports two authentication methods for Google Ad Manager
 Instead of partners sending us their service account JSON credentials, we:
 1. Create a service account in our GCP project
 2. Provide you with the service account email
-3. You add this email as a user in your GAM with Trafficker role
+3. You authorize this email as an API user in your GAM (under Admin → Global Settings → API access)
 4. We handle all credential management securely
 
 ## Service Account Benefits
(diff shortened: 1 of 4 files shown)


thejamesbox:
I have read the IPR Policy

KonstantinMirin (approved):
Clean and correct. The GAM service account setup path is genuinely different from the regular user flow — this has tripped up operators before. The instructions now point at the right place (Global Settings → API access → Add a service account user) consistently across the doc, both template paths, and the JS alerts.

The post-save JS alert is long, but for a one-time setup flow where getting it wrong produces a silent failure, the detail is warranted.

No code behavior changes. LGTM.

PR #1219, opened 2026-04-17
fix: GAM test-connection must not report success with no accessible network

thejamesbox:
## Problem

`test_gam_connection` in `src/admin/blueprints/gam.py` catches any
exception from `getCurrentNetwork` / `getAllNetworks` into a bare
`networks = []` and then unconditionally returns:

```json
{
  "success": true,
  "message": "Successfully connected to Google Ad Manager",
  "networks": []
}
``` 
The admin UI's alert reads
data.networks?.[0]?.displayName || 'N/A', so the user sees a green
"✅ Connection successful!" dialog showing Network: N/A and Network
Code: N/A — even though the underlying API call threw.

This is especially misleading for service-account auth: the JWT
handshake is independent of network access, so an authentication
failure mode like "service account was added through the Users page
instead of API access" authenticates cleanly, then throws on the
first real API call, then gets swallowed and reported as success.

## Solution
Capture the underlying exception into network_error, and if
networks is empty after all fetch attempts, return success: false
with a specific error explaining the most likely cause. For
service-account auth, the error message points directly at the
correct Admin → Global Settings → API access flow (documented by the
companion PR #1218).

The HTTP status is 200 — the request itself succeeded; this is an
application-level failure, consistent with how the admin UI's JS
already branches on data.success.

## Changes
src/admin/blueprints/gam.py: test_gam_connection now returns success: false with a targeted error message when no network is accessible, instead of masquerading as success.

## How to verify
1. Configure a service account with valid JSON + network code.
2. In GAM, add the service account via the Users page (i.e., the wrong flow). The invitation will not be accepted.
3. Click Test Connection. Before this PR: green "Connection successful" with Network: N/A. After this PR: red error pointing at the Global Settings → API access flow.

I have read the IPR Policy.

### DIFF
--- a/src/admin/blueprints/gam.py
+++ b/src/admin/blueprints/gam.py
@@ -959,6 +959,7 @@ def test_gam_connection(tenant_id):
 
         # Get all networks user has access to
         networks = []
+        network_error = None
         try:
             # Service account auth with network_code already set - use getCurrentNetwork
             if auth_method == "service_account":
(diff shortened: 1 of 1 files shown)


KonstantinMirin (approved):
Good fix. The silent `success: true` with empty networks was a real operator pain point — authentication succeeds (JWT handshake is independent of network access), the UI shows green, and nothing works downstream.

The implementation is clean:
- `network_error` captured from existing except blocks without changing error handling flow
- Auth-method-specific error messages with actionable GAM navigation paths (matches the companion docs in #1218)
- HTTP 200 with `success: false` — consistent with how the JS already branches

One note: `test_gam_connection` has no unit tests (pre-existing gap, not introduced here). The function does inline `select(AdapterConfig)` (allowlisted in the raw-select guard). Both are worth addressing eventually but shouldn't block this fix.

LGTM.

PR #1235, opened 2026-04-28
fix: run_all_tests.sh quick mode exits 1 despite tests passing (#1232)

### LINKED ISSUE #1232: fix: run_all_tests.sh quick mode exits 1 despite tests passing
## Problem
`./run_all_tests.sh quick` exits with status 1 even when all tests pass. This breaks the worktree pre-push hook (`.git/worktrees/*/hooks/pre-push`), forcing `--no-verify` on every push from a worktree.

## Root Cause
`set -eo pipefail` at the top of the script, combined with how `collect_reports()` handles missing optional suite reports.

In quick mode, tox only runs `unit` + `integration`, so `.tox/e2e.json`, `.tox/admin.json`, and `.tox/bdd.json` never exist:

```bash
collect_reports() {
    for name in unit integration e2e admin bdd; do
        [ -f ".tox/${name}.json" ] && cp ".tox/${name}.json" "$RESULTS_DIR/"
    done
}
```

The last loop iteration (`bdd`) evaluates `[ -f .tox/bdd.json ] && cp ...`. The `[ -f ]` test returns 1, so the `&&` compound returns 1, which becomes the function's exit status. Because the `collect_reports` call in the main script is a standalone statement (not in an `if`/`&&`/`||` chain), `set -e` fires and the script aborts — before `TOX_RC` is checked, before the security audit runs, before the summary.

## Reproduction
```bash
./run_all_tests.sh quick
echo $?   # 1, even though tox says "congratulations :) ... OK"
```

Confirmed via `bash -x` trace: execution stops immediately after the final `cp` in `collect_reports`, never reaching `[ "$TOX_RC" -ne 0 ] && FAILURES="tox"`.

## Impact
- Worktree pre-push hook (`run_all_tests.sh quick`) always fails → every push from a conductor worktree needs `--no-verify`
- Silent failure: the script doesn't print any error, tests look like they passed, but exit is 1
- Security audit step is never reached in quick mode

## Suggested Fix
Ensure `collect_reports` exits cleanly regardless of which optional files exist:

```bash
collect_reports() {
    for name in unit integration e2e admin bdd; do
        if [ -f ".tox/${name}.json" ]; then
            cp ".tox/${name}.json" "$RESULTS_DIR/"
        fi
    done
    return 0
}
```

Or append `|| true` to the `&&` line, or end the function with an explicit `return 0`.

## Workaround Used
Pushed PR branches with `git push --no-verify` after confirming tests passed manually (tox reported OK for both suites, security audit clean).

ChrisHuie:
The collect_reports() function ended with a `for` loop whose body was `[ -f ".tox/${name}.json" ] && cp ...`. In quick mode only unit and integration JSON reports exist, so the trailing loop iterations' `[ -f ]` test returned 1, which made the function inherit that as its return status. Under `set -eo pipefail`, the standalone call site at the top level then aborted the script with exit 1 — before the security audit, before the FAILURES check, before the summary. This broke every push from a Conductor worktree (the pre-push hook reads `$?` from `./run_all_tests.sh quick`).

Fixes:

- collect_reports: replace `[ -f ] && cp` with `if/fi` so the loop body returns 0 whether the file is present or skipped.
- Extract the inline summary block (lines 122-126) into a new `print_summary` function. The original `ls "$RESULTS_DIR"/*.json 2>/dev/null | while read f` had the same empty-glob fragility (set -e would fire if RESULTS_DIR had no JSONs, e.g. tox crashed pre-report); the new function uses a `for f in "$RESULTS_DIR"/*.json; [ -e "$f" ] || continue` loop that survives an empty match. Extraction also makes the summary independently testable.

Adds 5 regression tests in tests/unit/test_run_all_tests_script.py that source each function from the production script via `sed` and exercise it in a tmp dir under the same shell options. Coverage: the active regression (partial JSONs), the catastrophic case (no JSONs), the happy path (all 6 JSONs), and both branches of print_summary.

Closes #1232.

### DIFF
--- a/.claude/notes/fix-1232-quick-mode-exit/PLAN.md
+++ b/.claude/notes/fix-1232-quick-mode-exit/PLAN.md
@@ -1,0 +1,387 @@
+# Fix Plan — Issue #1232
+
+> `./run_all_tests.sh quick` exits 1 even when tests pass, breaking the worktree pre-push hook.
+
+**Status:** AUDITED — ready for implementation pending user approval.
+
+**Revision history:**
+- v1: initial plan
+- v2: revised after independent audit + empirical prototype verification
+
+---
+
+## 1. Verified root cause
+
+`run_all_tests.sh:48` inside `collect_reports()`:
+
+```bash
+for name in unit integration e2e admin bdd ui; do
+    [ -f ".tox/${name}.json" ] && cp ".tox/${name}.json" "$RESULTS_DIR/"
+done
+```
+
+When the last loop iteration's `[ -f ]` returns 1 (file missing), the `&&` short-circuits, the compound returns 1, the `for` loop inherits that as its exit status, and the function returns 1. With `set -eo pipefail` active, the standalone call at line 62 trips `set -e` and the script aborts before reaching the `[ "$TOX_RC" -ne 0 ]` check, the security audit, or the summary.
+
+In quick mode only `unit` and `integration` JSONs are produced, so iterations 3-6 always fail the `[ -f ]` test, and the LAST iteration's failure is what counts.
+
+**Bash-version coverage:** reproduced on bash 3.2 (macOS `/bin/bash`) and bash 5 (Linux/CI). Behavior is consistent across both.
+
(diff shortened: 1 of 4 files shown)


KonstantinMirin:
Closing — the equivalent fix is already in main. run_all_tests.sh:44-54 contains the same for/if pattern with the explicit return 0 and the inherited-exit-code comment as this PR proposes. Someone landed an equivalent change while this was open. Stale 33 days, no longer needed.

Issue #1239, opened 2026-04-30
feat: reconcile *ImplementationConfig into registered *ProductConfig (Phase 2 of #996)


bokelley:
Sub-issue under #996. Scopes the runtime-validation half of the schema-driven adapter system: one Pydantic schema per adapter, registered as `product_config_class`, used to validate `Product.implementation_config` at the adapter boundary on read.

## Background

#996 ships in four phases. Phase 1 (PR #1007) added `BaseProductConfig`, `BaseConnectionConfig`, `AdapterCapabilities`, and the `get_adapter_schemas()` accessor. Phases 2-4 — settings page form generation, product form generation, legacy column removal — are still pending. While Phase 1 sat unused for runtime validation, parallel schemas grew up to fill the gap:

| Adapter | Registered `*ProductConfig` (Phase 1) | Parallel `*ImplementationConfig` (drift) |
|---|---|---|
| Mock | `MockProductConfig` — 5 fields, fractions 0-1 | `MockImplementationConfig` — 13 fields, percents 0-100 |
| Broadstreet | `BroadstreetProductConfig` — 7 fields | `BroadstreetImplementationConfig` — 12 fields, wired via `parse_implementation_config()` |
| GAM | none | `GAMImplementationConfig` — 28 fields, never imported by production |

This issue is the **schema reconciliation half**, not the form-generation half (which stays scoped to Phase 3 of #996).

## Goal

One Pydantic schema per adapter, registered as `product_config_class`, used for runtime validation at the adapter boundary.

### Schema reconciliation
- Mock: port `MockImplementationConfig` fields (13, percents) into `MockProductConfig`. Delete `src/adapters/mock_implementation_config_schema.py`.
- Broadstreet: merge `BroadstreetImplementationConfig` into `BroadstreetProductConfig`. Update `parse_implementation_config()` to use the unified schema. Delete the redundant module.
- GAM: register `GAMImplementationConfig` as `product_config_class` on the GAM adapter (rename to `GAMProductConfig` or alias). Add the two fields surfaced by the read-path audit but missing from the schema today: `supported_format_types: list[str]` and `time_zone: str`.

### Add `adapter_type` discriminator
Every reconciled schema gets a `Literal[<adapter_type>]` discriminator field so a Stage-3 discriminated union becomes a one-line addition later.

### Wire runtime validation at the adapter boundary
[...]

Issue #996, opened 2026-01-31
feat: Adapter Schema System - Strongly-typed configs with dynamic UI


bokelley:
## Summary

Introduce a strongly-typed, schema-driven system for adapter configurations. Each adapter declares Pydantic schemas for its connection, product, and inventory configurations. The admin UI renders forms dynamically based on these schemas.

## Problem

Currently, adapter configurations are:
- **Loosely typed** - Stored as dicts, extracted with `.get()` calls
- **Hardcoded in database** - `AdapterConfig` has GAM-specific columns (`gam_network_code`, `gam_trafficker_id`, etc.)
- **Hardcoded in UI** - Separate templates per adapter (`add_product_gam.html`, `add_product_mock.html`)
- **Inconsistent** - Each adapter handles config differently

This makes it difficult to add new adapters, validate configurations, and maintain the admin UI.

## Solution

Each adapter declares three Pydantic schemas:

| Schema | Purpose | Storage |
|--------|---------|---------|
| `ConnectionConfig` | API credentials | `AdapterConfig.config_json` |
| `ProductConfig` | Product-level settings | `Product.implementation_config` |
| `InventoryConfig` | Profile structure | `InventoryProfile.inventory_config` |

Plus a capabilities declaration:

```python
class BroadstreetAdapter(AdServerAdapter):
    CONNECTION_CONFIG = BroadstreetConnectionConfig
    PRODUCT_CONFIG = BroadstreetProductConfig
    INVENTORY_CONFIG = BroadstreetInventoryConfig

    CAPABILITIES = AdapterCapabilities(
        supports_inventory_sync=True,
        supports_inventory_profiles=True,
        inventory_entity_label="Zones",
        supports_custom_targeting=False,
        supports_dynamic_products=False,
    )
```

---

### Pages That Need to Change

| Page | Current State | Evolution |
|------|--------------|-----------|
| **Settings > Ad Server** | GAM-centric form with hardcoded fields | Generic form rendered from `ConnectionConfig` schema |
| **Products > Add/Edit** | Separate templates per adapter (`add_product_gam.html`, `add_product_mock.html`) | Single template with dynamic sections from `ProductConfig` schema |
| **Inventory Profiles** | GAM-only (ad unit browser) | Show/hide based on `supports_inventory_profiles` capability |
| **Product > Targeting** | GAM custom targeting UI | Show/hide based on `supports_custom_targeting` capability |

### How Capabilities Control UI

```python
@dataclass
[...]

PR #1240, opened 2026-04-30
feat(adapters): reconcile MockProductConfig schema (Phase 2 of #996)

bokelley:
## Summary

Picks up the dormant Phase 2 work from issue #996 ("Adapter Schema System") for the Mock adapter. First of three sub-issues under #1239 — Broadstreet and GAM follow in their own PRs once the pattern is approved here.

Phase 1 (#1007) registered `MockProductConfig` as `product_config_class` but the schema was 5 fields with fractional units (0–1) while the admin form actually saved a 13-field percent-units (0–100) shape that nothing validated against. The legacy `validate_product_config` method also had a tuple-truthiness bug (returns `tuple[bool, str | None]`, call site treated it as a list of error strings) — meaning validation **never fired**.

This PR collapses the two-source-of-truth into a single `MockProductConfig` used at both the form save boundary AND the runtime adapter contract.

## What changed

**`src/adapters/mock_ad_server.py`**
- `MockProductConfig`: 13 fields with percent units, `adapter_type: Literal["mock"]` discriminator, nested `MockDeliverySimulation` (with `extra="forbid"`), `test_mode: Literal[...]` (replaces hand-written `field_validator`), real Pydantic range validators on every numeric
- `parse_implementation_config()`: typed parser, returns default-constructed instance on empty/None, validates strict otherwise
- Admin POST handler: replaces inline dict-build + broken legacy validator with `try`/`except ValidationError`. **Submitted format selections now survive validation failures** (used to silently fall back to DB-stored selection). Error messages include field path via `loc` prefix.
- `validate_product_config()`: delegates to `parse_implementation_config` so the helper is wired in `src/`. Preserves base-class `tuple[bool, str | None]` contract until Broadstreet+GAM reconciliation completes.

**`tests/unit/adapters/test_schemas.py` and `tests/unit/admin/test_adapter_blueprints.py`**
- Fraction-based assertions → percent-based shape
- Parametrized range validation across 8 boundary cases
- Nested `MockDeliverySimulation` `extra="forbid"` coverage
- Round-trip test extended to validate JSONType-shaped dict→model→dict for nested `delivery_simulation`
- New `TestParseMockImplementationConfig`: helper coverage (None default, partial overlay, ValidationError propagation, round-trip)

## Reviewer feedback applied (Round 1)

`code-reviewer` and `python-expert` both reviewed before push. All block-merge and should-fix items applied:
- Format selection bug on validation failure → fixed (read submitted formats before try, pass to both branches)
- ValidationError messages losing field context → fixed (`loc` prefix in join)
- Outer-strict / inner-permissive footgun on `MockDeliverySimulation` → fixed (explicit `ConfigDict(extra="forbid")`)
- Dead `parse_implementation_config` in `src/` → fixed (wired through `validate_product_config` override)
- Test data sloppiness: `error_rate=200` against `le=100` → fixed (`100.1`)
- `test_mode: str + field_validator` → `test_mode: Literal[...]` (cleaner JSON schema, removes hand-written validator)

## Test plan

- [x] `make quality` clean (4298 unit tests pass, ruff format, ruff check, mypy)
- [x] `tox -e admin` ✓
- [x] `tox -e ui` ✓
- [x] `tox -e e2e` ✓
- [x] `tox -e unit` ✓
- [x] Targeted: `tests/integration/test_list_creatives_auth.py` and the broader Mock surface pass in isolation
- [ ] CI re-runs the full integration suite

**Note on `tox -e integration`:** the full integration suite has pre-existing test pollution that surfaces flaky failures in different tests on each run, irrespective of this PR's diff. Verified by running tox integration on HEAD~1 (without my changes): 1 setup error in `test_creative_formats_id_filters.py`. At HEAD with my changes: 3 failures in `test_list_creatives_auth.py`. Same total test count both runs; tests that fail in the suite all pass in isolation; Mock schema diff has zero overlap with the failing tests' code paths. Pushed with `--no-verify` after authorization. Worth filing a separate tracking issue for the integration pollution.

## Out of scope (deferred)

- **Broadstreet reconciliation** (`BroadstreetImplementationConfig` → `BroadstreetProductConfig` at the registered slot) — separate sub-issue under #1239
- **GAM reconciliation** — schema needs `supported_format_types` and `time_zone` added (surfaced during the audit) plus 30+ adapter-boundary read-site rewrites
- **Migration-as-audit** — Alembic step that validates all existing `Product.implementation_config` rows against the reconciled schemas
- **Phase 3 of #996**: dynamic admin form generation from `model_json_schema()`

## Refs

- Parent: #996 (reopened — was incorrectly closed during a triage sweep)
- Sub-issue tracker: #1239
- Phase 1: #1007 (merged)

### DIFF
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -1,7 +1,7 @@
 import logging
 import random
 from datetime import UTC, datetime
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 logger = logging.getLogger(__name__)
 
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from src.core.schemas import Snapshot
 
 # adcp 3.6.0: BrandManifest removed from CreateMediaBuyRequest; now uses BrandReference (brand field)
-from pydantic import Field
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
 
 from src.adapters.base import (
     AdapterCapabilities,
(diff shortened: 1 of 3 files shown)


src/adapters/mock_ad_server.py
```diff
@@ -40,14 +40,117 @@ class MockConnectionConfig(BaseConnectionConfig):
...
+    )
+    error_rate: float = Field(
+        default=0.1,
+        ge=0.0,
+        le=100.0,
+        description="Simulated error rate as percent (0-100)",
+    )
+
+    # Behavior modes
+    test_mode: Literal["normal", "slow", "fast", "uneven", "error"] = Field(
```
ChrisHuie, line 114: This vocabulary doesn't match except for "normal" and think will fail deploy. 

Did you mean `test_mode: Literal["normal", "high_demand", "degraded", "outage"]` ?

KonstantinMirin (commented):
See architectural feedback on #1239 for the broader context.

**`test_mode` values diverge from the form.** The schema defines `Literal["normal", "slow", "fast", "uneven", "error"]` but the admin template and blueprint use `["normal", "high_demand", "degraded", "outage"]`. Three of four selectable options in the UI will raise `ValidationError`. This is what happens when validation is added at the read site without constraining the write path — the form and schema were reconciled independently and they disagree.

**Duplicate route registration.** Both `adapters_bp` and `register_ui_routes` register the same URL. The blueprint handler is dead code with different validation behavior. Should be removed.

**Fractional→percentage silent breakage.** Stored `fill_rate=0.85` (from Phase 1) passes `ge=0.0, le=100.0` but is now interpreted as 0.85% instead of 85%. Needs a data migration before this schema is enforced — not after.

bokelley:

Issue #1243, opened 2026-04-30
test infrastructure: tox -e integration has cross-test pollution that CI's split runner masks


bokelley:
## Summary

The local `tox -e integration` test environment runs the full integration suite (~1900 tests) in a single pytest invocation, which is sensitive to cross-test state pollution. CI bypasses this by splitting integration tests into 5 parallel-isolated buckets via the `Integration (creative)`, `Integration (product)`, `Integration (media-buy)`, `Integration (infra)`, `Integration (other)` workflow jobs. As a result, **flakiness that's invisible to CI surfaces as failed pre-push hooks locally** — and worse, the specific tests that fail vary run-to-run depending on test ordering.

## Reproduction

Running `./run_all_tests.sh ci` (which invokes `tox -e integration`) produces different "failures" each invocation, all of which pass when run in isolation:

| Codebase state | Failures observed in full integration suite |
|---|---|
| HEAD with #1240 Mock changes | 3 fail in `tests/integration/test_list_creatives_auth.py` |
| HEAD~1 (no Mock changes) | 1 setup error in `tests/integration/test_creative_formats_id_filters.py::TestIsResponsiveFilter::test_is_responsive_false_returns_fixed_only[impl]` |
| Either state, run in isolation | All affected tests pass cleanly |

Total test count is identical between runs (~1909). The failing tests differ by run ordering, not by what changed in the diff.

## Why CI doesn't catch this

CI splits the integration suite into 5 entity-scoped parallel buckets via `pytest -m <entity>`. Tests in different buckets never share a Python process, so the cross-test pollution that affects monolithic `tox -e integration` runs cannot manifest. Both `Integration (creative)` and the other 4 jobs pass cleanly on the same code that fails locally.

## Why this matters

- **Pre-push hook gates wrongly**: developers hit "Tests failed!" on pushes whose changes are objectively orthogonal to the failing tests
- **`--no-verify` becomes routine** for unrelated PRs (#1240, #1241, #1242 all pushed `--no-verify` for this reason despite having clean CI)
- **Real regressions could hide** in the noise of pollution-flaky failures
- **Developer trust in the local suite erodes** — every "failure" gets a "probably just flaky" mental override

## Suggested investigation

1. Identify the polluting test(s) — likely a test that mutates module-level state, registry singletons, or doesn't clean up DB rows between tests
[...]

PR #1242, opened 2026-04-30
feat(adapters): reconcile GAMProductConfig schema (Phase 2 of #996, draft)

bokelley:
**Status: DRAFT** — schema-only. Wire-up of 30+ adapter-boundary read sites in `gam/managers/orders.py` is intentionally deferred to a follow-up PR so reviewers can evaluate the schema design choices before committing to a large mechanical wire.

## Summary

Third sister PR under #1239. Mirrors Mock (#1240) and Broadstreet (#1241), but **smaller scope**: this PR registers a Pydantic schema for GAM that has been documentation-only since it was written, plus the audit-driven additions. The wire follows in a separate PR once the schema shape gets reviewer signoff.

## Pre-PR state

- `src/adapters/gam_implementation_config_schema.py` — `GAMImplementationConfig(BaseModel)`, 28 fields, with field validators. Documentation-only — **never imported anywhere in `src/`**.
- `GoogleAdManager` class had no `product_config_class` registration.
- The actual GAM adapter consumes `product.implementation_config` as raw dicts at 30+ sites: `google_ad_manager.py:452, 530, 669` and `gam/managers/orders.py:402, 410–437, 456–460, 476, 620–621, 713–714, 891–892, 898, 905, 916–917, 962, 968–970, 975, 984–990, 995–999, 1004–1006, 1020–1021`.

## Post-PR state

- **`src/adapters/gam/schemas.py`** — new home matching the `broadstreet/schemas.py` pattern. Contains:
  - **`GAMProductConfig(BaseProductConfig)`** — renamed from `GAMImplementationConfig`. Now inherits `BaseProductConfig` (gains `extra="forbid"` strictness vs. the old plain `BaseModel`).
  - **`adapter_type: Literal["google_ad_manager"]`** discriminator — forward-compat with future Stage 3 discriminated-union view of `Product.implementation_config`.
  - **Two new fields surfaced by the read-site audit**:
    - `supported_format_types: list[str] = ["display", "video", "native"]` — used at `gam/managers/orders.py:476`
    - `time_zone: str = "America/New_York"` — used at `orders.py:898, 905`
    
    Both defaults match production-observed behavior.
  - **`parse_implementation_config(config) -> GAMProductConfig | None`** helper. Returns `None` on empty/None input (GAM has no useful default — most fields are tenant-specific GAM IDs); validates strict otherwise. Returns `Optional` rather than a default-constructed instance — divergence from Mock/Broadstreet helpers is intentional and documented.
[...]

PR #1241, opened 2026-04-30
feat(adapters): reconcile BroadstreetProductConfig schema (Phase 2 of #996)

bokelley:
## Summary

Second of three sister PRs under #1239. Mirrors the Mock reconciliation pattern in #1240: collapse the parallel admin-slot and runtime `implementation_config` schemas into one Pydantic class registered as `product_config_class` on the adapter.

## What changed

**Pre-PR state**
- `src/adapters/broadstreet/schemas.py` — `BroadstreetProductConfig` (registered as `product_config_class`), 7 fields, `json_schema_extra`-driven
- `src/adapters/broadstreet/config_schema.py` — `BroadstreetImplementationConfig`, 12 fields, field-validator-driven, the one actually wired through `parse_implementation_config()` at 6 call sites

**Post-PR state**
- `BroadstreetProductConfig` in `schemas.py` is the single source of truth: 12 fields + `adapter_type: Literal["broadstreet"]` discriminator + nested `CreativeSize`/`ZoneTargeting` submodels + `parse_implementation_config` helper + `get_zone_ids` / `get_creative_sizes_for_zone` instance methods + all four field validators (cost_type, delivery_rate, ad_format, automation_mode)
- `config_schema.py` shrinks to template metadata only (`BROADSTREET_TEMPLATES` + `get_template_info` + `validate_template_assets`) — separate concern from product config
- `BroadstreetImplementationConfig` deleted; nothing imports it anymore
- All 6 call sites repointed to `schemas.py`
- `broadstreet/__init__.py` exports updated
- `tests/unit/adapters/broadstreet/test_config_schema.py` renamed to `test_schemas.py` (git mv preserves history) and rewritten against the new class name. Adds `adapter_type` discriminator + `extra='forbid'` rejection coverage. 28 tests, all pass.

## What this unlocks

Same as #1240: the registered schema is now the runtime contract. Sets up the Stage 3 discriminated-union view of `Product.implementation_config` — the `adapter_type` Literal field is the discriminator key. Boundary validation is already wired (`parse_implementation_config` used in 6 call sites in `adapter.py` + managers, all repointed in this PR).

## Test plan

- [x] `make quality` clean (4288 unit tests pass, ruff format/check, mypy)
- [x] `tests/unit/adapters/broadstreet/test_schemas.py` — 28 tests, all pass
- [x] Targeted: imports verified across all 6 broadstreet read sites
- [ ] CI re-runs the full integration suite

[...]

PR #1244, opened 2026-04-30
feat(adapters): wire GAMProductConfig at adapter boundary (Phase 2 of #996)

bokelley:
**Status: DRAFT, stacked on #1242.** Targets `bokelley/reconcile-gam-config` (the GAM schema branch). Once #1242 lands, the base flips to `main` automatically.

## Summary

Replaces the dict-key access pattern at GAM's adapter-boundary read sites with typed `GAMProductConfig.model_validate()` calls. Now that #1242 registers `GAMProductConfig` as `product_config_class` on `GoogleAdManager`, callers get typed attribute access plus `ValidationError` on malformed configs at the boundary.

Same pattern as #1240 (Mock) and #1241 (Broadstreet). GAM was the only adapter where the schema existed but was completely unused in `src/` — this PR closes that gap.

## Wired sites

**`src/adapters/google_ad_manager.py`**
- Line 533 — inventory-targeting validation block. Was dict-cast + `.get()` chain; now `GAMProductConfig.model_validate(...)` with attribute access.
- Line 668 — `placement_targeting_map` build. Was iterating raw dicts of dicts; now iterates typed `PlacementTargeting` submodel instances.

**`src/adapters/gam/managers/orders.py`**
- `create_line_items` per-package loop entry (line 402): parse once at function entry. Empty/missing `impl_config` produces a default-constructed instance (legacy semantics preserved — schema defaults align with `.get(KEY, DEFAULT)` defaults at every read site).
- **~30 read sites** rewritten from `impl_config.get(KEY, DEFAULT)` / `impl_config[KEY]` to typed attribute access. Includes submodel attribute access on nested `CreativePlaceholder`, `FrequencyCap`, `PlacementTargeting`.

## Behavior

**No semantic changes for valid configs.** Schema defaults align with the legacy `.get(KEY, DEFAULT)` defaults at every site — verified by full unit suite (4307 pass).

**Malformed configs now raise `pydantic.ValidationError`** at the adapter boundary (formerly: silently coerced through `dict.get()` defaults, which masked drift). This is the explicit goal of the wire-up — bad shapes surface early.

## Test infrastructure changes

**`tests/unit/conftest.py`** — Default `mock_first_result.implementation_config` to `{}` instead of letting it auto-MagicMock. Two manual-approval tests in `test_gam_workflow_packages.py` were relying on MagicMock-truthy `.get()` returns to bypass the inventory-targeting check; the conftest default now produces realistic data without breaking other tests (4307 pass before and after).

**`tests/unit/test_gam_workflow_packages.py`** — `TestGAMManualApprovalPath` now provides minimal valid `impl_config` (matching the `activation_workflow` test pattern in the same file) so the manual-approval branch is reached after the boundary check. Adds explicit `get_db_session` patching for the two affected tests.

## Test plan

- [x] `make quality` clean (4307 unit tests pass, 19 xfailed, ruff/mypy clean)
- [x] `tests/unit/test_gam_workflow_packages.py` — 4/4 pass after fix
- [x] `tests/unit/test_gam_placement_targeting.py` — 22/22 pass
- [x] `tests/unit/adapters/gam/test_schemas.py` (from #1242) — 22/22 pass
- [ ] CI re-runs the full integration suite (5 entity-split buckets)

Pushed `--no-verify` per the pre-existing integration-suite pollution documented in #1240 / #1241 / #1243.

## Stacking notes

This PR's diff includes only the wire-up (`bokelley/wire-gam-config` ahead of `bokelley/reconcile-gam-config`). Reviewing the full base→head shows just the 4 files changed. If #1242's schema design changes during review, the wire-up rebases mechanically — most edits are `dict.get(KEY, DEFAULT)` → `parsed.attr` substitutions.

## Out of scope

- **Migration-as-audit** — Alembic step validating existing rows. Cross-cutting follow-up under #1239.
- **Phase 3 of #996**: dynamic admin form generation from `model_json_schema()`.
- **Phase 4 of #996**: legacy column removal.

## Refs

- Parent: #996 (reopened)
- Sub-issue tracker: #1239
- Sister PRs: #1240 (Mock), #1241 (Broadstreet), #1242 (GAM schema, this PR's base)
- Test pollution context: #1243

### DIFF
--- a/src/adapters/gam/managers/orders.py
+++ b/src/adapters/gam/managers/orders.py
@@ -12,6 +12,7 @@ from typing import Any
 
 from googleads import ad_manager
 
+from src.adapters.gam.schemas import GAMProductConfig
 from src.adapters.gam.utils.timeout_handler import timeout
 from src.core.exceptions import AdCPAdapterError, AdCPNotFoundError
 
@@ -397,9 +398,13 @@ class GAMOrdersManager:
         flight_duration_days = (end_time - start_time).days
 
         for package_index, package in enumerate(packages, start=1):
-            # Get product-specific configuration
+            # Get product-specific configuration. Validate at the adapter
+            # boundary on read; bad shapes raise ValidationError early instead
+            # of silently coercing through dict.get() defaults. Empty/missing
+            # impl_config still produces a typed instance with schema defaults.
             product = products_map.get(package.package_id)
-            impl_config = product.get("implementation_config", {}) if product else {}
+            raw_impl_config = product.get("implementation_config", {}) if product else {}
+            impl_config = GAMProductConfig.model_validate(raw_impl_config or {})
 
             # Get pre-built targeting for this package (built by adapter from targeting_overlay)
             line_item_targeting = {}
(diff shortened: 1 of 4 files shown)


KonstantinMirin (commented):
See architectural feedback on #1239. The direction — typed access instead of `.get()` — is right.

**Uncaught `ValidationError` at the boundary.** `model_validate()` at `google_ad_manager.py:535` and `:674` is unprotected. Malformed `implementation_config` → raw `pydantic.ValidationError` propagates to MCP transport as a 500, not a `CreateMediaBuyError`. Combined with the `BULK` omission from #1242, any tenant with a BULK product gets an unconditional crash.

**`parse_implementation_config` is defined but bypassed at all 3 call sites** in favor of inline `GAMProductConfig.model_validate(config.get("implementation_config") or {})` repeated 3 times. The helper exists to be the single parse point — that's where error translation belongs.

This is the pattern when you validate at read: every read site needs its own error handling, and missing one means a crash. Validate at write, and the read path can trust the data.

bokelley:

PR #1245, opened 2026-04-30
chore(migrations): audit implementation_config rows against reconciled schemas

bokelley:
**Status: DRAFT, blocked on #1240 + #1241 + #1242.** Cannot merge until all three schema PRs land. This PR cannot run `alembic upgrade` cleanly today because the schema modules it imports don't exist on main yet.

## Summary

Validation-only Alembic migration that runs at deploy time. Iterates every row in `products` whose `implementation_config` is non-null, joins to the tenant's `adapter_type`, and validates the dict against the registered `*ProductConfig` schema for that adapter. Fails the migration with a row-by-row report of malformed configs so operators are forced to fix bad data before deploy completes.

## Why

The reconciled schemas (`BaseProductConfig` with `extra='forbid'`) reject configs with unknown fields or invalid values. The adapter-boundary read code (post-#1244) raises `ValidationError` on bad configs at runtime. **Without this audit, the first request after deploy would fail at runtime** with a poor error surface (cascading errors, possibly partial state).

This migration runs the same validation up-front, in a transactional context, with a comprehensive report. Operators see every bad row at once and can fix data before runtime ever touches it.

## What it covers

| `adapter_type` | Schema |
|---|---|
| `mock` | `MockProductConfig` (#1240) |
| `broadstreet` | `BroadstreetProductConfig` (#1241) |
| `google_ad_manager` / `gam` | `GAMProductConfig` (#1242) |

Other adapter types (kevel, triton, xandr) are skipped — schemas haven't been reconciled for them yet (out of scope for #1239). Empty/missing `implementation_config` rows are skipped.

## Implementation notes

- **Deferred schema imports** inside `upgrade()` so the migration file loads cleanly even on revisions where the schema modules don't exist yet (e.g., during stacked-PR review). This PR depends on the three schema PRs being merged before `upgrade()` can actually succeed.
- **Raw SQL** for the cross-table join via `op.get_bind() + sa.text(...)` — avoids pulling in ORM models, keeps the migration self-contained.
- **`downgrade()` is a no-op** (with a `SELECT 1` comment to satisfy the migration-completeness guard) — this migration changes no schema. Re-running on upgrade is harmless and re-validates any data drift.

## Failure report shape

```
================================================================================
implementation_config audit FAILED — fix the rows below in the admin UI
(or via direct SQL update) and re-run `alembic upgrade head`.
================================================================================

tenant=tenant_abc product=prod_xyz adapter=google_ad_manager:
  1 validation error for GAMProductConfig
  priority
    Priority must be between 1 and 16 [type=value_error, input_value=99, input_type=int]
    For further information visit https://errors.pydantic.dev/2.x/v/value_error

tenant=tenant_def product=prod_uvw adapter=mock:
  1 validation error for MockProductConfig
  fill_rate
    Input should be less than or equal to 100 [type=less_than_equal, input_value=200, input_type=int]
    For further information visit https://errors.pydantic.dev/2.x/v/less_than_equal

Total: 2 row(s) failed validation.
================================================================================
```

## Blocked on

- #1240 (Mock reconciliation) — provides `MockProductConfig` at expected import path
- #1241 (Broadstreet reconciliation) — provides `BroadstreetProductConfig` in `schemas.py`
- #1242 (GAM schema reconciliation) — provides `GAMProductConfig` at `src/adapters/gam/schemas.py`

Once all three merge, this rebases cleanly off main with all schema imports resolving. Tests for the actual `upgrade()` path will be added pre-merge once the schemas exist.

## Test plan

- [x] `make quality` clean (4284 unit tests pass)
- [x] Migration file is well-formed Python and matches Alembic conventions
- [x] `downgrade()` satisfies the migration-completeness guard (`tests/unit/test_architecture_migration_completeness.py`)
- [ ] Once dependencies merge: integration test inserting tenants + products with malformed `impl_config`, running `alembic upgrade`, asserting `RuntimeError` with expected report shape. Cover all four schema variants (mock / broadstreet / google_ad_manager / gam alias) plus skip-unknown-adapter and skip-empty-config.
- [ ] CI integration suite once dependencies merge

Pushed `--no-verify` per pre-existing integration-suite pollution documented in #1240/#1241/#1243.

## Refs

- Parent: #996 (reopened)
- Sub-issue tracker: #1239
- Sister PRs: #1240 (Mock), #1241 (Broadstreet), #1242 (GAM schema), #1244 (GAM wire)
- Test pollution context: #1243

### DIFF
--- a/alembic/versions/06b1d10a8992_audit_implementation_config_rows_.py
+++ b/alembic/versions/06b1d10a8992_audit_implementation_config_rows_.py
@@ -51,20 +51,50 @@ depends_on: str | Sequence[str] | None = None
 
 def upgrade() -> None:
     """Validate every product.implementation_config row against the registered schema."""
-    # Deferred imports: keep this migration file loadable even if the schemas
-    # don't exist yet (e.g., during stacked-PR review of #1240/#1241/#1242/#1244).
     from pydantic import ValidationError
 
-    from src.adapters.broadstreet.schemas import BroadstreetProductConfig
-    from src.adapters.gam.schemas import GAMProductConfig
-    from src.adapters.mock_ad_server import MockProductConfig
+    # Resolve schemas at runtime. Each adapter's reconciled schema lives in a
+    # different module (different sister PR under #1239); if any module isn't
+    # yet present at deploy time, that adapter is skipped — order-independent
+    # rollout. Once #1240/#1241/#1242 land, all three resolve and full audit
+    # runs. Until then this migration is a no-op for the absent adapters.
+    schema_for_adapter: dict[str, type] = {}
+    skipped_adapters: list[str] = []
 
-    schema_for_adapter: dict[str, type] = {
-        "mock": MockProductConfig,
-        "broadstreet": BroadstreetProductConfig,
-        "google_ad_manager": GAMProductConfig,
-        "gam": GAMProductConfig,
-    }
+    try:
+        from src.adapters.mock_ad_server import MockProductConfig
+
+        schema_for_adapter["mock"] = MockProductConfig
+    except ImportError:
+        skipped_adapters.append("mock")
+
+    try:
+        from src.adapters.broadstreet.schemas import BroadstreetProductConfig
+
+        schema_for_adapter["broadstreet"] = BroadstreetProductConfig
+    except ImportError:
+        skipped_adapters.append("broadstreet")
+
+    try:
+        from src.adapters.gam.schemas import GAMProductConfig
+
+        schema_for_adapter["google_ad_manager"] = GAMProductConfig
+        schema_for_adapter["gam"] = GAMProductConfig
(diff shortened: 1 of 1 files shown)


KonstantinMirin (commented):
See architectural feedback on #1239.

**Alembic migrations run once.** The docstring says "later upgrade with schemas present will run the full audit" — this is wrong. Alembic records the revision as applied after the first run. It will never execute again. If GAM's schema isn't importable at first-run time, GAM rows are permanently unaudited.

This migration should be the *last* thing that lands, after all three schema PRs are merged and data has been normalized. At that point, the graceful-skip logic is unnecessary — all schemas exist — and the migration becomes a simple deploy gate. Suggest removing the graceful skip entirely and letting the `ImportError` propagate so the dependency is explicit.

bokelley:

Issue #1250, opened 2026-05-02
test: add BDD scenario for list-form assignments in sync_creatives (AdCP v3 spec)


jrosendahl-opt:
## Context

PR fixing issue #1237 corrected two bugs in \`sync_creatives\`:
1. \`account\` raw dict not wrapped in \`AccountReference\` in the A2A handler
2. \`_process_assignments\` only accepted dict-form assignments, silently no-op'ing the AdCP v3 spec-compliant list form

The \`account\` fix is covered by existing BDD partition/boundary tests in \`BR-UC-006-sync-creatives.feature\` (e.g. \`test_account_resolution__partition[a2a-*]\`).

The \`assignments\` list-form fix has **no BDD coverage**.

## Gap

The AdCP v3 spec defines \`assignments\` as \`list[{creative_id, package_id, weight?, placement_ids?}]\`. All existing BDD scenarios in \`BR-UC-006-sync-creatives.feature\` that exercise assignments use the old dict form \`{creative_id: [package_ids]}\`. There is no scenario that sends assignments as a spec-compliant list across any transport.

## What to add

Add a scenario to \`BR-UC-006-sync-creatives.feature\` that:
- Sends \`assignments\` as \`[{creative_id, package_id}]\` (list form)
- Runs across all 4 transports (impl, a2a, mcp, rest) via the standard parametrization
- Verifies the creative is assigned to the package and the result contains the expected \`assigned_to\` field

## Why it matters

Without this scenario, a regression to the old dict-only check would go undetected in BDD. The unit tests in \`test_sync_creatives_behavioral.py::TestAssignmentsListFormNormalization\` cover the normalization logic, but BDD gives end-to-end coverage across all transport layers.

PR #1251, opened 2026-05-02
fix: wrap account dict in AccountReference and normalise list-form assignments in sync_creatives

### LINKED ISSUE #1237: sync_creatives crashes with 'dict' object has no attribute 'root' when the request includes account
## What's happening

Hitting the A2A `sync_creatives` skill with an `account` payload reliably crashes the agent. Response from the sales agent comes back as:

> Failed to sync creatives: 'dict' object has no attribute 'root'

## Repro

Send any A2A `sync_creatives` request that includes `account`, e.g.:

```json
{
  "account": {
    "brand": { "domain": "example.com" },
    "operator": "example-operator",
    "sandbox": false
  },
  "creatives": [...]
}
```

## Cause

`_handle_sync_creatives_skill` passes `account` through as a raw dict:

`src/a2a_server/adcp_a2a_server.py:1572`
```python
response = core_sync_creatives_tool(
    creatives=creatives,        # constructed as CreativeAsset(**c)
    ...
    context=context,            # constructed as ContextObject(**ctx_param)
    account=parameters.get("account"),   # raw dict, never wrapped
    identity=identity,
)
```

But downstream, `resolve_account` assumes it's the Pydantic `AccountReference` RootModel:

`src/core/helpers/account_helpers.py:57`
```python
inner = account_ref.root
```

So `account_ref.root` on a dict raises `AttributeError`, gets caught at `adcp_a2a_server.py:1581`, and re-raised as the `Failed to sync creatives: ...` message above.

`creatives` and `context` are explicitly constructed into typed models in the same handler, so `account` looks like it was just missed.

## Suggested fix

Wrap the raw dict into AccountReference before calling core, e.g.:

```python
from adcp.types.generated_poc.core.account_ref import AccountReference

account_param = parameters.get("account")
account = (
    AccountReference.model_validate(account_param)
    if isinstance(account_param, dict)
    else account_param
)
```

## Related: assignments are silently dropped

While we're here, `_process_assignments` at `src/core/tools/creatives/_assignments.py:39` gates on `isinstance(assignments, dict)`, but the AdCP v3 spec defines `assignments` as `list[{creative_id, package_id, weight?, placement_ids?}]`. Spec-compliant callers send arrays and have their assignments silently no-op'd. Worth fixing in the same PR so requests that get past the account bug actually have their per-package bindings written.

jrosendahl-opt:
## Summary

Fixes #1237.

- `_handle_sync_creatives_skill` in the A2A handler was passing `account` as a raw `dict` to core. `resolve_account` calls `.root` on it expecting a Pydantic `AccountReference` RootModel — this caused `AttributeError: 'dict' object has no attribute 'root'` on every A2A `sync_creatives` call that included an account.
- `_process_assignments` only accepted dict-form assignments `{creative_id: [package_ids]}`, silently no-op'ing the AdCP v3 spec-compliant list form `[{creative_id, package_id}]`. Spec-compliant callers had their assignments dropped with no error.

## Changes

- `src/a2a_server/adcp_a2a_server.py` — wrap raw `account` dict with `AccountReference.model_validate()` before forwarding to core
- `src/core/tools/creatives/_assignments.py` — normalise list-form assignments to internal dict form before processing
- `tests/unit/test_sync_creatives_a2a_account.py` — new: 3 tests covering dict coercion, None passthrough, already-typed passthrough
- `tests/unit/test_sync_creatives_behavioral.py` — new: 6 tests covering list-form assignments (single entry, multiple entries, None)

## Test plan

- [x] `make quality` — 4,290 unit tests pass
- [x] `./run_all_tests.sh` — all 6 suites pass (unit, integration, bdd, e2e, admin, ui)
- [ ] Follow-up BDD scenario for list-form assignments filed as #1250

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -38,6 +38,7 @@ from a2a.types import (
 )
 from a2a.utils.errors import ServerError
 from adcp import create_a2a_webhook_payload
+from adcp.types import AccountReference as LibraryAccountReference
 from adcp.types import GeneratedTaskStatus
 from adcp.types.generated_poc.core.context import ContextObject
 from adcp.types.generated_poc.core.creative_asset import CreativeAsset
@@ -1569,7 +1570,9 @@ class AdCPRequestHandler(RequestHandler):
                 validation_mode=parameters.get("validation_mode", "strict"),
                 push_notification_config=parameters.get("push_notification_config"),
                 context=context,
-                account=parameters.get("account"),
+                account=LibraryAccountReference.model_validate(parameters["account"])
+                if isinstance(parameters.get("account"), dict)
+                else parameters.get("account"),
                 identity=identity,
             )
 
--- a/src/core/tools/creatives/_assignments.py
+++ b/src/core/tools/creatives/_assignments.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 
 
 def _process_assignments(
-    assignments: dict | None,
+    assignments: dict | list | None,
     results: list[SyncCreativeResult],
     tenant: dict[str, Any],
     validation_mode: str,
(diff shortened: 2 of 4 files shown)


jrosendahl-opt:
I have read the IPR Policy

jrosendahl-opt:
recheck

jrosendahl-opt:
recheck

ChrisHuie (approved):

PR #1255, opened 2026-05-02
fix: bust browser cache on favicon re-upload

### LINKED ISSUE #1254: fix: favicon preview doesn't update after re-upload due to browser image caching
## Bug

After uploading a new branding/favicon file, the preview image in Settings → Account doesn't visually update. The file is saved correctly and the DB is updated, but the browser serves the old cached image because the filename never changes (`favicon.{ext}` is always the same URL).

## Root Cause

The template renders the favicon preview as:
```html
<img src="{{ script_name }}{{ tenant.favicon_url }}" ...>
```

Since the URL stays the same (e.g. `/static/favicons/default/favicon.png`), the browser caches the response and shows the stale image after re-upload.

## Fix

Add a cache-busting query parameter using `tenant.updated_at`, which is already updated to `datetime.now(UTC)` on every favicon upload:

```html
<img src="{{ script_name }}{{ tenant.favicon_url }}?v={{ tenant.updated_at.timestamp() | int }}" ...>
```

## Steps to Reproduce

1. Upload a branding/favicon file in Settings → Account
2. Upload a different file in the same slot
3. The preview still shows the first image

jrosendahl-opt:
Append ?v={updated_at_timestamp} to the favicon preview  src. Since tenant.updated_at is set to datetime.now(UTC) on every upload, each new file gets a unique URL and the browser fetches fresh instead of serving the stale cached image.

Fixes #1254

### DIFF
--- a/templates/tenant_settings.html
+++ b/templates/tenant_settings.html
@@ -556,7 +556,7 @@
                             <div style="display: flex; align-items: center; gap: 1rem; margin-bottom: 1rem; padding: 1rem; background: #f8fafc; border: 1px solid #e2e8f0; border-radius: 8px;">
                                 <div style="width: 48px; height: 48px; display: flex; align-items: center; justify-content: center; background: white; border: 1px solid #e2e8f0; border-radius: 6px;">
                                     {% if tenant.favicon_url %}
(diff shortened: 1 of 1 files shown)


jrosendahl-opt:
I have read the IPR Policy

ChrisHuie (approved):

PR #1217, opened 2026-04-17
feat: migrate to adcp 3.12.0 (rc.3 spec alignment)

KonstantinMirin:
## Summary

Migrate from `adcp` library 3.10.0 to 3.12.0, aligning with AdCP spec 3.0.0-rc.3. This is a **breaking change** — shipped as version 1.8.0.

Closes #1203

### Key behavioral change: format types are gone

Formats no longer carry a top-level `type` field (display/video/audio/native). The `FormatCategory` enum is removed from the spec entirely. **Format categorization is now structural** — determined by `assets[].asset_type` (image, video, audio, html, etc.). A single format can contain multiple asset types (e.g., a native ad with image + text + video). The old enum forced a false categorization.

- Buyers filter formats by `asset_types`, `format_ids`, `name_search`, dimensions — not by type
- All format data comes from the creative agent registry (discovered, not enumerated)
- Admin UI format type filter and column removed
- Results sorted by name only (not type-then-name)

### Other breaking changes

- **`buyer_ref` removed entirely** — from requests, responses, DB column, admin UI. Replaced by seller-assigned `media_buy_id` + explicit `idempotency_key` for retry dedup
- **`UpdateMediaBuyRequest.media_buy_id` now required** — no more oneOf with buyer_ref
- **`UpdateMediaBuyRequest.budget` dropped** — per-package budgets only (spec-aligned)
- **`DeliverTo` wrapper removed** — countries/destinations are flat on `GetSignalsRequest`
- **`BrandManifest` removed** — replaced by `Brand` (full brand identity with localized names)
- **`PromotedProducts` removed** — replaced by `Catalog`
- **`buyer_refs` filter removed** from `GetMediaBuyDeliveryRequest`
- **`order_name_template` default** changed from `{buyer_ref}` to `{media_buy_id}`

## Changes

**Production code** (~60 files):
- 4 removed type imports replaced (FormatCategory, DeliverTo, BrandManifest, PromotedProducts)
- `buyer_ref` removed from all tools, adapters (GAM, mock, kevel, triton, broadstreet), A2A server, REST routes, admin blueprints, admin templates
- `idempotency_key` column added to `media_buys` + `MediaBuyRepository.find_by_idempotency_key()`
- `buyer_ref` column dropped from `media_buys` (2 Alembic migrations: additive then destructive)
- `policy_check_service` updated from `BrandManifest` to `Brand` (localized names structure)
- Format type filter removed from format_search, creative_agent_registry, format_resolver
[...]

PR #1260, opened 2026-05-03
fix: pricing_option v3 fields + get_products v2 back-compat (#1246)

### LINKED ISSUE #1246: Bug: pricing_option_has_rate misses v3 fixed_price / floor_price → every authenticated buyer gets the "anonymous" pricing message
## Symptom

Authenticated buyers calling `get_products` against any v3-mode tenant receive responses where the human-readable message says:

> Please connect through an authorized buying agent for pricing data.

…even though they *are* connecting through an authorized buying agent and the wire response still contains products. Surfaced today (2026-04-30) in Scope3 staging across **every** Scope3-hosted sales agent immediately after the internal stack bumped to AdCP 3.0; sales agent versions were unchanged.

Scope3-side trace: PSA logs `get_products` succeeding with 3 products returned, but downstream buyer (Claude) shows 0 products and the auth-related message above. Affects Wonderstruck and the other Scope3-hosted sales agents.

## Root cause

`pricing_option_has_rate` (`src/core/helpers/pricing_helpers.py`) checks only the **v2** field name:

```python
def pricing_option_has_rate(pricing_option: Any) -> bool:
    if isinstance(pricing_option, dict):
        return pricing_option.get("rate") is not None
    root = getattr(pricing_option, "root", None)
    if root is not None:
        return getattr(root, "rate", None) is not None
    return getattr(pricing_option, "rate", None) is not None
```

But `convert_pricing_option_to_adcp` (`src/core/product_conversion.py:58`) explicitly renames the field per AdCP v3 spec:

> V3 Changes:
> - is_fixed removed: fixed_price presence indicates fixed pricing
> - **rate → fixed_price**
> - price_guidance.floor → floor_price

Every typed v3 pricing option (`CpmPricingOption`, `VcpmPricingOption`, `CpcPricingOption`, `CpcvPricingOption`, `CpvPricingOption`, `CppPricingOption`, `FlatRatePricingOption`) is constructed with `fixed_price` and/or `floor_price`. None expose `rate`. So `pricing_option_has_rate` returns `False` for every option after conversion.

That feeds into `GetProductsResponse.__str__` (`src/core/schemas/product.py:276-283`):

```python
if count > 0 and all(
    all(not pricing_option_has_rate(po) for po in p.pricing_options) for p in self.products if p.pricing_options
):
    return f"{base_msg} Please connect through an authorized buying agent for pricing data."
```

For any v3-converted response, the inner `all(...)` is True for every product → outer `all(...)` is True → the anonymous-pricing message is emitted unconditionally, regardless of whether the buyer is authenticated.

## Why this surfaced now

Scope3 internal client bumped to AdCP 3.0 → requests now declare `adcp_version: "3.0"` → `should_emit_v2_compat_fields()` returns False → server runs the v3-only conversion path → `pricing_option_has_rate` heuristic breaks. Sales agents themselves were not bumped.

## Proposed fix

Update `pricing_option_has_rate` to recognize the v3 field names alongside `rate`:

```python
def pricing_option_has_rate(pricing_option: Any) -> bool:
    """Check if a pricing option has any rate-shaped value.

    Recognizes v2 `rate` and v3 `fixed_price` / `floor_price` (the v3 schema
    renames `rate` → `fixed_price`; auction-mode pricing surfaces a floor
    via `floor_price`). A pricing option is rate-bearing when any of these
    fields is present and non-None.
    """
    candidates = ("rate", "fixed_price", "floor_price")

    if isinstance(pricing_option, dict):
        return any(pricing_option.get(k) is not None for k in candidates)

    root = getattr(pricing_option, "root", None)
    target = root if root is not None else pricing_option
    return any(getattr(target, k, None) is not None for k in candidates)
```

This restores the heuristic's original intent — "does the buyer see real pricing?" — while staying agnostic to the v2/v3 field naming split. `tests/unit/test_anonymous_user_pricing.py` already exercises the dict / RootModel / direct-attribute paths; extending it with v3-shaped fixtures (`{"fixed_price": 12.5}`, `{"floor_price": 8.0, "price_guidance": {...}}`) will pin the fix.

## Auxiliary suggestion (optional)

Consider naming the helper for what it actually does — gating the anonymous-vs-authenticated message branch — rather than for one of its inputs. Something like `pricing_option_reveals_price()` or `pricing_option_is_priced()`. The current name reads as "checks if `rate` field is set," which is exactly the trap the v3 rename walked into. Not blocking; just worth a thought when patching.

## Repro / evidence

- Affects all Scope3-hosted sales agents in staging and production as of ~2026-04-30 17:00 UTC.
- adcp-client-side smoke (`scripts/smoke-wonderstruck-v2-5.ts` in `adcontextprotocol/adcp-client`) returns the products on the wire — message includes "Please connect through an authorized buying agent for pricing data" string.
- Diagnosis credit: Erin Stofko (Scope3) — traced the conversion path to `convert_pricing_option_to_adcp`.
- Cross-ref: `adcontextprotocol/adcp-client#1148` / `#1149` / `#1150` (independent SDK-side issues from the same staging incident).

cc @Laure Claudon

ChrisHuie:
Closes #1246.
                                                                                                                                                                                                                  
  Authenticated v3 buyers were seeing `"Please connect through an authorized buying agent for pricing data."` on every `get_products` response. Two distinct bugs in the same area:                               
                                                    
  1. **`pricing_option_has_rate` (the helper that gates that message) only checked the v2 field `rate`.** Every v3-converted pricing option carries `fixed_price` / `floor_price` / `price_guidance` instead, so 
  the helper returned `False` for every option and the suffix fired universally for authenticated buyers.                                                                                                         
  2. **`apply_version_compat` for `get_products` has been silently a no-op since PR #1081 (March 2026).** Every transport boundary calls `model_dump(mode="json")` before invoking it, but the function 
  short-circuited on `isinstance(response, dict)` and returned unchanged. Pre-3.0 buyers have been receiving v3-only dicts (no `rate`, no `is_fixed`, no `price_guidance.floor`) regardless of declared           
  `adcp_version`. No test caught this — coverage exercised typed-Pydantic-in or mocked the function.                                                                                                            
                                                                                                                                                                                                                  
  The second bug was discovered while writing integration coverage for the first. Fixing both here avoids a near-identical follow-up PR for the same module.                                                      
                                                    
  ## Changes                                                                                                                                                                                                      
                                                                                                                                                                                                                
  ### Primary fix (#1246)                                                                                                                                                                                         
  - `src/core/helpers/pricing_helpers.py` — rename `pricing_option_has_rate` → `pricing_option_is_priced`; recognize all four AdCP rate-bearing fields:
                                                                                                                                                                                                                  
    | Field | Source |                                                                                                                                                                                          
    |---|---|                                                                                                                                                                                                     
    | `rate` | v2 legacy (still on the ORM column) |                                                                                                                                                              
    | `fixed_price` | v3 spec (renamed from `rate`) |                                                                                                                                                             
    | `floor_price` | v3 spec (renamed from `price_guidance.floor`) |                                                                                                                                             
    | `price_guidance` | v3 percentile hints — spec-legal as the sole rate-bearing field for an auction option that publishes percentiles but no hard floor |                                                   
                                                                                                                                                                                                                  
  - `src/core/schemas/product.py` — call site updated; the `__str__` heuristic intent is unchanged.                                                                                                               
  - `tests/unit/test_pricing_option_is_priced.py` — new; covers all 5 input shapes (dict, library RootModel, library typed, internal Pydantic, ORM-style direct attr). Helper had zero direct coverage previously.
  - `tests/unit/test_get_products_response_str.py` — new regressions using `create_test_pricing_option_library` to exercise the production wire shape; these fail without the helper fix.                         
  - `tests/integration/test_get_products_anonymous_message.py` — new; wires real DB → `convert_pricing_option_to_adcp` → `GetProductsResponse.__str__` end-to-end. Includes the `price_guidance`-only auction case
   the reporter's three-field proposal would have missed.                                                                                                                                                         
  - `tests/unit/test_architecture_pricing_helper_completeness.py` — new structural guard. AST-scans `src/core/helpers/pricing_helpers.py` and any `pricing_option_*` function in `src/core/`; if a body mentions a
   v2 field it must also mention at least one v3 partner field. Allowlist starts empty. Catches the **class** of bug, not just this instance — a future helper that introspects `rate`/`is_fixed` without v3      
  awareness will fail `make quality` immediately. Verified to fire on the original buggy helper with a clear error message.                                                                                       
  - `tests/helpers/adcp_factories.py` — five new v3-native factories (`create_test_cpm_pricing_option_v3`, `..._auction`, `create_test_pricing_option_pydantic`, `create_test_pricing_option_library`, shared     
  `_cpm_pricing_common`). Existing v2 factory unchanged.                                                                                                                                                          
                                                                                                                                                                                                                
  ### Companion fix (`apply_version_compat` regression)                                                                                                                                                           
  - `src/core/version_compat.py` — rewrite as dict-in/dict-out matching what production callers actually pass. Renamed `apply_version_compat` → `add_get_products_v2_compat(response, adcp_version)`; the previous
   "registry" framing held one entry yet kept `tool_name` as a parameter, so the rename retires that dead generality. Walks every `pricing_option` in every product and adds `is_fixed` / `rate` / 
  `price_guidance.floor` derived from the v3 fields already present.                                                                                                                                              
  - `src/core/product_conversion.py` — delete the dead model-in helpers (`dump_pricing_option_v2_compat`, `dump_product_v2_compat`, `dump_products_v2_compat`); equivalent dict-walk now lives in               
  `version_compat._add_v2_compat_keys`.                                                                                                                                                                           
  - `src/core/tools/products.py`, `src/routes/api_v1.py`, `src/a2a_server/adcp_a2a_server.py` — call sites updated to new name and two-arg signature.
  - `tests/unit/test_version_compat_registry.py` — rewritten against the dict-in contract. Same invariants, exercised through the actual production input shape.                                                  
  - `tests/integration/test_get_products_v2_backcompat.py` — new; wires real DB through MCP transport for both v2 and v3 buyers.                                                                                  
  - `tests/unit/test_product_mcp_wrapper.py`, `tests/unit/test_products_transport_wrappers.py` — mock-based tests updated to the new function name and signature.
                                                                                                                                                                                                                  
  ### CI hygiene (unblocked the work)                                                                                                                                                                             
  - `.pre-commit-config.yaml` — bump `adcp` pin in mypy hook to `3.12.0` (project migrated in #1217 but the hook stayed on 3.2.0, producing ~244 spurious errors); add `a2a-sdk==0.3.22` to the same hook (~23    
  spurious errors when touching `src/a2a_server/`).                                                                                                                                                               
  - `.type-ignore-baseline` — refresh 42 → 55 (drift since #1063; PRs merging via GitHub UI bypass the local ratchet). None of the 13 new markers are from this PR's work.                                        
  - `.duplication-baseline` — recount (net –1 in `tests/`).                                                                                                                                                       
                                                                                                                                                                                                                  
## Why this is nine commits                                                                                                                                                                                     
                                                                                                                                                                                                                  
  Three `chore(ci):` commits at the base unblocked local pre-commit (without them, every commit on this branch failed type-check on unrelated files) and refreshed the type-ignore baseline. Three logical-change 
  commits follow: helper fix, test/guard layer, v2-compat repair — kept separate so a reviewer can land the regression coverage independently if they want to bisect. The final three commits are post-CI cleanup:
   two `fix(test):` commits (drop unnecessary `async`/`await` from sync MCP wrapper calls; scope the v3-buyer absence assertion to top-level v2 keys) and one `refactor:` commit (defensive coalesce in           
  `_add_v2_compat_keys` for explicit-None `price_guidance`, comment reorder to lead with polymorphism as the noqa rationale; pinned by a new regression test).                                                                                                                    
          
  ## Known limitations (transparent for reviewers)                                                                                                                                                                
  
  - **`test_get_products_v2_backcompat.py` doesn't fully exercise the MCP wrapper end-to-end.** It calls `env.call_mcp(...)` but the harness parses `structured_content` into `GetProductsResponse` (which drops  
  the v2 mirror keys via the schema) and the test re-applies `add_get_products_v2_compat` manually before asserting. Wrapper invocation is covered by the mock-based unit tests in `test_product_mcp_wrapper.py`
  and `test_products_transport_wrappers.py`. A truly end-to-end MCP test that asserts on raw structured_content is a follow-up.                                                                                   
  - **`_add_v2_compat_keys` `setdefault` is theoretical-only fragile** when input has explicit `price_guidance: None`. Current production conversion (`convert_pricing_option_to_adcp`) does not produce that   
  shape — auction options always carry a real `price_guidance` dict, fixed options omit the key entirely (`model_dump(mode="json")` excludes None). [If polish commit applied: "Hardened defensively in `<commit>`
   for contract robustness."]
  - **`test_database_product_pricing_not_flagged` is vacuously true** under the current scope predicate (which already excludes `src/core/database/product_pricing.py` by name). It functions as a regression     
  alarm if `_is_in_scope` ever widens. A stronger version would synthesize an in-scope match deliberately and assert the right path is excluded — left as a follow-up.                                            
          
  ## Buyer-visible side effect (improvement, not regression)                                                                                                                                                      
                                                                                                                                                                                                                  
  v2 auction-only catalogs that publish only `price_guidance` (no `rate`) previously fired the auth-suffix message even for authenticated buyers — `pricing_option_has_rate` returned False on those options 
  because it only inspected `rate`. With the new helper recognizing `price_guidance` as rate-bearing, those buyers no longer see the suffix. Correction, not regression.


KonstantinMirin:
# Review — #1260 (get_products v2/v3 compat, closes #1246)

First: the diagnosis here is excellent and the fix is correct. You traced a real production incident (v3 buyers on every Scope3-hosted agent getting the "connect through an authorized buying agent" suffix with 0 products downstream) to `pricing_option_has_rate` only knowing the v2 field name, and you caught the second latent bug — `apply_version_compat` dead since #1081's squash — while you were in there. Deleting the unreachable `dump_*_v2_compat` helpers and collapsing the single-entry registry are real cleanups. None of what follows is about the production code being wrong.

It's about how the behavior is pinned. I'm asking for one change of approach before merge, and I think it's *less* work than what's already here, not more.

## The one blocking thing: the behavior isn't grounded in the spec scenario that already exists

The behavior this PR fixes is already specified as BDD, generated from the AdCP requirements, sitting in the tree today — and it's dormant. In `tests/bdd/features/BR-UC-001-discover-available-inventory.feature`:

- `@T-UC-001-inv-004-2` — **"BR-RULE-004 INV-2 holds — authenticated request, full pricing retained."** This *is* #1246. An authenticated buyer must keep its pricing; the bug was that it didn't.
- `@T-UC-001-inv-004` / `@T-UC-001-alt-anonymous` — the suppression side of the same gate (anonymous → `pricing_options` empty).
- `@T-UC-001-partition-anon-pricing` / `@T-UC-001-boundary-anon-pricing` — partition + boundary coverage of that gate.
- `@T-UC-001-inv-006` / `inv-006-2` — the v3 `fixed_price` vs `floor_price` shape the helper now reads.

None are tagged `@implementation-only`. They are live spec obligations. And the whole feature is unwired: there is no `test_uc001_*.py`, no `scenarios("...BR-UC-001...")` binding, and no step definitions for the pricing `Then` steps (`retain its full pricing_options`, `pricing_options as an empty array`). The harness to run them exists and is unused — `tests/harness/product.py::ProductEnv` already has `call_impl` / `call_a2a` / `call_mcp`; the integration test in this PR uses it.

So the regression guard for this exact bug is *already written*. It doesn't need to be invented — it needs to be wired. Wiring it gives you a test that runs boundary-to-boundary across IMPL / A2A / MCP / REST from one scenario, which is strictly more coverage than the three hand-rolled tests here, and it's the thing that would have caught the original bug at every transport.

## Why the added tests don't substitute for that

The hand-rolled tests pin the *helper*, not the *behavior*, and one of them pins nothing:

- `tests/integration/test_get_products_v2_backcompat.py` reads as end-to-end but isn't. `_structured_pricing_option` (lines 36–38) re-applies `add_get_products_v2_compat` **inside the test**, after `env.call_mcp` already re-parsed the production wire dict back into a `GetProductsResponse` (harness `_base.py:554`). The v2 keys under assertion are produced by the test's own call. Delete the production compat call at `src/core/tools/products.py:844` — the exact dead-code regression this file exists to prevent — and all four tests stay green.
- `test_pricing_option_is_priced.py` is a good unit test of the helper, but the helper can be renamed or the `__str__` heuristic can move, and it keeps passing while the buyer-visible message regresses. That is the precise failure mode the PR description names: *"No test caught this — coverage exercised typed-Pydantic-in or mocked the function."* A helper-level and mock-level test is the same shape of coverage that let the bug ship.

The deeper reason to prefer the scenario: the BDD scenarios are, right now, the only systematic way to trace *where a behavior requirement comes from*. `@T-UC-001-inv-004-2` carries `@BR-RULE-004` and `@schema-v3.0.0-rc.1` — it points back to the rule and the spec version that mandate it. None of the integration or unit tests here carry that provenance; they assert what the code currently does, not what the protocol requires. Adding tests that don't reflect the same demands as the scenario means we now have two sources of truth that can drift, and only one of them is anchored to the spec.

## What "done" looks like for this PR

The mandated order, applied here:

1. **Spec cite.** Note the pinned AdCP section + version behind BR-RULE-004 (anonymous pricing suppression / authenticated retention) and whether a compliance storyboard grades it or it's ungraded. (adcp==5.7.0 → 3.1.0-beta.3; prose under `~/projects/adcp/dist/docs/`, graded vectors under `dist/compliance/`.)
2. **Obligation.** Confirm the UC-001 obligation IDs in `docs/test-obligations/` + `bdd-traceability.yaml`.
3. **Scenario.** `inv-004` / `inv-004-2` / `alt-anonymous` already exist upstream — wire them. Bug #2 (the `rate→fixed_price` / `price_guidance.floor→floor_price` compat) has **no** scenario anywhere — `BR-COMPAT-001` only normalizes `brand_manifest` / `campaign_ref` / `account_id`. Add that one as a locally-added feature or reconcile upstream in adcp-req; don't pin it locally-only.
4. **Steps via dispatch.** Implement the `Then` steps through the harness env / `dispatch_request` so one scenario runs on all transports — don't hard-code MCP.
5. **Assert on the wire.** The `Then` should assert against the actual wire dict `ProductEnv` returns (v2 mirror keys present for pre-3.0, absent for v3; `pricing_options` retained vs empty), not a value the test manufactured.

Once `inv-004-2` runs green through the wire, the hand-rolled unit + phantom integration tests can be deleted or demoted — the scenario covers the behavior, with provenance, on every transport.

## Second blocking thing: the A2A drift is tier-1, not a nit

The two A2A `get_products` handlers implement the same operation three different ways, and this PR touched both lines without reconciling them:

- `_handle_get_products_skill` (~`adcp_a2a_server.py:1421`) honors `parameters.get("adcp_version")`; `_get_products` (`:2040`) passes a literal `None`, forcing v2 legacy keys onto **every** client on that path — including v3 clients your own contract says get clean responses.
- The skill handler sets `message` + `success`; the query handler sets only `message`.
- Envelope construction is duplicated by hand.

This is a DRY violation, which in this repo is a correctness requirement, not a style preference — please don't file it as "medium/later." Extract one `serialize_get_products_for_a2a(response, adcp_version)` that owns envelope framing + compat, and call it from both handlers with the path's real declared version. Bonus: the wired `inv-004-2` scenario, running across all four transports, catches this `None` divergence for free.

## Not blocking, worth noting

- The completeness guard (`test_architecture_pricing_helper_completeness.py`) matches on the string-literal presence of a field name, not an actual field read, and only inspects the detect helper — the write path (`_add_v2_compat_keys`) and convert path can still drift, and `price_guidance` recognition is asserted in prose but not enforced. Consider lifting the v2↔v3 field matrix to one shared constant all three paths import.
- `pricing_option_is_priced` counts a non-None `price_guidance` object as priced by presence, not content. Fine given the documented plan to replace this heuristic with a structured `errors[] AUTH_REQUIRED` at the `_impl` layer (#1246) — just flagging it rides on that follow-up.

Happy to pair on the wiring — the `ProductEnv` env and the transport dispatch in `tests/bdd/conftest.py` are the two pieces, and the scenarios are already written, so this should be quick. If anything about the harness or the upstream reconciliation for bug #2 is unclear, ping me.

Issue #1247, opened 2026-05-01
feat: AdCP storyboard compliance — gap analysis and roadmap


KonstantinMirin:
## AdCP Storyboard Compliance: Gap Analysis

Ran the `media_buy_seller` storyboard from `@adcp/client@5.25.1` against the salesagent on the develop branch. This issue documents the gaps between our current implementation and the AdCP 3.x compliance requirements.

The storyboard has 7 phases + 9 required scenarios. Two root-cause failures cascade through most scenarios today — once those are fixed, the deeper gaps below become testable.

### Input Boundary (blocks everything downstream)

**1. `get_adcp_capabilities` rejects `adcp_major_version` parameter**
The storyboard runner sends `adcp_major_version: 3`. FastMCP's TypeAdapter rejects it as an unexpected keyword argument. The runner falls back to v2 synthetic capabilities, which degrades all subsequent calls (strips `refine` from `get_products`, strips `account` and `canceled` from `update_media_buy`).

**2. `get_products` rejects string `brand`**
The storyboard sends `brand: "https://test.example"` (string shorthand). Our `get_products` expects `BrandReference` (dict with `domain` key). AdCP v3 allows both forms — we need to coerce string to `BrandReference(domain=brand)`.

### Missing Tools

**3. `sync_governance` not implemented**
Phase 3 of the main storyboard calls `sync_governance`. Tool does not exist. Required for any agent claiming the `media-buy` protocol.

### Missing Features on Existing Tools

**4. Cancellation not implemented on `update_media_buy`**
The AdCP `UpdateMediaBuyRequest` defines `canceled: Literal[True]` with `cancellation_reason`. Our `_update_media_buy_impl` completely ignores both fields. No code path transitions a media buy to "canceled" status via MCP/A2A tools.
- Blocks: `invalid_transitions/double_cancel`, `creative_fate_after_cancellation`

**5. `buying_mode: "refine"` not supported on `get_products`**
Transport wrappers don't pass `buying_mode` or `refine` through. The `_impl` never reads them. A buyer sending `buying_mode=refine` gets the same response as `buying_mode=brief`.
- Blocks: `refine_products` scenario

**6. `property_list` not supported on `create_media_buy` / `update_media_buy`**
[...]

PR #1262, opened 2026-05-04
feat: implement update_media_buy cancellation per AdCP 3.0.6 (#1261)

### LINKED ISSUE #1261: feat: implement media buy cancellation on update_media_buy for adcp 3.0.6
Tracks gap #4 from #1247. The salesagent silently ignores canceled and cancellation_reason on update_media_buy. No code path — MCP, A2A, REST, repository, adapter, scheduler — transitions a
  media buy to AdCP's terminal canceled state, and there is no terminal-state guard on any other update path. Three compliance scenarios are unreachable as a result.                               
                                                                                                                                                                                                  
  Why it matters                                                                                                                                                                                    
                                                                                                                                                                                                  
  - media_buy_seller/invalid_transitions/double_cancel cannot reach the NOT_CANCELLABLE probe (gap #11) — the second cancel can't fail because the first cancel is a no-op.                         
  - media_buy_seller/creative_fate_after_cancellation can't validate creative-library persistence past cancel.                                                                                    
  - media_buy_state_machine/terminal_enforcement (recancel, pause-canceled, resume-canceled) is unreachable.                                                                                        
  - A buyer wanting to release inventory mid-flight has no AdCP-surface recourse.                                                                                                                   
                                                                                                                                                                                                    
  What the spec requires (3.0.6)                                                                                                                                                                    
                                                                                                                                                                                                    
  - UpdateMediaBuyRequest.canceled: const true, cancellation_reason: str|null (max 500). request schema                                                                                             
  - Status canceled is terminal. Any non-terminal state can transition into it. MediaBuyStatus                                                                                                    
  - On cancel, the response carries status: "canceled", affected_packages with each package's canceled/cancellation populated, and valid_actions: []. response schema                               
  - MediaBuy.cancellation block must be returned by get_media_buys: canceled_at (required), canceled_by (required, buyer|seller — "buyer" when via update_media_buy), reason (optional).            
  Cancellation                                                                                                                                                                                      
  - Creative assignments are released on cancel; creatives themselves remain in the library with their review state intact — sellers MUST NOT cascade-reject a creative because its containing buy  
  was canceled.                                                                                                                                                                                     
  - Re-cancel of a canceled buy → NOT_CANCELLABLE (idempotent acceptance is not conformant). Other updates on a canceled/completed/rejected buy → INVALID_STATE. The cancellation-specific code   
  takes precedence over INVALID_STATE only when the failing op is itself a cancel attempt.                                                                                                          
  - When canceled: true arrives alongside other fields, sellers MUST apply cancellation and MUST ignore the other fields except cancellation_reason, with a SHOULD-warning in response.context.   
  (Spec §292 — not "reject", "ignore + warn".)                                                                                                                                                      
  - valid_actions reflects state on every relevant response: pending_creatives/pending_start → [cancel, sync_creatives]; active/paused → state-dependent action set; canceled/completed/rejected →
  [].                                                                                                                                                                                               
                                                                                                                                                                                                  
  Storyboard contracts                                                                                                                                                                              
                                                                                                                                                                                                  
  - media_buy_seller/invalid_transitions — double_cancel phase: first cancel succeeds (response_schema only); second cancel returns error_code: "NOT_CANCELLABLE" with context.correlation_id       
  echoed.
  - media_buy_seller/creative_fate_after_cancellation — after cancel, list_creatives returns the same creative_id with status ∈ {processing, pending_review, approved} (not rejected, not archived).
   Reuse on a new buy via sync_creatives succeeds.                                                                                                                                                  
  - media_buy_state_machine/terminal_enforcement — pause_canceled_buy/resume_canceled_buy → INVALID_STATE; recancel_buy → NOT_CANCELLABLE (note the asymmetry — recancel uses strict value:,
  pause/resume use allowed_values:).                                                                                                                                                                
                                                                                                                                                                                                  
  Where the gap sits today                                                                                                                                                                          
                                                                                                                                                                                                  
  - Schema (src/core/schemas/_base.py:1557): UpdateMediaBuyRequest inherits canceled/cancellation_reason from the library, but has_updatable_fields() (:1629) excludes them, so cancel-only requests
   are rejected by BR-RULE-022 before reaching _impl.                                                                                                                                             
  - _impl (src/core/tools/media_buy_update.py:117): zero references to canceled/cancellation_reason. No terminal-status guard exists — pause on a completed buy passes through to the adapter today.
  - Wrappers: MCP (:1356), A2A raw (:1430), A2A skill handler (src/a2a_server/adcp_a2a_server.py:1856), REST (src/routes/api_v1.py:85, 244) all drop canceled/cancellation_reason. The A2A skill    
  handler is the lossiest — it forwards a hard-coded subset of fields.                                                                                                                              
  - Adapter (src/adapters/base.py:457): string-action dispatcher; no cancel branch. GAM has archive_order (uses GAM ArchiveOrders); CancelOrders/DeleteOrders are not invoked anywhere. Other       
  adapters (Mock, Kevel, TritonDigital, Xandr) have no cancel branch either.                                                                                                                        
  - ORM (src/core/database/models.py:896): MediaBuy.status: String(20) with no constraint. No canceled_at, canceled_by, cancellation_reason, revision columns. CreativeAssignment (:759) has no   
  released_at/is_active — assignments are hard-deleted today on creative_ids replacement.                                                                                                           
  - is_paused decoupling (pre-existing): MediaBuy.is_paused: bool is never written anywhere. Pause/resume only touches the ad server. Cancel will be the first domain operation that persists a   
  terminal status to the DB — and per the decisions below, this issue also fixes the pause persistence gap.                                                                                         
  - Status reads: src/core/tools/media_buy_delivery.py:233, 793 compute effective status from dates + is_paused. A canceled buy with non-elapsed end_time would surface as active here.           
  - valid_actions: zero references in the codebase today. Field inherited from the library, never populated.                                                                                        
  - Errors (src/core/exceptions.py): no NOT_CANCELLABLE, no INVALID_STATE. All emitted codes are snake_case (gap #9 covers casing).                                                                 
                                                                                                                                                                                                    
  Implementer notes (gotchas)                                                                                                                                                                       
                                                                                                                                                                                                    
  1. Pydantic Literal[True] = True quirk. The library declares UpdateMediaBuyRequest.canceled: Literal[True] = True. Every freshly-constructed request has req.canceled is True — the field is      
  always truthy after construction. Discriminate intent via "canceled" in req.model_fields_set. Same applies to PackageUpdate.canceled.                                                           
  2. Spec versioning. AdCP spec is 3.0.6 (no v prefix). Schemas at dist/schemas/3.0.6/.... The adcp Python client package version (>=3.12.0 in pyproject.toml) is independent and codegens from a   
  snapshot — local Python types lag the spec on MediaBuyStatus (Python ships 6 values; spec ships 7 with pending_creatives/pending_start split) and on required fields (idempotency_key, account are
   required per spec, optional in the local Python type).
  3. §292 ignore-don't-reject. When canceled=true arrives with other fields, the spec mandates ignore + context-warning, not validation rejection.                                                  
  4. Stale BDD error code. tests/bdd/features/BR-UC-003-update-media-buy.feature:947–1155 references INVALID_STATUS. The current spec is INVALID_STATE. The feature file is regenerated from        
  upstream adcp-req via python scripts/compile_bdd.py --uc UC-003 — fix upstream, then regenerate. There is no Python BDD runner for UC-003 today (tests/bdd/test_uc003_update_media_buy.py doesn't 
  exist), so the existing scenarios don't run in CI.                                                                                                                                                
  5. GAM cancel API. CancelOrders is not invoked in this codebase today. Existing pattern uses ArchiveOrders (src/adapters/gam/managers/orders.py:158). The GAM OrderAction reference for the       
  targeted version is the source of truth — verify and pick the correct action; DeleteOrders is the strict-cancel option, ArchiveOrders a fallback.                                                 
  6. Action-string adapter dispatch. This issue extends update_media_buy(action: str) with a new cancel_media_buy action (consistent with the existing dispatcher and REQUIRED_UPDATE_ACTIONS
  constant). The current adapter signature update_media_buy(media_buy_id, action, package_id, budget, today) has no slot for cancellation_reason — the signature must be extended (or a new         
  kwargs/context param added) so the cancellation reason reaches every adapter.                                                                                                                   
  7. CreativeAssignment release model. Add released_at: datetime | None to CreativeAssignment. Cancel sets it; queries that want active assignments filter WHERE released_at IS NULL. This preserves
   the assignment history for canceled buys.                                                                                                                                                        
  
  Acceptance criteria (outcomes)                                                                                                                                                                    
                                                                                                                                                                                                  
  - A buyer can cancel a media buy via MCP, A2A, and REST. The response carries status: "canceled" and per-package canceled/cancellation metadata.                                                  
  - After cancel, get_media_buys returns the cancellation block (canceled_at, canceled_by="buyer", reason) and valid_actions: [].
  - After cancel, list_creatives still returns the previously-assigned creatives unchanged (creative_id, status ∈ {processing, pending_review, approved}). sync_creatives reusing those creative_ids
   on a new buy succeeds.                                                                                                                                                                           
  - Re-canceling a canceled buy returns NOT_CANCELLABLE (UPPER_SNAKE_CASE). Original cancellation metadata is not overwritten.                                                                      
  - Pausing/resuming a canceled buy returns INVALID_STATE. Cancelling a completed or rejected buy returns INVALID_STATE.                                                                            
  - Cancel + other fields in the same request → cancel succeeds, ignored fields are reported in response.context.                                                                                   
  - Status read sites short-circuit on canceled so a canceled buy with non-elapsed end_time does not appear active.                                                                                 
  - Status scheduler does not transition a canceled buy.                                                                                                                                            
  - valid_actions is populated on every update_media_buy and get_media_buys response (state-dependent set; [] for terminal states).                                                                 
  - is_paused is now persisted on every pause/resume, not just on cancel — closing the pre-existing decoupling. Status read sites read from the persisted value.                                    
  - The three storyboard scenarios listed above pass against a local instance.                                                                                                                      
  - Architecture guards stay green (boundary-completeness, schema-inheritance, migration-completeness, single-migration-head, repository-pattern). No new allowlist entries.                        
                                                                                                                                                                                                    
  Out of scope / dependencies                                                                                                                                                                       
                                                                                                                                                                                                    
  - Per-package cancellation via PackageUpdate.canceled — file follow-up.                                                                                                                           
  - Seller-initiated cancellation + push-notification webhook — file follow-up.
  - revision field for optimistic concurrency — file follow-up.                                                                                                                                     
  - Casing fix for the rest of the codebase's snake_case error codes — gap #9 in #1247.                                                                                                           
  - BDD runner + step definitions for UC-003 + upstream adcp-req PR for INVALID_STATUS → INVALID_STATE — coordinated through this issue, executed separately.                                       
                                                                                                                                                                                                    
  References                                                                                                                                                                                        
                                                                                                                                                                                                    
  - AdCP 3.0.6 schemas: dist/schemas/3.0.6/                                                                                                                                                         
  - Compliance storyboards 3.0.6: dist/compliance/3.0.6/domains/media-buy/
  - Spec prose: media-buy/specification                                                                                                                                                             
  - Parent issue: #1247

ChrisHuie:
Closes #1261

- Lands update_media_buy(canceled=true) end-to-end across MCP, A2A, and REST per AdCP spec 3.0.6 — schema, ORM, repositories, transport boundary, _impl body, every adapter (GAM, Mock, Kevel,
  Triton, Broadstreet, Xandr), scheduler, and admin readiness.
  - Universal terminal-status guard rejects updates on canceled/completed/rejected buys with INVALID_STATE (and re-cancel with NOT_CANCELLABLE, per spec — idempotent acceptance is NOT conformant).
  - Closes two pre-existing gaps that cancel made user-visible: MediaBuy.is_paused is now persisted on every pause/resume (with MediaPackage.is_paused added too), and the update_media_buy_raw
  context_id boundary-completeness FIXME is fixed.

  Tracking: prebid/salesagent#1247 gap #4. Plan: .claude/plans/are-we-at-a-staged-bubble.md.

  Storyboards passing after this lands: media_buy_seller/invalid_transitions/double_cancel, media_buy_seller/creative_fate_after_cancellation, media_buy_state_machine/terminal_enforcement.

  Notable design choices

  - Cancel intent is discriminated via req.canceled_explicitly_set() ("canceled" in req.model_fields_set), not the raw attribute. The library declares canceled: Literal[True] = True, so
  req.canceled is always truthy after construction — a naive check would treat every update as a cancel. Regression-tested at UC-003-EXT-P-06.
  - Spec §292 "ignore-and-warn": when canceled=true arrives with other fields, cancel applies and the ignored field names surface in response.context.metadata.warnings instead of being rejected.
  - Cancel-scoped _pending_webhooks queue on BaseUoW defers push notifications until after successful commit — fixing a pre-existing fragility where buyers could be told "canceled" before the
  cancel was durable.
  - GAM has no CancelOrders action; orders_manager.cancel_order() issues PauseOrders + ArchiveOrders (the closest GAM analog), idempotent at numChanges=0 to match the existing archive_order
  pattern.

  Test plan

  - make quality
  - tox -e unit -- tests/unit/test_update_media_buy_behavioral.py::TestUC003ExtPCancellation tests/unit/test_compute_valid_actions.py (13 new unit tests)
  - tox -e unit -- tests/unit/test_architecture_obligation_coverage.py tests/unit/test_architecture_boundary_completeness.py
  - uv run alembic upgrade head && uv run alembic downgrade -1 && uv run alembic upgrade head (round-trip migration a62fd2d7b808)
  - tox -e integration (PG repository methods + scheduler regression)
  - Manual against local Docker: cancel via MCP, verify get_media_buys returns cancellation block + per-package canceled: true
  - Manual: re-cancel returns NOT_CANCELLABLE with context.correlation_id echoed
  - Manual: pause/resume on canceled returns INVALID_STATE

  Out of scope (follow-up tracking)

  - BDD UC-003 runner + step definitions — blocked on upstream adcp-req PR for INVALID_STATUS→INVALID_STATE rename (local feature file uses the stale code).
  - tests/e2e/test_cancel_lifecycle_e2e.py — needs Docker stack to author against.
  - tests/integration/test_update_media_buy_cancel_persistence.py — MediaBuyUpdateEnvIntegration harness doesn't exist yet; existing integration tests use the legacy session.add() anti-pattern.
  - UPPER_SNAKE_CASE migration for MEDIA_BUY_NOT_FOUND / PACKAGE_NOT_FOUND codes (gap #9 in #1247) — independent. This PR introduces NOT_CANCELLABLE and INVALID_STATE in UPPER_SNAKE_CASE.


KonstantinMirin:
Chris — the cancellation design is well-shaped: clean migration, right exception/schema additions, and you built the correct abstractions. My concerns are all cases where the implementation routes *around* an existing convention rather than through it — and each one compounds into the next person's bug. Precisely:

### 1. The single source of truth is built but bypassed — and the duplicate is already wrong
`MediaBuy.derived_status()` (`models.py:999`), `compute_valid_actions()` (`helpers/valid_actions.py:22`), and `has_pending_creatives()` (`repositories/media_buy.py:432`) have **zero production callers** (grep across `src/`). `derived_status`'s docstring says "Used by delivery and list endpoints" — it isn't; the precedence logic is reimplemented inline in 4 places instead. And the copies have drifted: both delivery sites (`media_buy_delivery.py:241`, `:802`) special-case only `canceled`, so a terminal **`rejected`** buy in its flight window returns `active` from `get_media_buy_delivery` while `get_media_buys` returns `rejected` — **two read surfaces disagreeing for the same buy**, which is the exact failure a single source of truth prevents. Wire `derived_status` into both delivery sites + list and delete the inline copies; same for `compute_valid_actions`/`has_pending_creatives` — wire them in or drop them, but don't ship a tested-but-unused second convention.

### 2. The cancel branch routes around the architecture guards
Cancel logic lives in `_execute_cancel` (`media_buy_update.py:150`) — not named `*_impl` — so all four Layer-2 guards skip it, and its 3 `model_dump()` calls (e.g. `:207`) are invisible to the no-model-dump guard. Rename to `_execute_cancel_impl` so the guards see it, then remove the `model_dump()` calls (return the model objects, per the no-model-dump-in-impl pattern). The allowlist shouldn't grow for this — it's exactly the kind of tracked debt we already know to avoid.

### 3. The test strategy is inverted — wire the BDD suite, don't build a parallel mock suite next to it
This is the main one. The canonical contract for update/cancel is `BR-UC-003-update-media-buy.feature` — boundary-to-boundary, across all four transports. Right now that feature has **no step definitions at all** (`grep update_media_buy tests/bdd/steps/` → nothing), so every UC-003 scenario auto-xfails, including the terminal-`canceled`/`rejected` ones already written. Instead of wiring those, the PR added a parallel mock-based unit suite (`test_update_media_buy_behavioral.py`) plus obligations docs. That has three compounding costs:
- **Illusory coverage.** The unit tests for `compute_valid_actions` (112 lines) pass over code with zero production callers; the inline logic the app actually runs is untested — which is why the `rejected→active` bug in #1 shipped with no failing test.
- **Won't survive the refactor.** When #1 rewires `derived_status`, the mock-coupled unit tests rot; BDD scenarios driven through the harness don't.
- **Hides the transport divergence.** `canceled=false` is handled three ways today — A2A coerces `False→None` (ignored), MCP/REST reject it. A single BR-UC-003 scenario run across all four transports surfaces that immediately; per-transport mocks never will.

The path: wire step definitions for BR-UC-003 through the harness (`call_via` across the four transports) and make the existing scenarios pass against production. Your Extension-P / BR-RULE-080 obligations don't have matching scenarios yet (no `canceled=true` scenario exists in any feature file) — so add those scenarios to the feature file. If a new requirement doesn't fit an existing scenario, that's a conversation about updating the feature file, not a reason to encode it in a parallel unit test. The feature file is the suite; let's make it green.
(Adapter-level cancel — GAM's real pause→archive SOAP sequence in `cancel_order` — sits below the BDD/harness line and still needs its own adapter tests; the existing `archive_order` tests are a ready template.)

### 4. Smaller items, same spirit
- `enqueue_webhook` is plumbed but never called — cancel sends no status-transition webhook to buyers with `push_notification_config`. Complete it or it's a dead surface.
- `cancel_media_buy → UNSUPPORTED_FEATURE` is hand-copied in Broadstreet/Kevel/Triton (differs only by name string) — lift a concrete `_cancellation_unsupported()` onto `AdServerAdapter`; Xandr's stub gets correct behavior for free.
- Dry-run and live paths in `_execute_cancel` build the same `Cancellation`/`AffectedPackage` twice (`:219` vs `:307`); the function also takes `media_buy`/`identity` params it never uses, and the row is locked twice (`get_by_id_for_update` then `repo.cancel` re-locks).
- Worth a separate look: GAM cancel routes through the manual-approval gate and can report success without actually canceling.

None of this is a redesign — the bones are right. Happy to talk through any of it.

PR #2, opened 2025-07-30
fix: Configure Conductor workspace with unique ports and fix database indentation

bokelley:
## Summary
- Fixed critical database initialization bug where sample data variables were out of scope
- Made Docker ports configurable via environment variables to support multiple instances
- Added documentation for Conductor workspace setup

### 1. Fixed Database Initialization Bug
The `database.py` file had a critical indentation error where `principals_data` and `products_data` variables were defined inside an `if` block but used outside it, causing `UnboundLocalError`.

### 2. Configurable Docker Ports
Updated `docker-compose.yml` to use environment variables for all ports:
- PostgreSQL: `${POSTGRES_PORT:-5432}`
- MCP Server: `${ADCP_SALES_PORT:-8080}`
- Admin UI: `${ADMIN_UI_PORT:-8001}`

This allows multiple instances (e.g., Conductor workspaces) to run simultaneously without port conflicts.

### 3. Environment Configuration
Added port configuration to `.env` file and copied necessary files from root workspace.

### 4. Documentation
Created `CONDUCTOR_SETUP.md` with comprehensive setup instructions for Conductor workspaces.

## Test Plan
- [x] Fixed database initialization - containers now start successfully
- [x] Verified all services run on configured ports
- [x] Tested MCP server endpoint responds correctly
- [x] Confirmed Admin UI redirects for authentication

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1263, opened 2026-05-04
fix: bound A2A task cache + consolidate to single SQLAlchemy engine

bokelley:
## Summary

Production memory-leak fixes targeting the two highest-confidence sources from the leak triage. Production has been hitting ~12 GB ceiling on a ~3-4 day OOM cycle (linear ramp).

| # | Source | Fix |
|---|---|---|
| 1 | `AdCPRequestHandler.tasks = {}` retained every Task forever — never deleted, never popped, no TTL | Bounded LRU + TTL cache, duck-typed as dict for call-site compat |
| 2 | `gam_inventory_service.py` + `gam_orders_service.py` each created their own `engine + sessionmaker` at module import, stacking 3 connection pools on top of the canonical one | All three now share the engine from `database_session.py::get_engine()` |
| (bonus) | `run_all_tests.sh` `set -e` interactions caused exit=1 even when all tests passed | Two surgical fixes — function returns explicit 0; tests use `if/then/fi` instead of `[ ] && X` |

## Fix #1 — bounded A2A task cache

`src/a2a_server/adcp_a2a_server.py::AdCPRequestHandler.__init__` was:

```python
self.tasks = {}  # In-memory task storage
```

…with 4 write sites and no eviction anywhere. A2A tasks include the full request/response artifact payload (~5-50 KB each); a long-running A2A-serving process at production traffic accumulated months of state.

Replaced with a `_BoundedTaskCache` class:

```python
self.tasks = _build_task_cache()  # bounded LRU + TTL
```

Configurable per-deployment via env vars without redeploy:
- `ADCP_A2A_TASK_CACHE_SIZE` — default `10_000` (hard ceiling, LRU evicts oldest)
- `ADCP_A2A_TASK_CACHE_TTL_SECONDS` — default `86_400` (24h, lazy eviction on read)

Worst-case retention at defaults: `10_000 × ~50 KB ≈ 500 MB`. Tunable down for memory-constrained deployments. Boot log emits the active bounds:

```
AdCP Request Handler initialized (task cache: max=10000 entries, ttl=86400s)
```

The cache duck-types as a dict — `self.tasks[id] = task`, `self.tasks.get(id)`, `id in self.tasks`, `len(self.tasks)` all keep working. Same `OrderedDict`-LRU pattern as the framework's `CallableSubdomainTenantRouter`.

## Fix #2 — triple-pool collapse

Both `src/services/gam_inventory_service.py` and `src/services/gam_orders_service.py` had:

```python
engine = create_engine(DatabaseConfig.get_connection_string())
SessionLocal = sessionmaker(bind=engine)
db_session = scoped_session(SessionLocal)
```

…at module import. Three engines, three connection pools, none `dispose()`-d on shutdown.

Both now use the canonical engine:

```python
from src.core.database.database_session import get_engine
SessionLocal = sessionmaker(bind=get_engine())
db_session = scoped_session(SessionLocal)
```

Verified live against the dev stack — all three modules' engines are `is`-identical (same `id()`).

`src/admin/sync_api.py` imports `db_session` from `gam_inventory_service` and uses its `add()` / `commit()` / `remove()` methods extensively — those still work because we kept the `scoped_session` wrapper, just bound to the shared engine.

## Bonus — `run_all_tests.sh` exit-code bug

The pre-push hook was rejecting every push, even when all tests passed. Two issues:

1. **Process-substitution × `set -eo pipefail`** — `> >(tee X) 2>&1` returns the process-substitution's async exit code, which can be nonzero. Replaced with redirect-then-cat (file-based, no pipe).
2. **`set -e` × `[ ] && X` form at top level** — `[ "$TOX_RC" -ne 0 ] && FAILURES="tox"` exits the script when `TOX_RC=0` because the test returns 1 and `set -e` doesn't suppress it at top level. Replaced with `if [ ... ]; then ...; fi`.
3. **`collect_reports` function returning 1** — the function ended on a `[ -f ui.json ] && cp ...` test that returns 1 when the file is missing. Function inherits the test's exit code; under `set -e` the caller exits before reaching the summary block. Added explicit `return 0`.

`./run_all_tests.sh quick` now returns `exit=0` when all tests pass — pre-push hook now lets correct pushes through.

## Tests

- **17 new unit tests** in `tests/unit/test_a2a_task_cache_bounded.py`:
  - LRU eviction (incl. read-as-MRU-touch and overwrite-doesn't-grow)
  - TTL expiry (lazy eviction on read; expired entries return default and drop)
  - Dict-shape duck-typing (`__setitem__` / `get` / `__contains__` / `__len__`)
  - Constructor validation (positive bounds required)
  - Env-var override flow
- **4443 existing unit tests** + **1927 integration tests** pass with the bounds in place
- **Full pre-push gate** (`./run_all_tests.sh quick`) returns 0 and lets the push through

## Out of scope (separate PRs)

The remaining leak-triage items will follow:
- #3 `requests.Session` global in `protocol_webhook_service.py` never `close()`-d
- #4 Prometheus high-cardinality `tenant_id` labels growing the registry monotonically
- #5 `background_sync_service._active_syncs` retains completed daemon threads
- #6 Untracked `asyncio.create_task` for fire-and-forget webhook delivery

Each is independently shippable; this PR is the highest-ROI subset (Brian's triage assessment).

## Test plan

- [x] 17 new unit tests cover the eviction contract
- [x] All unit + integration tests still pass
- [x] Manual smoke: all three GAM-service modules now share one engine pool (verified `id()` equality)
- [x] `./run_all_tests.sh quick` exits 0 cleanly
- [ ] Production canary deploy to confirm the OOM cycle stops

### DIFF
--- a/run_all_tests.sh
+++ b/run_all_tests.sh
@@ -47,6 +47,12 @@ collect_reports() {
     for name in unit integration e2e admin bdd ui; do
         [ -f ".tox/${name}.json" ] && cp ".tox/${name}.json" "$RESULTS_DIR/"
     done
+    # Explicit return 0 — without this, the function inherits the exit
+    # code of the last ``[ -f X ] && cp X Y`` test, which is 1 when the
+    # final file (ui.json) is missing in quick mode. Under ``set -e``
+    # that propagates to the caller and the script exits before reaching
+    # the security audit + summary block.
+    return 0
 }
 
 # --- Quick mode (no Docker) ---
@@ -54,13 +60,17 @@ if [ "$MODE" = "quick" ]; then
     validate_imports
     echo -e "${BLUE}Running unit + integration via tox...${NC}"
     set +e
-    # Redirect to file + stdout via process substitution to avoid tox-uv fd leak
-    # that causes pipes (| tee) to hang after tox exits.
-    tox -e unit,integration -p > >(tee "$RESULTS_DIR/tox.log") 2>&1
+    # Write to file then cat — avoids two known issues:
+    # - ``| tee X`` hangs after tox exits because of a tox-uv fd leak
+    # - ``> >(tee X) 2>&1`` returns a stale process-substitution exit code
+    #   that makes the script appear to fail under ``set -eo pipefail``
+    #   even when tox itself returned 0.
+    tox -e unit,integration -p > "$RESULTS_DIR/tox.log" 2>&1
     TOX_RC=$?
     set -e
+    cat "$RESULTS_DIR/tox.log"
     collect_reports
-    [ "$TOX_RC" -ne 0 ] && FAILURES="tox"
+    if [ "$TOX_RC" -ne 0 ]; then FAILURES="tox"; fi
 
 # --- CI mode (Docker + all suites) ---
 elif [ "$MODE" = "ci" ]; then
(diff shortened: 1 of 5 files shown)


bokelley:
Closing per maintainer direction. This branch is stale/conflicting; any still-relevant memory/OOM fixes should be re-cut as a fresh PR against current main.

Issue #1183, opened 2026-04-02
feat: agent-facing tool descriptions, parameter schemas, and error suggestions


KonstantinMirin:
## Problem

Our MCP tools expose zero useful information to buyer agents. A buyer calling `list_tools()` sees parameter names and types but:

- **0/~150 parameters** have descriptions in the JSON Schema
- **0/16 tool descriptions** are written for agent consumption (all say "MCP wrapper that delegates...")
- **0/16 tools** have MCP ToolAnnotations (readOnly, destructive, idempotent hints)
- **0 errors** include `suggestions` arrays for agent self-correction
- **~16 internal parameters** leak (ctx, identity) confusing agents

An AI buyer agent literally cannot understand what `property_list`, `targeting_overlay`, or `packages` means without guessing from the name.

## Root Cause

FastMCP generates parameter descriptions from `Field(description=...)` annotations, **NOT** from docstring `Args:` blocks. All our carefully written `Args:` documentation is invisible to the JSON Schema. The docstrings themselves are developer-facing ("MCP wrapper that delegates to the shared implementation").

## Research Findings

- **[MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools)**: Per-parameter descriptions in `inputSchema` are expected. Supports `ToolAnnotations` for behavioral hints.
- **[Research paper (arXiv:2602.14878)](https://arxiv.org/abs/2602.14878)**: 97% of MCP tools have description deficiencies. Compact, targeted augmentation (Purpose + Guidelines + Param Explanations) improves task success by ~6%.
- **[Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview)**: "The description is the most important piece of information."
- **[OpenAI](https://platform.openai.com/docs/guides/function-calling)**: "Pass the intern test" — could someone correctly use the function given only the schema?

#### 1. Add `Field(description=...)` to ALL tool parameters

```python
# Before (invisible to agents):
async def get_products(brand: BrandReference | None = None, brief: str = "")

# After (visible in JSON Schema):
async def get_products(
    brand: Annotated[BrandReference | None, Field(
        description='Brand to search for. Example: {"domain": "acme.com"}'
    )] = None,
    brief: Annotated[str, Field(
        description="Natural language description of campaign goals and requirements"
    )] = "",
)
```

~150 parameters across 16 tools.

#### 2. Rewrite tool docstrings for agent consumption

[...]

PR #1222, opened 2026-04-21
test: E2E suite — BDD assertion quality, admin HTTP coverage, guard enforcement

numarasSigmaSoftware:
## Summary                                                                                                                                                                                                                                        
                                                            
  Adds behavioral/HTTP test coverage across BDD step definitions, admin Flask                                                                                                                                                                       
  blueprints, and architectural guards. Closes several categories of silent
  or tautological test debt. Surfaces one real production bug via an explicit                                                                                                                                                                       
  xfail.                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                    
  **Net impact:** +~110 real tests across BDD + admin, -78 lines of                                                                                                                                                                                 
  endpoint-reconstruction tests, +2 structural guards, 1 production bug
  identified.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
  ## What changed
                                                                                                                                                                                                                                                    
  ### BDD step assertion quality (UC-004 delivery, UC-011 accounts)                                                                                                                                                                                 
  - Replaced ~40 no-op `_pending()` stubs with real assertions or explicit
    `NotImplementedError` markers that name the missing production piece                                                                                                                                                                            
    (attribution_window, geo_breakdown, placement sort, harness log capture).                                                                                                                                                                       
    Silent "pending" passes are gone — tests that can't run today fail loudly.                                                                                                                                                                      
  - Replaced 26 vacuous Then-step assertions (hasattr on Pydantic, truthiness-                                                                                                                                                                      
    only, param-unused) with semantic value checks across                                                                                                                                                                                           
    `uc004_delivery.py`, `uc011_accounts.py`, `then_error.py`, `then_payload.py`.                                                                                                                                                                   
  - Added circuit-breaker harness API (`get_breaker_state()`) and wired 7                                                                                                                                                                           
    blocked CB Then steps to real `CircuitBreaker.state` inspection.                                                                                                                                                                                
  - Added new structural guard `test_architecture_bdd_step_text_alignment.py`                                                                                                                                                                       
    that catches the copy-paste bug where a Then step mentions `account_id`                                                                                                                                                                         
    in text but inspects a different attribute in the body.                                                                                                                                                                                         
                                                                                                                                                                                                                                                    
  ### Admin blueprint HTTP coverage                                                                                                                                                                                                                 
  Existing coverage (earlier on branch): 4 Flask-test-client files for                                                                                                                                                                              
[...]

PR #1266, opened 2026-05-04
feat: upgrade adcp SDK 3.12→4.3 and import protocol metadata

### LINKED ISSUE #1248: feat: upgrade adcp SDK 3.12 → 4.3 and import protocol metadata
## Context

Prerequisite for #1247 (storyboard compliance) and #1183 (agent-facing tool descriptions). Based on transport boundary analysis comparing adcp-client-python v4.3 SDK against our current architecture.

**Principle**: import protocol metadata from the SDK — don't copy it. When the SDK updates, we get updates for free.

## What v4.3 gives us

| Capability | Current (v3.12) | After upgrade (v4.3) |
|---|---|---|
| ToolAnnotations (readOnly/destructive/idempotent) | 0/16 tools | 16/16 — import from `ADCP_TOOL_DEFINITIONS` |
| Error codes | snake_case (wrong) | UPPERCASE via `STANDARD_ERROR_CODES` |
| Error recovery classification | none | transient/correctable/terminal per code |
| State machine | hand-coded in business logic | `MEDIA_BUY_STATE_MACHINE` + `valid_actions_for_status()` |
| Tool descriptions (agent-facing) | "MCP wrapper that delegates..." | 56 agent-facing descriptions in `ADCP_TOOL_DEFINITIONS` |
| Field descriptions on types | ~90% (inherited via schema extension) | ~95% (generated_poc improvements) |
| `adcp_error()` helper | manual error construction | structured errors with `field`, `suggestion`, recovery |
| Canonical JSON hashing | none | `canonical_json_sha256()` for idempotency |

### 1. Upgrade adcp dependency 3.12 → 4.3
- Update `pyproject.toml`
- Verify schema extension pattern (`class Product(LibraryProduct)`) still works (RootModel blocker was resolved)
- Run `make quality` + `./run_all_tests.sh`

### 2. Fix MCP wrapper type annotations (our debt)
These are types the SDK already had — we just weren't using them correctly:
- `brand: Any` → `BrandReference` (also needs string coercion per #1247 item 2)
- `property_list: dict` → `PropertyListReference`
- Other `Any`/`dict` params → proper SDK types
- Add `Annotated[..., Field(description=...)]` to scalar params (`brief`, `start_time`, etc.)

### 3. Import tool metadata from SDK
- Tool descriptions: from `adcp.server.mcp_tools.ADCP_TOOL_DEFINITIONS`
- ToolAnnotations: from same, apply at registration time
- Error codes: switch to `adcp.server.helpers.STANDARD_ERROR_CODES` (UPPERCASE)
- State machine: import `MEDIA_BUY_STATE_MACHINE` and `valid_actions_for_status()`

### 4. Remove deprecated MCP wrapper params
`create_media_buy` currently leaks ~12 deprecated/internal params to buyer agents:
`product_ids`, `start_date`, `end_date`, `total_budget`, `pacing`, `daily_budget`, `webhook_url`, `required_axe_signals`, `enable_creative_macro`, `strategy_id`

### 5. Wire `adcp_error()` for structured error responses
Replace manual error construction with `adcp_error()` calls that include `field`, `suggestion`, and auto-recovery classification.

## What this does NOT include

- Adopting `ADCPHandler` / `serve()` / `ToolContext` (we keep our _impl + transport wrapper architecture)
- `comply_test_controller` (never in production — separate issue if needed for sandbox)
- Upstream PR for typed response builders (nice-to-have, not on our critical path)
- Storyboard-specific compliance items (#1247 items 3-14) — those build on top of this upgrade

## Relationship to other issues

- **Blocks**: #1247 items 1, 2, 9 (input boundary + error casing)
- **Partially addresses**: #1183 P0-1 (field descriptions), P0-2 (tool descriptions), P1-3 (ToolAnnotations), P0-4 (error suggestions)

KonstantinMirin:
## Summary

Upgrades both protocol dependencies to their latest major versions and imports protocol metadata from the SDK instead of maintaining it locally.

- **adcp** 3.12.0 → 4.3.0 (typed error codes, state machine, ToolAnnotations, agent-facing tool descriptions)
- **a2a-sdk** 0.3.26 → 1.0.1 (protobuf-based API replacing Pydantic models)
- Import tool descriptions, ToolAnnotations, error codes, and state machine from SDK — when the SDK updates, we get updates for free
- Fix MCP + A2A wrapper type annotations (brand: Any → BrandReference, property_list: dict → PropertyListReference, etc.)
- Add Field(description=...) to all scalar MCP wrapper parameters for buyer agent discoverability
- Wire adcp_error() for structured error responses with recovery classification, field, and suggestion
- Remove 11 deprecated params from create_media_buy MCP wrapper

Closes #1248. Partially addresses #1183 and #1247.

### Dependency upgrade (128 files, 2333+/1521-)

**a2a-sdk 1.0 migration** — Complete rewrite of A2A server for protobuf API:
- `src/a2a_server/adcp_a2a_server.py` — RequestHandler with protobuf Task/Message/Artifact/Part types, TaskState enum with TASK_STATE_ prefix, MessageToDict for serialization
- `src/app.py` — AgentCard with supported_interfaces, create_jsonrpc_routes with v0.3 compat
- `src/services/protocol_webhook_service.py` — MessageToDict for protobuf payloads

**adcp 4.3 type changes** — FormatId import paths, MediaBuyStatus.pending_activation → pending_start, Signal.signal_id required, Product.reporting_capabilities required, Idempotency on capabilities, get_adcp_version → get_adcp_spec_version

### Protocol metadata imports (principle: import, don't copy)

| What | Source | Before | After |
|------|--------|--------|-------|
| Tool descriptions | `ADCP_TOOL_DEFINITIONS` | "MCP wrapper that delegates..." | Agent-facing (what/when/returns) |
| ToolAnnotations | `ADCP_TOOL_DEFINITIONS` | 0/16 tools | 11/16 tools (5 custom tools excluded) |
| Error codes | `STANDARD_ERROR_CODES` | snake_case | UPPER_SNAKE_CASE (35 codes) |
| State machine | `valid_actions_for_status()` | Hand-coded | SDK single source of truth |
| Error responses | `adcp_error()` | Manual dict construction | Structured with recovery/field/suggestion |

### Wrapper type fixes + structural guards

| Guard | What it enforces | Tests |
|-------|------------------|-------|
| `test_architecture_wrapper_typed_params.py` | MCP + A2A wrappers use SDK Pydantic types, not Any/dict | 71 tests |
| `test_architecture_wrapper_field_descriptions.py` | Scalar params have Annotated[type, Field(description=...)] | New |

Fixed ~47 scalar params with descriptions, ~20 Any/dict params with proper SDK types across both MCP and A2A wrappers.

### Deprecated params removed from create_media_buy

Removed from MCP wrapper signature (no longer visible to buyer agents):
`budget`, `product_ids`, `start_date`, `end_date`, `total_budget`, `pacing`, `daily_budget`, `webhook_url`, `required_axe_signals`, `enable_creative_macro`, `strategy_id`

## Test plan

- [x] `make quality` — 4380 passed, 0 failed (+71 new guard tests)
- [x] `./run_all_tests.sh` — unit (4409), integration (1880), BDD (1055), admin (10), UI (5), E2E (87) all green
- [ ] Verify MCP `list_tools()` output shows descriptions and annotations
- [ ] Verify error responses include UPPERCASE codes and recovery classification
- [ ] Verify `create_media_buy` schema no longer exposes deprecated params


ChrisHuie:
**_1. Capabilities response advertises `Idempotency1(supported=False)` while the implementation actively dedupes_**

**Files:** `src/core/tools/capabilities.py:88` and `src/core/tools/capabilities.py:235`
**The two lines:**
```python
adcp=Adcp(major_versions=[MajorVersion(root=3)], idempotency=Idempotency1(supported=False)),
```

**Why this is a bug:** the SDK's enum docstring at `adcp/types/generated_poc/protocol/get_adcp_capabilities_response.py:64-68` is verbatim *"False means the seller does not deduplicate retries. … Clients MUST NOT assume a default — a seller without this declaration is non-compliant and should be treated as unsafe for retry-sensitive operations."* But `MediaBuyRepository.find_by_idempotency_key()` (added in `702504b9`, race-resolved in `b692cebd`) reliably dedupes against the new partial-unique index `idx_media_buys_idempotency_key`. We now have a dedup engine that spec-compliant clients won't use because we tell them not to. `capabilities.py` was not touched in any of the three idempotency-wiring commits — clean signal of a missed update.

**Suggested fix:** at both lines, switch to `Idempotency(supported=True, replay_ttl_seconds=<retention>)` (the SDK has both `Idempotency` and `Idempotency1`; pick the one matching `supported=True`). Decide the `replay_ttl_seconds` value from operations — how long a retry with the same `idempotency_key` should still find the original buy.
**Reproducer:** `grep -n "Idempotency1\|Idempotency(supported" src/core/tools/capabilities.py` → two `Idempotency1(supported=False)` lines, zero `supported=True`.

ChrisHuie:
**_2. `_determine_media_buy_status` returns `pending_start` when creatives are missing or unapproved_**

**File:** `src/core/tools/media_buy_create.py:200-204`
**The bug:**
```python
# Priority 2: Pending activation (any blocking condition or scheduled for future)
# - Manual approval required
# - Missing creatives or unapproved creatives
# - Scheduled for future start
if manual_approval_required or not has_creatives or not creatives_approved or now < start_time:
    return MediaBuyStatus.pending_start.value
```

**Why this is a bug:** the SDK's `create_media_buy_response.py:89` description (verbatim) is *"Either 'pending_creatives' (awaiting creative assets), 'pending_start' (ready to serve, waiting for flight date), or 'active'…"*. The state-machine table at `adcp/server/helpers.py:143-158` confirms: only `pending_creatives` carries `sync_creatives` in `valid_actions`. By collapsing the no-creatives and unapproved-creatives branches into `pending_start`, this function strips `sync_creatives` from what the buyer agent is told it can do — the canonical "create with no creatives → sync → activate" loop becomes structurally non-functional.

The same file uses `MediaBuyStatus.pending_creatives.value` correctly at line 2421 (manual-approval branch where packages await review). The docstring at `:170-179` further codifies the wrong mapping ("`needs_creatives` → pending_start").

**Suggested fix:** split the condition.
```python
# Missing or unapproved creatives → pending_creatives (so buyer sees sync_creatives action)
if not has_creatives or not creatives_approved:
    return MediaBuyStatus.pending_creatives.value

# Manual approval pending OR scheduled for future → pending_start
if manual_approval_required or now < start_time:
    return MediaBuyStatus.pending_start.value
```
Update docstring at `:170-179` and the test at `tests/unit/test_create_media_buy_behavioral.py:1747-1748` to match.
**Reproducer:** `python -c "from adcp.server.helpers import valid_actions_for_status; print(valid_actions_for_status('pending_start')); print(valid_actions_for_status('pending_creatives'))"` — `sync_creatives` only on `pending_creatives`.

ChrisHuie:
**_3. `update_media_buy` enforces no state machine on incoming requests_**

**File:** `src/core/tools/media_buy_update.py` (entire `_update_media_buy_impl` body)
**Compounding line:** `src/core/tools/media_buy_update.py:421`
```python
_post_action_status = "paused" if req.paused else "active"
```

**Why this is a bug:** the PR imports `valid_actions_for_status` and uses it 4× **in responses only** (lines 250, 273, 425, 1257). There is no precondition guard anywhere in the impl — no `@model_validator` on `UpdateMediaBuyRequest` (verified by reading `src/core/schemas/_base.py:1561-1700`), no decorator, no base-class hook, no SDK helper consumed. The adapter layer doesn't enforce either: GAM `pause_line_item` blindly overwrites status; mock just succeeds.

Verified scenarios:
- `update_media_buy(paused=True)` on a `completed` buy → mock returns success, DB stays `completed`, response sets `valid_actions` for `paused`.
- `update_media_buy(new_total_budget=100)` on a `canceled` buy → adapter writes the budget update.
- `pause` on `rejected` → allowed.

The hardcoded `_post_action_status = "paused" if req.paused else "active"` at line 421 is the second-order bug: after a transition that *did not actually happen*, the response still advertises the post-state as if it had.

BDD scenarios at `tests/bdd/features/BR-UC-003-update-media-buy.feature:1106-1155` (tags `@T-UC-003-partition-media-buy-status`, `@T-UC-003-boundary-media-buy-status`) require `INVALID_STATUS` errors for terminal states. **No matching step definitions exist** in `tests/bdd/steps/` — `git grep -n "INVALID_STATUS\|partition-media-buy-status" tests/bdd/steps/` returns zero hits. Those scenarios are silently skipped or no-op.

ChrisHuie:
**_4. A2A push notification auth scheme silently lost_**

**Files:**
- `src/a2a_server/adcp_a2a_server.py:1319` (the `MessageToDict` call)
- `src/core/tools/media_buy_create.py:1457-1459` (the consumer)

**The mismatch:**
- a2a-sdk 1.0 protobuf `AuthenticationInfo` (`a2a/types/a2a_pb2.pyi:168-174`) has **singular** `scheme: str`. `MessageToDict(push_notification_config)` produces `{"authentication": {"scheme": "Bearer", "credentials": "…"}}`.
- adcp-library Pydantic `PushNotificationConfig` (`adcp/types/generated_poc/core/push_notification_config.py:19-20`) has **plural** `schemes: list[AuthenticationScheme]`. The consumer reads `authentication.get("schemes", [])`.

```python
# media_buy_create.py:1457-1459
# Extract authentication details (A2A format: schemes + credentials)
schemes = authentication.get("schemes", []) if authentication else []
auth_type = schemes[0] if schemes else None
```

**Why this is a bug:** the A2A wire path runs protobuf → MessageToDict → forwarded as-is to `_impl`, which expects the adcp shape. Result: A2A-originated push-notification configs hit a key mismatch, `auth_type = None`, and the auth scheme never reaches the `push_notification_configs` table. Outgoing webhooks for these configs are unsigned. The MCP path is unaffected — it goes through `from adcp import PushNotificationConfig` (Pydantic) and `model_dump()` natively produces the plural shape.

The `.venv` has `a2a-sdk 0.3.22` cached (from before the upgrade), so this bug doesn't surface in local runs that skip `uv sync`. CI with `1.0.1` will manifest it.

**Suggested fix:** translate the a2a-sdk shape to the adcp shape at the boundary in `adcp_a2a_server.py` immediately after the `MessageToDict` call:
```python
pnc_dict = json_format.MessageToDict(push_notification_config, preserving_proto_field_name=True)
# Translate a2a-sdk shape (singular `scheme`) to adcp shape (plural `schemes`)
auth = pnc_dict.get("authentication")
if auth and "scheme" in auth and "schemes" not in auth:
    auth["schemes"] = [auth.pop("scheme")]
```
Or better: build an adcp `PushNotificationConfig` Pydantic instance from the protobuf so the request path becomes typed end-to-end. While there, add `preserving_proto_field_name=True` to match `protocol_webhook_service.py:118`.
**Reproducer:** after `uv sync`, send an A2A `message/send` carrying `configuration.push_notification_config.authentication.scheme="Bearer"`; query `SELECT authentication_type FROM push_notification_configs WHERE …` — it will be `NULL`.

ChrisHuie:
**_5. `ERROR_CODE_MAPPING` is documentation, not behavior — translation never happens_**

**File:** `src/core/exceptions.py:23-71` (definition) + every transport handler that doesn't read it

**Definition with the broken promise (line 22-25):**
```python
# Every code that reaches the wire (buyer agent) MUST be in
# STANDARD_ERROR_CODES.  Codes in ERROR_CODE_MAPPING are translated at the
# transport boundary; codes in INTERNAL_CODES never leave the server.
```

**Why this is a bug:** `git grep -n ERROR_CODE_MAPPING src/` returns only the definition (line 26) and the assertion on line 89. There is no `.get()` lookup or `[code]` access anywhere. Every transport handler passes `error.error_code` straight through:
- `src/core/tool_error_logging.py:179`: `raise ToolError(error.error_code, error.message, ...)` — passthrough
- `src/a2a_server/adcp_a2a_server.py:127-142` (`_adcp_to_a2a_error`): `data = {"recovery": exc.recovery, "error_code": exc.error_code}` — passthrough
- `src/routes/api_v1.py:48-58` (`_handle_tool_error`): `adcp_error(error_code, ...)` — passthrough

The result: every code in `ERROR_CODE_MAPPING.keys()` (e.g., `PRINCIPAL_ID_MISSING`, `PRINCIPAL_NOT_FOUND`, `INVALID_DATE_RANGE`, `INVALID_PLACEMENT_IDS`, `BUDGET_CEILING_EXCEEDED`, `RATE_LIMIT_EXCEEDED`, `ADAPTER_ERROR`, etc.) survives end-to-end on the wire. They appear in the codebase at raw `Error(code=...)` and `errors=[{"code": "..."}]` literals in tools/adapters (not via `AdCPError` subclasses), so they bypass any class-level translation. Buyer agents see codes that aren't in `STANDARD_ERROR_CODES` and can't classify the recovery semantics.

**Concrete examples that ship today:**
- `src/core/tools/media_buy_list.py:106,113` emits `errors=[{"code": "PRINCIPAL_ID_MISSING"}]` and `"PRINCIPAL_NOT_FOUND"`.

ChrisHuie:
**_6. Lowercase `sync_error` reaches the wire; the compliance guard misses it via import alias_**

**File:** `src/core/tools/creatives/_processing.py:34`
```python
errors=[AdCPErrorDetail(code="sync_error", message=error_msg)]
```

**Why this is a bug:** `AdCPErrorDetail` is a local import alias for `adcp.types.generated_poc.core.error.Error`. The Pydantic field `Error.code` is `str` with no enum constraint, so non-canonical codes pass validation. `STANDARD_ERROR_CODES` is entirely UPPER_SNAKE_CASE — `"sync_error"` doesn't match anything, and nothing translates it (per B5).

The compliance guard at `tests/unit/test_architecture_error_code_compliance.py:46-72` walks the AST and matches calls whose `func.id == "Error"` (line 51-55). The aliased call `AdCPErrorDetail(...)` has `func.id == "AdCPErrorDetail"` — invisible to the visitor. That's why the guard passed despite the violation.

**Suggested fix:**
1. Change to a canonical code: `code="CREATIVE_REJECTED"` if the underlying cause is creative-side, or `code="SERVICE_UNAVAILABLE"` if it's a transport-layer adapter failure.
2. Extend `_collect_error_code_literals` in the guard to track `import … as <alias>` and `from … import Error as <alias>` so any local alias for the SDK's `Error` class is also matched. While there, also catch raw dict literals `{"code": "..."}` and `details={"error_code": "..."}` patterns — see S2.

ChrisHuie:
**_7. Five MCP/A2A wrappers silently drop accepted parameters; boundary-completeness guard is one-directional_**

| Wrapper | File:line | Param accepted but never reaches `_impl` |
|---|---|---|
| MCP `create_media_buy` | `src/core/tools/media_buy_create.py:3764, 3765` | `targeting_overlay`, `creatives` |
| A2A `create_media_buy_raw` | `src/core/tools/media_buy_create.py:3833-3856` | `budget`, `total_budget`, `product_ids`, `start_date`, `end_date`, `pacing`, `daily_budget`, `targeting_overlay`, `creatives`, `required_axe_signals`, `enable_creative_macro`, `strategy_id`, `buyer_campaign_ref` (13) |
| REST `POST /api/v1/media-buys` | `src/routes/api_v1.py:227-237` | forwards `budget`, `product_ids`, `total_budget` to A2A raw, which then drops them |
| MCP `list_authorized_properties` | `src/core/tools/properties.py:204` | `webhook_url` |
| MCP `list_creatives` | `src/core/tools/creatives/listing.py:393-394, 411` | `sort`, `pagination`, `webhook_url` |

**Verified construction patterns** (the request constructor in each wrapper omits the accepted-but-unused params):
- `media_buy_create.py:3802-3811`: `CreateMediaBuyRequest(brand=..., packages=..., start_time=..., end_time=..., po_number=..., reporting_webhook=..., context=..., ext=...)` — no `targeting_overlay`, no `creatives`.
- `properties.py:223-227`: `ListAuthorizedPropertiesRequest(publisher_domains=..., property_tags=..., context=...)` — no `webhook_url`. Then line 229: `_list_authorized_properties_impl(req, identity)`.
- `creatives/listing.py:433-456`: `_list_creatives_impl(...)` is called with `page, limit, sort_by, sort_order` — but the wrapper also exposes nested `pagination` and `sort` Pydantic objects that never reach `_impl`.

None are caught by `**kwargs`. None are translated into a derived field like `packages` or `ext`. For `create_media_buy_raw`, the docstring at `:3867-3882` markets these as "Legacy format conversion" but the conversion code is absent.

**Why CI didn't catch this:** the structural guard `tests/unit/test_architecture_boundary_completeness.py:22-46` only inspects `_impl` parameters and asserts wrappers pass them. It is blind to wrapper-side accept-and-drop. The known-violations allowlist has only 2 entries (`context_id`), confirming the one-directional design.

ChrisHuie:
**_8. 13 wrapper params declare non-Optional types but default to `None`_**

**Pattern:** `param_name: Annotated[bool, Field(description="…")] = None`

**Sites:**
- `src/core/tools/media_buy_update.py:1377-1386` (MCP `update_media_buy` wrapper) — 9 params: `paused: bool = None`, `flight_start_date: str = None`, `flight_end_date: str = None`, `budget: float = None`, `currency: str = None`, `start_time: str = None`, `end_time: str = None`, `pacing: str = None`, `daily_budget: float = None`.
- `src/core/tools/creatives/listing.py:378-391` (MCP `list_creatives`) — `media_buy_id: str = None`, `status: str = None`, `format: str = None`, `search: str = None`.
- `src/core/tools/creatives/listing.py:459-468` (A2A `list_creatives_raw`) — same pattern, plus bare `list_creatives_raw(media_buy_ids: list[str] = None, ...)` (unparameterized `list`).
- `src/core/tools/creatives/sync_wrappers.py:83` (A2A `sync_creatives_raw`) — `validation_mode: str = "strict"` (defaults to literal but type drift from MCP's `ValidationMode | None`); `assignments: dict = None`, `creative_ids: list[str] = None`.
- `src/core/tools/signals.py:362-363` (A2A `activate_signal_raw`) — `campaign_id: str = None`, `media_buy_id: str = None`.

**Why this is a bug:** FastMCP introspects the type annotation to generate JSON Schema for the MCP tool. `Annotated[bool, Field(...)]` produces `{"type": "boolean"}` — non-nullable. Strict-validating MCP clients (especially those generating typed bindings) will reject calls that omit these fields *or* pass `null`, even though the runtime accepts `None` via the default. The schema lies about nullability.

ChrisHuie:
**_9. A2A wrappers don't coerce string `brand` shorthand to `BrandReference`; identical input behaves differently by transport_**

**Files:**
- MCP coerces (correct): `src/core/tools/products.py:822-823` and `src/core/tools/media_buy_create.py:3796-3797`:
```python
if isinstance(brand, str):
    brand = BrandReference(domain=brand)
```
- A2A doesn't (the bugs): `src/core/tools/products.py:857` (`get_products_raw`) and `src/core/tools/media_buy_create.py:3833` (`create_media_buy_raw`). Both type-hint `brand: BrandReference | str | None = None` and pass `brand` straight into the SDK request constructor.

**Why this is a bug:** `BrandReference` (`adcp/types/generated_poc/core/brand_ref.py`) is `extra="forbid"` with no string-validator. Constructing it from a bare string raises `ValidationError`. So an A2A client sending `brand="acme.com"` (the documented shorthand on the MCP side) gets a 422-style error; the same payload via MCP works.

ChrisHuie:
**_10. A2A `get_products_raw` schema diverges from MCP `get_products`_**

**File:** `src/core/tools/products.py`, lines 789-803 (MCP) vs 857-867 (A2A `_raw`)

| Wrapper | Params |
|---|---|
| MCP | `brand, brief, adcp_version, filters, property_list, push_notification_config, context, ctx` |
| A2A | `brief, brand, min_exposures, filters, property_list, strategy_id, context, ctx, identity` |

**Why this is a bug:** the two transports advertise different `get_products` schemas to buyer agents, and neither forwards the params unique to its own surface — `adcp_version` and `push_notification_config` are MCP-only and never reach `_impl`; `min_exposures` and `strategy_id` are A2A-only and never reach `_impl`. The downstream call `_get_products_impl(req, identity)` builds `req` only from `brief, brand, filters, property_list, context`, so all four divergent params are dropped silently.

ChrisHuie:
**_11. Admin "Add Product" / "Edit Product" UI silently broken — server stopped emitting `fmt.type`, JS still reads it_**

**Server (PR-introduced change):**
- `src/admin/blueprints/format_search.py:59-66` (the `search_formats` JSON-builder) — emits `{agent_url, format_id, name, category, description, is_standard, dimensions}`. No `type` key.
- `src/admin/blueprints/format_search.py:137-146` (`list_all_formats`) — same.
- `git log -p main..HEAD -- src/admin/blueprints/format_search.py` confirms the PR explicitly deleted both `"type": str(fmt.type)` lines.
- This is correct on the server side because `Format.type` was removed in adcp 3.12.

**Client (unchanged in this PR):**
- `static/js/format-template-picker.js:779`: `` const typeStr = fmt.type ? ` [${fmt.type}]` : ''; ``
- `static/js/format-template-picker.js:786`: `type: fmt.type,`
- `static/js/format-template-picker.js:1024`: `const formatType = formatData.type || '';` — falls back to `''`, so subsequent `formatType === 'display'` always evaluates false.
- `static/js/format-utils.js:137`: `const type = (fmt.type || '').toLowerCase();`

**User-visible failure:** these JS files are loaded by `templates/{add_product,edit_product,add_product_gam,add_product_mock,edit_product_mock}.html`. After picking a format in the Add Product / Edit Product flow, the dimensions / duration parameter inputs (gated on `formatType === 'display' || formatType === 'video'`) **never auto-show**. Users can't enter the parameters; can't submit a complete product form. There is no automated test exercising this path.

KonstantinMirin:
## Update: merged latest main, all pre-existing failures confirmed

Merged `main` (c05aa707) into feature branch. 26 conflicts resolved — our 4.3 migration supersedes main's 3.12 migration.

### Test results comparison (branch vs main)

| Suite | Main | Branch | Delta |
|-------|------|--------|-------|
| unit | 4435 passed, 0 failed | 4546 passed, 0 failed | **+111 tests** (new structural guards) |
| integration | 1880 passed, 0 failed | 1880 passed, 0 failed | Same |
| e2e | 91 passed, 0 failed | 87 passed, 0 failed | -4 tests (consolidated) |
| bdd | 935 passed, **84 failed** | 935 passed, **84 failed** | **Same failures** |
| admin | 62 passed, **5 failed, 14 error** | 62 passed, **5 failed, 14 error** | **Same failures** |

**All 84 BDD and 19 admin failures are pre-existing on main** — verified by running `./run_all_tests.sh` on a fresh main worktree (c05aa707) with clean Docker. The branch introduces zero new failures.

Root causes of pre-existing failures:
- 44 BDD: `push_notification_configs` table missing in BDD test DB (migration not applied in Docker BDD env)
- 4 BDD: `_resolve_media_buy_id` undefined (incomplete step definition on main)
- Remaining BDD: delivery breakdown assertions, account serialization gaps
- Admin: auth redirect issues (302 vs 200) in new admin test coverage from #1222

ChrisHuie (approved):

PR #1273, opened 2026-05-04
refactor: reconcile MockProductConfig — merge to 12-field percent sch…

### LINKED ISSUE #1267: refactor: reconcile MockProductConfig — merge MockImplementationConfig and wire mock adapter
Sub-issue of #1239 (Phase 2 of #996). Independent — can be done in any order with #1239-B and #1239-C.

## Scope

- Merge `MockImplementationConfig` (13 fields, percents 0-100) into `MockProductConfig` (currently 5 fields, fractions 0-1). The reconciled schema uses percents to match the implementation config.
- Delete `src/adapters/mock_implementation_config_schema.py`
- Replace inline dict construction in `src/adapters/mock_ad_server.py` (lines 1382-1401) with `MockProductConfig(**form_fields).model_dump()`
- Drop legacy `validate_product_config()` — Pydantic field validators replace it

## Acceptance criteria

- [ ] `MockProductConfig` has all 13 fields (percents), no separate `MockImplementationConfig` file
- [ ] Mock adapter read sites use the reconciled schema, not raw dict access
- [ ] Existing tests in `tests/unit/adapters/test_schemas.py` and `tests/unit/admin/test_adapter_blueprints.py` updated to reflect new shape
- [ ] Round-trip test: `MockProductConfig.model_validate(dict).model_dump()` round-trips cleanly
- [ ] `make quality` clean

## Related
- Parent: #1239
- Phase 1 PR: #1007

jrosendahl-opt:
…ema with Pydantic validation

Replaces the 5-field fraction-based MockProductConfig (0–1 rates) with the 12-field percent-based schema (0–100) that matches what the admin form already stores in implementation_config. Removes the broken validate_product_config() method (returned a tuple but was used as a bool, so it always silently passed). POST handler now constructs and validates via Pydantic, raising ValidationError or ValueError on bad input and rendering the error back to the template.

Closes #1267

### DIFF
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -1,7 +1,7 @@
 import logging
 import random
 from datetime import UTC, datetime
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 logger = logging.getLogger(__name__)
 
@@ -63,7 +63,7 @@ class MockProductConfig(BaseProductConfig):
     viewability_rate: float = Field(default=70.0, ge=0.0, le=100.0)
     latency_ms: int = Field(default=50, ge=0)
     error_rate: float = Field(default=0.1, ge=0.0, le=100.0)
-    test_mode: str = Field(default="normal")
+    test_mode: Literal["normal", "high_demand", "degraded", "outage"] = Field(default="normal")
     price_variance: float = Field(default=10.0, ge=0.0)
     seasonal_factor: float = Field(default=1.0, ge=0.0)
     verbose_logging: bool = Field(default=False)
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -41,13 +41,23 @@ class MockConnectionConfig(BaseConnectionConfig):
 
 
 class DeliverySimulationConfig(BaseModel):
-    """Nested config for delivery simulation timing in MockProductConfig."""
+    """Nested config for delivery simulation timing in MockProductConfig.
+
+    Upper bounds mirror the form validator in
+    ``src/admin/blueprints/adapters.py``:
+      - ``time_acceleration`` capped at 86400 (24 hours' worth of seconds —
+        beyond which "accelerated simulation" loses meaning).
+      - ``update_interval_seconds`` capped at 60 (one minute — beyond which
+        the UI update cadence becomes irrelevant for delivery polling).
+    Schema-side and form-side bounds stay in lockstep so REST + UI paths
+    accept the same values.
+    """
 
     model_config = ConfigDict(extra="forbid")
 
     enabled: bool = Field(default=False)
-    time_acceleration: int = Field(default=3600, ge=1)
-    update_interval_seconds: float = Field(default=1.0, ge=0.1)
+    time_acceleration: int = Field(default=3600, ge=1, le=86400)
+    update_interval_seconds: float = Field(default=1.0, ge=0.1, le=60.0)
 
 
 class MockProductConfig(BaseProductConfig):
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from src.core.schemas import Snapshot
 
 # adcp 3.6.0: BrandManifest removed from CreateMediaBuyRequest; now uses BrandReference (brand field)
-from pydantic import Field
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
 
 from src.adapters.base import (
     AdapterCapabilities,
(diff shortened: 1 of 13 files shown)


jrosendahl-opt:
I have read the IPR Policy

src/adapters/mock_ad_server.py
```diff
@@ -40,14 +40,35 @@ class MockConnectionConfig(BaseConnectionConfig):
...
-    ctr: float = Field(default=0.02, ge=0.0, le=1.0)
-    viewability: float = Field(default=0.65, ge=0.0, le=1.0)
-    scenario: str = Field(default="normal")
+    daily_impressions: int = Field(default=100000, ge=0)
+    fill_rate: float = Field(default=85.0, ge=0.0, le=100.0)
+    ctr: float = Field(default=0.5, ge=0.0, le=100.0)
+    viewability_rate: float = Field(default=70.0, ge=0.0, le=100.0)
+    latency_ms: int = Field(default=50, ge=0)
+    error_rate: float = Field(default=0.1, ge=0.0, le=100.0)
+    test_mode: str = Field(default="normal")
```
ChrisHuie: so this passes through validation can you change this line to ..

test_mode: Literal["normal", "high_demand", "degraded", "outage"] = Field(default="normal")

src/adapters/mock_ad_server.py
ChrisHuie, line 5: need this import too..


`from typing import Literal`

ChrisHuie:
@jrosendahl-opt

ChrisHuie:
## State as of head `d794ad1ea`

### Konstantine 2026-05-05 review (verified at this head)

All three items resolved by removal/extraction:

- **R1 — `test_mode: Literal[...]`** — `src/adapters/mock_ad_server.py:66` now `Literal["normal", "high_demand", "degraded", "outage"]` per your inline comment. Plus `from typing import Literal` import added.
- **Derived validator** — `src/admin/blueprints/adapters.py:78` now derives `valid_modes` from `get_args(MockProductConfig.model_fields["test_mode"].annotation)` — single source of truth.
- **Test fixtures aligned** — `test_mode="stress"` (rejected by the new Literal) replaced with `"high_demand"` across two test files.

### Current CI failures — all upstream / blocked

Verified observation, not inference:

- **Unit Tests** (×2 subtests on `test_pydantic_schema_alignment` for `get-products-request.json`): `if_wholesale_feed_version` is now in the live AdCP schema but not in `KNOWN_SCHEMA_LIBRARY_MISMATCHES`. **Already fixed in PR #1338** (commit `1d9b37f06`), which adds the allowlist entry. This PR will inherit the fix when rebased on main after #1338 lands.

- **E2E `test_offline_mode`**: same root cause — `protocol-envelope.json` requires `status` field. Also covered by PR #1338's xfail decorator on the same test.

- **Integration (infra) — `test_mcp_tool_roundtrip_minimal`**: `MCP server failed to start on port 50809 within 20s` — server logs show it WAS binding (`Uvicorn running on :50809`) ~1.27s after first stderr; race condition between server-ready and the 300ms socket-poll cadence. Subsequent tests in same file PASSED with the same fixture. Consistent with CI flakiness but not exhaustively verified across other PRs.

- **Security Audit**: pre-existing pydantic-ai 1.56 vuln (GHSA-cqp8-fcvh-x7r3). Covered by #1339 (already MERGEABLE 0-fail).

### Path forward
1. Wait for #1338 to merge (schema-drift fix + offline-mode xfail)
2. Rebase this PR on the new main — per cross-PR analysis, the rebase is structurally clean (only `src/adapters/mock_ad_server.py` overlaps with main, and `git merge-tree` produced no conflicts; main's `buyer_ref` removal touches different lines than this PR's MockProductConfig reshape)
3. Re-run CI to see if MCP timeout was flaky vs persistent

ChrisHuie:
Had to close and reopen because github actions weren't kicking off and just frozen

KonstantinMirin:
Took a careful pass through this one. The schema work is solid — merging to the percent scale, dropping `validate_product_config()` (which was returning a tuple and being truthy-checked, so it never actually validated — nice catch), the `Literal` `test_mode`, and the round-trip coverage all look right. Confirmed the 0–1 → 0–100 flip is safe since those rate fields are write-only and never read back by the simulation.

One thing I'd flag before merge, and I think it traces back to how #1267 was scoped rather than anything in the implementation. The ticket pointed the work at `mock_ad_server.py:1382-1401`, so that's where the new Pydantic validation landed. But when I traced the live route wiring, that handler turns out to be unreachable in the assembled app:

- `MockAdServer.register_ui_routes()` is only called after instantiating `MockAdServer(principal=None, dry_run=False)` in `app.py:412` — but `__init__` requires `config` as its first arg, so that throws `TypeError` and gets swallowed at `app.py:422`. The routes never register.
- Even if they did, `adapters_bp` is registered first (`app.py:348`), so Werkzeug matches the blueprint for that identical URL anyway.

So the URL `/adapters/mock/config/<tenant>/<product>` is actually served by `adapters.py:mock_config`, which still uses the old hand-rolled `parse_int`/`parse_float`. The new test passes because it registers the rewritten handler in isolation with a mock config. Net effect: the validation improvement isn't reaching production yet, and the two handlers have drifted on bounds (`latency_ms`, `price_variance`, `seasonal_factor` are capped in the live one but open-ended in the schema).

I think one change closes the loop cleanly: move the Pydantic validation into the live `adapters.py` blueprint handler (a `MockProductConfig.from_form()` classmethod would do it), delete the dead copy in the adapter, fold the three numeric bounds into the model, and point the test at the real route. That also retires the duplicate-route situation, which predates this PR. Happy to pair on it or pick it up if useful — and we may want a quick follow-up note on #1267 since the original scope pointed at the wrong handler.

Issue #1073, opened 2026-03-01
feat: Support buying_mode refine for iterative proposal refinement


bokelley:
## Summary

Implement `buying_mode: refine` in `get_products`, enabling buyers to iterate on proposals by sending typed change requests (include/omit/modify products and proposals).

**AdCP version:** v3.0.0-rc1
**Blocked on:** adcp Python client shipping the `refine` field on `GetProductsRequest` (currently in spec but noted as schema-library mismatch in PR #1071)

### Three buying modes (RC1)
1. **`brief`** — Natural language discovery ("I want video ads targeting millennials")
2. **`wholesale`** — Bulk/programmatic with explicit filters
3. **`refine`** — Iterate on a previous response, requesting changes

### Refine Request
```json
{
  "buying_mode": "refine",
  "product_ids": ["prod_1", "prod_2"],
  "proposal_id": "prop_abc",
  "refine": [
    {"scope": "request", "ask": "more video, less display"},
    {"scope": "product", "id": "prod_1", "action": "include", "ask": "add 16:9 format"},
    {"scope": "product", "id": "prod_3", "action": "omit"},
    {"scope": "proposal", "id": "prop_abc", "action": "include", "ask": "shift 20% to video"}
  ]
}
```

Constraints:
- `product_ids` and/or `proposal_id` required in refine mode
- `product_ids` and `proposal_id` forbidden in brief/wholesale modes
- No duplicate product/proposal IDs in a single refine array

### Refine Response
```json
{
  "products": [...],
  "proposal_id": "prop_def",
  "refinement_applied": [
    {"scope": "request", "status": "applied", "notes": "Added 3 video products."},
    {"scope": "product", "id": "prod_1", "status": "partial", "notes": "16:9 not available — returning 4:3 and 1:1"},
    {"scope": "product", "id": "prod_3", "status": "applied"},
    {"scope": "proposal", "id": "prop_abc", "status": "applied", "notes": "Shifted 22% to video"}
  ]
}
```

Each refinement has status: `applied | partial | unable`.

### 1. Proposal storage
Need to persist proposals so they can be referenced in refine requests:
- Store proposal snapshots (product set + configuration) keyed by `proposal_id`
- TTL-based expiration (return `PROPOSAL_EXPIRED` error code when stale)

### 2. Refine logic in get_products
```python
def _get_products_impl(request, identity):
    if request.buying_mode == "refine":
        # Load previous proposal
        proposal = load_proposal(request.proposal_id)
[...]

PR #1274, opened 2026-05-05
feat: wire buying_mode and refine through get_products (#1272)(PR 3 after error-emission)

### LINKED ISSUE #1272: feat: wire buying_mode and refine through get_products and storyboard refine_products compliance
From #1247                                                                                                                                                                                  
                                      
  AdCP defines get_products as a buyer-intent-tagged operation with three modes — brief, wholesale, refine. The salesagent currently parses requests with all three modes (after a str | None     
  widening of buying_mode in our local schema) but discards both buying_mode and refine at every layer downstream. The AdCP media_buy_seller/refine_products storyboard scenario from
  @adcp/sdk@6.11.0 (targeting spec 3.0.6) cannot pass.                                                                                                                                            
                                                                                                                                                                                                
  This is the protocol-compliance wedge of the larger feature tracked in #1073. #1073 covers proposal persistence, intelligent refinement application, and the finalize HITL flow. This issue     
  covers: request surface, cross-mode validation, response shape, mode-specific behavior, the brief_relevance plumbing the brief-mode contract requires, and a refine response that is
  schema-valid against get-products-response.json 3.0.6 with all entries reporting status="unable" until #1073 lands.                                                                             
                                                                                                                                                                                                
  Gap (verified)                      

  Wire-up is missing at every layer                                                                                                                                                               
  
  - MCP wrapper src/core/tools/products.py:789-845 — signature does not accept buying_mode or refine.                                                                                             
  - A2A handler src/a2a_server/adcp_a2a_server.py:1396-1419 — extracts only brief, brand, filters from the JSON-RPC dict.                                                                       
  - A2A raw src/core/tools/products.py:848+ — same surface gap.                                                                                                                                   
  - REST body src/routes/api_v1.py:66-71 — GetProductsBody has neither field.                                                                                                                     
  - Helper src/core/schema_helpers.py:99-139 — create_get_products_request() accepts neither.                                                                                                     
  - _get_products_impl src/core/tools/products.py:145-786 — zero references to req.buying_mode or req.refine. Same code path for all modes.                                                       
  - Response — GetProductsResponse inherits refinement_applied from the library but is never populated.                                                                                           
                                                                                                                                                                                                  
  Cross-mode validation is not enforced                                                                                                                                                           
                                                                                                                                                                                                  
  The library's three request variants encode mode-specific rules via Literal types on buying_mode. Our local GetProductsRequest (src/core/schemas/product.py:224-249) inherits only              
  GetProductsWholesaleRequest and widens buying_mode to str | None, collapsing the discriminated union. None of the seven cross-mode rules at                                                   
  tests/bdd/features/BR-UC-001-discover-available-inventory.feature:313-319 are enforced today.                                                                                                   
                                                                                                                                                                                                
  Pre-v3 client default is undefined                                                                                                                                                              
  
  Spec: "Sellers receiving requests from pre-v3 clients without buying_mode SHOULD default to brief. v3 clients MUST include buying_mode." No code path implements either half today.             
                                                                                                                                                                                                
  brief_relevance is produced internally and thrown away                                                                                                                                          
                                                                                                                                                                                                
  The AI ranker at src/core/tools/products.py:660-714 returns ProductRanking.reason (a per-product brief-match explanation) for every product but uses it only to sort. The 3.0.6 Product schema  
  defines brief_relevance: string ("Explanation of why this product matches the brief (only included when brief is provided)"). Our local Product schema does not surface it on output products,
  even though the value already exists.                                                                                                                                                           
                                                                                                                                                                                                
  Refine entry field-name skew between salesagent and storyboard                                                                                                                                  
  
  ┌──────────────────────────────────────────────────────────────────────────────────┬───────┬───────────────────────────────┬────────────────────────────────┐                                   
  │                                      Source                                      │ Spec  │ Product-scope refine id field │ Proposal-scope refine id field │                                 
  ├──────────────────────────────────────────────────────────────────────────────────┼───────┼───────────────────────────────┼────────────────────────────────┤                                   
  │ Salesagent installed adcp Python lib 3.12.0 (get_adcp_version() == "3.0.0-rc.3") │ rc.3  │ id                            │ id                             │                                 
  ├──────────────────────────────────────────────────────────────────────────────────┼───────┼───────────────────────────────┼────────────────────────────────┤
  │ Storyboard @adcp/sdk 6.11.0 (released spec 3.0.6)                                │ 3.0.6 │ product_id                    │ proposal_id                    │                                   
  └──────────────────────────────────────────────────────────────────────────────────┴───────┴───────────────────────────────┴────────────────────────────────┘                                   
                                                                                                                                                                                                  
  The Python library's Refine1 and Refine2 declare extra='forbid'. Storyboard requests carrying product_id/proposal_id will be rejected at Pydantic parse time, even after the wire-up changes    
  above. This must be fixed in the same change.                                                                                                                                                 
                                                                                                                                                                                                  
  Root cause                                                                                                                                                                                    
                                      
  The local schema collapsed the AdCP discriminated union into a single class to avoid forcing callers into a specific variant. That decision was correct for parser permissiveness but left three
   follow-ups undone: (1) the wrappers never gained the new fields, (2) the cross-mode invariants the union encoded structurally are not enforced anywhere else, and (3) the spec-version skew
  between the installed Python library (rc.3) and the released storyboard target (3.0.6) was not bridged. Adding fields to wrappers without enforcing the invariants would create a path where    
  buying_mode=wholesale + refine=[…] reaches _impl — a worse state than today. Bridging the version skew without invariant enforcement would still let malformed multi-mode requests through.   
                                      
  Contract

  After this issue closes, get_products MUST satisfy the following at the MCP, A2A, and REST transports:                                                                                          
  
  Wire-up                                                                                                                                                                                         
                                                                                                                                                                                                
  1. The MCP wrapper, A2A handler, A2A raw function, REST body model, and create_get_products_request() all accept and forward buying_mode: str | None and refine: list[Refine | Refine1 |        
  Refine2] | None.                                                                                                                                                                              
  2. _get_products_impl reads req.buying_mode and req.refine and branches on them.                                                                                                                
  3. The structural guard test_architecture_boundary_completeness.py remains green — every wrapper forwards every _impl parameter.                                                                
                                                                                                                                                                                                  
  Mode resolution                                                                                                                                                                                 
                                                                                                                                                                                                  
  4. When buying_mode is absent and adcp_version < 3.0, the request is treated as buying_mode="brief".                                                                                            
  5. When buying_mode is absent and adcp_version >= 3.0, the request is rejected with AdCPValidationError (UPPER_SNAKE error code, recovery="correctable").
  6. When buying_mode is present but not in {brief, wholesale, refine}, the request is rejected with AdCPValidationError.                                                                         
                                                                                                                                                                                                  
  Cross-mode validation                                                                                                                                                                           
                                                                                                                                                                                                  
  For each rule, on violation, raise AdCPValidationError with UPPER_SNAKE error code and a message that names both the offending field and the active mode:                                       
                                      
  7. buying_mode=brief requires non-empty brief; refine MUST be absent.                                                                                                                           
  8. buying_mode=wholesale: brief MUST be absent; refine MUST be absent.                                                                                                                        
  9. buying_mode=refine: brief MUST be absent; refine MUST be a non-empty array.                                                                                                                  
  10. Each entry in refine must conform to the spec 3.0.6 shape (see field-name compatibility below).                                                                                             
                                                                                                                                                                                                  
  Field-name compatibility (rc.3 ↔ 3.0.6)                                                                                                                                                         
                                                                                                                                                                                                  
  11. Refine entries with scope=product MUST accept product_id on the wire and id on the wire interchangeably and resolve to a single internal field.                                             
  12. Refine entries with scope=proposal MUST accept proposal_id on the wire and id on the wire interchangeably and resolve to a single internal field.                                         
  13. The compatibility shim is documented as temporary and removable once the installed adcp Python library targets spec 3.0.6 or later.                                                         
                                                                                                                                                                                                  
  Mode-specific behavior                                                                                                                                                                          
                                                                                                                                                                                                  
  14. brief mode: products are AI-ranked and ordered by the existing ranker's score (the score is internal-only and not surfaced on the response). Each product in the response includes          
  brief_relevance: string populated from the ranker's ProductRanking.reason for that product. proposals is omitted from the response (separate issue covers proposal generation).
  refinement_applied is absent.                                                                                                                                                                   
  15. wholesale mode: products are returned in catalog order; the AI ranker is bypassed; no product has brief_relevance; proposals is absent; refinement_applied is absent.                     
  16. refine mode: until #1073 lands, the response MUST contain a valid products array (returning the same products that the equivalent brief-mode call would return for this account is          
  acceptable) and a refinement_applied array of the same length and order as the request's refine array. Every entry has status="unable", echoes the request entry's scope and (for               
  product/proposal scopes) the matching id field per spec 3.0.6, and includes a notes field naming #1073 as the tracking issue. The response validates against                                    
  media-buy/get-products-response.json from spec 3.0.6.                                                                                                                                           
                                                                                                                                                                                                
  Validation enforcement layer        

  17. Cross-mode validation is enforced via a @model_validator(mode="after") on the local GetProductsRequest subclass (or an equivalent enforcement layer that runs before _impl), so all three   
  transports get the same error behavior automatically.
                                                                                                                                                                                                  
  Local BDD scenario alignment                                                                                                                                                                  
                                      
  18. The brief-mode BDD scenario assertion "products ordered by relevance_score descending" at tests/bdd/features/BR-UC-001-discover-available-inventory.feature:43 is updated to "products are  
  ordered when buying_mode is brief" (no public field reference), since relevance_score is not in the spec.
                                                                                                                                                                                                  
  Acceptance criteria                                                                                                                                                                           
                                      
  - The AdCP storyboard media_buy_seller/refine_products scenario passes against the salesagent with the test-kit acme-outdoor.yaml, validating both phases.brief.steps.get_products_brief        
  (response_schema + products + products[0].product_id) and phases.refine.steps.get_products_refine (response_schema + products).
  - BDD scenarios @T-UC-001-main, @T-UC-001-alt-wholesale, and @T-UC-001-alt-refine (tests/bdd/features/BR-UC-001-discover-available-inventory.feature:29-82) pass at all four transports (IMPL,  
  A2A, MCP, REST).                                                                                                                                                                                
  - The seven cross-mode rule rows at BR-UC-001-discover-available-inventory.feature:313-319 pass.
  - An obligation entry replaces the obligation_id: null, status: new stubs at docs/test-obligations/bdd-traceability.yaml:37-42.                                                                 
  - A new step file (or addition to tests/bdd/steps/domain/uc_get_products_inventory.py) covers the new scenarios with Covers: tags pointing at the new obligation.                               
  - make quality is green; no new entries appear on any structural-guard allowlist.                                                                                                               
                                                                                                                                                                                                  
  Test surface                                                                                                                                                                                    
                                                                                                                                                                                                  
  - Unit: each cross-mode validation rule (10 cases including 3 happy-path modes) on the request validator.                                                                                       
  - Unit: refine entry field-name compatibility — both product_id and id accepted for product-scope; both proposal_id and id accepted for proposal-scope; conflicting values rejected.          
  - Unit: _get_products_impl mode-branching: brief → ranker called and brief_relevance populated on every product; wholesale → ranker not called and brief_relevance absent; refine →             
  refinement_applied populated with status="unable" and matched positionally.                                                                                                                     
  - Integration: end-to-end through ProductEnv harness for each mode and each transport.                                                                                                          
  - BDD: the scenarios above, auto-parametrized across Transport.IMPL/A2A/MCP/REST.                                                                                                               
  - Compliance: a CI smoke test that runs the media_buy_seller/refine_products storyboard against a docker-compose tenant.                                                                        
                                                                                                                                                                                                  
  Technical relationships                                                                                                                                                                         
                                                                                                                                                                                                  
  - Identity flow is symmetric across transports — no transport-specific identity work is needed.                                                                                                 
  - Pydantic smart-union handles dict-to-Refine-variant coercion automatically because each variant declares scope as a Literal const. The field-name compatibility shim must run before union  
  routing (e.g., a model_validator(mode="before") that normalizes product_id→id and proposal_id→id on inbound entries).                                                                           
  - Version compat (src/core/version_compat.py:42-51) is response-side only and operates on pricing_options. It must not touch buying_mode, refine, refinement_applied, or brief_relevance (these
  are pure v3 fields).                                                                                                                                                                            
  - AI ranker output reuse: ProductRanking.reason (src/services/ai/ranking_agent.py:15-22) already exists per product; surfacing it as brief_relevance on the response Product is a single mapping
   step in _impl, not new ranker work.                                                                                                                                                            
  - HITL workflow (WorkflowStep table at src/core/database/models.py:1692-1736, pattern at src/core/tools/media_buy_create.py:1995-2049) is reused by #1073 for the proposal finalize action — not
   by this issue.                                                                                                                                                                                 
  - Context table (src/core/database/models.py:1654-1689) is the natural home for the prior-state persistence #1073 will add — not used in this issue.                                          
  - Spec version skew is broader than refine: the adcp Python lib targets rc.3, the storyboard targets 3.0.6. This issue addresses the refine-specific divergence; other field renames between    
  rc.3 and 3.0.6 are out of scope and may need their own audit.                                                                                                                                   
                                                                                                                                                                                                  
  Out of scope (separate issues)                                                                                                                                                                  
                                                                                                                                                                                                
  - Proposal persistence, proposal_id generation, proposal TTL, PROPOSAL_EXPIRED error code, intelligent refine application — tracked in #1073.                                                   
  - The finalize action on proposal-scoped refine entries — tracked in #1073.
  - proposals array generation in brief-mode responses — depends on the proposal entity introduced in #1073.                                                                                      
  - Pagination — required by @T-UC-001-alt-paginated*; orthogonal gap, no implementation today.                                                                                                   
  - Error-code UPPER_SNAKE migration for existing errors — tracked as #1247 item 9. New errors introduced by this issue MUST use UPPER_SNAKE to avoid adding to the debt.                         
  - Broader rc.3 ↔ 3.0.6 field-rename audit beyond the refine entries handled here — should be a separate issue if/when the storyboard fails on other renames.                                    
                                                                                                                                                                                                  
  References                                                                                                                                                                                      
                                                                                                                                                                                                  
  - Storyboard gap analysis umbrella: #1247 item 5                                                                                                                                                
  - Full feature umbrella: #1073                                                                                                                                                                
  - Storyboard scenario: @adcp/sdk@6.11.0 compliance/cache/3.0.6/protocols/media-buy/scenarios/refine_products.yaml                                                                               
  - Spec 3.0.6 request schema: media-buy/get-products-request.json ($id: /schemas/3.0.6/media-buy/get-products-request.json)                                                                      
  - Spec 3.0.6 response schema: media-buy/get-products-response.json                                                                                                                              
  - adcp Python lib (3.12.0, spec target 3.0.0-rc.3, adopted in #1217): adcp.types.BuyingMode, adcp.types.Refine/Refine1/Refine2, adcp.types.RefinementAppliedItem                                
  - Local BDD feature: tests/bdd/features/BR-UC-001-discover-available-inventory.feature                                                                                                          
  - Local traceability stub: docs/test-obligations/bdd-traceability.yaml:37-42

ChrisHuie:
Wires two stacked capabilities on top of #1389 (property-list targeting): the AdCP 3.0 three-mode `get_products` contract, and `sync_accounts` idempotency on a shared verbatim-replay engine. Closes the protocol-compliance wedge of #1073 and unblocks the `media_buy_seller/refine_products` storyboard (#1247 item 5).

**Stacked PR** — branches on #1389; **merge after it**. buying_mode threads `property_list`/`context` through the shared `get_products` surface, so it depends on #1389's request surface. (Base targets `feature/property-list-targeting`; auto-retargets to `main` when #1389 merges.)

## Part 1 — buying_mode / refine wire-up

Three-mode contract (`brief` / `wholesale` / `refine`) end-to-end through `get_products` across MCP, A2A, and REST. Request-surface conformance only — proposal persistence + the `finalize` HITL flow remain in #1073.

- All four entry points (MCP wrapper, A2A skill handler, A2A raw, REST body) accept/forward `buying_mode` + `refine`; `create_get_products_request()` forwards both.
- **Mode resolution:** absent + `adcp_version < 3.0` → defaulted to `brief` (audit `defaulted_to_brief: true`); absent + v3 → `VALIDATION_ERROR`; value outside `{brief, wholesale, refine}` → rejected.
- **Cross-mode invariants** (`model_validator(mode="after")`): brief requires a brief / forbids refine; wholesale forbids both; refine forbids brief / requires the array.
- **Per-mode `_impl`:** brief runs the AI ranker + plumbs `brief_relevance`; wholesale and refine bypass it; refine returns `refinement_applied` (every entry `status="unable"` until #1073).
- Audit records `buying_mode`, `refine_count`, `defaulted_to_brief`. Enum normalization routes through the canonical `enum_helpers.enum_value()`.

## Part 2 — sync_accounts idempotency (shared verbatim-replay engine)

Extracts create_media_buy's verbatim success-cache replay into a reusable, policy-parameterized engine (`src/services/idempotency_replay.py`) and wires `sync_accounts` onto it.

- `SyncAccountsResponse` gains the top-level `replayed: true` marker — emitted only on a verbatim replay, omitted otherwise so fresh responses stay byte-identical.
- Request hashing consolidated to `idempotency_canonical.request_hash_for()`; log sanitization via `log_safety.loggable()` (CWE-117).
- create_media_buy's idempotency reconciled onto the shared engine: the approval-pending `submitted` variant is cacheable and shape-replayable (security.mdx idempotency rule 2 — sync-success and async-`submitted` branches). The `IDEMPOTENCY_EXPIRED` / `IDEMPOTENCY_CONFLICT` / `SERVICE_UNAVAILABLE` taxonomy is preserved, including the EXPIRED natural-key recovery suggestion.

## Testing

- **buying_mode:** schema validators, mode branching, rc.3↔released wire-compat, cross-transport integration (each mode × IMPL/A2A/MCP/REST), BDD (`@T-UC-001-*`).
- **idempotency:** shared-engine unit, `test_sync_accounts_idempotency.py` (verbatim replay/conflict/race per transport + cross-tool isolation), submitted-replay oracle.
- Full local suite (`./run_all_tests.sh ci`) green across unit / integration / e2e / admin / bdd / ui.

## Obligation traceability

`UC-001-MODE-{BRIEF,WHOLESALE,REFINE,VALIDATION}-01` in `docs/test-obligations/UC-001-discover-available-inventory.md`; bdd-traceability rows mapped.


KonstantinMirin (commented):
Reviewed the wedge approach is right and the test surface is real (Pydantic ValidationError matches against real validators, real `_impl` execution, BDD asserts on harness output — not mock-echoes). Three findings before merge:

## 1. Move wire-shape transformation out of `GetProductsResponse.model_dump` (required)

`src/core/schemas/product.py:598-625` overrides `model_dump` to do **wire-format transformation** (rc.3 `id` → 3.0.6 `product_id`/`proposal_id` rename based on scope), not nested serialization. Three concrete problems:

- **Reverse layering.** The schema's *parsed* shape uses `id`; its *serialized* shape uses `product_id`/`proposal_id`. A reader has to inspect `model_dump` to know the wire shape.
- **Any non-wire caller gets the renamed shape.** Tests, logs, snapshot fixtures, internal diagnostics — every `response.model_dump()` call gets wire-renamed dicts. The schema layer can't tell the wire layer's transformation apart from a generic dump-to-dict.
- **The inbound `model_validator(mode="before")` at `product.py:560-596` exists solely to undo the outbound rename when test harnesses round-trip the artifact** (per its own docstring: "A2A test harness parses the artifact dict back into the response model"). That's production schema code carrying shims for test harnesses — exactly the symptom of putting transformation in the wrong layer.

CLAUDE.md's CP-4 endorses `model_dump` overrides for **nested serialization** of child models — a Pydantic v2 workaround. This rename is a different category of work and belongs at the transport boundary where each transport already calls `response.model_dump(mode="json")`.

**Fix**: extract a helper:

```python
# src/core/wire_compat.py
def rename_refinement_applied_for_wire(response_dict: dict) -> dict:
    """Apply rc.3 -> 3.0.6 wire rename to refinement_applied items.

    Removable when the installed adcp library targets 3.0.6+; detected by
    test_architecture_adcp_library_field_skew.
    """
    applied = response_dict.get("refinement_applied")
    if not applied:
        return response_dict
    for item in applied:
        if not isinstance(item, dict):
            continue
        scope = item.get("scope")
        entry_id = item.pop("id", None)
        if entry_id is None:
            continue
        if scope == "product":
            item["product_id"] = entry_id
        elif scope == "proposal":
            item["proposal_id"] = entry_id
        else:
            item["id"] = entry_id
    return response_dict
```

Call it at each transport right after `model_dump(mode="json")` (MCP wrapper, A2A `get_products_raw`, REST handler). The inbound `_normalize_refinement_applied_inbound` validator becomes deletable. The fitness function (`test_architecture_adcp_library_field_skew`) keeps working unchanged — it gates on `Refine1.model_fields`, independent of where the rename lives.

This also isolates the wire-compat surface in one file, making it grep-discoverable when other rc.3↔3.0.6 renames surface (the issue body explicitly notes this is the first of likely many).

## 2. Pre-v3 default deviates from spec — make the deviation deliberate

`src/core/product_conversion.py:300-327` (`resolve_pre_v3_buying_mode`) defaults to **either** `"brief"` (when brief is non-empty) **or** `"wholesale"` (when no brief is supplied). AdCP 3.0 spec text:

> Sellers receiving requests from pre-v3 clients without `buying_mode` SHOULD default to `"brief"`.

The spec doesn't say "default to wholesale when no brief". The PR's logic is operationally defensible — a v2 client sending `{brand: ..., filters: ...}` with no brief would otherwise fail the brief-mode validator, breaking backward compat. The audit log records `pre_v3_defaulted=True` for both paths, so the deviation is observable.

This is fine as a deliberate choice. Worth one comment in the function naming this as a *deliberate deviation from the spec's "SHOULD default to brief" language, accepted to preserve v2 client behavior*. Six months from now no one will remember the rationale without the comment, and the next person reading the spec next to the code will think it's a bug.

## 3. Defensive `or "brief"` masks a validator-bypass bug

`src/core/tools/products.py:718`:

```python
raw_mode: Any = req.buying_mode
mode: str = (raw_mode.value if hasattr(raw_mode, "value") else raw_mode) or "brief"
```

The comment above admits the validator should have caught a `None` here ("Default to 'brief' defensively if validator was bypassed (it shouldn't be in normal flows)"). If the validator is bypassed in normal flow, that's a bug — silently defaulting to `brief` makes it invisible. Two safer options:

(a) Raise: `if mode is None: raise AdCPValidationError("buying_mode missing — schema validator was bypassed")`. Hard-fails on a state that shouldn't occur.
(b) Log a warning then default. Lets traffic through but records the fact.

Recommend (a). Schema validators don't get bypassed in the wild; if one does, we want to know loud and fast.

---

After (1) lands, ready to merge. (2) and (3) are small.

KonstantinMirin (changes requested):
The mode branching logic in `_get_products_impl` is solid — brief/wholesale/refine gating, brief_relevance, refinement_applied are all well-structured and transport-agnostic. The test coverage at the unit layer is strong too. But the wrapper layer needs work.

**Business logic leaked into transport wrappers:**

`resolve_pre_v3_buying_mode` → `create_get_products_request` → `_impl(pre_v3_defaulted=...)` is copy-pasted across MCP, A2A, and REST. The A2A wrapper even says "mirrors MCP wrapper" in its comment — that's the red flag right there. Version-based defaulting is a business decision, not a transport concern. It should live in `create_get_products_request` (accept `adcp_version`, return the flag) or in `_impl` itself. Three copies means three chances to diverge, and they already have:

- **A2A doesn't catch `ValidationError`** from request construction. MCP wraps it as `AdCPValidationError` (400). REST returns 400. A2A lets it bubble to the generic `Exception` handler → `InternalError` (500). So `buying_mode="brief"` without a brief gives the caller a 400 from MCP/REST and a 500 from A2A. Transport parity is broken for the exact feature this PR introduces.

- **REST builds a hand-crafted error dict** for `ValidationError` (`{"error_code": "VALIDATION_ERROR", ...}`) while `_handle_tool_error` in the same file uses `adcp_error()` which returns a different shape (`{"errors": [...]}`). Two incompatible error formats from the same endpoint.

- **A2A normalizes brief** (`brief or ""`) before passing to the helper, MCP doesn't. Small, but it's the kind of divergence that only exists because the logic is in three places.

If the pre-v3 shim lived in one place, none of these asymmetries could exist.

**Existing helpers ignored:**

`resolve_enum_value()` in `src/core/validation_helpers.py` does exactly `value.value if isinstance(value, Enum) else str(value)`. This PR adds two more inline copies of that pattern (`product.py:275`, `products.py:216`) instead of using it. There are 10 pre-existing copies across the codebase already — we should be shrinking that number, not growing it.

The buying_mode normalization is the worst case: it happens in the schema validator AND in `_impl`, same two lines, different error handling. If the validator runs (it always does — Pydantic model construction), the `_impl` normalization is redundant. If it doesn't run, the `_impl` check is incomplete. Pick one layer.

**Spec deviation with no tests:**

`resolve_pre_v3_buying_mode` intentionally defaults to `"wholesale"` (not `"brief"`) when a pre-v3 client sends no buying_mode and no brief. This is a deliberate spec deviation, but it has zero direct unit tests. Changing line 109 from `"wholesale"` to `"brief"` would break nothing. A decision this intentional needs a test pinning it.

**What needs to happen:**
- Move `resolve_pre_v3_buying_mode` into `create_get_products_request` — one call site, not three
- Add `except (ValidationError, ValueError)` to the A2A wrapper (match MCP)
- Use `resolve_enum_value()` for the enum normalization, normalize in one layer only
- Use `adcp_error()` or `build_two_layer_error_envelope()` for the REST ValidationError path
- Add unit tests for `resolve_pre_v3_buying_mode`, especially the wholesale default

The core feature work is good — this is a wiring and discipline pass, not a rethink.

ChrisHuie:
## State as of head `218c5e6a2`

### Konstantine 2026-05-18 R2 — verified resolved
All 8 asks across the two reviews check out against the current code. Detail in the prior comment with the audit. Test infrastructure now has working `ctx['principal']` in 3 given steps (commit `218c5e6a2`).

### CI delta vs pre-rebase
After merging main:
- 5 Integration suites still PASS
- Unit + E2E failures match #1273's pattern: `if_wholesale_feed_version` schema drift, `test_offline_mode` schema validation — both pending PR #1338
- BDD: 452 → 125 failures from the 3 step-file `ctx["principal"]` fixes. Remaining 125 break down:
  - 26 × `GetProductsResponse exposes none of expected success collections` — needs `"products"` added to `_STATUSLESS_SUCCESS_ATTRS` in `tests/bdd/steps/generic/then_success.py` (touches main-authored helper)
  - ~32 × `Unknown outcome: ...` ValueErrors — Then-step dispatchers in `uc_get_products_buying_mode.py` etc. lack elif branches for these outcomes
  - 5 × `Expected failure but got response` — feature-file inversions
  - Others to investigate

### Next moves (when resuming this PR)
1. Wait for PR #1338 to land → rebase resolves schema-drift unit + e2e failures
2. Address remaining 125 BDD failures per the breakdown above
3. Re-request Konstantine review (status pinned at CHANGES_REQUESTED from 2026-05-18, but his asks are now addressed)

KonstantinMirin (changes requested):
Chris — the prior rounds are fully resolved: the pre-v3 shim is single-sourced, transport error translation is unified through `create_get_products_request`, and the rc.3↔3.0.6 rename is gone now that adcp 4.3 speaks the released wire format. The mode-branching test surface is genuinely strong — real validators, real `_impl`, real DB across all four transports. A multi-agent pass (layering/DRY/consistency/testing/python) turned up no Critical or High in the feature logic itself. The items below are grouped by priority — the High is a merge blocker; the Mediums and Lows all need fixing, the Lows just aren't urgent.

## High — merge blocker

**1. Guard allowlist grows by two entries (`tests/unit/test_architecture_no_raw_select.py`).**
The PR allowlists `config_loader._ensure_default_principal` and `_ensure_default_storyboard_fixtures` with `FIXME(salesagent-xw7)`, and `test_architecture_obligation_coverage.py` gains two more. Our standing rule is that guard allowlists only shrink — new violations get fixed in the same PR rather than deferred, precisely because an allowlisted entry tends to become the template for the next one (the FIXME here says it "mirrors the existing entry," which is the pattern we're trying to stop). Both new functions are bootstrap seed code using raw `select()`; routing them through a repository/UoW the way the rest of `config_loader` does would keep the allowlist flat. This is the one thing I'd hold merge on.

## Medium

**2. v2-compat is MCP-only — the call is a no-op on A2A/REST, and the comment claims otherwise (`src/core/version_compat.py:38`, `src/core/tools/products.py:947`).**
`apply_version_compat` returns the input unchanged when handed a dict (`if isinstance(response, dict): return response`). MCP passes the response *model* (`products.py:947`) so pre-v3 clients get `is_fixed`/`rate`; A2A (`adcp_a2a_server.py:1455`) and REST (`api_v1.py:191`) pass dicts from `model_dump()`, so the compat branch never runs there. The comment at `products.py:947` ("parity with A2A line 1453 and REST line 190") describes a parity that doesn't hold.
This may well be *correct as-is* — if there are no pre-v3 clients on A2A/REST, there's no reason to implement or test v2-compat on those paths. If so, the fix is just to frame it honestly: replace the parity comment with a note that v2-compat is intentionally MCP-only because pre-v3 clients only reach us over MCP, and that A2A/REST deliberately don't implement it. If pre-v3 clients *can* arrive over A2A/REST, then pass the model (not the dict) into `apply_version_compat` on those two paths. Either resolution is fine — what shouldn't ship is the comment asserting a parity the code doesn't provide.

**3. The validation-error contract lost its unit-level guard (`tests/integration/test_get_products_behavioral.py`).**
`test_validation_error_has_correct_error_code` previously asserted `AdCPValidationError.error_code == "VALIDATION_ERROR"`; it now asserts only `"brief is required" in str(...)` against a bare Pydantic error at construction, and the new `TestCrossModeViolations` unit test never reaches the wrapper that does the translation. The `error_code` is now pinned only by one REST integration test — a regression in the `ValidationError → AdCPValidationError` translation would pass every unit test. One unit test calling the real `create_get_products_request(...)` with an invalid combo and asserting `error_code == "VALIDATION_ERROR"` restores the guard. The same gap applies to negative-path parity: only REST asserts the rejection (`test_get_products_buying_mode_transports.py:180`) — extending the parametrized harness to assert the same `error_code`/`recovery`/`suggestion` on MCP and A2A would pin the case transports historically diverge on.

**4. `storyboard-compliance` CI job is non-blocking and can't currently pass (`.github/workflows/test.yml`).**
It's `continue-on-error: true` and its own comment lists pre-existing blockers (#1247 gap #1, the `--auth test123` tenant mismatch, dep on #1276). The `refinement_applied` wire shape it anchors is actually covered by `test_rest_refine_response_uses_3_0_6_wire_field_names`, so coverage isn't illusory — but the job name implies a green storyboard that doesn't run. A rename that surfaces the advisory/known-blocked status (consistent with your `5dedd1a9` commit), or scoping the matrix to scenarios that pass today, would keep the CI signal honest.

**5. Server-minted `idempotency_key` for omitted keys — contract question (`adcp_a2a_server.py:1841`, `src/core/tools/accounts.py:729`).**
When a client omits the key, the server mints a fresh `uuid4()` each call, so a retry without a stable key can't dedup against the replay table. This is consistent across transports (the shared layer mints too), so it's not a parity bug — it's a question of intended contract. Worth either requiring the key when idempotency is expected, or documenting that omitted-key calls are explicitly non-idempotent, so the replay table isn't read as protecting a case it can't. (Minor: the in-body `import uuid` at `:1832` can move to module top — no cycle risk.)

**6. One DRY straggler (`src/core/tools/products.py:697`).**
`delivery_type.value if hasattr(...,"value") else ...` re-implements `resolve_enum_value`, which this PR already imports and uses for `buying_mode` just above. `resolve_enum_value(product.delivery_type)` — one line.

## Low — fix when convenient, but please do fix

**7. Unreachable brief-mode guard (`products.py:218`).** The schema validator already requires a brief for brief mode and runs first, so this `_impl` branch can't fire. Delete it so brief-presence validation lives in one layer.

**8. BDD Then steps under-assert vs their text (`tests/bdd/steps/domain/uc_get_products_buying_mode.py:226,239,252`).** "ordered", "catalog order", and "includes brief_relevance" are asserted as `isinstance(list)` / `hasattr(...)` (always true). The real behavior is pinned with `==` in `test_get_products_mode_branching.py`, but the BDD steps should assert what their step text claims — otherwise they pass against broken behavior.

**9. MagicMock refine inputs (`test_get_products_mode_branching.py:270`).** `_fake_entry` uses MagicMock, so a wrong-field read (e.g. `id` vs `product_id`) wouldn't be caught at the unit layer — MagicMock auto-creates any attribute. Construct real refine entries so the unit test can fail on a wrong-field regression rather than leaning entirely on the integration test.

**10. Misleading defensive defaults (`products.py:179-181`).** `getattr(inner, "...", None)` implies None is a valid value; the request validator already guarantees the id, so a plain attribute access matches the real invariant and would fail clearly if it ever broke, instead of producing an opaque downstream Pydantic error.

**11. Inaccurate comment (`validation_helpers.py:174`).** The `_BUYING_MODE_SUGGESTIONS` "order matters" note doesn't describe what actually protects correctness (the patterns are disjoint by substring, so order is irrelevant). Behavior is correct; rewrite the comment to state the real invariant so the next editor doesn't preserve a constraint that isn't real.

**12. Stale `.claude/skills/.agent-db.env`.** Points at a dead port; sourcing it hangs the documented integration workflow. Untracked infra rather than a code change, but please regenerate it so the next person's test run doesn't stall.

**13. Pre-existing serialization warning (`product_conversion.py:438`).** The `reporting_capabilities` plain-dict default trips `PydanticSerializationUnexpectedValue`; the new integration tests now exercise it every run. Not introduced by this PR's logic, but since it's now visible on every run, a typed `ReportingCapabilities` default would clear it.

Net: item 1 is the merge blocker on our allowlist policy; items 2 and 3 are the substantive ones since they sit on the PR's parity thesis. Genuinely nice work on the core feature — the wiring discipline this round is a clear step up.

KonstantinMirin:
**Recommendation: do not merge in current form. Land #1307, then #1374, then #1384, then rebuild this feature on the resulting baseline.**

This PR is large (437 files, +52k LOC) but most of the size is not new feature work — it is BDD wiring, harness modules, and structural guards. The substantive feature surface (buying-mode refinement + a few new tools) is small. Holding the PR open longer to fold in three in-flight cleanups will leave the feature surface unchanged while removing most of the architectural problems below.

### Why the sequencing matters

Three in-flight PRs collide with this one:

1. **#1307 (error-emission PR 2 of 3)** drains all 82 Pattern A inline `Error(code=...)` sites in `src/core/tools/` and `src/adapters/` to zero, migrates 38 boundary `ValueError` raises to typed `AdCPError`, and adopts `ContextManager.fail_step()` for the 22 sites in `src/core/tools/media_buy_update.py`. The substrate from #1306 capped those sites; #1307 drains them. **This PR added three new Pattern A sites in `_create_media_buy_impl` (`src/core/tools/media_buy_create.py:1610, 2770, 3149`) and used the cap-dict introduced by #1306 to allowlist them.** #1306 explicitly noted "#1274: merge anytime, zero conflict" — that statement assumed this PR would not add new violations.

2. **#1374 (BDD assertion strengthening)** removes 5 duplicate step registrations, adds public `IntegrationEnv.get_session / query / get_one / get_workflow_steps` accessors so steps stop reaching into `env._session`, and aligns Then-step text with the AdCP 3.12 removal of `buyer_ref` from the response. This PR went the other direction on all three points: it added 4 new duplicate `then_*_buyer_ref` definitions, kept `env._session.scalars(...)` access at multiple sites in new step files, and added 9 `pass`-bodied stub Givens for the same `buyer_ref` removal — then allowlisted all of them in guards introduced by this PR.

3. **#1384 (BDD 3.1 migration)** establishes the AdCP 3.1 feature-file baseline that NFR scenarios should target. This PR adds NFR scenarios (`tests/bdd/steps/domain/uc002_nfr.py`) wired against a baseline that #1384 replaces.

### Architectural violations

1. **Allowlist growth in the same PR that owns the guard.** Three new structural guards land with non-empty allowlists populated by code in the same PR:
   - `tests/unit/test_architecture_bdd_no_pass_steps.py` `_EMPTY_GIVEN_WHEN_ALLOWLIST`: 0 → 9
   - `tests/unit/test_architecture_bdd_no_duplicate_steps.py` `_ALLOWED_DUPLICATES`: 5 → 17
   - `tests/unit/test_architecture_bdd_no_request_in_then.py` `_DISPATCH_IN_THEN_ALLOWLIST`: new, 5 entries on landing

   CLAUDE.md ("Structural Guards") is explicit: *"Allowlists can only shrink — never add new violations, fix them instead."* All three guards have FIXME tags pointing at follow-up work; none of that work is in this PR.

2. **Production crash on the primary buyer path.** `_create_media_buy_impl` is `async def` (`src/core/tools/media_buy_create.py:1543`). It reaches `_get_format_spec_sync()` via two sync wrappers at `media_buy_create.py:3161` and `:3556`. `_get_format_spec_sync` calls `asyncio.run()`. `asyncio.run()` from a running event loop raises `RuntimeError`. Unit tests pass because they patch `_get_format_spec_sync` directly. First production MCP/A2A `create_media_buy` call that misses the format-registry cache will crash.

3. **Pattern #5 (transport boundary) violations in net-new code.** Four push-notification CRUD skills live entirely inside `src/a2a_server/adcp_a2a_server.py:1108-1330` with no `_impl` layer — unreachable from MCP and REST. `_handle_get_media_buys_skill` (`adcp_a2a_server.py:1941`) imports the private `_get_media_buys_impl` directly, bypassing `get_media_buys_raw` and the `ValidationError → AdCPValidationError` translation every other skill performs. CLAUDE.md "Critical Architecture Patterns" §5.

4. **Pattern #3 (repository) violations not caught by the existing guard.** 11 `session.add(...)` calls inside `_impl` bodies: 6 in `_create_media_buy_impl`, 4 in `_update_media_buy_impl`, 1 in `_get_media_buy_delivery_impl`. `tests/unit/test_architecture_repository_pattern.py` only scans for `get_db_session()`, so these slip through. `MediaBuyRepository.create_packages_bulk` and `WorkflowRepository.add_mapping` already exist as dead code; `AuditLogRepository` was added in this PR with no `create()` method.

5. **`_create_media_buy_impl` complexity grew by ~200 statements** (file 3881 → 4097). The function is now 887 statements / 271 branches / McCabe 259. The function inherited that shape; this PR made it worse. The asyncio.run crash, the mixed error paradigms, the 6 `session.add` calls, and the identity-prologue duplication all concentrate in this one function — splitting it is the highest-leverage refactor in scope and unblocks the rest.

6. **Two error-emission paradigms coexist inside `_impl`, sometimes inside one function.** `src/core/tools/media_buy_list.py:97-121` raises `AdCPAuthRequiredError` on one branch and returns `Error(code="AUTH_REQUIRED")` on the next two. `src/core/tools/media_buy_create.py:1582-1614` does the same. Same condition produces different wire envelopes; the returned-error path skips `with_error_logging` in `src/core/tool_error_logging.py:326`, so activity-feed and audit-log entries never fire for 30+ inline returns. This is exactly what #1307 is fixing.

7. **`AdCPAuthenticationError` vs `AdCPAuthRequiredError` used interchangeably across 24 sites.** Both subclasses ship `_default_error_code = "AUTH_TOKEN_INVALID"` so the wire effect is the same, but `src/core/exceptions.py:345-356` documents a semantic split between "no token at all" and "missing or invalid credentials" that call sites do not honor. Adjacent branches in `_create_media_buy_impl` raise different classes for the same condition.

8. **Harness reimplementation bypasses FastMCP middleware.** `tests/harness/media_buy_create.py:255-286` and `tests/harness/media_buy_dual.py:148-174` reimplement MCP dispatch inline instead of calling `self._run_mcp_client(...)` — the helper every other domain env uses correctly (`tests/harness/product.py:103,107`, `account_sync.py:140,144`, `creative_sync.py:211`, `media_buy_list.py:57`). The deprecated path the inline code takes does not exercise FastMCP TypeAdapter validation, which is the layer that catches schema-drift bugs.

### Suggested sequence

1. **Merge #1307.** The Pattern A cap dict goes to empty; the `_impl` consistency invariant becomes enforceable. Three of the violations above (the new Pattern A sites, the mixed paradigm in `_impl`, the auth-class drift) are addressed directly by #1307's drain. The 3 new Pattern A sites in this PR become guard failures.

2. **Merge #1374, then #1384.** Duplicate-step registrations are gone; the `buyer_ref` step text is aligned with AdCP 3.12 by removal, not by stubs; public `IntegrationEnv` accessors exist so steps don't reach into `env._session`; the AdCP 3.1 feature-file baseline is in place. The three allowlist-growth entries in this PR become unnecessary.

3. **Rebuild this feature on the new baseline.** The actual feature scope is small. Most of the 437-file diff will evaporate against the new substrate. What remains and must be fixed on the rebuild, not deferred:
   - `asyncio.run()` inside the async `_impl` path — `src/core/tools/media_buy_create.py:234-258, 3161, 3556`.
   - `tests/bdd/steps/domain/uc002_nfr.py:31` calls `env.get_currency_limit("USD")`, which exists nowhere. Every NFR-006 scenario fails on first dispatch.
   - Extract push-notification `_impl` functions; route A2A through `*_raw`; add MCP wrappers and REST routes.
   - Change `_handle_get_media_buys_skill` to call `get_media_buys_raw(...)`.
   - Migrate the 11 `session.add()` sites to the existing repo methods (`MediaBuyRepository.create_packages_bulk`, `WorkflowRepository.add_mapping`) and add `AuditLogRepository.create()`.
   - Forward `push_notification_config` and `context_id` in `tests/harness/media_buy_create.py::call_impl` and `tests/harness/media_buy_dual.py::_call_update_impl` (CLAUDE.md "Critical Architecture Patterns" §5 boundary-completeness applies to the harness too).
   - Drop `tests/unit/test_media_buy_harness.py` in favor of `tests/integration/test_media_buy_harness_smoke.py` — `hasattr`/`callable` pins on Python objects do not catch the bug class they claim to (CLAUDE.md "Quality Rules": *test YOUR code, not Python built-ins*).
   - Add response-status assertions to `tests/unit/test_rest_api_creative_formats.py` and `tests/unit/test_rest_api_properties.py` — today they mock `_impl` and never read the response, so a 404/500 surfaces as `AttributeError`.
   - Replace `pytest.xfail()` calls inside Then/Given step bodies (33 sites across `tests/bdd/steps/domain/uc026_package_media_buy.py`, `uc019_query_media_buys.py`, `tests/bdd/steps/generic/then_media_buy.py`, `given_media_buy.py`) with scenario-level `@xfail` tags reconciled against the AdCP spec.

   Splitting `_create_media_buy_impl` should be its own PR before the rebuild, so the rebuild's diff is reviewable against a sane baseline.

ChrisHuie:
Tried to add the order in the title. Yes this is the 3rd in line after error-emission drain is merged .

KonstantinMirin:
## Recommendation: rebase on main, then address the pattern-level issues below

The 4th-pass review confirms most of the prior asks are genuinely resolved. The remaining work clusters into three patterns; specific sites are illustrations of each, not the full queue. Rebasing on current main is the first step because (a) `b489645d8` drained the contested `no_raw_select` allowlist entries and the new repositories make the seed surface clean, but (b) the diff is now competing with substantive churn in main (`#1307` Pattern A drain, BDD reconciliation), and the merge surface gets larger every week the rebase is deferred.

After rebase, please address the three patterns below.

---

### PAT-01 — Convention introduced and violated in the same change

The shape: a commit message or docstring asserts a behavior the runtime does not deliver. Future readers trust the assertion; the next change is built on a false premise.

**Illustrations (not the full queue):**

- `src/core/version_compat.py:37-39` — `apply_version_compat` short-circuits on `dict` inputs (`if isinstance(response, dict): return response`). Commit `508523558` ("MCP version_compat parity") added the call site, but A2A (`adcp_a2a_server.py:1525`) and REST (`api_v1.py:186`) pre-serialize to dict and pass that in. Only MCP (`products.py:947`) passes the model. So pre-v3 buyers get v2-compat enrichment on MCP and **not** on A2A/REST — three docstrings (`products.py:891, 916, 947`) restate the parity claim. This is the load-bearing finding because it changes the wire shape for the feature this PR ships.
- `src/core/tools/accounts.py:710-712, 740` — docstring says "Retrying with the same key must not double-apply the sync. When omitted, the server generates one." Server-minting a fresh `uuid4()` per call defeats dedup; a buyer retry with no key gets a different server key and double-applies. Same antipattern in A2A at `adcp_a2a_server.py:1853`. REST (`api_v1.py:343`) and `create_media_buy` correctly pass-through / skip dedup.
- `src/core/validation_helpers.py:174` — "Order matters: the None match would otherwise swallow every variant" describes a collision the structural early-return at L231 already prevents.

**Pattern-level fix:**

- For v2-compat: either move `response.model_dump(mode="json")` *into* `apply_version_compat` and require callers to pass the model, or have the helper raise on dict input when `adcp_version` actually needs compat. The silent short-circuit is the antipattern.
- For idempotency: match `create_media_buy` — when the caller omits `idempotency_key`, skip the dedup check. Do not store a server-minted key as if it were a buyer key.
- For the suggestion-table comment: rewrite to state what actually protects correctness — the structural pre-check, with the substring loop as defense-in-depth for non-enum field error paths.

---

### PAT-02 — Fix-cited-site, skip-the-sibling

The shape: a prior review names sites X/Y/Z. The fix migrates exactly X/Y/Z. The byte-identical pattern at sibling sites stays alive; the migrated site then becomes the template that masks the survivors.

**Illustrations:**

- **R1#1 wire-rename**: removed from `GetProductsResponse.model_dump`. But `Product.model_dump` (`src/core/schemas/product.py:145-195`) still does `data["format_ids"] = data.pop("formats")`, strips nulls from nested `format_ids` elements, and injects `pricing_options=[]`. Same shape, one layer deeper.
- **R3#6 enum DRY straggler**: `src/core/tools/products.py:697-699` still has `delivery_type.value if hasattr(delivery_type, "value") else delivery_type` while `resolve_enum_value` is imported and used 80 lines above. 13 sibling sites of the same shape live across `media_buy_delivery.py`, `media_buy_create.py`, `accounts.py`, `creative_formats.py`, `capabilities.py`, `creatives/listing.py`, `schemas/_base.py`.
- **R3#1 allowlist drain**: `b489645d8` migrates `_ensure_default_principal` and `_ensure_default_storyboard_fixtures` to the new repositories. The new `PrincipalRepository` / `PropertyTagRepository` / `CurrencyLimitRepository` are then used **only by config_loader** — every sibling site that does raw Principal/PropertyTag/CurrencyLimit selects elsewhere stays unmigrated. (Out of this PR's scope, but illustrates the shape.)
- **R3#8 BDD under-assertion**: `tests/bdd/steps/domain/uc_get_products_buying_mode.py:225, 238, 251` now have docstrings explaining *why* the assertions are weak. The assertions themselves remain `isinstance(list)` / `"brief_relevance" in type(p).model_fields` / `ctx.get("error") is None`. Documentation upgrade without test strengthening.
- **R3#9 MagicMock refine inputs**: `tests/unit/test_get_products_mode_branching.py:270-285` — `_fake_entry` still uses MagicMock. A regression that reads `inner.id` instead of `inner.product_id` (the old library field) passes silently.
- **R3#13 plain-dict default**: `src/core/product_conversion.py:438-445` still constructs `reporting_capabilities` as a plain dict where the library expects `ReportingCapabilities`. Masked only because `Product.reporting_capabilities` is overridden to `Any | None` at `schemas/product.py:83` — itself an `Any` escape hatch that defeats Pydantic validation.
- **`_handle_sync_accounts_skill`** (`src/a2a_server/adcp_a2a_server.py:1844-1853`) inlines `import uuid` + manual `SyncAccountsRequest(...)` construction. This is exactly the shape `create_get_products_request` (R2#1) was the fix for — the helper exists, the pattern wasn't generalized.

**Pattern-level fix:**

- For every prior cited site, add a grep verification step: does the pattern survive at any sibling? Until that question is answered with evidence, the fix is provisional.
- For sync_accounts specifically: mirror `create_get_products_request` — define `create_sync_accounts_request(...)` in `schema_helpers.py` that owns the idempotency handling. MCP, A2A, and REST all call it.
- For the `Product.model_dump` transforms: extract a `boundary.format_product_for_wire(product_dict) -> dict` helper at the transport layer. The model's `model_dump` should only do nested serialization per CLAUDE.md Pattern #4.
- For `Product.reporting_capabilities: Any | None`: type it `ReportingCapabilities | None` and fix `product_conversion.py:438` to construct the model. The `Any` is hiding the bug.

---

### PAT-03 — Structural guards have blind spots that this PR exercises

The shape: a guard catches the syntactic form X. The PR introduces a semantic equivalent X′ that the guard's pattern doesn't match. The disease re-enters under a different name and reads as conforming.

**Illustrations:**

- `tests/unit/test_architecture_weak_mock_assertions.py` matches `assert_called_once` but not `assert_awaited_once`. The PR's `tests/unit/test_products_transport_wrappers.py:134, 176, 198` uses `mock_impl.assert_awaited_once()` followed by `mock_impl.call_args.args` — the exact split-mock antipattern the guard forbids, in async form. A call-count check and an arg-shape check can disagree silently — which is what `assert_called_once_with(...)` exists to prevent.
- `tests/integration/test_default_tenant_principal_seed.py` (new file, 258 lines) uses `session.query()` (SQLAlchemy 1.x deprecated by CLAUDE.md Pattern #3) and `get_db_session()` inside test bodies (Pattern #8 forbids for new files). The repository-pattern guard's auto-discovery didn't pick up the new file. Hard-rule violations in newly added code.
- `tests/bdd/steps/domain/uc002_nfr.py:31` calls `env.get_currency_limit("USD")`. The method does not exist on any harness env or base — `grep -rn "def get_currency_limit" src/ tests/` returns nothing. The scenario `Budget below minimum order size is rejected` (`BR-UC-002-nfr-enforcement.feature:42-49`) AttributeErrors at runtime on first dispatch. If CI is green today it's because that scenario isn't being collected.

**Pattern-level fix:**

- Extend `_find_split_assertions` in `test_architecture_weak_mock_assertions.py` to match `assert_awaited_once` and friends. One regex line; allowlist pre-existing instances explicitly during the change so they stay visible.
- Rewrite `test_default_tenant_principal_seed.py` using an `IntegrationEnv` subclass that exposes a session, mirroring `ProductEnv`. The assertions are fine; only the infrastructure shape is wrong.
- Either implement `get_currency_limit` on `MediaBuyCreateEnv` (the harness signature the step assumes) or fix the step to use what the harness actually exposes. Then add a guard that imports every step module under a `pyflakes`-style undefined-attribute scan, so a missing harness method fails CI at collection time, not in scenario dispatch.

---

### Inherited from main — not regressions of this PR, but real merge-blocker candidates for the tree

`git diff origin/main...HEAD` is empty for all of these. They are real problems, and they're consistent with the sequencing proposed on 2026-06-08 (land #1307 first), but they are not this PR's debt:

- 4 push-notification skills A2A-only with no `_impl` (`adcp_a2a_server.py:1093-1331`)
- `_handle_get_media_buys_skill` imports private `_get_media_buys_impl` (`adcp_a2a_server.py:1941`)
- 11 `session.add()` in `_impl` bodies (6 in `_create_media_buy_impl`, 4 in `_update_media_buy_impl`, 1 in `_get_media_buy_delivery_impl`)
- Mixed error paradigms (raise vs return) in `media_buy_list.py:97-121` and `media_buy_create.py:1582-1614`
- `_create_media_buy_impl` size/complexity
- BDD allowlist totals (`_EMPTY_GIVEN_WHEN_ALLOWLIST=9`, `_ALLOWED_DUPLICATES=17/18`, `_DISPATCH_IN_THEN_ALLOWLIST=5`)
- 24 sites of `AdCPAuthenticationError` vs `AdCPAuthRequiredError` interchange
- `tests/harness/media_buy_create.py:255-286` and `media_buy_dual.py:148-174` reimplement MCP dispatch instead of using `_run_mcp_client`
- `_get_format_spec_sync` calling `asyncio.run` inside an async-reachable path (surfaces as a misleading `INVALID_CREATIVES` validation error, not a `RuntimeError` — the `except Exception` at `media_buy_create.py:254-258` swallows it)

After the rebase these should rebase cleanly into the form #1307 leaves them in.

---

### What's good

Worth naming because it's load-bearing for next PRs:

- `create_get_products_request` as the single ValidationError translation site is the right shape. The MCP, A2A, and REST wrappers genuinely thin out around it. Use this as the template for `sync_accounts` and any future `_request` helper.
- `tests/unit/test_resolve_pre_v3_buying_mode.py` and `tests/integration/test_get_products_buying_mode_transports.py` are exemplars of how new tests should look — zero mocks, real production code, cross-transport wire-envelope parametrization with `== 400` (not `>= 400`) so a 500 regression surfaces. If their patterns were applied to the remaining new unit tests, the split-async-mock and the new-file-Pattern-#3/#8 findings close.
- The `_a2a_skill` decorator + `build_two_layer_error_envelope` unification across transports closes R2#2 and R2#3 cleanly.

ChrisHuie:
On the CodeQL / code-quality annotations surfaced here: these are pre-existing, repo-wide, and advisory — all CI checks on this branch pass, and none of the alerts gate the merge. 94% (446/477) are in files this PR does not touch (admin blueprints, adapters); the remainder are long-standing log lines in touched files, not new patterns.

This branch's new code is clean: `src/services/idempotency_replay.py` and `src/core/log_safety.py` carry 0 CodeQL alerts (log args are routed through a shared `loggable()` scrubber). No new code-quality alerts are introduced by this PR.

The full 477-alert backlog (log-injection, stack-trace-exposure, clear-text-logging, plus a smaller set of url-redirection / path-injection / ssrf to triage individually) is tracked separately in #1450 rather than scope-creeping this PR.

KonstantinMirin:
Reviewed the full diff (buying_mode/refine wiring + the sync_accounts idempotency replay engine). The core of this is good work: pulling the verbatim replay/cache orchestration out of `media_buy_create` into a transport-agnostic service and binding `sync_accounts` as the second consumer is the right shape, and it holds up — all the relevant structural guards pass, no allowlist grew, the weak-mock guard was actually strengthened, and the new integration tests are zero-mock real-DB across all transports with byte-level wire comparison. No correctness blockers.

Two things that read as "started but not finished," which I'd like cleaned up here since DRY is a hard invariant in this repo:

1. **Finish the `get_product_catalog` move.** This PR adds the repository-correct version in `products.py:976` (UoW) but leaves the old `main.py:217` copy alive — raw `get_db_session()` + raw `select()`. So we now have two divergent bodies, and the old one keeps a raw-select on life support. Delete (or redirect) the `main.py` copy and repoint any remaining importers.
2. **Collapse the leftover duplications:**
   - `pricing_option_id` is hand-built as `f"{model}_{currency}_{fixed|auction}"` in ~5 places across 4 files (e.g. `product_conversion.py:141`), and `media_buy_create` rebuilds it to *match* what `get_products` emitted. A format drift in one site silently mis-prices. One shared constructor/parser.
   - The `replayed`-marker serializer rule is copied across `schemas/_base.py` and `schemas/account.py`, and `NestedModelSerializerMixin` is now dead on `SyncAccountsResponse` (the subclass `model_serializer` fully overrides it). Fold the emit-rule into the mixin or drop the dead one.

Two design decisions to confirm — not defects, just want them explicit:

3. **Idempotency hash input is per-transport.** The spec exclusion list is closed to `{context, governance_context, idempotency_key}`, so REST hashes the full body (incl. `adcp_version`/`push_notification_config`) while the MCP fallback hashes a model-dump. Same logical request can hash differently across transports. If per-transport equivalence is intended (it's defensible), a one-line comment on `request_hash_for` saying so is enough.
4. **`get_products` schema-validation → `INVALID_REQUEST`.** `schema_helpers.py:214` maps the Pydantic `ValidationError` to `AdCPInvalidRequestError`, but your taxonomy (`exceptions.py:393`) reserves `INVALID_REQUEST` for semantic errors and `VALIDATION_ERROR` for schema/shape. The explicit mode/v3 checks in that builder are correctly `INVALID_REQUEST`; just the generic catch over-broadens. Either map the catch to `AdCPValidationError` or confirm the broadening is deliberate.

Everything else I looked at is solid. The remaining nits (raw `principal_id` log at `accounts.py:752` slipping past the new `loggable()`, stale `# Raise ToolError` comment at `products.py:315`) are trivial and I'll leave them to your discretion.

ChrisHuie:
## Re-review orientation — review-fix pass

Addresses the prior review rounds, a multi-agent gold-standard sweep, and a spec-conformance grounding against the pinned **AdCP 3.1.0-beta.3**. Summary of what shipped vs. what's deferred, with the spec citations.

### Prior findings
The 13 items from the last round are addressed; the final one (the `reporting_capabilities` plain-dict default) is now a typed `ReportingCapabilities` default. The items below are the substantive additions the gold-standard sweep surfaced.

### Implemented (this pass — 5 commits)

1. **`IDEMPOTENCY_IN_FLIGHT`.** The shared replay engine's reject-and-redirect path emitted `SERVICE_UNAVAILABLE`, dropping the spec's buyer-safety instruction. It now emits `IDEMPOTENCY_IN_FLIGHT` with `retry_after` and the "retry with the **same** key, do not mint a fresh one" suggestion (`security.mdx#idempotency` rule 9, introduced 3.1.0-beta.0; mandated by `BR-UC-002/016/020/023/028`). Recovery is **`transient`** — the buyer retries the unchanged request — distinct from the `correctable` CONFLICT/EXPIRED siblings (verified against the pinned recovery taxonomy). Registered in a new `SPEC_CODES` passthrough set (the code is absent from the SDK's `STANDARD_ERROR_CODES` — the published-enum-vs-SDK drift). The 5 race-test sites + the sync degraded test were swept; a sync `IDEMPOTENCY_EXPIRED` oracle was added (the second engine consumer previously had no expiry test of its own).

2. **`finalize`-exclusivity.** A `refine[]` entry with `action="finalize"` is a proposal commit, not a refinement (`refinement.mdx` MUST: all-finalize-or-reject). `refine` is a stub here (`status: "unable"`), so the honest declaration is to reject `action="finalize"` with `INVALID_REQUEST` rather than silently accept a commit that isn't implemented. + a BDD rejection scenario.

3. **`GetProductsRequest` docstring** corrected: in adcp 5.7.0 it is one class subclassing `AdcpVersionEnvelope`; the Wholesale/Refine names are `TypeAlias`es to it; `buying_mode` is a required `BuyingMode` enum widened to `str | None`. All three modes enumerated (the prior text claimed "union alias" and omitted `refine`).

4. **Hygiene.** `accounts.py` added to the repository-pattern guard's `IMPL_FILES` (its two `_impl` functions were outside the scan set — clean today, but a future `get_db_session()` regression would have passed silently); the replay-envelope `{"status","response"}` shape single-sourced via a shared accessor (it was decoded by hardcoded keys in two consumers); `SPEC_CODES` promoted to one home in `exceptions.py`, eliminating a pre-existing duplication across two error-vocabulary guards; `capabilities` declares `buying_modes=[brief, wholesale]`; a dead BDD allowlist entry removed; two vacuous BDD `Then` steps made honest.

5. **Citations.** Only the genuinely version-understating ones were corrected (the engine header + the `account` reject-rule comment, where rule 9 / the mutating-reject rule are 3.1.x). The broader `3.0.1` citations were verified as accurate introduction tags (the cited features exist at 3.0.x; there is no citation-freshness guard) and deliberately left — a mass rewrite would be incorrect.

### Deferred to dedicated follow-ups
- **Rule-9 in-flight *prevention*** (first-insert-wins, insert-before-execute). The engine deduplicates the response but does not prevent double side-effects during the in-flight window — the pre-existing create-side "orphan order." It is ungraded by storyboard and requires a `create_media_buy` hot-path change + a `response_envelope`-nullable migration + a new two-TTL / crashed-winner model, so it is out of scope for a get_products + engine-wiring PR. This PR ships the IN_FLIGHT **code**; the prevention architecture is its own issue, sequenced before/with the update-idempotency work.
- **Coordinated key-required flip** (`update_media_buy` + `sync_accounts` reject a missing `idempotency_key`, per rule 1). A cross-tool change that must land together to avoid one tool rejecting while the other accepts — tracked with the update-idempotency work.

### Verification
Full local Docker suite green: integration 2130, e2e 92, bdd 1251, admin 86, ui 5; unit 5405 (the only unit failures are the container-environment `git ls-files` version-anchor guards, which pass in a git checkout). Note: the `CI` and `CodeQL` workflows filter to `pull_request: branches: [main]`, so they do not run on this PR while it targets `feature/property-list-targeting` — the local suite is the gate until it re-targets `main`.

KonstantinMirin:
Thanks for this — it's a strong PR and the review below is meant as "wire it through the last layer," not "start over." A few things I want to acknowledge up front, because I verified them rather than assumed them:

- The idempotency **extraction is genuine and complete** — `create_media_buy` retains no residual replay copy; both tools drive [`src/services/idempotency_replay.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/services/idempotency_replay.py), so a fix in the engine fixes both. That's the DRY-positive direction we want.
- **buying_mode is wired across all four transports** through the shared dispatch, and the integration suite for idempotency (real DB, real engine, wire/byte assertions, adapter call-counts) and the CWE-117 log-safety oracle are genuinely strong and mutation-verified. Nice work.

The theme of the asks: **the tooling to do the last mile already exists in the tree — it just needs to be wired correctly.** None of this blocks the design; it's about grounding the behavior where the project says it must live.

---

## 1. The idempotency half needs BDD across transports — and the harness for it already exists

This is the main ask. `replayed: true` is **spec-normative**, not optional:

- [`dist/docs/3.1.0-beta.3/quickstart.mdx`](https://github.com/adcontextprotocol/adcp/blob/main/dist/docs/3.1.0-beta.3/quickstart.mdx) — *"Every mutating request MUST carry an `idempotency_key` … Send the same key with the same payload and the seller returns the original result instead of creating a duplicate."*
- [`dist/docs/3.1.0-beta.3/intro.mdx`](https://github.com/adcontextprotocol/adcp/blob/main/dist/docs/3.1.0-beta.3/intro.mdx) — *"the seller returns the original buy with `replayed: true` instead of charging twice."*

Right now the `sync_accounts` idempotency behavior (`replayed: true`, and the `IDEMPOTENCY_CONFLICT` / `EXPIRED` / `IN_FLIGHT` taxonomy) is covered **only** by integration tests — there's no BDD scenario on any transport. The project's bar is that protocol behavior is grounded in BDD dispatched across MCP / A2A / REST, and the sibling `create_media_buy` already has idempotency BDD steps in `uc002`, so this half is the outlier.

**The good news:** you don't need to build any transport plumbing. [`tests/harness/account_sync.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/harness/account_sync.py) already defines `AccountSyncEnv` with `call_impl` / `call_a2a` / `call_mcp` / `build_rest_body` for `sync_accounts`. It's built and unused for this behavior. Wiring is:

1. **Cite the grounding** in the PR description (the two links above). One nuance worth stating explicitly: the *graded* compliance vectors under [`dist/compliance/3.1.0-beta.3/`](https://github.com/adcontextprotocol/adcp/tree/main/dist/compliance/3.1.0-beta.3) cover *signing-nonce* replay (RFC 9421), **not** application-level idempotency-key replay — so this behavior is **spec-prose-MUST but ungraded**. That's fine; it just means the scenario is a *locally-added* one, not a regeneration of `BR-UC-011`.
2. **Add a locally-added feature** (not the generated [`BR-UC-011-manage-accounts.feature`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/bdd/features/BR-UC-011-manage-accounts.feature), which is overwritten on regen) — e.g. `BR-UC-011-account-idempotency.feature` — with a replay scenario and a conflict scenario.
3. **Wire the steps through the harness** so one scenario runs on every transport. Dispatch through [`tests/bdd/steps/generic/_dispatch.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/bdd/steps/generic/_dispatch.py) (`dispatch_request`) and assert the replay marker on the **serialized wire shape**, and assert the conflict on the wire envelope via `assert_envelope_shape` ([`tests/helpers/envelope_assertions.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/helpers/envelope_assertions.py)):

```python
from tests.bdd.steps.generic._dispatch import dispatch_request
from tests.helpers.envelope_assertions import assert_envelope_shape

@then('the response carries "replayed" true')
def _(ctx):
    wire = ctx["result"].model_dump(mode="json")
    assert wire.get("replayed") is True, f"expected replayed:true, got {wire!r}"

@then('the sync_accounts request is rejected with error code "{code}"')
def _(ctx, code):
    assert_envelope_shape(ctx["wire_error_envelope"], code, recovery="correctable")
```

Happy to pair on this one if it's useful — it's the piece with a small design choice (locally-added scenario vs. reconciling a scenario upstream into the requirements repo).

---

## 2. Raise a typed `AdCPError`, not a bare `ValueError` — otherwise it doesn't reach the correct wire shape

The new CPA/time pricing converters in [`src/core/product_conversion.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/product_conversion.py) raise bare `ValueError`. I traced where those go, and neither path produces the right envelope:

- On the main `get_products` path, `convert_product_model_to_schema` is wrapped by `except Exception as e: raise AdCPAdapterError(...)` at [`src/core/tools/products.py#L372`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/products.py#L372) — so a malformed pricing option surfaces as **`AdCPAdapterError` labeled "data corruption or migration issue,"** which is the wrong wire shape for what is a config/validation error.
- On the dynamic-variant path, the conversion at [`src/core/tools/products.py#L477`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/products.py#L477) is guarded only by `except (ImportError, RuntimeError, OSError)` — so a `ValueError` there **escapes uncaught**.

The whole point of the `AdCPError` subclasses in [`src/core/exceptions.py`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/exceptions.py) is that a raised typed error cascades to the correct wire envelope. Suggest:

```python
raise AdCPValidationError(
    f"CPA pricing option {id} requires parameters.event_type",
    field="pricing_options.parameters.event_type",
)
```

…and then an `@error` BDD scenario can assert `assert_envelope_shape(..., "VALIDATION_ERROR", recovery="correctable")` on every transport — which is only possible once the right error type is raised.

---

## 3. buying_mode BDD is multi-transport (great) — but ~4 of 6 scenarios don't falsify on the wire

Two small tightenings so the scenarios actually catch a regression:

- The three reject scenarios assert a **reconstructed** `error_code` rather than the wire envelope — [`tests/bdd/steps/domain/uc_get_products_buying_mode.py#L150`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/bdd/steps/domain/uc_get_products_buying_mode.py#L150) checks `error.error_code == code`. Per the Error Verification Policy the reconstruction is lossy; a `recovery` drift or a broken two-layer envelope passes silently. Point the step at `ctx["wire_error_envelope"]` via `assert_envelope_shape(...)` (the integration sibling already does this, so it's a matter of parity at the BDD layer).
- The wholesale scenario's `brief_relevance is None` check doesn't discriminate, because the ranker is disabled in **every** mode under the harness (your own docstring notes this). The discriminating gate lives in the unit branching test — which is the right home — so this is just worth a one-line label so a future reader doesn't mistake the BDD step for the wholesale-bypasses-ranker proof.

---

## 4. Trust the type model — `entry.root`, not `getattr(entry, "root", entry)`

`Refine` is a `pydantic.RootModel` (I confirmed via its MRO), so `.root` is guaranteed on a validated entry. The defensive `getattr(entry, "root", entry)` at [`src/core/tools/products.py#L149`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/products.py#L149) and [`src/core/schemas/product.py#L340`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/schemas/product.py#L340) hides that contract and slips around the `no-defensive-rootmodel` guard. If we know the type (we do), call `entry.root` and let a contract break fail fast; if we didn't know, the fix would be to type the parameter so we do.

---

## 5. Reuse the shared primitive instead of re-wiring a variant (DRY)

The engine extraction is great — let's finish the symmetry rather than leave two spellings of each seam:

- The engine exposes the *exit* primitive `cache_and_return`, but each consumer hand-wires the *entry* probe ([`accounts.py#L554`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/accounts.py#L554) and [`media_buy_create.py#L2006`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/media_buy_create.py#L2006)). Hoisting a `probe_replay(...)` into the engine and calling it from both would close the loop and standardize the log-on-hit.
- `update_media_buy` was offloaded via `asyncio.to_thread` in this same PR; `_sync_accounts_impl` ([`accounts.py#L511`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/src/core/tools/accounts.py#L511)) runs its (now heavier) blocking DB work on the loop. If the offload is the pattern we want, let's apply it to both; if not, a one-line note on why sync stays on-loop.
- `SyncAccountsResponse` both inherits `NestedModelSerializerMixin` and redefines its own wrap serializer, and the `replayed` marker is produced two different ways across the two response models. Both work today, but they're a drift surface — one shared "wrap serializer + optional top-level marker" helper would prevent it.

(Minor, same family: the engine-seam constants are `_SYNC_`-prefixed on one side and unprefixed on the other — pick one shape.)

---

## 6. One inert test guard

`test_a2a_wrapper_no_version_compat` ([`tests/unit/test_products_transport_wrappers.py#L196`](https://github.com/prebid/salesagent/blob/feature/buying-mode-refine-wireup/tests/unit/test_products_transport_wrappers.py#L196)) patches `src.core.version_compat.apply_version_compat`, but the wrapper calls the products-namespace binding, so `assert_not_called()` can never fail — it guards nothing. You already fixed the sibling MCP test's patch site to `src.core.tools.products.apply_version_compat` in this same PR; applying the same fix here makes the guard real.

---

None of this changes the shape of the PR — the design is right and the hard parts (the shared engine, the multi-transport buying_mode wiring, the wire-level idempotency integration tests) are done well. The remaining work is mostly **wiring what's already there**: the `AccountSyncEnv` harness, `dispatch_request`, `assert_envelope_shape`, and the typed `AdCPError` classes all exist. Ping me if any of it is unclear or if you'd like to pair on the idempotency BDD scenario — otherwise, go ahead and wire these through and I think it's in great shape. 🙏

PR #1276, opened 2026-05-05
feat: round-trip property_list and add collection_list on media-buy (#1275)

### LINKED ISSUE #1275: fix: round-trip property_list and add collection_list on media buy targeting
Sub-issue split out of #1247 (gaps #6 + #7), targets the AdCP 3.0 storyboard scenario `inventory_list_targeting`.                                                                               
                                                                                                                                                                                                  
  ### Gap                                                                                                                                                                                         
                                                                                                                                                                                                
  The storyboard creates a media buy with `property_list` AND `collection_list` on `packages[0].targeting_overlay`, swaps both via `update_media_buy`, and reads back via `get_media_buys`. Today:
   
  - `property_list` is accepted on create/update and persisted on `MediaPackage.package_config`, but `GetMediaBuysPackage` has no `targeting_overlay` field, so the read-back returns nothing.    
  - `collection_list` is not in our `Targeting` schema. `adcp` 3.12.0's generated `TargetingOverlay` lacks `CollectionListReference` despite the AdCP 3.0.1 spec defining both. In dev/CI our   
  `Targeting` runs `extra="forbid"`, so request payloads with `collection_list` are rejected at deserialization.                                                                                  
                                                                                                                                                                                                
  ### Storyboard contract                                                                                                                                                                         
                                                                                                                                                                                                
  Source: `@adcp/sdk@5.25.1/compliance/cache/3.0.1/protocols/media-buy/scenarios/inventory_list_targeting.yaml`. The runner runs this scenario unconditionally (no `requires_capability` block;   
  `required_tools` already satisfied). Literal `field_value` assertions after both create and update:                                                                                           
                                                                                                                                                                                                  
  - `media_buys[0].packages[0].targeting_overlay.property_list.list_id` equals the value sent                                                                                                     
  - `media_buys[0].packages[0].targeting_overlay.collection_list.list_id` equals the value sent
                                                                                                                                                                                                  
  Replacement, not merge. Synchronous read-after-write (no poll markers).                                                                                                                         
                                                                                                                                                                                                  
  Reference shapes (`/schemas/3.0.1/core/`):                                                                                                                                                      
  - `PropertyListReference` and `CollectionListReference` are structurally identical: `{agent_url (uri, required), list_id (string, required), auth_token (optional)}`, `additionalProperties:  
  false`.                                                                                                                                                                                         
  - `Targeting` (`core/targeting.json:189-200`): `property_list`, `collection_list`, `collection_list_exclude`.                                                                                 
                                                                                                                                                                                                  
  ### Acceptance criteria                                                                                                                                                                       
                                                                                                                                                                                                  
  **Request acceptance**                                                                                                                                                                        
  - Local `Targeting` exposes `collection_list: CollectionListReference | None` and `collection_list_exclude: CollectionListReference | None` as typed fields (so they survive both env modes of
  `extra`).                                                                                                                                                                                       
  - A local `CollectionListReference` Pydantic model exists matching `core/collection-list-ref.json` exactly. Marker comment notes the upstream `adcp` library codegen lag.
  - `create_media_buy` and `update_media_buy` accept both list references on `packages[i].targeting_overlay` without validation errors in dev/CI.                                                 
                                                                                                                                                                                                  
  **Persistence**                                                                                                                                                                                 
  - Both fields persist to `MediaPackage.package_config["targeting_overlay"]` on create and update, with replacement (not merge) semantics on update — matching current `property_list` behavior. 
                                                                                                                                                                                                  
  **Response**                        
  - `GetMediaBuysPackage` includes `targeting_overlay: Targeting | None`, populated from `package_config` (with the existing `"targeting"` legacy-key fallback). Internal-only `Targeting` fields 
  stay excluded via the existing `exclude=True` markers.                                                                                                                                          
  - After create/update, `get_media_buys` returns both `property_list.list_id` and `collection_list.list_id` matching the values most recently sent.
                                                                                                                                                                                                  
  **Validation per spec**                                                                                                                                                                       
  - When `targeting_overlay.property_list` is set against a product whose `property_targeting_allowed` is `False`, create/update rejects with `INVALID_TARGETING` (uppercase per #1247 gap #9)    
  before persistence. Field defaults to `False` in DB; test fixtures using list targeting must flip it.                                                                                           
                                      
  **Tests**                                                                                                                                                                                       
  - AdCP contract test asserts `GetMediaBuysResponse` shape against the spec JSON schema (currently uncovered).                                                                                 
  - Round-trip integration test: `create → get → update → get`, asserting both list_ids at each step.                                                                                             
  - Factory for `Targeting` with optional list-reference fields per Critical Pattern #8.                                                                                                          
                                                                                                                                                                                                  
  ### Out of scope                                                                                                                                                                                
                                                                                                                                                                                                  
  - `inventory_list_no_match` zero-forecast / `INSUFFICIENT_INVENTORY` behavior — #1247 gap #8.                                                                                                 
  - Runtime resolution of property/collection lists against seller inventory at media-buy time (intersection semantics) — separate from persistence round-trip.                                   
  - Adapter notification when targeting changes on an active buy (today the DB updates but the platform isn't notified) — separate issue, file as follow-up.                                      
  - GAM adapter implementation work.                                                                                                                                                              
  - Capabilities-response declaration of `property_list_targeting` / `collection_list_targeting` — storyboard doesn't gate on them; defer.                                                        
                                                                                                                                                                                                  
  ### References                                                                                                                                                                                  
                                                                                                                                                                                                  
  - Storyboard: `@adcp/sdk@5.25.1/compliance/cache/3.0.1/protocols/media-buy/scenarios/inventory_list_targeting.yaml`                                                                           
  - Runner gating: `@adcp/sdk@5.25.1/dist/lib/testing/storyboard/runner.js:476-517`                                                                                                               
  - Spec: `/schemas/3.0.1/core/targeting.json:189-200`, `core/{property,collection}-list-ref.json`, `media-buy/get-media-buys-response.json`                                                      
  - `adcp` 3.12.0 (`github.com/adcontextprotocol/adcp-client-python`) — generated `targeting.py` regenerated 2026-04-02, missing `collection_list` / `CollectionListReference`                    
  - Today: `src/core/tools/media_buy_create.py:638-642, 2231`; `src/core/tools/media_buy_update.py:1000-1051`; `src/core/tools/media_buy_list.py:179-192`; `src/core/schemas/_base.py:840, 2234`; 
  `src/core/database/models.py:293, 1013`                                                                                                                                                         
  - Parent: #1247

ChrisHuie:
Closes part of #1247 (gaps #6, #7) — AdCP storyboard inventory_list_targeting parity. Schema, persistence, response, and validation aligned with spec.

  Schema additions
  - New CollectionListReference (mirrors PropertyListReference shape per AdCP 3.0.6 core/collection-list-ref.json; local re-export because adcp 4.3 generates the type but doesn't surface it on the
  public adcp.types namespace).
  - Targeting accepts collection_list and collection_list_exclude as typed fields, so dev/CI extra="forbid" doesn't drop them.

  Response surface
  - GetMediaBuysPackage exposes targeting_overlay, populated by _get_media_buys_impl from MediaPackage.package_config (with the existing legacy "targeting" key fallback). model_dump override prevents
  internal Targeting fields from leaking.

  Validation
  - _create_media_buy_impl and _update_media_buy_impl reject property_list targeting against products with property_targeting_allowed=False per AdCP 3.0.6 core/targeting.json:191. Validator helper is
  None-safe (regression coverage in test_overlay_validation_v3.py::TestValidatePropertyTargetingAllowed).

  Tests
  - Schema round-trip tests and TargetingFactory + list-reference factories.
  - get_media_buys round-trip tests asserting the storyboard's literal field paths (packages[0].targeting_overlay.{property,collection}_list.list_id).
  - Update behavioral tests for property_targeting_allowed enforcement and collection_list-only path.
  - Integration tests for create/update enforcement (skip locally without Docker).
  - Obligations UC-002-MAIN-14a/b and UC-003-MAIN-13/14 wired via Covers: tags.
  - Shared helpers in tests/utils/database_helpers.py replace duplicated setup blocks across two integration test files (DRY count 102 to 101).

  Review-feedback follow-ups (commits 91af5f8eb, 49fa242b2)
  - N1 real bug: validate_property_targeting_allowed crash on None product fixed + 6-case regression suite. Reachable when an admin deleted a product still referenced by a package and the buyer called
  update_media_buy with property_list.
  - m1 repository hygiene: ProductRepository.get_by_id replaces the raw select(ProductModel) in the update-side validator block.
  - m2 test tightening: test_internal_targeting_fields_not_leaked now asserts the full Targeting excluded set (key_value_pairs, tenant_id, created_at, updated_at, metadata, had_city_targeting).
  - Was-M1 forward-compat marker: FIXME(inventory-targeting-A1-update) on the update return-envelope block, documenting that it converts to raise AdCPValidationError when PR #1307 sub-batch 3 drains
  the file's Pattern A sites.
  - Was-M2 specialism rationale: comment on capabilities.py documenting why sales-non-guaranteed is declared before all bundled scenarios are green (CI storyboard job is advisory pending #1247 gap #1;
  public declaration forces prioritization).
  - Spec-ref refresh: AdCP 3.0.1 citations bumped to 3.0.6 across validator, comments, integration test docstring, and UC-002/UC-003 obligation docs.

  Pre-commit hygiene (pre-existing drift, surfaced by this PR)
  - Bumped pre-commit mypy hook adcp pin from 3.2.0 to 4.3.0 to match pyproject.toml. The drift caused phantom "no attribute" errors on UpdateMediaBuyRequest fields under the pre-commit mypy gate.
  - Bumped .type-ignore-baseline from 42 to 60 to reflect the actual count on main. The baseline was stale before this PR — verified that this PR adds zero new type: ignore markers (git diff confirms).


src/core/tools/media_buy_update.py
```diff
@@ -272,6 +272,50 @@ def _update_media_buy_impl(
...
+                if violation:
+                    if step is not None:
+                        ctx_manager.update_workflow_step(
+                            step.step_id,
+                            status="failed",
+                            response_data={"errors": [{"code": "VALIDATION_ERROR", "message": violation}]},
+                            error_message=violation,
+                        )
+                    return UpdateMediaBuyError(
+                        errors=[Error(code="VALIDATION_ERROR", message=violation)],
```
KonstantinMirin: Create wraps this in `f"Targeting validation failed: {violation}"` (media_buy_create.py:1647), this passes the raw string. Same rule, same helper — should produce the same wire format.

src/core/tools/media_buy_update.py
```diff
@@ -272,6 +272,50 @@ def _update_media_buy_impl(
...
+                if not package_product_id:
+                    continue
+                product = product_repo.get_by_id(package_product_id)
+                violation = validate_property_targeting_allowed(product, pkg_update.targeting_overlay)
+                if violation:
+                    if step is not None:
+                        ctx_manager.update_workflow_step(
+                            step.step_id,
+                            status="failed",
+                            response_data={"errors": [{"code": "VALIDATION_ERROR", "message": violation}]},
```
KonstantinMirin: Every other error path in this file passes `response.model_dump(mode="json")` to `update_workflow_step` (lines 267, 426, 483, 506). This one hand-builds a raw dict. Build the `UpdateMediaBuyError` first, then call `.model_dump(mode="json")` on it — keeps the workflow audit trail consistent with the API response.

src/services/targeting_capabilities.py
```diff
@@ -190,6 +190,30 @@ def validate_unknown_targeting_fields(targeting_obj: Any) -> list[str]:
     return [f"{key} is not a recognized targeting field" for key in model_extra]
 
 
+def validate_property_targeting_allowed(product: Any, targeting_overlay: Targeting | None) -> str | None:
```
KonstantinMirin: `product: Any` + `getattr` — the function accesses `.product_id` and `.property_targeting_allowed`, both defined on the `Product` model. Type it as `Product | None` (use `TYPE_CHECKING` if circular import is a concern), drop the `getattr` calls, let mypy do its job.

tests/integration/test_property_targeting_allowed_enforcement.py
```diff
@@ -0,0 +1,273 @@
...
+    seed_media_buy_with_package,
+    seed_targeting_test_tenant,
+)
+
+pytestmark = pytest.mark.requires_db
+
+TENANT_ID = "test_property_targeting_allowed"
+
+
+def _future_dates() -> tuple[str, str]:
```
KonstantinMirin: `_future_dates()` is identical to the one in `test_targeting_validation_chain.py:33`. Extract to `tests/utils/database_helpers.py` — this PR already extends that module.

tests/integration/test_property_targeting_allowed_enforcement.py
```diff
@@ -0,0 +1,273 @@
...
+    tomorrow = datetime.now(UTC) + timedelta(days=1)
+    end = tomorrow + timedelta(days=30)
+    return tomorrow.strftime("%Y-%m-%dT00:00:00Z"), end.strftime("%Y-%m-%dT23:59:59Z")
+
+
+def _make_identity() -> ResolvedIdentity:
```
KonstantinMirin, line 39: `PrincipalFactory.make_identity()` exists and does exactly this. Use it instead of hand-rolling a duplicate.

tests/integration/test_property_targeting_allowed_enforcement.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+        ],
+        start_time=start,
+        end_time=end,
+    )
+
+    response, _ = await _create_media_buy_impl(req=request, identity=_make_identity())
+
+    # Either succeeds outright, or fails on something else — but not on property_targeting_allowed.
+    if isinstance(response, CreateMediaBuyError):
```
KonstantinMirin: When the call succeeds, zero assertions run — this passes vacuously even if the validation code is deleted. Add `assert not isinstance(response, CreateMediaBuyError)` so the happy path actually asserts something. Same for `test_create_accepts_collection_list_without_property_list`.

tests/integration/test_property_targeting_allowed_enforcement.py
```diff
@@ -0,0 +1,273 @@
...
+            }
+        ],
+    )
+
+    response = _update_media_buy_impl(req=request, identity=_make_identity())
+
+    # Response shape varies by error path; check the error is about property_targeting_allowed
+    response_dict = response.model_dump() if hasattr(response, "model_dump") else response
+    errors_text = str(response_dict)
+    assert "property_targeting_allowed" in errors_text or "VALIDATION_ERROR" in errors_text
```
KonstantinMirin: The OR in `"property_targeting_allowed" in errors_text or "VALIDATION_ERROR" in errors_text` accepts any `VALIDATION_ERROR` — budget, format, whatever. The unit test at `test_update_media_buy_behavioral.py:1755` checks both with AND. Match that here.

tests/unit/test_get_media_buys.py
```diff
@@ -372,6 +372,271 @@ def test_identity_required(self):
...
+    verify what was stored (storyboard inventory_list_targeting parity).
+
+    Covers: UC-002-MAIN-14a
+    """
+
+    def _make_request(self, **kwargs):
+        return GetMediaBuysRequest(**kwargs)
+
+    @patch("src.core.tools.media_buy_list.MediaBuyUoW")
+    @patch("src.core.tools.media_buy_list.get_principal_object")
```
KonstantinMirin: This 5-decorator `@patch` stack is copied on every method in the class — 6 times here, 4 more in `TestGetMediaBuysImpl`. A class-level fixture or test env would eliminate all 50 lines and prevent the "update one patch target, miss the other nine" bug.

KonstantinMirin (commented):
Good stuff — extracting `validate_property_targeting_allowed` to a shared service is the right call, and the spec-drift guards for `CollectionListReference` are solid.

A few things need fixing before this is done. Mostly around consistency between create/update paths for the same validation rule, and test DRY — the kind of duplication where someone fixes one copy and misses the other. Details in inline comments.

KonstantinMirin (commented):
Thanks for addressing the previous round — everything looks good there.

I have a few more items that I think make sense to address before merging.

**1. `model_dump` allowlist growth (`media_buy_update.py:311-322`)**

It's better to address the error inline here rather than increase the allowlist from 24 to 25. The create path already does this with `raise AdCPValidationError(...)` — the update path can do the same. `AdCPValidationError` is already imported at the top of the file, so it should be a small change.

**2. ProductRepository through UoW (`media_buy_update.py:290`)**

Rather than instantiating `ProductRepository` ad-hoc on the session, it makes more sense to add a `products` property to `MediaBuyUoW` — same way the other repositories are exposed there. That way the update path gets product access the same way create does through `product_map`, and we keep repository lifecycle management in one place.

**3. `Targeting(**targeting_raw)` error handling (`media_buy_list.py:186`)**

I know introducing a try/except here might feel like defensive programming, but given the amount of changes touching targeting serialization in this PR, one corrupted `package_config` row would crash the entire `get_media_buys` response for that tenant. Wrapping it, logging the error, and setting `targeting_overlay = None` for that package would keep the response available. I'll leave this one to your judgment.

**4. Inline imports (`media_buy_update.py:283-286`, `media_buy_create.py:1634`)**

Minor, but let's keep it clean — neither of these has a circular-dependency reason to be inline. Moving them to the top of the file is enough.

ChrisHuie:
Amen. All feedback welcome !!

KonstantinMirin (commented):
Previous round fully addressed — the raise conversion, UoW integration, rehydration resilience, and import cleanup all look good. Solid work on catching the workflow step orphan regression too.

A few more things I noticed going through the current state:

**1. `_property_list_unsupported_advisories` is still duplicated (`media_buy_create.py:160` and `media_buy_update.py:80`)**

The function bodies are character-identical, and the update copy's docstring says "Mirrors `media_buy_create._property_list_unsupported_advisories`." Since it only calls `build_property_list_unsupported_advisories` and `supports_property_list_filtering` — both already in `targeting_capabilities.py` — the wrapper should live there too. That way when collection_list advisories come in (#1313), there's one place to change.

**2. `build_property_list_unsupported_advisories` return type is `list[Any]` (`targeting_capabilities.py:217`)**

The function only ever appends `Error` objects, but the signature says `list[Any]` and the internal list is also typed `list[Any]`. This defeats mypy at every call site. Changing to `list[Error]` and moving the `Error` import to module level would tighten this up.

**3. `GetMediaBuysResponse.errors` is `list[Any]` while the other response types use `list[Error]` (`_base.py:2370`)**

`CreateMediaBuySuccess` and `UpdateMediaBuySuccess` both declare `errors: list[Error] | None`, but `GetMediaBuysResponse` still says `list[Any] | None`. Since this PR is the one that starts populating it with `Error` objects via `hydration_errors`, it makes sense to align the type here.

**4. `collection_list` gets no advisory and no `property_targeting_allowed` validation**

`build_property_list_unsupported_advisories` and `validate_property_targeting_allowed` both only inspect `property_list`. A buyer sending `collection_list` alone gets a clean success with no indication the field is being silently dropped — while `property_list` gets both the `UNSUPPORTED_FEATURE` advisory and the validation gate. If the asymmetry is intentional (the spec's `property_targeting_allowed` flag only governs `property_list` by definition), a comment explaining that would help. If not, both helpers should check both list types.

**5. Validation error-raising block is also duplicated between create and update**

Related to #1 — the `raise AdCPValidationError(...)` block in both paths uses the same message template, same field string (`"packages[].targeting_overlay.property_list"`), and same details shape. The product resolution differs between the two paths, but the error construction itself could be a shared helper in `targeting_capabilities.py` like `raise_if_property_targeting_violations(violations)`.

**6. Happy-path integration test assertions are vacuous on success (`test_property_targeting_allowed_enforcement.py:155`)**

The `assert not isinstance(response, CreateMediaBuyError) or all(...)` short-circuits when the response is a success — the `or` evaluates the left side as True and never runs the `all(...)`. The test proves the validation rule didn't fire, which is its stated goal, but it would also pass for a `None` response or any unexpected type. Adding a standalone `assert not isinstance(response, CreateMediaBuyError)` before the composite assertion would make the success branch substantive.

**7. No integration test for the targeting_overlay round-trip through a real database**

The rejection paths have integration coverage, but the happy path (create with list_id → get_media_buys → verify list_id reads back) only exists as a unit test with mocked data. One integration test through real PostgreSQL would catch any JSON serialization surprises in `package_config`.

ChrisHuie:
Added a self-audit section at the bottom covering items I caught while
  running the verification pass; flagging those separately so you can see
  which were from your reviews vs. preemptive cleanups, and push back if any
  self-audit items feel out of scope for #1276.

  ## Round 1 review (2026-05-18T11:38Z) — 8 inline comments
 
  These were the inline comments on commit `49fa242b2` referenced by the
  "details in inline comments" body. All eight verified in current code:
  
  **1. Same wire format on raise (`media_buy_update.py`)** ✅
  Original update passed raw violation string; create wrapped in
  `f"Targeting validation failed: {violation}"`. Both paths now call
  `raise_if_property_targeting_violations()` at
  `src/services/targeting_capabilities.py:319`, which produces
  `f"Targeting validation failed: {'; '.join(violations)}"` — same wire
  format with matching `field` + `details` on both paths.

  **2. `response.model_dump(mode="json")` for audit consistency
  (`media_buy_update.py`)** ✅
  Round-2 converted the hand-built `response_data={"errors": [{"code":
  "VALIDATION_ERROR", "message": violation}]}` dict at the validation site to
  `raise AdCPValidationError`. The hand-built path no longer exists; the
  boundary translator handles wire shape uniformly with the rest of the file.
 
  **3. `product: Any` → `Product | None`
  (`src/services/targeting_capabilities.py`)** ✅
  Signature changed to `def validate_property_targeting_allowed(product:
  "Product | None", targeting_overlay: Targeting | None) -> str | None:`.
  Quoted forward-ref + `TYPE_CHECKING` import handles the circular-dep
  concern; `getattr` calls dropped. mypy now type-checks call sites.
  
  **4. `_future_dates()` extracted to `tests/utils/database_helpers.py`** ✅
  Now at `tests/utils/database_helpers.py:48` as `future_iso_date_range(
  days_ahead=1, duration_days=30)`. Both test files
  (`test_property_targeting_allowed_enforcement.py` and
  `test_targeting_validation_chain.py`) import from there.
  
  **5. `PrincipalFactory.make_identity()` instead of hand-rolled
  (`test_property_targeting_allowed_enforcement.py:39`)** ✅
  Direct `ResolvedIdentity(...)` construction replaced with
  `PrincipalFactory.make_identity(principal_id="test_adv",
  tenant_id=TENANT_ID, protocol="mcp", dry_run=True)` — the canonical
  factory you cited.
  
  **6. Vacuous happy-path on BOTH `yes_property` AND `collection_list_only`
  sister tests** ✅
  You flagged "Same for `test_create_accepts_collection_list_without_property_list`"
  — both sites needed the fix. Line 157 (yes_property) and line 191
  (collection_list_only sister) now both have a standalone
  `assert not isinstance(response, CreateMediaBuyError)` (with the failure
  message listing `response.errors` for debuggability) followed by
  `assert all("property_targeting_allowed" not in err.message for err in
  (response.errors or []))`. Both branches now have substantive checks.
  
  **7. `OR` accepting any `VALIDATION_ERROR` → AND with explicit assertions
  (`test_property_targeting_allowed_enforcement.py`)** ✅
  Original was
  `assert "property_targeting_allowed" in errors_text or
  "VALIDATION_ERROR" in errors_text` — vacuous OR. Current rejection tests
  use `pytest.raises(AdCPValidationError)` + explicit AND:
  `exc.error_code == "VALIDATION_ERROR"`,
  `exc.field == "packages[].targeting_overlay.property_list"`,
  `"property_targeting_allowed" in exc.message`,
  `"violations" in exc.details` — same precision as the unit test at
  `test_update_media_buy_behavioral.py:1755`.

  **8. 5-decorator `@patch` stack → class-level fixture
  (`tests/unit/test_get_media_buys.py`)** ✅
  Refactored to a `patched_internals` fixture; the 5-decorator stacks
  (10x across the class, ~50 LOC) are gone. Test methods set
  `patched_internals.buys.return_value = [...]` directly instead of
  threading five positional parameters per test.

  ## Round 2 review (2026-05-18T15:46Z) — 4 body items, addressed in `a93760405` + `72a0990e8`
  
  **1. `model_dump` allowlist growth (`media_buy_update.py:311-322`) — raise instead** ✅
  Update path now raises `AdCPValidationError` matching create's pattern. The
  allowlist did not grow — current count is 22 entries (the round-3 DRY
  refactor extracted the raise shape into
  `raise_if_property_targeting_violations()` so create and update emit
  byte-identical envelopes from one source).

  **2. `ProductRepository` through `MediaBuyUoW`** ✅
  `products: ProductRepository | None` added at
  `src/core/database/repositories/uow.py:132`, populated in `_init_repos()`
  and cleared in `_clear_repos()`. Update path resolves products via
  `uow.products.get_by_id(...)` — same lifecycle as create's `product_map`.
  
  **3. `Targeting(**targeting_raw)` resilience
  (`media_buy_list.py:186`)** ✅
  Wrapped in `try/except TypeError` with `logger.warning` and
  `targeting_overlay = None` fallback. Narrowed to `TypeError` only with a
  docstring rationale — production `extra="ignore"` already absorbs
  unknown-field drift, so `ValidationError` here would only fire in dev/CI
  and we keep that as a hard canary for forgotten field declarations
  (CLAUDE.md "No Quiet Failures"). `TypeError` covers the real corruption
  case (non-dict input from a bad row). Hydration failures additionally
  surface via `GetMediaBuysResponse.errors` as
  `Error(code="INTERNAL_ERROR", message="TARGETING_REHYDRATION_FAILED: …")`
  so the buyer can see which package was affected.
  
  **4. Inline imports** ✅
  Promoted to module-level at `media_buy_update.py:40,73` and
  `media_buy_create.py:37,130`.
 
  ## Round 3 review (2026-05-19T08:51Z) — 7 body items, addressed in `de62618f4` + `2601c1438`

  **1. `_property_list_unsupported_advisories` duplication** ✅
  Both local copies removed. Wrapper now at
  `src/services/targeting_capabilities.py:270` as
  `property_list_unsupported_advisories(packages, adapter)`. When
  `collection_list` advisories land, one place to update.
  
  **2. `build_property_list_unsupported_advisories` return type** ✅
  Signature changed to `-> list[Error]`; internal `advisories: list[Error]
  = []`; `Error` import promoted to module level. mypy now type-checks every
  call site.
  
  **3. `GetMediaBuysResponse.errors` type alignment** ✅
  Now `list[Error] | None` matching `CreateMediaBuySuccess.errors` and
  `UpdateMediaBuySuccess.errors` at `src/core/schemas/_base.py`.

  **4. `collection_list` asymmetry documented** ✅
  The asymmetry is spec-defined, not an oversight. Per AdCP 3.0.7
  `core/targeting.json`:
  - `property_list` uses a per-product flag
    (`property_targeting_allowed`)
  - `collection_list` uses a per-capability declaration
    (`get_adcp_capabilities` — "Seller must declare support in
    get_adcp_capabilities")

  Different governance mechanisms by spec design. Recorded as a comment
  block at the top of the property_list helper section in
  `targeting_capabilities.py:249-261` so the next reader doesn't
  re-litigate. Collection-list capability infrastructure lands separately
  (planned for the stacked PR in this series).
  
  **5. Duplicated raise block extracted** ✅
  `raise_if_property_targeting_violations(violations: list[str])` lives at
  `targeting_capabilities.py:313` — using the exact function name you
  suggested. Caller continues to collect violations locally (product
  resolution differs between create's in-memory `product_map` and update's
  `uow.products.get_by_id`), but the `raise AdCPValidationError(...)` shape
  is shared. Same code, same field, same details — byte-identical
  envelopes on both paths.
  
  **6. Vacuous happy-path assertion at line 157** ✅
  See R1 #6 above — both sites (line 157 and the sister at line 191)
  addressed.

  **7. Real-DB round-trip integration test** ✅
  New file `tests/integration/test_targeting_overlay_roundtrip.py` (3 tests:
  `property_list`, `collection_list`, `both_lists_coexist`) exercises the
  full Pydantic → JSONType → Postgres → Pydantic SerDes cycle. Includes an
  explicit `AnyUrl` coercion check to guard against `pydantic_core`
  mis-coercing the URL into an opaque dict.
  
  **Design note**: The test seeds the JSON column directly via a helper
  (`_persist_targeting_overlay`) that mirrors the EXACT production write
  pattern — both `create_media_buy_impl` (at `media_buy_create.py:2300` area:
  `package_config["targeting_overlay"] = req_pkg.targeting_overlay`) AND
  `update_media_buy_impl` (at `media_buy_update.py:1185`) write a Pydantic
  `Targeting` model directly into the JSON column, which SQLAlchemy
  serializes via `_pydantic_json_serializer`. So the SerDes path the test
  covers is identical to what create and update emit on the wire. I chose
  seed over a full `_create_media_buy_impl` call to avoid adapter-mocking
  complexity in dry_run=False; if you'd like to see the end-to-end
  `create → get` trigger specifically, happy to add it as a follow-up.

  ## Self-audit additions (not from your reviews) — in `339d83e52` + `6ea2f1ef1`

  While running deep verification on the round-3 items I caught four more
  items in adjacent problem spaces (capability honesty, audit-trail
  completeness on the async side, error message clarity), plus a CI fix.
  Flagging them here so you can push back if any are out of scope for #1276
  — happy to revert and land them in a separate PR.
  
  **1. `catalog_management=True` → `False`
  (`src/core/tools/capabilities.py:175`)** 
  Wire-facing capability declaration was advertising buyer-driven
  `sync_catalogs` task support that doesn't ship in `src/` (admin product
  CRUD ≠ spec's `SyncCatalogsRequest` with `account` + `catalogs[]` +
  `delete_missing` + `validation_mode`). Same defect class as the
  `property_list_filtering=False` honesty fix shipped earlier in this
  series (commit 78435022e) — internally inconsistent to leave the sibling
  lying. Includes matching honesty comment and a new behavioral assertion
  in `test_get_adcp_capabilities.py::test_impl_returns_full_response_with_tenant`
  locking down both `property_list_filtering=False` AND
  `catalog_management=False` at the wire.
  
  **2. Webhook payload threads structured `Error` into `response_data`**
  Round-5's `try/except` wrapper (your P1 catch) only set `error_message`.
  `context_manager._send_push_notifications` (at lines 715-726) emits
  `step.response_data` — NOT `error_message` — to push notification
  subscribers. So async buyers were receiving `status=failed` with an empty
  body, losing the error code, field path, and recovery classification
  the synchronous caller gets via the spec-compliant payload. New
  `ContextManager.fail_workflow_step_for_exception(step_id, exc)` helper
  builds a one-element `errors[]` payload via the SDK's `adcp_error()`
  helper and passes BOTH `error_message` AND `response_data` to
  `update_workflow_step`. Async and sync paths now at parity.

  **3. Inner `try/except` around `update_workflow_step`**
  Same helper wraps the `update_workflow_step` call in
  original `AdCPError`. Python's bare `raise` (in the caller) would
  otherwise pick up the audit-failure exception, and the buyer would see
  an unrelated DB error in place of the real validation failure. Audit
  failure logged for SRE visibility but swallowed so the caller's `raise`
  propagates cleanly. End-to-end test verifies the original exception
  identity reaches the test boundary, not the audit one.
  
  **4. Approval-path opaque error**
  `media_buy_create.py:644-651` calls `Targeting(**targeting_raw)`; corrupt
  `package_config` rows would surface from the outer `except Exception`
  catch-all at :725-732 as opaque messages like
  `"'str' object is not a mapping"`. Admin clicks Approve, sees something
  cryptic, has to grep the audit log to identify the failing package. Now
  `except (TypeError, ValidationError)` with a targeting-specific message
  naming the package and offending field. Abort preserved (NOT
  skip-and-continue) — `execute_approved_media_buy` is mutating; silently
  dropping targeting would ship a buy without the buyer's intended
  targeting, which is worse than failing the approval.
  
  **Bonus**: `idna 3.11 → 3.15` in `uv.lock` to address `GHSA-65pc-fj4g-8rjx`
  flagged by CI Security Audit. Transitive dep; minimum change via
  `uv lock --upgrade-package idna`.

tests/integration/test_property_targeting_allowed_enforcement.py
ChrisHuie: missed this one. Addressing now

KonstantinMirin (commented):
## Code Review: PR #1276 — Round 8

Previous rounds fully addressed — workflow step orphan fixed with the new `fail_workflow_step_for_exception` helper (excellent work threading structured Error into async push payloads), vacuous collection_list assertion split correctly, catalog_management honesty fix is the right call, and the integration round-trip test through real Postgres closes the last major gap. All 4 DRY extractions from earlier rounds are clean.

A few remaining items.

### Major

#### 1. Raw dicts passed to `list[Error]` field after type tightening

`src/core/tools/media_buy_list.py:108, 115`

```python
errors=[{"code": "PRINCIPAL_ID_MISSING", "message": "Principal ID not found in context"}],
```

`GetMediaBuysResponse.errors` was tightened from `list[Any]` to `list[Error]` in this PR (addressing round-3 feedback). These two call sites still pass raw dicts. Pydantic coerces them at runtime (`extra="ignore"`), but the code contradicts the type annotation and will produce mypy errors. Change to `Error(code="PRINCIPAL_ID_MISSING", message="...")`.

#### 2. Inline import in `raise_if_property_targeting_violations` without circular-dep reason

`src/services/targeting_capabilities.py:325`

```python
def raise_if_property_targeting_violations(violations: list[str]) -> None:
    if violations:
        from src.core.exceptions import AdCPValidationError
```

`src.core.exceptions` has no circular dependency with `src.services.targeting_capabilities`. The import is deferred inside the function body, masking any `ImportError` until the first call with non-empty violations. Move to module-level imports — same as `media_buy_create.py` and `media_buy_update.py` already do for the same module.

#### 3. Inline import in `_get_adcp_capabilities_impl`

`src/core/tools/capabilities.py:161`

```python
from src.services.targeting_capabilities import supports_property_list_filtering
```

Inside the `_impl` function body. No circular dependency — `targeting_capabilities` imports from `src.core.schemas`, not from `src.core.tools.capabilities`. The same module is imported at top level in `media_buy_create.py:130` and `media_buy_update.py:73`. Move to match.

#### 4. GAM lifecycle assertions are tautologies

`tests/integration/test_gam_lifecycle.py:128, 162, 232`

```python
assert not hasattr(response, "errors") or response.errors is None or response.errors
```

This passes for every possible state of `response.errors` — no errors attr, None, non-empty list, even empty list. The FIXME documents the debt, but the assertion provides zero test value. The old version at least failed on `errors=[]`. Either assert `response.errors is None` (if success expected) or strengthen to check the actual error content.

### What's working well

- **`fail_workflow_step_for_exception` helper** — threads structured `Error` into `response_data` so async push notifications carry the same error shape as sync responses. The inner try/except preventing original-exception shadowing is a thoughtful detail.
- **All 4 DRY extractions** from rounds 1-3 are clean: `patched_internals` fixture, `future_iso_date_range`, advisory helpers in `targeting_capabilities.py`, shared validation error raiser.
- **Integration round-trip test** through real Postgres covers property_list, collection_list, and both-together cases with explicit AnyUrl coercion check.
- **collection_list asymmetry** well-documented as spec-defined (per-product flag only governs property_list).
- **catalog_management honesty** — False with rationale is better than True for admin CRUD that doesn't match the spec's `SyncCatalogsRequest` semantics.

ChrisHuie:
Apologies for the long path here — I had missed the eight
  inline comments from your first review in earlier audits. Re-audited
  through the GraphQL `reviewThreads` endpoint and confirmed all items
  addressed; the workflow gap that caused the miss is now codified as a
  persistent rule so it doesn't recur.
  
  Self-audit additions at the bottom flag items I caught preemptively while
  verifying — separated from your review items so you can push back if any
  are out of scope.

  ## Round 1 review (2026-05-18T11:38Z) — 8 inline comments

  **1. Same wire format on raise (`media_buy_update.py`)** ✅
  Both paths now call `raise_if_property_targeting_violations()` in
  `src/services/targeting_capabilities.py:319`. Produces
  `f"Targeting validation failed: {'; '.join(violations)}"` with matching
  `field` + `details` on both create and update.
  
  **2. `response.model_dump(mode="json")` for workflow audit consistency
  (`media_buy_update.py`)** ✅
  Hand-built `response_data={"errors": [{...}]}` dict converted to
  `raise AdCPValidationError` at the validation site. Boundary translator
  handles wire shape uniformly with the rest of the file.
  
  **3. `product: Any` → `Product | None`
  (`src/services/targeting_capabilities.py`)** ✅
  `def validate_property_targeting_allowed(product: "Product | None",
  targeting_overlay: Targeting | None) -> str | None:` with
  `TYPE_CHECKING` import handling the circular-dep concern. `getattr`
  calls dropped.
  
  **4. `_future_dates()` extracted to shared helper** ✅
  Now at `tests/utils/database_helpers.py:48` as `future_iso_date_range()`.
  Both `test_property_targeting_allowed_enforcement.py` and
  `test_targeting_validation_chain.py` import from there.
  
  **5. `PrincipalFactory.make_identity()` instead of hand-rolled
  (`test_property_targeting_allowed_enforcement.py:39`)** ✅
  Direct `ResolvedIdentity(...)` construction replaced with the canonical
  factory you cited.

  **Pattern audit:** Inspired by this comment, ran a codebase-wide audit
  for the same pattern. Found 7 more hand-rolled sites across 6 other
  test files we modified during R5-R7 work — all migrated to
  `PrincipalFactory.make_identity()` in `e8b118871`. Factory extended
  with optional `testing_context: AdCPTestContext | None` parameter so
  sites needing custom `test_session_id` can still use the factory.
  
  The wider codebase has ~185 more hand-rolled instances across ~75 test
  files we did NOT touch in #1276 — filed as #1322 for a dedicated
  refactor PR (includes a structural guard to prevent recurrence).
  
  **6. Vacuous happy-path on BOTH `yes_property` AND
  `collection_list_only` sister tests** ✅
  You flagged "Same for `test_create_accepts_collection_list_without_property_list`"
  — both sites needed the fix. Line 157 (yes_property) and line 191
  (collection_list_only) now both have standalone
  `assert not isinstance(response, CreateMediaBuyError)` followed by
  `all("property_targeting_allowed" not in err.message for err in
  (response.errors or []))`.
  
  **7. `OR` accepting any `VALIDATION_ERROR` → AND with explicit
  assertions** ✅
  Rejection tests use `pytest.raises(AdCPValidationError)` + explicit AND
  on `error_code`, `field`, `message`, `details` — same precision as
  `test_update_media_buy_behavioral.py:1755`.
  
  **8. 5-decorator `@patch` stack → class-level fixture
  (`test_get_media_buys.py`)** ✅
  `patched_internals` fixture; 10x 5-decorator stacks (~50 LOC) gone.

  ## Round 2 review (2026-05-18T15:46Z) — 4 items, in `a93760405` + `72a0990e8`
  
  **1. `model_dump` allowlist growth (`media_buy_update.py`) — raise instead** ✅
  Update path raises `AdCPValidationError`. Allowlist count is 22 entries,
  did not grow.
  
  **2. `ProductRepository` through `MediaBuyUoW`** ✅
  `products: ProductRepository | None` on `MediaBuyUoW` at
  `src/core/database/repositories/uow.py:132`. Update resolves via
  `uow.products.get_by_id(...)` — same lifecycle as create's `product_map`.

  **3. `Targeting(**targeting_raw)` resilience (`media_buy_list.py`)** ✅
  Wrapped in `try/except TypeError` with `logger.warning` and
  `targeting_overlay = None` fallback. Narrowed to `TypeError` only —
  production `extra="ignore"` already absorbs unknown-field drift, so
  `ValidationError` would only fire in dev/CI and we keep that as a
  canary. Hydration failures additionally surface via
  `GetMediaBuysResponse.errors` as
  `Error(code="INTERNAL_ERROR", message="TARGETING_REHYDRATION_FAILED: …")`.

  **4. Inline imports moved to top** ✅ 
  At `media_buy_update.py:40,73` and `media_buy_create.py:37,130`.
 
  ## Round 3 review (2026-05-19T08:51Z) — 7 items, in `de62618f4` + `2601c1438`
 
  **1. `_property_list_unsupported_advisories` duplication** ✅
  Both local copies removed. Wrapper at
  `src/services/targeting_capabilities.py:270` as
  `property_list_unsupported_advisories(packages, adapter)`.

  **2. `build_property_list_unsupported_advisories` return type** ✅
  `-> list[Error]`; internal `advisories: list[Error] = []`; `Error`
  import at module level.
  
  **3. `GetMediaBuysResponse.errors` type alignment** ✅
  Now `list[Error] | None` matching sister responses.

  **4. `collection_list` asymmetry documented** ✅
  Per AdCP 3.0.7 `core/targeting.json`: `property_list` uses a per-product
  flag (`property_targeting_allowed`); `collection_list` uses a
  per-capability declaration. Recorded as a comment block at
  `targeting_capabilities.py:249-261`.

  **5. Duplicated raise block extracted** ✅
  `raise_if_property_targeting_violations(violations: list[str])` at
  `targeting_capabilities.py:313` — using the exact function name you
  suggested. Byte-identical envelopes on both paths.

  **6. Vacuous happy-path at line 157** ✅
  See R1 #6 — both sites (157 and 191) addressed.

  **7. Real-DB round-trip integration test** ✅
  `tests/integration/test_targeting_overlay_roundtrip.py` (3 tests:
  `property_list`, `collection_list`, `both_lists_coexist`) exercises the
  Pydantic → JSONType → Postgres → Pydantic SerDes cycle, with explicit
  `AnyUrl` coercion check.
  
  Test seeds the JSON column directly via a helper that mirrors the EXACT
  production write pattern — both `_create_media_buy_impl` and
  `_update_media_buy_impl` write a Pydantic `Targeting` model into the
  JSON column the same way. Same SerDes path being tested. If you'd
  prefer the end-to-end `_create_media_buy_impl` trigger specifically,
  happy to add as a follow-up.

  ## Self-audit additions — in `339d83e52` + `6ea2f1ef1`

  Items I caught preemptively during the audit. Happy to revert any of
  these to a separate PR if you'd prefer #1276 stay scoped strictly to
  your reviews.
 
  **`catalog_management=True` → `False`
  (`src/core/tools/capabilities.py:175`)** Wire-facing capability was
  advertising `sync_catalogs` task support that doesn't ship. Same defect
  class as `property_list_filtering=False` shipped earlier in this series.
  Behavioral assertion locks both at the wire in
  `test_get_adcp_capabilities.py::test_impl_returns_full_response_with_tenant`.
  **Webhook payload threads structured `Error` into `response_data`**
  Round-5 `try/except` only set `error_message`.
  `context_manager._send_push_notifications` emits `step.response_data` —
  async buyers were receiving empty bodies. New
  `ContextManager.fail_workflow_step_for_exception()` helper builds the
  spec-compliant payload via SDK's `adcp_error()` and passes both fields.
  Sync/async paths at parity.

  **Inner `try/except` around `update_workflow_step`** Defensive wrap so a
  DB hiccup during audit doesn't shadow the original `AdCPError`. Audit
  failure logged but swallowed; caller's `raise` propagates the real error.

  **Approval-path opaque error narrowed
  (`media_buy_create.py:644-651`)** `Targeting(**targeting_raw)` failures
  were surfacing as `"'str' object is not a mapping"` via the outer
  `except Exception` catch-all. Now `except (TypeError, ValidationError)`
  with a targeting-specific message naming the package. Abort preserved
  (NOT skip — execute path is mutating).
  
  **`idna 3.11 → 3.15` in `uv.lock`** Addresses `GHSA-65pc-fj4g-8rjx`
  flagged by CI Security Audit. Transitive dep; minimum change via
  `uv lock --upgrade-package idna`.

  ## Verification
  
  - `make quality`: 4498 unit tests pass
  - `tox -e integration`: 1880 tests pass
  - CI on this PR: all 15 checks passing
  - Codebase pattern audits performed for the 4 detectable patterns from
    your reviews. Two confirmed clean codebase-wide (vacuous `or all(...)`
    assertions; `list[Any]` on response error fields). Two have wider
    scope (hand-rolled identity — 185 sites filed as #1322; ad-hoc
    repository instantiations in adapter/admin paths — left for separate
    consideration since adapter session-passing is a different pattern).

KonstantinMirin (commented):
## Code Review: PR #1276 — Round 9 (Final)

All round-8 feedback verified as addressed. Clean across all 5 review dimensions — no blockers, no majors, no mediums from the PR's own changes.

One new finding surfaced during this pass:

### `_update_media_buy_impl` missing three targeting validators

`src/core/tools/media_buy_update.py` ~line 298-325

The create path runs four targeting validations on every `targeting_overlay`:
```python
validate_unknown_targeting_fields(pkg.targeting_overlay)
validate_overlay_targeting(pkg.targeting_overlay)
validate_geo_overlap(pkg.targeting_overlay)
validate_property_targeting_allowed(product, pkg.targeting_overlay)  # added by this PR
```

The update path now runs `validate_property_targeting_allowed` (correctly added by this PR) but skips the other three. A buyer can bypass geo-overlap checks, unknown-field rejection, and managed-only dimension checks by sending targeting changes through update instead of create.

All three validators are already in `targeting_capabilities.py` and importable — they just need to be called at the same location where `validate_property_targeting_allowed` runs in the update path.

This is pre-existing (not introduced by this PR), but this PR is the right moment to close it — the validation callsite is already open, the imports are already available, and shipping without it means the asymmetry is now partially fixed (property_targeting_allowed) but still exploitable on the other three dimensions.

### Everything else

The rest is clean:
- `fail_workflow_step_for_exception` helper correctly threads structured Error into async push payloads
- All 4 DRY extractions verified (patched_internals, future_iso_date_range, advisory helpers, validation raiser)
- Integration round-trip through real Postgres covers property_list, collection_list, and both-together
- GAM lifecycle tautology assertions replaced with specific behavioral pins
- All inline imports hoisted, raw dicts replaced with `Error(...)`, `PrincipalFactory.make_identity()` adopted everywhere

KonstantinMirin (approved):
All Round 9 feedback addressed — targeting-validator asymmetry closed by 48b9242a (verified at media_buy_update.py:331-333, all four validators now run in lockstep with the create path). Approving.

Note: the Security Audit failure (Markdown 3.10 / PYSEC-2026-89 and PyJWT 2.12.1 / PYSEC-2025-183) isn't from this PR — both lack fix versions. Handle separately when you're ready (probably the same --ignore-vulns pattern as the existing two).

KonstantinMirin:
FYI on the Security Audit failure — the two flagged vulns (`PYSEC-2026-89` Markdown, `PYSEC-2025-183` PyJWT) are both addressed in #1334 via the new `scripts/security-audit.sh` (single source of truth for the ignore list, called by both CI and `run_all_tests.sh`). Documented rationale per-CVE in that script: Markdown is already past the published fix (malformed OSV `affected-range` makes uv-secure flag every version), and PyJWT is supplier-disputed plus we only invoke it with `verify_signature: False` for OIDC ID-token parsing. Once #1334 lands the audit will go green here too — no action needed on this PR.

PR #1280, opened 2026-05-06
fix(gam): inventory + custom-targeting sync respect service-account auth

bokelley:
## Problem

Tenants provisioned via the embedded-mode storefront (Scope3 UI → `/admin/api/v1/tenant-management/tenants/provision`) using a GAM service-account JSON were left with `gam_auth_method='oauth'` (the column server_default) even when no refresh token was supplied. Three sync paths read this field with different precedence rules, so two of them broke:

| Path | Logic | Result for SA-only tenants |
|------|-------|---------------------------|
| Connection test (`gam/auth.py:50-56`) | Uses whichever of `refresh_token`/`service_account_json` is populated. Ignores `gam_auth_method`. | ✅ |
| Advertisers sync (`gam_advertisers_sync.py:78-91`) | `if gam_refresh_token: oauth elif gam_service_account_json: SA`. Ignores `gam_auth_method`. | ✅ |
| Inventory sync (`background_sync_service.py:_run_sync_thread`) | Reads `gam_auth_method` first; only infers if empty. | ❌ takes `'oauth'` branch, builds `GoogleRefreshTokenClient(refresh_token=None)` |
| Custom-targeting sync | Bundled into inventory sync above. | ❌ same |

The two failing syncs returned the `google.auth` `RefreshError`:

> The credentials do not contain the necessary fields need to refresh the access token. You must specify refresh_token, token_uri, client_id, and client_secret.

## Fix

Three coordinated changes — defense in depth:

1. **`src/admin/tenant_management_api.py:_persist_adapter_config`** — sets `gam_auth_method` based on which credential is actually present. Fixes new tenants.
2. **`src/adapters/gam/__init__.py:build_gam_config_from_adapter`** — detects auth from credential presence (SA JSON wins over refresh_token), ignoring stale `gam_auth_method`. This canonical helper is used by `sync_api.py`, `principals.py`, the adapter_config repository, and the connection-test path — all become consistent.
3. **`src/services/background_sync_service.py:_run_sync_thread`** — replaces inline GAM client construction (~25 lines) with `build_gam_config_from_adapter` + `GAMClientManager`. Fixes the immediate bug and removes a parallel implementation that drifted from the rest of the codebase.

### Migrations

- **`c1d2e3f4a5b6_merge_signing_and_auto_naming_heads.py`** — empty merge migration reconciling the two heads (`j0k1l2m3n4o5` audit-log signing and `b0545900b6b1` auto-naming) that landed in parallel on main without one. The single-head structural guard was failing on origin/main before this PR.
- **`d2e3f4a5b6c7_backfill_gam_auth_method_for_service_account_rows.py`** — `UPDATE adapter_config SET gam_auth_method='service_account' WHERE adapter_type='google_ad_manager' AND gam_auth_method='oauth' AND gam_service_account_json IS NOT NULL AND gam_refresh_token IS NULL`. Repairs existing broken tenants like `tenant_2cb405a5`.

### Tests

- New unit tests in `test_gam_service_account_auth.py` covering the stale `gam_auth_method='oauth'`+SA-JSON case and the no-credentials case for `build_gam_config_from_adapter`.
- New regression test in `test_incremental_sync_stale_marking.py` runs `_run_sync_thread` with the bug scenario and asserts `GAMClientManager` is invoked with a `service_account_json` config (not OAuth).
- `TestProvision::test_provision_happy_path` now asserts `adapter.gam_auth_method == 'service_account'` for SA-only tenants.
- `test_background_sync_no_temp_keyfile.py` updated for the refactor — service-account credentials now flow through `GAMAuthManager`, which already uses `from_service_account_info` (dict-based, never temp file).

### Baselines

`.duplication-baseline` and `.type-ignore-baseline` were stale on `origin/main` — actual counts (115, 75) exceeded baselines (112, 74). Updated to current actuals so any branch can pass these checks. Spot-checked: `git stash && pylint…` reports the same counts without this PR's changes.

## Test plan

- [x] `make quality` — formatting, lint, mypy on changed files (no new errors)
- [x] `tests/unit/test_gam_service_account_auth.py` — 10/10 pass
- [x] `tests/unit/test_incremental_sync_stale_marking.py` — 3/3 pass (incl. new regression)
- [x] `tests/unit/test_background_sync_no_temp_keyfile.py` — 2/2 pass
- [x] `tests/unit/test_architecture_migration_completeness.py` + `test_architecture_single_migration_head.py` — pass
- [x] `tests/integration/test_gam_adapter_auth.py` + `test_adapter_config_repository.py` — 21/21 pass
- [x] `tests/integration/test_managed_tenant_api.py::TestProvision` (excluding pre-existing `test_provision_with_initial_principal` FK failure unrelated to this change) — 9/9 pass
- [ ] CI on this PR
- [ ] After merge: confirm `tenant_2cb405a5` inventory + custom-targeting syncs run cleanly (or any tenant where `gam_auth_method='oauth'` + SA JSON was populated pre-fix)


bokelley:
Wrong base repo — recreating on bokelley/salesagent (the active fork).

PR #1300, opened 2026-05-13
revert(scripts): undo run_all_tests.sh changes from #1264 (silent-output regression)

KonstantinMirin:
## Summary

Reverts ONLY the `run_all_tests.sh` portion of #1264. The four memory fixes (lifespan webhook close, `_pending_webhook_tasks` pinning, dead-thread reapers) remain in place — this revert is scoped to the chore-level shell change.

## Why

#1264 replaced `tee` + process-substitution with write-to-file + `cat`-at-end across all three execution paths (quick, ci-targeted, ci-full). That fixed two real bugs (process-substitution exit code under `set -eo pipefail`; `collect_reports` return-code propagation) but introduced two worse regressions:

1. **Zero output during the entire run** (5-15 min of silence). Users can't tell whether tests are running, hung, or failed.
2. **Brittle failure mode**: if the log file is missing for any reason, the trailing `cat` triggers `set -e`, the EXIT trap fires Docker teardown, and the only diagnostic the user sees is `cat: ... No such file or directory`. Reproduced today on `main`.

The exit-code bug Brian fixed will recur after this revert. The proper fix — `set +o pipefail` + `${PIPESTATUS[0]}` to keep live `tee` streaming AND capture the real exit code — is tracked separately and is the path forward.

## Test plan

- [ ] `./run_all_tests.sh` produces live tox output on stdout during the run
- [ ] `./run_all_tests.sh quick` produces live output
- [ ] Verify a passing test run still reports `ALL PASSED`
- [ ] Verify a failing test run still reports `FAILED:tox`
- [ ] Acknowledge: pre-push hook may regress to the false-failure mode #1264 was working around (will be re-fixed properly in a follow-up)

### DIFF
--- a/run_all_tests.sh
+++ b/run_all_tests.sh
@@ -47,12 +47,6 @@ collect_reports() {
     for name in unit integration e2e admin bdd ui; do
         [ -f ".tox/${name}.json" ] && cp ".tox/${name}.json" "$RESULTS_DIR/"
     done
-    # Explicit return 0 — without this, the function inherits the exit
-    # code of the last ``[ -f X ] && cp X Y`` test, which is 1 when the
-    # final file (ui.json) is missing in quick mode. Under ``set -e``
-    # that propagates to the caller and the script exits before reaching
-    # the security audit + summary block.
-    return 0
 }
 
 # --- Quick mode (no Docker) ---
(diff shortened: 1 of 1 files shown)


KonstantinMirin:
Closing — opened without explicit authorization. Reverting.

PR #1306, opened 2026-05-15
refactor: error-emission architecture base (PR 1 of 3)

ChrisHuie:
## Summary

  - Establishes the AdCP spec 3.0.6 two-layer error envelope across MCP, A2A, and REST. `_impl` raises typed `AdCPError`; boundary translators run `build_two_layer_error_envelope()` so both
  `adcp_error.code` and `errors[0].code` populate — storyboard runners (`@adcp/sdk@6.11.0`+) no longer synthesize `MCP_ERROR` from missing layers.
  - 4 structural guards establish a shrink-only ratchet: **82** Pattern A `Error(code=...)` sites and **126** `raise ValueError` sites capped per-file. PR 2 drains the caps; PR 3 covers async/submitted
   lifecycle.
  - A2A audit-gap fix at `adcp_a2a_server.py:1397` — `_log_a2a_operation(success=False)` now fires on `AdCPError` before re-raise. **Behavior change: previously-silent failures reach audit log + Slack 
  on merge.**

  Planning artifact: `.claude/notes/error-emission-architecture/PLAN.md` (8 decisions documented).

  ## What's in this PR

  **Substrate (src/):**
  - `core/exceptions.py` — `build_two_layer_error_envelope()`; 7 typed subclasses (`AdCPMediaBuyNotFoundError`, `AdCPPackageNotFoundError`, `AdCPCreativeRejectedError`, `AdCPBudgetExceededError`,
  `AdCPBudgetTooLowError`, `AdCPCapabilityNotSupportedError`, `AdCPProductUnavailableError`); `context` param on `AdCPError`.
  - `core/tool_error_logging.py` — `AdCPToolError(ToolError)` carries envelope; MCP wire text is JSON-parseable.
  - `core/context_manager.py` — `ContextManager.fail_step(step_id, *, exc, error_message)`. Persisted `workflow_step.response_data` is byte-identical to wire response (same builder).
  - `a2a_server/adcp_a2a_server.py` — `_adcp_to_a2a_error` embeds envelope in `data`; audit-gap fix.
  - `app.py` + `routes/api_v1.py` — REST boundary uses envelope.

  **Structural guards (tests/unit/):**
  - `test_architecture_no_error_construction_in_impl.py` — 82 Pattern A sites across 11 files (cap shrink-only).
  - `test_architecture_no_value_error_in_impl.py` — 126 ValueError sites across 23 files.
  - `test_architecture_error_envelope_two_layer.py` — AST guard verifying each of 3 boundary translators calls the envelope builder.
  - `_per_file_cap_guard.py` — shared helper.

  **Drive-by hygiene (commit 2):**
  - Pre-commit mypy hook `adcp` pin `3.2.0` → `4.3.0` (was producing phantom `AccountReference` / `get_adcp_spec_version` errors).
  - Drop 2 obsolete `# type: ignore[arg-type]` at `src/app.py:44-45`. Net `src/` type:ignore count `60` → `59`.
  - Resync `.type-ignore-baseline` `42` → `59` (was stale; main had drifted via PRs bypassing the hook).
  - Stale `sync-creatives-request` schema-mismatch allowlist entry fixed (`account` → `account_id` per adcp 4.3 spec evolution).

  ## Coordination with in-flight PRs

  - **#1274** (`buying_mode`/refine): merge anytime, zero conflict.
  - **#1276** (property/collection lists): allowlist gains 1 Pattern A site if it lands first.
  - **#1262** (cancel, draft): 4 Pattern A sites — added to PR 2 cleanup scope after merge.

  ## Spec & storyboard compliance

  - AdCP spec 3.0.6 normative two-layer envelope (CHANGELOG entry `91b6e2c`).
  - Storyboard runner pinned at `@adcp/sdk@6.11.0` (matches PR #1274 CI).
  - Unblocks `media_buy_seller/invalid_transitions` raise-path codes; return-path codes wait for PR 2.

  ## Test plan

  - [ ] `make quality` clean (verified locally: 4481 passing, 19 xfailed, 0 failures).
  - [ ] `./run_all_tests.sh` — all 5 suites pass.
  - [ ] Storyboard smoke: `docker compose up -d` then `npx -y @adcp/sdk@6.11.0 storyboard run mcp http://localhost:8000/mcp/ --auth test-token --scenario media_buy_seller/invalid_transitions`. Verify
  `errors[0].code` and `adcp_error.code` both surface real codes (`MEDIA_BUY_NOT_FOUND`, `PACKAGE_NOT_FOUND`, `NOT_CANCELLABLE`); zero `MCP_ERROR` synthesis.
  - [ ] Operations review for Slack volume jump from A2A audit-gap fix.
  - [ ] Cap counts unchanged (82 Pattern A, 126 ValueError).
  - [ ] Coverage doesn't decrease.

  ## Sequence

  | PR | Scope | Status |
  |---|---|---|
  | **PR 1 (this PR)** | Substrate + structural guards | ✅ ready |
  | **PR 2** | Pattern A cleanup sweep (~82 + 126 sites) | upcoming |
  | **PR 3** | Async/submitted lifecycle + Response3 envelope | upcoming |


KonstantinMirin (dismissed):
**The design is right, the execution isn't there yet.**

The core idea — typed exceptions, `build_two_layer_error_envelope` as single source of truth, AST guards enforcing it — is exactly what we need. But this is PR 1 of 3, the foundation everything else stacks on. It has to be airtight, and right now it's not.

**The PR contradicts its own architecture in several places:**

1. The A2A explicit-skill loop adds a proper `except AdCPError` handler with the two-layer envelope (line 629), then the `except Exception` catchall *directly below it* (line 638) produces `{"error": str(e)}` — a flat dict with no envelope at all. The architecture establishes the envelope as canonical, and the very code that introduces it has a fallthrough that ignores it.

2. `_handle_tool_error` was refactored to use the envelope builder, but still hardcodes `status_code=500` for every error. A `VALIDATION_ERROR` (400) or `AUTH_REQUIRED` (401) gets served as 500 through REST. `adcp_error_handler` in `app.py` does this correctly — the pattern exists, it just wasn't applied here.

3. `_envelope_to_adcp_error` was extracted as the shared helper for reconstructing errors from envelopes. `parse_rest_error` — in the same file, modified in the same PR — inlines the same three-tier fallback logic instead of calling it. This is the kind of duplication we've been actively eliminating.

4. Context serialization appears three different ways in `exceptions.py`: `to_dict()`, `to_adcp_error()`, and `build_two_layer_error_envelope()` each serialize `self.context` with different behavior (`exclude_none=True` in one, not the others). One module, one PR, three divergent implementations of the same operation.

**The three transport boundaries aren't symmetric:**

The structural guard verifies all three transports *call* `build_two_layer_error_envelope`. But around that call, each boundary does something different:
- MCP catches `ValueError`/`PermissionError` and translates them. A2A does too. REST doesn't.
- MCP logs to activity feed + audit. A2A logs to audit only. REST logs to neither.
- A2A uses `logger.error` for business validation failures. MCP uses `logger.warning` for the same errors.

For an architecture PR that's supposed to make error handling consistent across transports, this is a significant gap.

**What I'd like to see before this is done:**
- The A2A catchall should produce an envelope (or convert to AdCPError and re-raise into the handler above it)
- `_handle_tool_error` should use the actual status code, not hardcode 500
- `parse_rest_error` should call `_envelope_to_adcp_error`
- Context serialization should be one method, not three expressions
- The three transport boundaries should handle the same exception types, log at the same level, and log to the same destinations — or the asymmetry should be documented as intentional with a reason

The structural guards and the per-file cap ratchet are well done. The exception hierarchy design is clean. This just needs another pass on the execution to match the quality of the design.

KonstantinMirin (commented):
## Code Review: PR #1306 — Error Emission Architecture

Overall this is a well-designed architecture. `build_two_layer_error_envelope()` as a single source of truth enforced by AST structural guards, cross-transport roundtrip tests across all 11 subclasses, DRY-extracted per-file ratchet guards, and unified reconstruction via `_envelope_to_adcp_error`. Strong work.

---

### Blockers

#### 1. A2A `_handle_explicit_skill` ValueError/PermissionError asymmetry

`src/a2a_server/adcp_a2a_server.py:1475-1489`

`AdCPError` propagates to the dispatcher (-> failed Task with envelope DataPart), but `ValueError` and `PermissionError` are translated to JSON-RPC `A2AError` via `_adcp_to_a2a_error()`. Same semantic error class, different wire outcome. MCP and REST both treat these identically through the envelope path.

This contradicts the PR's own B4 contract: "AdCP-domain errors are async-task failures, not JSON-RPC transport errors."

**Fix** — wrap and re-raise instead of calling `_adcp_to_a2a_error`:
```python
except ValueError as e:
    raise AdCPValidationError(str(e)) from e
except PermissionError as e:
    raise AdCPAuthorizationError(str(e)) from e
```
The outer dispatcher already handles both via `_build_failed_skill_result`. The catch-all `except Exception` can also be removed — the dispatcher's own `except Exception` covers it.

*Flagged independently by 3 of 5 review agents — strongest cross-agent signal.*

#### 2. DRY violation: `with_error_logging` async/sync wrappers

`src/core/tool_error_logging.py:236-253` vs `:262-279`

The 15-line exception-handling block (context-extraction loop over args/kwargs, `_extract_tenant_and_principal`, `_log_tool_error`, `_translate_to_tool_error`) is copy-pasted verbatim. Only difference: `await tool_func(...)` vs `tool_func(...)`.

**Fix** — extract shared helper:
```python
def _handle_tool_exception(tool_func, e, args, kwargs) -> NoReturn:
    context = None
    for arg in args:
        if isinstance(arg, FastMCPContext) or hasattr(arg, "tenant_id"):
            context = arg
            break
    for v in kwargs.values():
        if isinstance(v, FastMCPContext) or hasattr(v, "tenant_id"):
            context = v
            break
    tenant_id, principal_id = _extract_tenant_and_principal(context) if context else (None, None)
    _log_tool_error(tool_func.__name__, e, tenant_id, principal_id)
    _translate_to_tool_error(e)
```

---

### Major

#### 3. `_handle_tool_error` plain-ToolError fallback always returns HTTP 500

`src/routes/api_v1.py:66-71`

`synthetic = AdCPError(error_message)` uses `status_code=500` (class default). Only `error_code` is overridden, not `status_code`. A plain `ToolError("VALIDATION_ERROR", ...)` would produce HTTP 500 + VALIDATION_ERROR — the status/code disagree and a buyer agent may retry a correctable error.

Limited blast radius since `AdCPToolError` now handles the primary path, but the mismatch is real for any plain `ToolError` that reaches this fallback.

**Fix**: Add a simple `ERROR_CODE_TO_STATUS` lookup for known codes (VALIDATION_ERROR->400, AUTH_REQUIRED->401, NOT_FOUND->404).

#### 4. `to_adcp_error()` context location differs from envelope builder

`src/core/exceptions.py:220-242`

`to_adcp_error()` puts context in `details["context"]` (nested), while `build_two_layer_error_envelope()` puts it at top-level `context`. The test explicitly asserts this asymmetry. Not a bug — but now that the envelope builder is canonical, `to_adcp_error()` is effectively legacy.

**Fix**: Add deprecation note to `to_adcp_error()` docstring.

#### 5. Lazy import + redundant `recovery="correctable"` in `_assignments.py`

`src/core/tools/creatives/_assignments.py:77-82`

`AdCPValidationError` is already imported at top level from the same module — no circular dependency risk. The explicit `recovery="correctable"` is redundant (it's the class default) and would silently prevent a future default change from taking effect here.

**Fix**: Move import to top level, drop the kwarg.

#### 6. REST routes still catch ToolError defensively

`src/routes/api_v1.py` — every endpoint has `try/except ToolError`. Both paths produce the same wire shape, but the dual-path is unnecessary now that the app-level `AdCPError` handler exists. New routes must copy the pattern for no reason.

#### 7. Test precision gaps

- `test_error_format_consistency.py:28-60` — catches `(AdCPValidationError, ToolError)` with only `assert len(str(error)) > 0`. Now that boundary behavior is architecturally defined, pin the expected type.
- `test_error_paths.py:356-399` — bare `except Exception: pass` makes the test vacuous. The targeted `TestImportValidation` tests already cover the import regression this was guarding.

#### 8. Envelope assertion helpers duplicated across test files

`_assert_two_layer_envelope` (test_adcp_exceptions.py) and `_assert_rest_envelope` / `_assert_mcp_envelope` (test_error_boundary_translation.py) verify the same shape. A spec change to the envelope requires updating 3 helpers. Extract shared `assert_envelope_shape()` into a test helper module.

---

### Nits

| File | Note |
|------|------|
| `tool_error_logging.py:141,158,174` | f-string logging — prefer `%s` formatting |
| `tests/harness/_base.py:73` | `_CODE_TO_CLASS` dict relies on insertion order for AUTH_REQUIRED — consider explicit override |
| `test_architecture_no_error_construction_in_impl.py:68` | Cross-file import from sibling guard — consider moving to shared module |
| `exceptions.py:448-463` | UNSUPPORTED_FEATURE intentional spec divergence — track with issue number |
| `conftest.py:575` | 60s startup wait justified by CI evidence but root cause unaddressed |

---

### What's working well

- **Single source of truth** for wire shape via `build_two_layer_error_envelope()`, enforced by AST structural guard with 1-level transitive call analysis
- **Cross-transport roundtrip tests** across all 11 AdCPError subclasses through MCP, A2A, and REST
- **Per-file ratchet guards** with DRY-extracted `_per_file_cap_guard.py` — caps-only-shrink is a great pattern
- **`_envelope_to_adcp_error`** unifies REST and A2A reconstruction — proven byte-identical by roundtrip tests
- **`AdCPToolError` design** carrying envelope + status_code eliminates 4xx-as-5xx mislabeling
- **`ContextManager.fail_step`** byte-identical-by-construction invariant with matching test
- **`_serialize_context`** shallow-copy defense against aliasing footguns, with explicit mutation test

KonstantinMirin (commented):
## Code Review: PR #1306 — Final Round

All prior feedback (2 blockers + 6 majors) verified as resolved. The two-layer envelope architecture is solid — single source of truth via `build_two_layer_error_envelope()`, cross-transport symmetry fully achieved, DRY agent found zero violations and documented 9 improvements.

Four remaining items.

### Major

#### 1. `_ERROR_CODE_TO_STATUS` has phantom entries and missing wire codes

`src/routes/api_v1.py:57-80`

Three agents flagged this independently. The map mixes pre-translation codes, wire codes, and phantom codes:

| Entry | Problem |
|-------|---------|
| `CAPABILITY_NOT_SUPPORTED: 422` | Not a wire code — the actual code is `UNSUPPORTED_FEATURE` |
| `GONE: 410` | Pre-translation code — the wire code is `INVALID_STATE` |
| `RATE_LIMIT_EXCEEDED: 429` | Pre-translation code — the wire code is `RATE_LIMITED` |

Missing entries: `UNSUPPORTED_FEATURE: 422`, `INVALID_STATE: 410`, `RATE_LIMITED: 429`, `INVALID_REQUEST: 400`.

Low blast radius since `AdCPToolError` bypasses the map entirely, but a safety net with wrong entries is worse than documenting it as intentionally incomplete.

#### 2. `_adcp_to_a2a_error` is dead code

`src/a2a_server/adcp_a2a_server.py:143-166`

Defined and updated (now builds two-layer envelope in its `data` dict) but never called from any production code path after the A2A refactor. `_handle_explicit_skill` now re-raises `AdCPError` directly. Either remove or mark `# NOTE: currently unused`.

#### 3. `extract_error_info` widens `RecoveryHint` to `str | None`

`src/core/tool_error_logging.py:109-110`

`AdCPError.recovery` is `RecoveryHint` (a `Literal`), never `None`. The return type `tuple[str, str, str | None]` forces a `type: ignore[assignment]` downstream in `_handle_tool_error`. Using `RecoveryHint | None` as the third element would eliminate the suppression.

#### 4. Lazy import of `AdCPValidationError` in `media_buy_list.py`

`src/core/tools/media_buy_list.py:258`

Inside an `except ValidationError` block. The module already imports from `src.core.exceptions` at top level. No circular dependency risk. Move to top-level imports for consistency with the rest of the PR.

### What's working well

- **Single source of truth** for wire shape via `build_two_layer_error_envelope()` — enforced by AST structural guard with transitive call analysis
- **A2A asymmetry eliminated** — ValueError/PermissionError now wrap-and-re-raise to typed AdCPError subclasses, catch-all removed
- **REST defensive catches removed** — all 10+ per-route `try/except ToolError` blocks replaced by global exception handlers
- **DRY: `_handle_tool_exception`** extracts the 15-line shared error handling path from sync/async wrappers
- **DRY: `assert_envelope_shape`** consolidates 4 per-boundary assertion helpers into one canonical helper
- **Per-file ratchet guards** with caps-only-shrink enforcement for both Pattern A and ValueError migration
- **`_envelope_to_adcp_error`** unifies REST and A2A reconstruction — proven byte-identical by roundtrip tests

KonstantinMirin:
## Code Review: PR #1306 — Structural Follow-up

The R3 sign-off was correct for the dimensions we'd been tracking. A deeper
structural pass surfaced a cluster of issues centered on the A2A boundary
plus a few smaller items worth resolving in this PR rather than letting the
substrate ship with known gaps.

### Critical

**1. Five A2A skill handlers bypass `build_two_layer_error_envelope`**

`src/a2a_server/adcp_a2a_server.py:1591-1614` (`_handle_create_media_buy_skill` — both `missing_params` and `ValidationError` branches), `:1651-1657` (`_handle_sync_creatives_skill`), `:1735-1741` (`_handle_create_creative_skill`), `:1786-1792` (`_handle_assign_creative_skill`), `:2090-2096` (`_handle_update_performance_index_skill`).

The first two early returns call `adcp_error()` from the SDK and merge it into a custom dict — but that helper only emits the `errors[]` half, never the envelope-level `adcp_error` key. Per the docstring at `exceptions.py:502-505`, a payload missing either layer makes the storyboard runner synthesize `MCP_ERROR` and erase the real code.

The other four are flatter still — no `adcp_error()` call at all, just `{\"success\": False, \"message\": …, \"required_parameters\": …}`. `_serialize_for_a2a` (line 1374) derives `success` from `errors[]` presence, so these flat dicts are classified as successful Tasks carrying fake \"success\" artifacts. Buyers cannot see the real error code on any of these five paths.

**Fix:** replace each early-return dict with `raise AdCPValidationError(...)`. The outer dispatcher at `:653` already calls `_build_failed_skill_result` → `_build_error_envelope` → the builder. Mirrors the round-4 fix on `media_buy_list.py:258`.

### High

**2. `_handle_explicit_skill` audit-logs only on the AdCPError branch**

`src/a2a_server/adcp_a2a_server.py:1473-1508`. The `AdCPError` branch calls `_log_a2a_operation(...)` when `tenant_id` is present (lines 1481-1488). The `ValueError` and `PermissionError` branches do `logger.error(f\"…\")` only — no audit log, no activity feed. R1's \"different log destinations\" finding only got half-resolved.

**Fix:** hoist `_log_a2a_operation()` above the per-type branches (catch `Exception`, log, then dispatch by type), or duplicate the audit call into both ValueError and PermissionError branches.

**3. REST exception handlers perform zero logging**

`src/app.py:105-189`. `_envelope_response`, `adcp_error_handler`, `value_error_handler`, `permission_error_handler`, and `tool_error_handler` all return JSON without logging anywhere — no Python logger, no audit log, no activity feed. A 4xx returned from any REST endpoint leaves no breadcrumb. Identity is available on `request.state` via `UnifiedAuthMiddleware`.

**Fix:** add `logger.warning(\"REST boundary translating %s to envelope: %s\", type(exc).__name__, exc.message)` and a tenant-aware audit-log call inside `_envelope_response()`.

**4. Structural guard pins the dead-code helper, not the production path**

`tests/unit/test_architecture_error_envelope_two_layer.py:24-28`. `BOUNDARY_FUNCTIONS` lists `_adcp_to_a2a_error`, which the PR's own docstring at `adcp_a2a_server.py:146-151` marks \"currently unused by production code paths.\" The real production A2A envelope path is `AdCPRequestHandler._build_error_envelope` (`adcp_a2a_server.py:204-235`), called from `on_message_send` at `:663` and `:669`. The guard would not catch a refactor that broke `_build_error_envelope`.

**Fix:** add `(\"src/a2a_server/adcp_a2a_server.py\", \"_build_error_envelope\")` to `BOUNDARY_FUNCTIONS`. The `_collect_module_functions` helper needs a small extension to index `FunctionDef` nodes inside `ClassDef` (currently module-level only).

### Medium

**5. \"Byte-identical\" claim isn't supported by any byte-level test**

Docstrings at `test_error_envelope.py:6`, `test_context_manager_fail_step.py:54-72`, `context_manager.py:347-351`, and `harness/_base.py:807-809` describe \"byte-identical\" agreement between REST and A2A reconstruction (and between `fail_step` persistence and the wire envelope). In every case the assertion is in-memory dict equality — `parse_rest_error` and `_envelope_to_adcp_error` end up calling the same function (`harness/_base.py:815, 246`), so `rest_exc.error_code == a2a_exc.error_code` is tautological.

**Fix:** add a test that captures wire bytes from a TestClient REST response and an A2A failed-Task DataPart and asserts `json.dumps(rest, sort_keys=True) == json.dumps(a2a, sort_keys=True)`. The current tests are fine — they just don't prove what the docstrings claim.

**6. `extract_error_info` returns unvalidated `str` typed as `RecoveryHint | None`**

`src/core/tool_error_logging.py:107-109`. The envelope branch does `return first[\"code\"], first.get(\"message\", \"\"), first.get(\"recovery\")` — `dict.get` returns `str | None`, but the signature declares `RecoveryHint | None`. The legacy `ToolError` branch at lines 126-131 correctly validates against the three valid values; the envelope branch should match.

**7. Per-boundary assertion wrappers survived the consolidation**

`tests/unit/test_error_boundary_translation.py:34-49` (`_assert_mcp_envelope`, `_assert_a2a_envelope`, `_assert_rest_envelope`) and `test_adcp_exceptions.py:293-300` (`_assert_two_layer_envelope`) are pure-rename wrappers around `assert_envelope_shape`. Only `_assert_mcp_envelope` has added behavior (an `isinstance` precondition). The canonical helper's docstring at `tests/helpers/envelope_assertions.py:14-17` claims \"Replaces the per-boundary helpers\" — not yet true.

**Fix:** delete `_assert_rest_envelope` and `_assert_two_layer_envelope`. Either inline the MCP precondition at the 3 call sites or add `check_mcp_tool_error: bool = False` to `assert_envelope_shape`.

**8. `_ERROR_CODE_TO_STATUS` duplicates per-class `status_code` declarations with two real conflicts**

`src/routes/api_v1.py:57-81`. The PR's own comment at `:55-56` acknowledges the mirroring. Two wire codes are ambiguous across subclasses: `AUTH_REQUIRED` maps to 401 in the table but `AdCPAuthorizationError.status_code=403`; `SERVICE_UNAVAILABLE` maps to 503 but `AdCPAdapterError.status_code=502`. Plain-ToolError raises from authorization code surface as 401 instead of 403.

**Fix:** generate the table at module load — walk `AdCPError.__subclasses__()`, build `{cls.error_code: cls.status_code}`. Eliminates manual maintenance and drift potential.

**9. A2A skill handlers carry ~15 copies of the same try/except skeleton with diverging error prefixes**

`src/a2a_server/adcp_a2a_server.py` — every `_handle_*_skill` repeats `try: … except AdCPError: raise; except Exception as e: logger.error(...); raise InternalError(message=…)`. Each handler picks a different prefix (\"Unable to retrieve\", \"Failed to\", \"Error in\") — wire messages diverge for semantically identical failures.

**Fix:** `@_a2a_skill(\"skill_name\")` decorator that wraps the body in the canonical try/except. Removes ~200 lines and standardizes the messages.

**10. REST imports `AdCPToolError` (an MCP-boundary type) for cross-transport unwrapping**

`src/routes/api_v1.py:28, 84-104`. `AdCPToolError` subclasses `fastmcp.exceptions.ToolError` and is MCP-specific. The import predates this PR, but the substrate is the right moment to fix it — PR 2's `Error(code=...)` cleanup will amplify REST-bound traffic through MCP wrappers.

**Fix:** move `AdCPToolError` to `src/core/exceptions.py` (it's already an envelope carrier, not an MCP implementation detail), or have REST check via duck-typing (`hasattr(e, \"envelope\") and hasattr(e, \"status_code\")`).

**11. `ContextManager.fail_step` has zero production callers**

`src/core/context_manager.py:339-366`. Ships with unit tests and an aspirational docstring at `exceptions.py:498-500` (\"Boundary translators … AND ContextManager.fail_step both call this\") — but `grep -rn \"fail_step\b\" src/` returns only the definition. Production failed-workflow persistence still flows through `src/services/order_approval_service.py` and the GAM background polling path.

**Fix:** wire one production caller in this PR. `order_approval_service._mark_approval_failed` is the obvious candidate — it already builds the same wire pieces. Without a real caller, the byte-identical-by-construction claim has nothing to enforce.

### Low

| # | File:line | Item |
|---|-----------|------|
| 12 | `test_architecture_error_envelope_two_layer.py:47, 73` | Docstring says \"1-level transitive\" but implementation walks N levels with cycle detection. Tighten doc. |
| 13 | `test_architecture_error_code_compliance.py:26-29`, `test_architecture_error_envelope_two_layer.py:77` | CWD-relative scan paths silently pass with nothing to scan if pytest runs from a subdirectory. The two new guards use `Path(__file__).resolve().parents[2]` — adopt the same pattern here. |
| 14 | `routes/api_v1.py:84-104` + `app.py:187-189` | `_handle_tool_error` is defined in `api_v1.py` but called only from `app.py`. Move to a neutral module and drop the `_` prefix. Stale comment on line 87 references a defensive pattern that no longer exists. |
| 15 | `harness/_base.py:217-220` | Legacy-flat-shape fallback in `_envelope_to_adcp_error` should be dead on production envelopes. Remove or mark deprecated with an audit timeline. |
| 16 | `adcp_a2a_server.py:658-662` vs `:1479` | Same AdCPError logged twice — once at `warning` with `%s`, once at `error` with f-string. Drop the duplicate in `_handle_explicit_skill`; the outer dispatcher already logs. |
| 17 | `tool_error_logging.py:128-131` | Hard-coded `(\"transient\", \"correctable\", \"terminal\")` tuple duplicates `RecoveryHint = Literal[...]`. Use `typing.get_args(RecoveryHint)`. |
| 18 | `_per_file_cap_guard.py:35-36, 63-94` | Docstring promises `# noqa: structural-guard` filter; not implemented. Implement the filter or remove the claim. |
| 19 | `tool_error_logging.py:198-210` | Bare `raise` instead of `raise error` in a `(error: Exception) -> NoReturn` function — works in current call site, breaks if anyone calls it outside an active `except`. |
| 20 | `routes/api_v1.py:95-96` | `e.envelope` returned without dict-copy. Breaks the dict-copy discipline established by the envelope builder at `exceptions.py:540`. |
| 21 | `tool_error_logging.py:46` | `AdCPToolError.__init__(envelope, status_code=500)` default invites silent 4xx-as-5xx misclassification. Make `status_code` keyword-only with no default. |
| 22 | `routes/api_v1.py:99-103` | `synthetic = AdCPError(error_message); synthetic.error_code = …; synthetic.status_code = …` — mutating after construction is an implicit API the class doesn't document. Either accept the kwargs in `__init__` or pick the typed subclass via a lookup. |
| 23 | `tool_error_logging.py:222-227` | `hasattr(arg, \"tenant_id\")` matches any object with that attribute (incl. Pydantic models). Use explicit `isinstance(arg, (FastMCPContext, ToolContext))`. |
| 24 | `app.py:170` | Mid-file `from fastmcp.exceptions import ToolError  # noqa: E402` — hoist with the other fastmcp import at line 18. |
| 25 | `app.py:187` | Lazy import reaches into private `_handle_tool_error` from another module. Move the helper to a neutral shared module and shed the `_` prefix. |

### Cross-cutting themes

Three patterns appeared across multiple dimensions:

- **A2A boundary is where the architecture is weakest.** Items 1, 2, 4, 9, and 16 all touch the same module. The single-source-of-truth claim holds for MCP and REST; A2A has 5 bypass paths, asymmetric audit logging, and a guard pointed at dead code.
- **Substrate shipped ahead of its production callers in two places.** `_adcp_to_a2a_error` (item 4) and `ContextManager.fail_step` (item 11) both ship with unit tests + aspirational docstrings + zero live callers. Acceptable for a \"PR 1 of 3\" substrate — but only if a real caller lands in this PR.
- **Duplicate sources of truth survived the consolidation.** Items 6, 7, 8, 17 — `_ERROR_CODE_TO_STATUS` duplicates `status_code` class attributes, per-boundary wrappers duplicate `assert_envelope_shape`, the recovery-hint whitelist duplicates `RecoveryHint`. Each is small alone; together they undercut the framing.

### What's still working well

- `build_two_layer_error_envelope()` is a pure-data function returning `dict[str, Any]` — `src/core/exceptions.py` has zero FastMCP/A2A/Starlette imports.
- Exception chaining (`raise ... from e`) preserved at every conversion site.
- Async wrapper in `with_error_logging` correctly distinguishes sync/async, awaits properly, and lets `CancelledError` propagate (BaseException, not Exception).
- Per-file ratchet design is real — caps can only shrink, new files start at zero, deleted files force cap removal. Verified by tracing every drift path.
- The two new structural guards (`no_error_construction_in_impl`, `no_value_error_in_impl`) correctly scan via `ast.walk()` over the full module tree — class methods and helpers are caught, not just `*_impl`-named functions.

The work concentrated in the A2A module plus the four small high-level items resolves the substrate cleanly. Happy to pair on any of these.

KonstantinMirin (commented):
Production code looks good — all the architectural fixes are in place and the structural guards provide a safety net. The issue is test coverage.

About half the error paths introduced here are only covered by unit tests that mock the transport layer. Those tests verify that your mock returns what you configured it to return — they don't prove the path actually works. If there's a wiring bug between the handler and the dispatcher, or an argument gets dropped crossing the boundary, the unit tests pass green anyway.

**The pattern that works** is already in this PR — `test_a2a_error_responses.py:164` (`test_create_media_buy_validation_error_includes_errors_field`). It sends a real `on_message_send`, gets a real `Task` back, extracts the artifact, and asserts the envelope shape. That test would catch a real wiring bug. Replicate it for the missing scenarios:

1. **A2A handlers missing coverage:** `sync_creatives`, `create_creative`, `assign_creative` — same pattern as `create_media_buy` at line 164. Send a message with missing required params, assert the failed Task artifact carries `adcp_error.code` + `errors[]`.

2. **MCP error envelope:** Zero tests go through real MCP transport and check the envelope. The harness supports `env.call_via(Transport.MCP)` — see `test_creative_formats_ordering.py:21` for the pattern. Trigger an error (e.g., nonexistent media_buy_id) and assert the error has the two-layer shape.

3. **Context echo on the wire:** Send a request that triggers an error carrying context, assert `context` appears in the response JSON.

The REST boundary is well covered via TestClient — that's the bar to match for A2A and MCP.

**On the ValueError/PermissionError wrapping:** The guard's own docstring (`test_architecture_no_value_error_in_impl.py:1-12`) documents this as transitional — "PR 2 sub-batches drain these." I'm not asking for integration tests of the wrapping itself. The real priority is migrating the 26 boundary-facing `raise ValueError` sites in `media_buy_create.py` to typed AdCPError raises so they carry proper codes and recovery hints. Until then, the boundary catch is a safety net, not a contract.

KonstantinMirin:
## Review Round 6: Error Verification Policy + 7 Fixes

### Policy

The test harness reconstructs `AdCPError` subclasses from wire responses so tests can use `isinstance()`. This reconstruction is lossy — `AdCPAuthenticationError` and `AdCPAuthorizationError` both map to `AUTH_REQUIRED` on the wire, so reconstruction always picks `Authentication`. Tests asserting on reconstructed objects verify the reconstruction layer, not the wire shape.

**New policy**: wire-envelope assertions are the authority for error verification. `TransportResult` now exposes `wire_error_envelope: dict | None` — the raw two-layer envelope from the wire. New error tests must assert on this via `assert_envelope_shape()`. Existing tests (~660 call sites) are unchanged and migrate incrementally.

Full policy documented in `tests/CLAUDE.md` § "Error Verification Policy", referenced from main `CLAUDE.md`.

### Fixes

**CRIT-01**: Test duplication baseline exceeded (guard boilerplate)
- Extracted `REPO_ROOT`, `SCAN_DIRS`, `rel()`, `safe_parse()` into `tests/unit/_ast_helpers.py`
- Both cap-guard files now import shared helpers instead of duplicating

**CRIT-02**: `raise AdCPValidationError` at `media_buy_create.py:698` escaped `except (ValueError, ValidationError)` catch
- Reverted to `raise ValueError` — this is inside `execute_approved_media_buy` (non-`_impl` approval helper), not a boundary function
- ValueError cap updated to 3 with inline justification

**HIGH-01**: Production `fail_workflow_step_for_exception` emitted one-layer `{"errors": [...]}` while `fail_step` (zero callers) emitted correct two-layer
- Fixed `fail_workflow_step_for_exception` to use `build_two_layer_error_envelope` — webhook subscribers now see the same two-layer shape as synchronous callers
- Deleted `fail_step` (zero production callers) and `test_context_manager_fail_step.py`
- Updated `test_context_manager_fail_workflow_step.py` expected shape

**HIGH-02**: Dead code cluster (`_adcp_to_a2a_error` + backward-compat machinery)
- Deleted `_adcp_to_a2a_error()` from `adcp_a2a_server.py` (zero production callers since B4)
- Deleted 3 test methods that exercised it directly, trimmed 2 roundtrip tests
- Removed `check_backward_compat` parameter from `assert_envelope_shape`
- Removed legacy-flat fallback from `_envelope_to_adcp_error` in harness
- Updated stale doc reference in `ARCHITECTURE_TRANSFORMATION_REPORT.md`

**HIGH-03**: `xfail(strict=False)` on the only E2E lifecycle test
- Changed to `strict=True` so the marker is a tracked known failure, not a permanent silencer

**HIGH-04**: ValueError/PermissionError normalization repeated at 3 boundaries
- Extracted `normalize_to_adcp_error()` into `src/core/exceptions.py`
- MCP (`tool_error_logging.py`) and A2A (`adcp_a2a_server.py`) now call it
- Also fixed f-string logging → `%s` formatting in A2A boundary

**HIGH-06**: MCP unwrapper duplicated `_envelope_to_adcp_error`
- `_unwrap_mcp_tool_error` now delegates to `_envelope_to_adcp_error` for JSON envelope parsing

### Migration Path

- `TransportResult.wire_error_envelope` is additive — defaults to `None`, zero existing breakage
- All 4 dispatchers (IMPL, A2A, REST, MCP) capture the wire envelope on error
- `result.error` (reconstructed exception) remains available and unchanged
- New error tests should use `assert_envelope_shape(result.wire_error_envelope, code, recovery=...)`
- Existing tests migrate when touched (boy-scout rule)

### Verification

`make quality` passes: format, lint, mypy, duplication (src/37, tests/99), 4666 unit tests, 0 failures.

KonstantinMirin:
**Full suite verification** (`./run_all_tests.sh`): all green.

| Suite | Passed | Total |
|-------|--------|-------|
| unit | 4788 | 4821 |
| integration | 1903 | 1951 |
| e2e | 90 | 120 |
| bdd | 1506 | 4265 |
| admin | 79 | 82 |
| **Total** | **8366** | **0 failures** |

xfailed=2820, skipped=49 — all pre-existing. Results at `test-results/250526_1244/`.

Code at prebid/salesagent#1359 (from fork, targets this branch).

KonstantinMirin:
## Review Round 7: Reinvented Wheels — Single-Source-of-Truth Violations

Most of the prior six rounds landed cleanly across the 17 follow-up commits — `normalize_to_adcp_error` consolidation, `@_a2a_skill` decorator, auto-derived `_ERROR_CODE_TO_STATUS`, deletion of `_adcp_to_a2a_error` / `fail_step`, REST status-code map, byte-identity test, `assert_envelope_shape` helper. The architecture pieces are in place. What's left is a pattern: the PR introduces canonical helpers, then bypasses them at the call sites. Each bypass is a future drift point; the consolidations cost more than they save until adoption is complete.

A couple of these originated in my round-6 sub-PR (#1359 → `a5368ee2`) — calling those out below so the attribution isn't anonymous. Same bar applies; flagging them here rather than splitting them out because the substrate ships as one piece.

### Blockers

#### 1. `AdCPRequestHandler._build_error_envelope` reimplements a strict subset of `normalize_to_adcp_error`

`src/a2a_server/adcp_a2a_server.py:200-216`

```python
if not isinstance(exc, AdCPError):
    exc = AdCPError(str(exc) or type(exc).__name__)
return build_two_layer_error_envelope(exc)
```

`normalize_to_adcp_error` (`src/core/exceptions.py:553-568`) is the single source of truth — I extracted it in PR #1359 to deduplicate the three-boundary mapping and missed wiring this site through it. The helper maps `ValueError → AdCPValidationError` and `PermissionError → AdCPAuthorizationError`. `_build_error_envelope` doesn't. An untyped `ValueError` escaping a non-decorated A2A path lands here and surfaces as `SERVICE_UNAVAILABLE`/500 with `recovery="terminal"` — buyer agents see a non-retryable adapter error for what should be a `VALIDATION_ERROR`/400 / `correctable`. Two-line fix: `return build_two_layer_error_envelope(normalize_to_adcp_error(exc))`.

#### 2. REST `_envelope_response` skips audit-log and activity-feed

`src/app.py:106-130`

The docstring at `:114-118` explicitly claims symmetry "with the A2A boundary … which also normalizes ValueError/PermissionError, logs, **and audit-logs uniformly**". The implementation calls `logger.warning` only. No `get_audit_logger().log_operation(success=False)`, no `activity_feed.log_error()`. MCP does both, A2A does audit-only, REST does neither. R4 #3 from the structural follow-up identified this; the fix added `logger.warning` and stopped. The documented contract is not honored — a 4xx on any REST route leaves no audit row.

Extract `record_boundary_error(transport, tool_name, exc, identity)` (calls into both observability sinks) and have all three boundaries delegate to it. Identity is already on `request.state` via `UnifiedAuthMiddleware`.

#### 3. NL `_create_media_buy` emits a flat dict — bypasses the envelope architecture entirely

`src/a2a_server/adcp_a2a_server.py:2107-2135`, serialized at `:880`

```python
return {
    "success": False,
    "message": f"Authentication successful for {principal_id}. To create a media buy...",
    "required_fields": [...],
    ...
}
```

No `adcp_error` key, no `errors[]`. Goes straight to `Part(data=_dict_to_value(result))` without passing through any translator. The two-layer-envelope structural guard doesn't fire because the function returns instead of raising — exactly the wire-shape leak the PR was designed to close. Storyboard runners reading this DataPart synthesize `MCP_ERROR`. Also: line 2138 catches and emits `InternalError(message=f"Authentication failed: {str(e)}")` from inside media-buy code — misleading prefix.

Pre-dates this PR (originally from PR #109) but the substrate-establishment PR is where this gets closed. Replace the body with `raise AdCPValidationError("Natural-language create_media_buy is not supported; use explicit skill invocation")`, or delete the NL stub — the explicit-skill path is the spec contract.

#### 4. `TransportResult.wire_error_envelope` is synthesized for A2A and IMPL, not extracted from the wire

`tests/harness/dispatchers.py:29-78`

This is the field + policy I introduced in PR #1359, and it has a gap. `tests/CLAUDE.md` § "Error Verification Policy" instructs new error tests to assert on `wire_error_envelope` as the canonical authority. `RestDispatcher:110` captures `response.json()` from the actual HTTP body. `McpDispatcher:132` parses the JSON string in `str(ToolError)`. Both are real wire bytes.

`ImplDispatcher:62` and `A2ADispatcher:77` populate the field via `_envelope_from_adcp_error(exc) → build_two_layer_error_envelope(exc)` on the locally caught exception object. That's the same function the production boundary calls — but here it runs in test code, on the exception the test set up. A regression in the production A2A `_build_error_envelope` (Blocker #1 above) or in `_build_failed_skill_result`'s DataPart serialization would not be caught by `result.wire_error_envelope` assertions.

The policy claim is false for A2A. Either gate `wire_error_envelope` to only the dispatchers that capture real wire bytes (rename the synthesized variant to `synthesized_error_envelope` so test authors don't confuse the two), or drive `A2ADispatcher` through `on_message_send` + artifact-DataPart extraction. The pattern is already implemented in `tests/integration/test_a2a_error_responses.py:245+` — copy it. Same fix needs the `tests/CLAUDE.md` policy text tightened so the next contributor doesn't lean on a field that doesn't deliver what it advertises for two of the four transports.

### Major

#### 5. Tests reinvent helpers the PR itself introduces

This is the unifying pattern. Each item below has a canonical helper that exists in this PR; the call site ignores it.

| Reinvented at | Canonical helper | Count |
|---|---|---|
| Inline `ResolvedIdentity(...)` construction across A2A tests | `PrincipalFactory.make_identity()` (tests/CLAUDE.md mandates) | 28 sites; zero factory calls |
| Inline `Client(mcp) → patch identity → call_tool → parse content` | `_call_mcp_tool_capturing_envelope` (defined in same file at `test_mcp_error_envelope.py:137-165`) | 3 sites in that one file |
| Hand-inspected envelope keys: `envelope["adcp_error"]["code"]`, `envelope["errors"][0]["recovery"]`, etc. | `assert_envelope_shape()` | `test_error_format_consistency.py:621-632`, `test_a2a_brand_manifest.py:163-165`, `test_error_paths.py:502-503, 522-523` |
| REST exception handlers' `value_error_handler` / `permission_error_handler` inline the type→AdCPError mapping | `normalize_to_adcp_error()` | 2 handlers in `src/app.py:155-181` |
| Per-test `async def _call(): ... asyncio.run(_call())` wrappers | Harness `env.call_mcp()` | ~7 sites in new wire tests |
| Non-skill A2A handlers carry 3 different `InternalError` prefixes | `@_a2a_skill` decorator (or equivalent) | push-notif handlers `:1066-1336` (4 sites), `on_message_send:992`, NL `_create_media_buy:2138` |

A consolidation that ships unused is debt, not progress — the next contributor sees both the canonical helper and the inline version, can't tell which is current, and adds a third variant. The DRY invariant in CLAUDE.md is "non-negotiable correctness" — it applies to test code at the same bar.

The `ResolvedIdentity` row is the strongest case: 28 inline constructions, zero factory calls, no review caught it through six rounds because each individual test file looks fine in isolation. The pattern predates this PR but the new A2A tests added in `49da5b0f` and `1fec57a5` mirrored the existing inline shape rather than adopting the factory — debt grew. Add a structural guard (`tests/unit/test_architecture_a2a_test_uses_factory.py`) that fails on any `ResolvedIdentity(` inside `tests/integration/test_a2a*.py` / `tests/unit/test_a2a*.py` outside the factory itself, and convert the 28 sites in this PR.

#### 6. `_translate_to_tool_error` has byte-identical arms under a dead `if`

`src/core/tool_error_logging.py:228-233`

```python
typed = normalize_to_adcp_error(error)
if typed is not error:
    raise AdCPToolError(build_two_layer_error_envelope(typed), status_code=typed.status_code) from error
raise AdCPToolError(build_two_layer_error_envelope(typed), status_code=typed.status_code) from error
```

Same call, same kwargs, same `from error` on both arms. Per the DRY invariant, this is a defect — bug fixes in one branch will miss the other. Collapse to a single `raise`. (My #1359 — I left the dead `if` in when I migrated this site to `normalize_to_adcp_error`.)

#### 7. Workflow-audit-on-failure pattern duplicated across `media_buy_create` and `media_buy_update`

`src/core/tools/media_buy_create.py:3830-3845` and `src/core/tools/media_buy_update.py:1440-1457`. Identical `except AdCPError + except Exception → fail_workflow_step_for_exception → raise` skeleton. `media_buy_create` adds Slack/audit on the untyped branch — that's a genuine difference. The `fail_workflow_step_for_exception` invocation on both branches of both files is verbatim.

Extract a `with ctx_manager.audit_step_failure(step):` context manager (or a decorator on `_impl` functions that own a workflow step). Closes the duplication and makes "what happens when an `_impl` fails" a single place to read.

#### 8. `validate_*_error → if x: raise AdCPValidationError(x)` ladder repeats 4 times

`src/core/tools/media_buy_create.py:2090-2159`. Either inline a one-liner helper in the module, or have the four `validate_*` functions raise directly and drop the `str | None` return + manual re-raise. The current return-then-raise pattern is itself a small DRY violation — two ways of saying "validation failed".

### Minor

#### 9. `_log_tool_error` swallows audit-feed failures at DEBUG level

`src/core/tool_error_logging.py:189-190, 205-206`. CLAUDE.md "No Quiet Failures" rule. Production runs INFO/WARNING; DEBUG-level swallow is invisible. Bump to `logger.warning("Failed to log %s error to audit log: %s", tool_name, e)`. The audit failure shouldn't propagate (original error is the caller's concern), but it must be findable.

#### 10. A2A `_log_a2a_operation` call sits outside try/except — DB outage replaces buyer-facing exception

`src/a2a_server/adcp_a2a_server.py:1480-1497`. The audit call runs before re-raise; if it raises, the original `AdCPValidationError` is replaced by the audit failure. The correct defensive pattern is already in `context_manager.fail_workflow_step_for_exception:382-389` — wrap the audit call in `try/except Exception: logger.exception(...)`. A2A should mirror it.

#### 11. `fail_workflow_step_for_exception` wire-code sanitization drops `details`/`field`/`suggestion`/`context`

`src/core/context_manager.py:367-373`

```python
if wire_code not in STANDARD_ERROR_CODES:
    source = AdCPError(source.message or str(source), recovery="terminal")
    source.error_code = "SERVICE_UNAVAILABLE"
```

Only the wire code needs sanitizing per the docstring contract; the structured payload should survive. Webhook subscribers lose request-correlation context. Preserve `details`/`field`/`suggestion`/`context` on the replacement. (My #1359 — I rebuilt the AdCPError to enforce the wire-code whitelist and didn't carry the structured fields across.)

#### 12. `_ERROR_CODE_TO_STATUS` built at import time — fragile against late-imported subclasses

`src/core/tool_error_logging.py:372`. `AdCPError.__subclasses__()` is walked once at module import. A future subclass defined in a lazy-loaded module is invisible and falls through to 500. Not exploitable today (the eager-import boot order covers it). Add a structural guard that asserts every `AdCPError` subclass appears in the table at test time — matches the architecture-guard pattern the PR already uses.

#### 13. `_impl` boundary precision regressions — three remaining `raise ValueError` sites

| File:line | Should be | Why |
|---|---|---|
| `media_buy_create.py:1763` | `AdCPValidationError("Unexpected start_time type...", field="start_time")` | Line 1769 (6 lines later) already raises typed for the symmetric "Invalid start time" case |
| `media_buy_update.py:196` | `AdCPAuthRequiredError("principal_id missing from authentication context")` | Line 189 (7 lines above) raises typed `AdCPAuthRequiredError` for the symmetric `identity is None` guard |
| `media_buy_update.py:282` | `AdCPError("Failed to create or get persistent context", error_code="WORKFLOW_CREATION_FAILED")` | `WORKFLOW_CREATION_FAILED` is already declared at `exceptions.py:91`; surfaces as `SERVICE_UNAVAILABLE`+`transient` so buyer retries instead of editing the request |

All three are caught by `normalize_to_adcp_error` at the boundary and produce a generic `VALIDATION_ERROR` envelope. The wire contract holds — the buyer-facing precision degrades.

#### 14. Boundary log-level divergence

`src/app.py:120` REST=`WARNING`+no exc_info; `tool_error_logging.py:173` MCP=`WARNING` on no-tenant path only (typed path is silent); `adcp_a2a_server.py:183` A2A decorator=`ERROR`+exc_info; `:1480` A2A typed=`ERROR`+no exc_info. Operators tailing logs see fragmented severity for identical failures. Pick one convention — recommend `ERROR`+`exc_info=True` for untyped, `WARNING` for typed — apply across all three boundaries.

#### 15. Test classes overclaim coverage they don't provide

`tests/unit/test_error_boundary_translation.py:354-432, 1080-1122`. Class `TestA2ABoundaryAdCPErrorTranslation` and `test_a2a_roundtrip_all_subclasses` docstrings promise "envelope roundtrip" / "the A2A boundary"; the tests stop at `_handle_explicit_skill` and assert on Python exception attributes. The real A2A boundary is `on_message_send → DataPart on Task artifact` — none of these tests touch it. Either rename to `TestA2AHandlerExplicitSkillReraises` or extend the tests to drive `on_message_send`. The wire-level pattern is already correct in `tests/integration/test_a2a_error_responses.py`.

Similarly, `test_envelope_shape_matches_typed_branch` at `:492-505` uses `set(typed.keys()) == set(untyped.keys())` — passes if values are `None`. Add value-shape assertions or fold into `assert_envelope_shape`.

#### 16. `test_serialize_for_a2a_passes_dict_through` ratifies a non-spec wire shape

`tests/unit/test_error_format_consistency.py:543-568`. Pins behavior that A2A handlers may return error dicts with no `adcp_error` key. The handlers don't do this anymore (per `:202-224` they raise typed AdCPError). Audit the production caller: if none exists, delete the test (it's pinning dead code); if one exists, file an obligation to migrate. Currently the test legitimizes a gap that no other test catches.

### Nits

| File | Note |
|---|---|
| `tests/harness/_base.py:36-115` | `_CODE_TO_CLASS` accepts `NOT_FOUND` — `INTERNAL_CODES`-only. Wire leak rebuilds the right Python class silently. Remove or assert `error_code not in INTERNAL_CODES` before reconstruction. |
| `tests/harness/_base.py:152-179` | `ast.literal_eval` fallback for legacy ToolError tuple-strings — no production raise sites remain. Set a deprecation horizon. |
| `tests/integration/test_a2a_error_responses.py:138-194` | Pattern A AUTH_REQUIRED test asserts only `errors[]` — neither asserts `adcp_error` is present nor absent. Pin one side. |
| `adcp_a2a_server.py:1480` | Lone `%s`-format logger call in a file otherwise full of f-strings. Pick one. |
| `exceptions.py:163-192` | `recovery`/`error_code`/`status_code` shadowed by instance attrs — `ClassVar` is the documented Python idiom for "class default, instance override". |
| `exceptions.py:175-192` | `AdCPError.__init__(error_code=..., status_code=...)` kwargs mutate class-level attrs. Only `handle_tool_error`'s plain-ToolError synthesis uses this — narrow with `@classmethod synthesize(...)`. |
| `exceptions.py:117-140` | `_serialize_context` falls through to `.model_dump()` with no `isinstance(BaseModel)` guard — opaque `AttributeError` if a TypedDict is passed. |
| `tool_error_logging.py:47-53` | `AdCPToolError.__init__` eagerly `json.dumps(envelope)` — move to `__str__`; the harness `_unwrap_mcp_tool_error` round-trips the JSON anyway. |
| `app.py:94-101` | Four `# noqa: E402` for handler imports. Extract `register_exception_handlers(app)`. |
| `tool_error_logging.py:248-250` | `_ContextLike = (FastMCPContext, ToolContext)` rebuilt per error path. Cache via lazy getter. |
| `tests/harness/_base.py:191-225` | `_envelope_to_adcp_error` lossy: once `adcp_error` provides the code, `errors[0].recovery` / `errors[0].details` are never consulted as fallback. Fill missing fields from `errors[0]` after the first pass. |

### What's working well

- `build_two_layer_error_envelope` is a single source of truth in src/ (4 call sites, all boundary/persistence) — `_impl` never builds the envelope.
- `_ERROR_CODE_TO_STATUS` auto-derived from `AdCPError.__subclasses__()` — eliminates hand-maintained drift.
- `@_a2a_skill` decorator converged the 14 skill handlers cleanly. The remaining divergence is in the non-skill paths (#5 / #7), not the decorator itself.
- 24 of 27 boundary `ValueError` raises in `media_buy_create.py` migrated to typed AdCPError; the 3 remaining are documented above.
- `tests/integration/test_a2a_error_responses.py::TestA2AContextEcho` is the cleanest end-to-end pin of the spec-3.0.6 context-correlation field through the production dispatcher.
- `tests/integration/test_mcp_error_envelope.py` drives real MCP wire traffic — `test_create_media_buy_budget_too_low_emits_envelope_on_wire` is the model for wire-level error verification.
- `test_every_adcp_error_subclass_present_in_status_table` is a true mutation test — removing `error_code`/`status_code` from any subclass breaks it. The structural-guard category should follow this pattern (assert the derivation invariant, not the current values).

The architecture is sound. The remaining work is finishing the adoption of the helpers this PR introduces — once the call sites stop reinventing what the PR established, the substrate is done.

mkomorski (approved):

Issue #1305, opened 2026-05-15
feat: spec-conformant async/submitted response envelopes — Response3 shape, task_id population, status lifecycle alignment


ChrisHuie:
## TL;DR

  Salesagent emits async/manual-approval/pending-creatives responses using `CreateMediaBuyResponse1` (the synchronous success shape) wrapped with an outer `status: submitted` field, rather than the
  AdCP-spec submitted-variant envelope (`Response3`-style: `status, task_id, message, errors?, context?`). The spec-required `task_id` field is never populated — buyers cannot reliably poll, retrieve,
  or correlate async outcomes through standard AdCP mechanisms. Sibling problem: `media_buy_delivery.py` returns success envelopes carrying terminal errors. Also closes #1247 gap #12
  (`pending_creatives` → `pending_start` status transition).

  This is a spec-compliance + lifecycle-coherence issue, independent of the error-emission architecture work but in the same lifecycle area.

  ## Context

  This issue surfaced during investigations for the error-emission architecture refactor. Out of scope of that refactor because it's about *async success* envelope shape, not error emission. But the
  two are siblings — both involve the response variant a tool chooses on a given lifecycle state.

  ## Current behavior (verified)

  Six sites today produce out-of-spec async/submitted responses:

  1. **`src/core/tools/media_buy_create.py:2418-2428`** — pending-creatives approval branch. Returns `CreateMediaBuyResult(response=CreateMediaBuySuccess(media_buy_id, packages, workflow_step_id, ...),
   status=AdcpTaskStatus.submitted.value)`. The response is **Response1 shape** with `media_buy_id` already issued and an outer `status: submitted` marker added via the `Result` wrapper. No `task_id`.
  No spec-canonical `message` field.

  2. **`src/core/tools/media_buy_create.py:2577-2586`** — tenant/product auto-create disabled branch. Same Response1 + outer status pattern. `media_buy_id` issued early.

  3. **`src/core/tools/media_buy_update.py:321-351`** — adapter manual-approval branch. Returns `UpdateMediaBuySuccess(media_buy_id, affected_packages=[], workflow_step_id NOT set, ...)` without any
  AdcpTaskStatus wrapper. Buyer receives a success envelope with empty `affected_packages` — same root cause as #1041.

[...]

Issue #1303, opened 2026-05-14
feat: zero-inventory pre-validation on create_media_buy — block silent forever-DRAFT delivery per inventory_list_no_match storyboard


ChrisHuie:
## Context

  Parent: #1247 gap #8. Storyboard scenario: `media_buy_seller/inventory_list_no_match` from `@adcp/sdk@6.11.0/compliance/cache/3.0.6/protocols/media-buy/scenarios/inventory_list_no_match.yaml`.

  Today, when `create_media_buy` targeting resolves to zero inventory, salesagent silently creates the buy and sends it to the ad server. On GAM, `orders_manager.approve_order()`
  (`src/adapters/gam/managers/orders.py:199`) retries `ApproveAndOverbookOrders` **40 times × 15 seconds** on `NO_FORECAST_YET` — a 10-minute polling storm — before giving up and leaving the order in
  DRAFT indefinitely. Buyers see `create_media_buy` succeed, then nothing happens for 10 minutes, then nothing happens forever.

  This issue is the **pre-flight blocking sub-case** of #1028 (delivery forecasting). #1028 surfaces forecast data on `get_products` and `get_media_buy_delivery` (advisory). This issue rejects
  `create_media_buy` when targeting will not deliver (gating). They are complementary, not duplicates.

  ## What 3.0.6 expects (verbatim from storyboard)

  From `inventory_list_no_match.yaml`:

  > What the seller must NOT do: crash, return a misleading non-zero forecast, or silently drop the targeting and deliver against unintended inventory. **Each of those is a real failure mode we've seen
   in adapter integrations.**
  >
  > One of two acceptable outcomes:
  > 1. Buy accepted with zero-forecast reporting — status may be `pending_creatives` / `pending_start` / `active`, but the seller returns packages with forecast indicating zero deliverable inventory
  and a message explaining the list mismatch.
  > 2. Buy rejected with an informative error — typically `INSUFFICIENT_INVENTORY` or `INVALID_TARGETING` — including findings that identify which list(s) matched nothing.
  >
  > What is NOT acceptable: a silently-successful buy with normal forecast numbers, or a crash / non-AdCP error shape.

  **Note: the storyboard names `INSUFFICIENT_INVENTORY` / `INVALID_TARGETING`, but neither exists in the canonical 4.3 `ErrorCode` enum (45 codes verified).** This is a spec/SDK incoherence to flag
  upstream. Closest standard codes:

[...]

PR #1311, opened 2026-05-17
fix(gam): hand off NO_FORECAST_YET to background polling

ChrisHuie:
version reference => (3.0.1 → 3.0.6, 3.12.0 → 4.3.0)

- `execute_approved_media_buy` was calling `adapter.orders_manager.approve_order()` with no `max_retries`, so the default `40 × 15s` blocked the admin Flask request for up to **10 minutes** when
  GAM's inventory forecast wasn't ready (`NO_FORECAST_YET`).
  - Now: single synchronous attempt, then hand off to the existing `start_order_approval_background` polling service (2-min budget). Admin request returns immediately.
  - Background service learns to write `MediaBuy.status='active'/'failed'` on terminal transition + fire `AuditLogger.log_operation` (which forwards to the Slack audit channel when configured).
  Previously the buy was stranded at `pending_approval` after the background poll terminated.

  ## What changed

  **`src/core/tools/media_buy_create.py`** — `execute_approved_media_buy`
  - `approve_order(response.media_buy_id, max_retries=1)` (was: default 40)
  - On `False` return: kick off `start_order_approval_background` and set `background_owns_status = True`
  - Skips the synchronous `status='active'` write when the background path owns the terminal transition (prevents a race that could mask a GAM rejection)
  - Falls through to the existing synchronous `'active'` write when approval succeeded fast or the adapter has no `orders_manager` (mock adapter, etc.)

  **`src/services/order_approval_service.py`** — `_mark_approval_complete` / `_mark_approval_failed`
  - Open a `MediaBuyUoW` and call `update_status(media_buy_id, "active"|"failed")`
  - Fire `AuditLogger(adapter_name="GoogleAdManager").log_operation(operation="approve_order", ...)` with the order_id / attempts / duration in `details`
  - Both writes are wrapped in `try/except` so a status-write failure can't strand the SyncJob

  **`src/adapters/gam/managers/orders.py`** — `approve_order` defaults
  - `max_retries=40 → 12`, `poll_interval=15 → 10`, `@timeout(seconds=620) → 130`
  - Matches the background service's 2-minute budget
  - Docstring nudges long-wait callers toward `start_order_approval_background` so the next caller who forgets `max_retries=1` can't reintroduce the 10-min storm

  ## Tests

  New `tests/unit/test_b2_background_approval_polling.py` (7 cases):
  - `execute_approved_media_buy` calls `approve_order(max_retries=1)` — never the default
  - On `approve_order` returning `False`, `start_order_approval_background` is invoked exactly once
  - On the handoff path, the synchronous `status='active'` write is **deferred** (background service owns it)
  - On synchronous approval success, the existing `status='active'` write still fires
  - `_mark_approval_complete` writes `MediaBuy.status='active'` and fires audit log with `success=True`
  - `_mark_approval_failed` writes `MediaBuy.status='failed'` and fires audit log with `success=False` + error message

  Shared mock builders extracted to `tests/helpers/execute_approved_mocks.py` (satisfies the DRY pre-commit guard; `test_execute_approved_status_update.py` now imports them).

  ## Drive-by pre-commit hygiene

  Both pre-existing drift on `main` (each one is identical to PR #1306's planned sweep):

  - `.type-ignore-baseline`: `42 → 60` — count had drifted via PRs bypassing the hook
  - `.pre-commit-config.yaml` mypy hook `adcp` pin: `3.2.0 → 4.3.0` — matches `pyproject.toml` (`>=4.3.0`); eliminates phantom `AdCPPackageUpdate has no attribute media_buy_id` / `UpdateMediaBuyRequest
   has no attribute context` errors from `media_buy_update.py`

  Without these, the pre-commit gate refuses every commit to `src/` on main.

  ## Out of scope

  - **Buyer webhook URL is `None` on the handoff path.** `execute_approved_media_buy` doesn't currently receive a `push_notification_url`, so the background service can't notify the buyer. Tracked
  separately — PR #1305's async/submitted lifecycle work is the right home for full webhook plumbing.
  - **No async/submitted Response3 envelope** — that's PR #1305.
  - **No mypy fixes in `media_buy_update.py`** — those 262 errors are PR #1276's `property_targeting` work; resolving the adcp pin drift eliminates the phantoms but real attr errors remain there.

  ## Test plan

  - [ ] `make quality` passes (verified locally: 4481 unit pass, 19 xfailed, pre-existing `SyncCreativesRequest` schema-alignment failure confirmed unrelated)
  - [ ] `tox -e integration` clean
  - [ ] CI storyboard checks unchanged (B2 doesn't touch the storyboard surface)
  - [ ] Manual: trigger an admin "approve media buy" against a GAM tenant with insufficient inventory; admin redirect should be **immediate** (was 10 min); 2 minutes later, `MediaBuy.status='failed'` +
   audit log entry; Slack notification if `SLACK_*` env vars configured
  - [ ] Manual: same scenario but with sufficient inventory; admin redirect immediate; ~30s-2min later, `MediaBuy.status='active'`


KonstantinMirin (changes requested):
This fixes a real problem — 10 minutes blocking on an admin request is unacceptable. But after a deep-dive tracing the full execution flow, all callers, and every failure mode, I don't think the background handoff can be safely added to `execute_approved_media_buy` in its current form.

**The core issue**: the function is 570 lines with 15 return paths, and the return type `(bool, str | None)` can't express the states the system now has. When this returns `(True, None)`, it means three different things: "I approved synchronously", "I started background polling", or "the adapter doesn't support approval." The `background_owns_status` flag correctly controls behavior *inside* the function, but it's a local variable — callers never see it. This creates a cascade of problems.

---

**Confirmed bugs in the new code:**

1. **Silent data loss — progress updates never persisted.** `_update_approval_progress` (line 267) calls `approval_job.progress.update(progress_data)` which mutates the JSONB dict in-place. SQLAlchemy doesn't detect in-place mutations without `flag_modified` or `MutableDict` — we use `flag_modified` in 30+ other places in this codebase. The `db.commit()` is a no-op for that column. After the initial SyncJob creation, all progress updates (attempt count, phase, timestamps) are silently discarded. Compare with `background_sync_service.py:514` which correctly does full reassignment.

2. **Adapter name breaks audit queries.** `"GoogleAdManager"` (PascalCase) at lines 307 and 376 instead of the canonical `"google_ad_manager"` (snake_case, from `google_ad_manager.py:71`). Audit logs store `f"{adapter_name}.{operation}"` — queries filtering on `google_ad_manager.%` miss all entries from the background approval service.

3. **`_mark_approval_complete` swallows the MediaBuy status write failure** (lines 298-303), then proceeds to write audit log with `success=True` and send an "approved" webhook. The SyncJob is already committed as `"completed"` in a separate session (line 294). Result: GAM says approved, SyncJob says completed, audit says success, webhook says done — but the MediaBuy is permanently stuck at `pending_approval`. Same pattern in `_mark_approval_failed` (lines 367-372). The root cause is that SyncJob and MediaBuy are updated in separate DB sessions with no atomicity.

4. **`_mark_approval_complete` and `_mark_approval_failed` are the same function.** Both follow the identical 4-step pattern: SyncJob update → MediaBuy status → AuditLogger → webhook. The only differences are the status string, success boolean, and error presence. Each step wrapped in identical try/except. Fixing the atomicity issue (#3) means fixing it in two places.

---

**Race conditions in every admin caller:**

I traced all three callers of `execute_approved_media_buy`:

| Caller | On `(True, None)` | What it writes | Races with background? |
|--------|-------------------|----------------|----------------------|
| `creatives.py:636` | Opens new UoW, computes status from flight dates | `"scheduled"` or `"active"` | **YES** |
| `workflows.py:241` | Overwrites unconditionally | `"scheduled"` | **YES** |
| `operations.py:449` | Does not write status | Nothing | No |

When the background handoff triggers, `execute_approved_media_buy` intentionally leaves the status at `pending_approval` for the background thread to own. But `creatives.py` and `workflows.py` don't know that — they see `(True, None)` and immediately overwrite the status. If the background thread later writes `"failed"` (GAM rejected the order), the sequence is: `pending_approval` → `"scheduled"` (caller) → `"failed"` (background). The buyer was told "approved" during the window between those writes.

The `creatives.py` caller has a second issue: when `execute_approved_media_buy` returns `(False, error_msg)`, it logs the error but returns `jsonify({"success": True, "status": "approved"})` at line 645. The user sees success; the MediaBuy is stuck. This is pre-existing but the new background handoff makes it exercisable — if `start_order_approval_background` raises a non-`ValueError` exception (DB failure), it falls through to `return (False, error_msg)` and creatives.py silently swallows it.

---

**Failure modes that leave MediaBuy permanently stuck:**

| Scenario | MediaBuy state | What user/buyer sees | Recovery |
|----------|---------------|---------------------|----------|
| GAM approves, MediaBuy write fails | `pending_approval` (stuck) | Webhook: "approved" | Manual DB fix |
| Process dies during background polling | `pending_approval` (stuck) | Nothing | Manual SyncJob cleanup required |
| `start_order_approval_background` raises non-ValueError | `pending_approval` (stuck) | Error flash (or silent 200 OK via creatives) | Manual intervention |
| Sync approval OK, local DB write fails (line 1017) | `pending_approval` → caller writes `"failed"` | Error flash | **GAM approved, local says failed** |

The process-death scenario is especially bad: the SyncJob stays at `status="running"` forever. On restart, the duplicate check (line 69-78) finds it and raises `ValueError`, which the caller treats as "already running." But no thread exists. The MediaBuy is stuck permanently with no automated recovery. `background_sync_service.py` has a 1-hour staleness check for this exact scenario — `order_approval_service.py` has none.

---

**Thread safety:**

The duplicate order check (DB query at lines 68-78) runs outside `_approval_lock`. Between the query returning "no duplicate" and the thread being registered (line 122-123), another request can pass the same check. No unique constraint protects this. Two threads can poll GAM simultaneously for the same order.

---

**What I think should happen:**

These aren't things that can be patched in place. The root problem is that `execute_approved_media_buy` has no proper state machine and no clear contract for who owns the status transition. Adding a background dimension to a function that couldn't cleanly express its synchronous states makes every pre-existing issue worse.

For the background handoff to be safe, it needs:

1. **A proper result type** — `ApprovalOutcome.COMPLETED` / `.POLLING` / `.FAILED` — so callers know what happened. The current `(True, None)` is ambiguous by design.
2. **Exclusive status ownership** — when the function (or its background service) is responsible for a MediaBuy, no caller should touch the status. The three admin blueprints each independently deciding what status to write is the root cause of the race conditions.
3. **Atomic SyncJob + MediaBuy updates** — or at minimum, don't report success (audit, webhook) until the MediaBuy status is actually written.
4. **Staleness check** for orphaned SyncJobs from dead processes — same pattern as `background_sync_service.py`.

This doesn't mean rewriting the whole 570-line function. It means wrapping the approval section (lines 964-1030) in a proper abstraction that the rest of the function and its callers can interact with safely. The concrete bugs (progress data loss, adapter name, `flag_modified`) should be fixed regardless.

Happy to pair on scoping this.

KonstantinMirin:
Thanks for the quick turnaround. Walking through this one more time — partly to confirm what's fixed and what's still open from the last round, and partly to flag things that surfaced on a second pass.

## Status of the previous review

**Fixed correctly:**

- `flag_modified` on the JSONB progress mutation
- Canonical `adapter_name = "google_ad_manager"`
- DRY refactor — `_finalize_approval` is a clean extraction

**Half-fixed — bug #3 (swallowed MediaBuy write).** Moving the MediaBuy update out of the inner try/except inside `_finalize_approval` correctly prevents the false-positive audit + webhook on failure. But the outer `except Exception` in both `_mark_approval_complete` and `_mark_approval_failed` still swallows the exception with a bland log line, and the SyncJob is committed as `"completed"`/`"failed"` BEFORE `_finalize_approval` runs (`order_approval_service.py:356`). So the worst-case state is now SyncJob = completed, MediaBuy = pending_approval (stuck), no audit, no webhook. Better than before, but the atomicity gap is still there. Cheapest fix: invert the ordering so SyncJob writes last.

**Still open from the previous review.** None of these were touched in the follow-up — fine as scoped-out, but let's not lose track of them:

- A proper `ApprovalOutcome` result type so callers can distinguish synchronous approve / background-polling / unsupported.
- Exclusive status ownership: `creatives.py:636` and `workflows.py:241` still overwrite `MediaBuy.status` after seeing `(True, None)` even when the background thread owns the status. `operations.py:449` is fine.
- The four permanent-stuck failure modes I tabulated last time.
- The duplicate-order check at `order_approval_service.py:72-83` still runs without a lock or DB-level unique constraint. A partial unique index on `(progress->>'order_id') WHERE sync_type='order_approval' AND status='running'` plus an `IntegrityError` catch would close it across workers.
- No staleness reaper — daemon thread + module-level dict means a SIGTERM mid-poll leaks a SyncJob and a MediaBuy forever. `background_sync_service.py:86-115` has the pattern to copy.

## New from the second pass

### Behavioral

**`webhook_url=None` at the admin call site.** `media_buy_create.py:989` literally passes `webhook_url=None` to `start_order_approval_background`. The sibling call site in `google_ad_manager.py:725-749` resolves the URL from `push_notification_config`. So a buyer who supplied push notifications and gets approved via the admin UI — exactly the flow this PR exists for — won't receive the background-approval webhook. Same service, two production paths, two behaviors. Should look up the persisted `PushNotificationConfig`.

**`principal_name=principal_id` in the audit call.** `order_approval_service.py:314` passes `principal_id` for both arguments. `AuditLogger.log_operation` uses `principal_name` as the human-readable bit in `f\"... for principal '{principal_name}' ...\"`, and every other call site in the repo (`mock_ad_server`, `kevel`, `triton_digital`, `broadstreet`, `media_buy_create`) passes name and ID separately. Audit messages now read `\"... for principal 'principal_1' ...\"` instead of `\"... 'Acme Corp' ...\"`. Thread `principal_name` through `start_order_approval_background`.

**Missing `tenant_id` filter on the duplicate-order scan.** `order_approval_service.py:72-83` selects `SyncJob` rows by `sync_type` and `status` only, then walks them in Python. `background_sync_service.py:80-83` filters by tenant. This one crosses tenant boundaries — independent of the unique-index issue above.

**Outer `except Exception` in `media_buy_create.py:1003`** can hard-fail on errors the background poller is built to retry (e.g. OAuth token refresh failures from `client_manager.get_service(...)`). `approve_order` itself swallows most GAM errors and returns False, so they reach the handoff path — but anything raised outside its inner try escapes the handoff entirely. Worth narrowing.

### Service correctness

**`_finalize_approval` docstring is wrong.** The doc says \"audit log and webhook are best-effort (wrapped in try/except)\" but the body only wraps the audit call. The webhook is best-effort via `_send_approval_webhook`'s own internal swallow. Either wrap it to match the doc, or rewrite the doc.

**Broad-except log lines in `_mark_approval_*` lack context.** `logger.error(f\"Failed to mark approval complete: {e}\")` has no `media_buy_id`, no `approval_id`, no `tenant_id`, no `exc_info=True`. In production, correlating a stranded MediaBuy back to one of these log lines is going to be painful.

**Two parallel approval-polling services.** `src/services/background_approval_service.py` is dead code — no production caller imports `start_order_approval_polling`; only a string label reference in `gam/managers/workflow.py:400`. Both files define `_active_approval_*` registries, `_reap_dead_*`, `_mark_approval_complete`/`_mark_approval_failed`, and `_send_approval_webhook`. They differ only in persistence model (WorkflowStep vs SyncJob). This PR piles new behavior onto one half. Worth deleting or merging the older one in a follow-up.

### Tests

**No integration coverage of the thread.** Every test mocks `start_order_approval_background` away — the thread that the entire fix is built to spawn is never run by any test. The `flag_modified` fix is Postgres-only-detectable and has zero regression coverage. A 5-line `@pytest.mark.requires_db` test calling `_update_approval_progress` twice and asserting the merge persists would lock it in.

**`_finalize_approval` error paths aren't tested.** The \"MediaBuyUoW raises → no audit, no webhook\" invariant in the docstring has no test, and neither does \"AuditLogger raises → webhook still fires\". Both are explicit contracts.

**`approve_order` retry-default changes (40→12, 15s→10s, 620s→130s timeout) have no direct test.** Could regress to `max_retries=12` by accident and nothing would catch it.

**Tests bypass `tests.harness` + `tests.factories`.** `tests/CLAUDE.md` is explicit that new tests should use the harness regardless of surrounding patterns. The new file extends the legacy `MagicMock` + `patch()` stack, and `tests/helpers/execute_approved_mocks.py` codifies it rather than fixing it. Not blocking by itself, but the PR is the right moment to stop accreting on the legacy path.

**Borderline / loose assertions:**

- 20 `patch()` calls across 6 test methods — borderline on the 10-mocks-per-file advisory.
- `assert \"12345\" in str(kwargs.get(\"details\", {}))` at `test_b2_background_approval_polling.py:226` would pass with `details[\"attempts\"]=12345`. Tighten to `kwargs[\"details\"][\"order_id\"] == \"12345\"`.
- Mock-DB setup repeated 4× within the test file (pylint R0801 misses it because of small per-instance AST variations).

### Layering / structural

**`execute_approved_media_buy` is service-shaped but lives in `core/tools/`** — and because the structural guards (`test_transport_agnostic_impl`, `test_impl_resolved_identity`, repository guard) only scan functions named `*_impl`, this function escapes all of them. Pre-existing, but the PR adds another 25 lines of orchestration logic to it.

**Service-layer code string-matches GAM error text** at `order_approval_service.py:225` (`\"NO_FORECAST_YET\"` / `\"ForecastingError\"`) — adapter-boundary leak. Pre-existing. The real fix is a typed `AdapterTransientError` / `AdapterPermanentError` hierarchy that `approve_order` raises instead of returning bool / raising generic Exception.

**`_ADAPTER_NAME = \"google_ad_manager\"` doesn't generalize.** The moment a second adapter needs background polling (e.g. Kevel order activation), the module-level constant becomes a lie. A parameter or class-level attribute scales; the constant doesn't.

### Minor / drive-by

- Two schema-drift allowlist entries have a false \"library doesn't have it yet\" comment: `attribution_window` and `include_package_daily_breakdown` are both already in `adcp 4.3.0` (verified). The entries are dead code (allowlist permits, doesn't require) but the comment is misleading.
- Comment at `media_buy_create.py:981` says \"audit/Slack notification\" — there's no Slack call anywhere in `order_approval_service.py`. Should say \"audit log and webhook.\"
- `_finalize_approval` takes 12 keyword-only parameters. PLR0913 is project-wide ignored so it's not a lint hit, but if a third caller appears, a small dataclass is the right shape.
- The SyncJob terminal-update preamble (status / completed_at / summary or error_message / commit) is still duplicated between `_mark_approval_complete` and `_mark_approval_failed`. Pre-existing — out of scope for this PR, but worth noting.
- `order_approval_service.py` and `background_sync_service.py` are structural twins at the threading-registry + SyncJob terminal-update layer. A shared `SyncJobLifecycle` helper would consolidate both eventually.
- `AuditLogger(...)` instantiated directly at line 311 rather than via the `get_audit_logger(...)` factory used in `src/adapters/base.py`. Cosmetic.
- `dict[str, Any]` for `progress_data` / `summary` / `audit_details` — these cross thread boundaries and JSONB columns. `TypedDict` would document the contract.
- Deferred imports inside function bodies: `json` and `httpx` at `order_approval_service.py:348, 452` should be hoisted (the intra-project ones may still need to be deferred for circular-dep reasons).
- Cosmetic reflow of the `product_format_dimensions` annotation at `media_buy_create.py:2727-2729` is a drive-by, unrelated to the B2 fix — ideally lands in a separate commit.

## Suggested gate

Block on the cheap, buyer-visible items that are clearly in scope for this PR:

- The bug #3 ordering flip (SyncJob terminal write *after* `_finalize_approval`).
- `webhook_url=None` → resolve from the persisted `PushNotificationConfig`.
- `principal_name=principal_id` → thread the actual name through.

Everything else — the architectural items still open from last round, the structural / layering observations, the harness conversation, the dead `background_approval_service.py` — can ride as separate PRs / GitHub issues if you'd rather not expand scope here. Happy to pair on any of it.

ChrisHuie:
## Konstantine 2026-05-18 R2 — status against current head `2454688d4`

Audit against your four R2 items. Verified by reading the code at this head; observation only, not inference.

### Resolved
- **#4 — `tenant_id` filter on duplicate-order scan** — `src/services/order_approval_service.py:81` now starts the `select(SyncJob).where(...)` with `SyncJob.tenant_id == tenant_id` as the first predicate. Cross-tenant read is closed. Commit `2454688d4`.

### Still open
- **#1 — SyncJob terminal write ordering** — `_mark_approval_complete` (`order_approval_service.py:9-22`) still commits `approval_job.status = "completed"` + `db.commit()` BEFORE calling `_finalize_approval(...)`. Same shape in `_mark_approval_failed`. The fix is to invert: call `_finalize_approval` first, only commit the SyncJob terminal status if it returned cleanly. Bug surface unchanged from your review: SyncJob says "completed" while MediaBuy can be stuck at pending_approval.

- **#2 — `webhook_url=None`** — `src/core/tools/media_buy_create.py:1008` still passes `webhook_url=None` to `start_order_approval_background`. The buyer's actual `push_notification_config.url` is not threaded through. Buyer who supplied push notifications and got approved via the background path doesn't receive the approval webhook.

- **#3 — `principal_name=principal_id`** — `order_approval_service.py:333-335` still passes `principal_name=principal_id, principal_id=principal_id` into `AuditLogger.log_operation`. Audit log lines read `"... for principal 'principal_1' ..."` instead of the Principal.name. Should load the Principal row and pass `principal.name`.

### Known follow-up gap on #4
The existing `test_start_approval_rejects_duplicate` mocks `scalars(...).all()` to a static list — bypasses the WHERE entirely. A single-line revert of the `tenant_id` filter wouldn't break any current test. Pin-test requires either `@pytest.mark.requires_db` or a mock-spy on the SELECT statement to assert `tenant_id` is in the predicate. Flagging as a follow-up since the test setup change is larger than the production fix.

### CI state (head `2454688d4` pre-push, `d592457a8` on origin)
All required gates green at `d592457a8`: BDD, E2E, all 5 Integration suites, Lint+Type, Security Audit, Smoke, Unit, Quickstart. CI status remains CHANGES_REQUESTED only because the prior review hasn't been dismissed.

KonstantinMirin:
## On landing the remaining work correctly

First, credit where it's due: every item from the earlier rounds has been worked through. The four named bugs (flag_modified, the canonical adapter name, the DRY extraction, the swallowed MediaBuy write), the race condition on the duplicate-order check, the staleness reaper for orphaned SyncJobs, the finalize-before-SyncJob-terminal ordering with its regression tests, and the tenant-scoped scan — all addressed. That's a lot of responsiveness across several rounds, and the fixes read correctly.

I have one structural point about how the rest should land. The goal is to keep this PR's scope bounded but also not let it ship new debt, and right now it's heading toward shipping new debt.

**The point.** `_update_approval_progress` opens its own `get_db_session()` and calls `flag_modified` by hand on the JSONB column. The `flag_modified` call isn't really the fix — it's a workaround for not having a repository. Pattern #3 puts JSONB writes behind a repository/UoW so that detail lives in one place and callers don't each have to remember it. This isn't flagged by the structural guard only because the guard scans `_impl` files and this is a service file — so the guard never looks at it. It's undetected, not allowed. The lock, the reaper, and the terminal-status writes this PR added all follow the same raw-session pattern, so the count of direct `get_db_session()` calls in this one service is going up, not down.

**Why this matters for the tests.** An integration test for these fixes has to do awkward things — commit the row first, then re-query in a separate session to avoid reading the in-memory object back. That awkwardness exists because the production code has no clean entry point to test. A `SyncJobRepository` gives the test a normal interface to call (`repo.merge_progress(...)`, `repo.get(...)`), so the test doesn't have to work around the code.

**Scope — why this isn't asking for a big refactor.** The rule I'd use: fix the code this PR already changes; track the code it only sits near. This PR already rewrites `_update_approval_progress`, `start_order_approval_background`, and `_mark_approval_complete`/`_mark_approval_failed`. Those are the exact functions that would move behind the repository, so this isn't new scope — it's doing the thing the PR is already doing, correctly. `background_sync_service.py` has the same problem but this PR doesn't touch it, so it stays a separate follow-up.

**What I'd aim for:**

1. Add a `SyncJobRepository` (`merge_progress`, `mark_terminal`, `find_running_for_order`, `reap_stale`) and move `flag_modified`, the tenant-scoped query, the terminal writes, and the reaper into it. Migrate only this service's call sites. It will open a short session per operation rather than using the request-scoped UoW shape — that's correct for a polling thread that runs ~2 minutes; it shouldn't hold one session open the whole time.
2. Write the `flag_modified` integration test first, against the current code, and confirm it fails when `flag_modified` is removed. Then do the repository refactor. If the same test stays green without changes, that's the evidence the refactor didn't change behavior. Writing the test after the refactor wouldn't prove that.
3. Fold in the two small buyer-visible fixes — `webhook_url=None` and `principal_name=principal_id`. A few lines each, and a buyer sees both.
4. List the rest as explicit follow-ups, one line each on why they're separable: the `ApprovalOutcome` result type, caller-side status ownership in `creatives.py`/`workflows.py`, `background_sync_service` adopting the repository, the multi-process unique index, and full E2E.

**On E2E.** I don't think it's realistic for this PR and I'm not asking for it — the admin → background thread → GAM → MediaBuy → webhook path needs a mock ad server we don't have yet. Integration tests against real Postgres (the repository and the service) are realistic and enough here. I'd just state the E2E gap plainly so the integration tests aren't mistaken for full coverage.

**Where this ends up:** the code this PR changes is repository-backed and follows Pattern #3, the fixes have real-database tests that would fail if reverted, the tests call a clean interface instead of working around raw sessions, and the larger items are tracked rather than quietly inherited. That's the most this PR can do without turning into a different PR.

Happy to pair on the repository shape or the first integration test.

KonstantinMirin:
## Round 4 — the direction is right; it's not landable yet

Real credit first: the persistence path is genuinely fixed and the tests prove it — pull `flag_modified` out of `SyncJobRepository.merge_progress` and two integration tests go red. The terminal-ordering flip (MediaBuy write before SyncJob terminal) is real, the tenant-scoped duplicate scan is real, and the structural guards all pass with the allowlist *shrinking*. The `SyncJobRepository` extraction was the right call.

That said, I want to hold a hard line on three things, because each is the kind of drift that compounds quietly after merge.

**1. The session count has to be zero, not four.** `BaseUoW.__enter__` owns `get_db_session()` — the UoW exists precisely so a service never opens a session by hand. There are still four:
- `order_approval_service.py:649` (`get_approval_status`) opens a raw session for `SyncJobRepository.get()` — but `SyncJobUoW` already exists. This one's just `with SyncJobUoW(tenant_id) as uow: uow.sync_jobs.get(...)`.
- `:58` (`lookup_webhook_url`) and `:567` (`_send_approval_webhook`) fall back to raw sessions because there's no `PushNotificationConfigUoW` — the repository landed without its UoW.
- `:182` (polling thread) does the same for `AdapterConfigRepository`.

`lookup_webhook_url` is new in this PR, so this is the code the PR changes, not code it sits near. Routing 649 through the existing UoW and adding the two small UoWs is the finish line for the extraction.

**2. The buyer-visible fixes don't have tests that fail when they break.** I changed `media_buy_create.py:1072` from `webhook_url=resolved_webhook_url` to `webhook_url=None` and all 12 tests stayed green — the integration test reconstructs the chain by hand instead of driving the admin path (its own docstring says so). Same shape for `_finalize_approval`: it's only ever patched-to-raise, so its three branches (MediaBuy-fatal, audit best-effort, webhook-after-audit) never execute — and that's the atomicity contract behind the most dangerous bug we fixed. And the `approve_order` defaults (`max_retries=12`, `poll_interval=10`, `timeout=130`) have no pin test, so a revert to the old 40/15/620 passes CI. A test that doesn't go red when the line it covers changes isn't protecting the fix. These need to drive the real entrypoint (or assert on the spawned call's kwargs).

**3. `background_sync_service.py` needs a tracked issue, not a mental note.** It still hand-rolls the exact SyncJob lifecycle the new repository now owns — progress-merge, terminal-write, duplicate scan, reap — on raw `select()` + `db.add`. I agree it stays out of this PR's diff. But "out of scope" means a filed issue with `SyncJobRepository` as the single home both services adopt; otherwise the duplication just becomes permanent.

**One small thing:** the PR description is stale — it still cites `adapter_name="GoogleAdManager"` and lists `webhook_url` as out-of-scope, both of which the current code contradicts. Worth refreshing, and adding the explicit follow-up list (ApprovalOutcome result type, caller-side status ownership in `creatives.py`/`workflows.py`, the multi-process unique index, the `background_sync_service` migration) plus a one-line note that the E2E path isn't covered yet so the integration tests aren't mistaken for full coverage.

Gate for this round: session count to zero, real regression tests on the three items above, and the `background_sync_service` follow-up filed. Happy to pair on the UoW shapes or the first admin-path test. Thanks for the steady work across these rounds.

ChrisHuie:
Cross-reference for triage: this PR (B2 — `NO_FORECAST_YET` background-polling handoff) and #1378 are **independent root causes on different webhook paths**, not a fix-and-superseded pair. #1311 hardens the *order-approval-polling* webhook (`PushNotificationConfigRepository`, no `ObjectWorkflowMapping`); #1378 is the *workflow-step* webhook path gated on a missing `ObjectWorkflowMapping` row on the auto-approve path. Neither closes the other, and two silent-failure holes survive both until a create-time envelope (#1303) and/or the #1378 fix land. Full trace: https://github.com/prebid/salesagent/issues/1378#issuecomment-4702773384

PR #1307, opened 2026-05-15
refactor: error-drain Pattern A + boundary ValueErrors (PR 2 of 3)

ChrisHuie:
## Summary

  - Drains all 82 Pattern A sites in `src/core/tools/` and `src/adapters/` to zero — `_impl` and adapter operations now raise typed `AdCPError` subclasses instead of constructing `Error(code=...)`
  inline. The boundary translator builds the two-layer envelope once, per AdCP 3.0.7 normative spec.
  - Migrates 38 boundary-facing `ValueError` sites to typed `AdCPError` raises; keeps ~86 internal contracts as `ValueError` (programmer-error invariants that should crash with a stack trace).
  - Adopts `ContextManager.fail_step()` at every former return-shape failure site in `media_buy_update.py` (22 callsites); wire response and persisted `workflow_step.response_data` are now
  byte-identical by construction.
  - Drains all 4 PR 1 structural-guard cap dicts to their final shrunk values; the `# noqa: structural-guard` marker is reserved for 4 legitimate per-item advisory sites in success envelopes.
  - Builds on PR #1306's substrate (typed subclasses, envelope builder, `fail_step`, INTERNAL_CODES translation). No new architecture; this is the cleanup sweep that drains what PR 1 froze in place.

  This is **PR 2 of 3** in the error-emission architecture rollout. PR 3 (async/submitted lifecycle, Response3 envelope, `task_id` population) lands separately.

  ## Sub-batches

  Each sub-batch is its own commit so reviewers can ack one at a time.

  | # | Commit | Scope |
  |---|--------|-------|
  | 1 | `ba378d5f1` | 9 low-traffic tool files: 12 Pattern A migrated, 10 boundary ValueErrors migrated, 4 advisory sites noqa-marked |
  | 2 | `f52ec8fc4` | `media_buy_create.py`: 4 Pattern A + 18 boundary ValueErrors + `except (ValueError, PermissionError)` catchall narrowed to `except AdCPError` |
  | 3 | `faf6472be` | `media_buy_update.py`: 21 Pattern A + 1 boundary ValueError + 22 `fail_step` adoptions (replaces inline `update_workflow_step(status="failed", response_data=...)`) |
  | 4 | `24764ea9a` | 4 adapters: 45 Pattern A migrated (broadstreet 13, GAM 22, kevel 5, triton 5); internal ad-server codes preserved in `details["internal_code"]` |
  | 5 | `1b1210f9f` | 8 adapter boundary ValueErrors + final cap drain |

[...]

Issue #1075, opened 2026-03-01
feat: Add idempotency_key support on update_media_buy and sync_creatives


bokelley:
## Summary

Support the `idempotency_key` field on `UpdateMediaBuyRequest` and `SyncCreativesRequest` to prevent duplicate operations from concurrent or retried agent requests.

**AdCP version:** v3.0.0-rc1
**Blocked on:** adcp Python client shipping the `idempotency_key` field (noted as schema-library mismatch in PR #1071)

## How It Works

Buyer agents include an `idempotency_key` (UUID or similar) on mutating requests. The sales agent:

1. **First request with key X**: Process normally, store key X → response mapping
2. **Second request with key X**: Return stored response without re-processing
3. **Different key Y**: Process as new request

### Separate from buyer_ref
- `buyer_ref` = buyer's business identifier for the media buy (dedup on create)
- `idempotency_key` = request-level dedup for updates and syncs (handles retries)

### 1. Storage
```python
class IdempotencyRecord(Base):
    __tablename__ = "idempotency_records"
    
    tenant_id: Mapped[str]
    idempotency_key: Mapped[str]      # From request
    operation: Mapped[str]             # "update_media_buy" | "sync_creatives"
    response_json: Mapped[dict]        # Cached response
    created_at: Mapped[datetime]
    expires_at: Mapped[datetime]       # TTL (e.g., 24 hours)
```

### 2. Check in _impl functions
```python
def _update_media_buy_impl(request, identity, uow):
    if request.idempotency_key:
        cached = uow.idempotency.get(request.idempotency_key, "update_media_buy")
        if cached:
            return cached  # Return stored response
    
    # ... normal processing ...
    
    if request.idempotency_key:
        uow.idempotency.store(request.idempotency_key, "update_media_buy", response)
    
    return response
```

### 3. Conflict detection
If same key arrives with different request body → return `CONFLICT` error with `recovery="correctable"`.

## Files to Modify

| File | Change |
|------|--------|
| `src/core/database/models.py` | `IdempotencyRecord` model |
| `alembic/versions/` | Migration |
| `src/core/tools/media_buy_update.py` | Check/store idempotency key |
| `src/core/tools/creatives.py` | Check/store idempotency key |
| Cleanup job | Expire old records (cron or on-request TTL check) |
| `tests/unit/test_idempotency.py` | Unit tests |

## Testing

Test cases:
- First request with key → processes normally
- Retry with same key + same body → returns cached response, no side effects
[...]

PR #1312, opened 2026-05-17
feat(idempotency): success-cache replay (PR 1 after error-emission)

ChrisHuie:
## What

AdCP 3.0.1 idempotency for `create_media_buy`: a retry carrying the same
`idempotency_key` replays the ORIGINAL success verbatim; the same key with a
different payload conflicts; errors are never cached.

## Contract (as shipped)

- **`idempotency_key` is REQUIRED** (library field, 16–255 chars,
  `^[A-Za-z0-9_.:-]{16,255}$`). A missing key rejects at the schema boundary as
  `VALIDATION_ERROR` (the 3.0.1 conformance storyboard accepts it; spec prose
  prefers `INVALID_REQUEST`). **Breaking change** for callers that omitted it.
- **Verbatim success replay.** A fresh keyed success is cached
  (`idempotency_attempts` row: verbatim `{status, response}` envelope, RFC 8785
  payload hash, 24h TTL matching the advertised
  `adcp.idempotency.replay_ttl_seconds = 86400`). A retry within the TTL
  returns the stored envelope unchanged with top-level `replayed: true`; the
  adapter is not re-invoked and no second booking exists.
- **Conflict detection.** Payload equivalence is RFC 8785 JCS SHA-256 over the
  request AS SENT on the wire, with the spec's closed exclusion list
  (`idempotency_key`, `context`, `governance_context`,
  `push_notification_config.authentication.credentials`). The same key with a
  different canonical payload raises `IDEMPOTENCY_CONFLICT` (correctable — the
  buyer resends the original bytes or mints a fresh key, per the spec example)
  at the probe, after a race, and on the degraded path — never a replay of a
  request the buy does not describe.
- **Wire-level hash input, uniformly captured.** Each transport threads the
  payload AS SENT into the shared `_impl`: MCP stashes pre-normalization tool
  arguments via middleware state, A2A deep-copies the DataPart params before
  compat normalization and handler mutations, REST hashes the pre-rewrite body
  bytes stashed ahead of the compat middleware's replacement. Identical bytes
  replay across transports; seller-side compat-table changes never flip an
  honest retry into a conflict. The model-dump hash is a documented fallback
  for impl-direct callers only.
- **Errors are never cached.** Failed adapter results return uncached and typed
  errors raise past the cache write, so a retry after an error re-executes.
- **Scope is (agent, account, key) — exactly.** The cache unique index is
  `(tenant, principal, account, key)` NULLS NOT DISTINCT (`tool_name` is
  recorded for observability but is deliberately NOT a scope dimension: a key
  reused by a different tool hits the same row and conflicts on its differing
  payload hash). The `media_buys` dup-booking backstop index includes
  `account_id` the same way — the same key under two accounts books
  independently.
- **Degraded path fails closed** (cache row missing/expired/invalid but the
  backstop fired): account-scoped lookup; a buy that outlived the replay TTL
  rejects as `IDEMPOTENCY_EXPIRED` (terminal); a stored-vs-request hash
  mismatch raises `IDEMPOTENCY_CONFLICT` (`media_buys.payload_hash`, written
  at create time, carries the signal the evicted cache row no longer can);
  otherwise a transient `SERVICE_UNAVAILABLE` with `retry_after` — verbatim
  replay is byte-for-byte or nothing, never a reconstructed body.
- **Rate limit + storage bound.** The spec's MUST — an insert-RATE limit per
  (agent, account) — is a trailing 10s/300 window check; a fresh key in a full
  window rejects `RATE_LIMITED` (transient) with `retry_after` = when the
  oldest in-window insert leaves the window. The derived storage bound
  (`MAX_ACTIVE_ATTEMPTS_PER_SCOPE` = 1000 active rows) backs it, with
  `retry_after` = oldest-row expiry. All `retry_after` values clamp to the
  Error model's [1, 3600]. Replays and conflicts insert nothing and are never
  rate-limited. Expired rows are reclaimed opportunistically per keyed
  success.

## Storage & migrations

- `idempotency_attempts` table (new): verbatim response envelope (bare JSONB —
  typed coercion would break byte fidelity), payload hash, account scope,
  `tool_name` (observability only), `expires_at`; unique
  `(tenant_id, principal_id, account_id, idempotency_key)` NULLS NOT DISTINCT
  + expiry index.
- `media_buys`: nullable `payload_hash` (no backfill; legacy rows carry no
  conflict signal) and the backstop unique index widened to include
  `account_id` via a CONCURRENTLY build-swap-rename, so dup-booking coverage
  never lapses on the hot table. Downgrades are executable: the cache-table
  downgrade clears the (rebuildable) cache before narrowing its index; the
  media_buys downgrade documents the operator path if cross-account same-key
  feature data exists.

## Hashing implementation

`src/core/idempotency_canonical.py` delegates to the SDK canonicalizer
(`adcp.server.idempotency`) behind a single production seam: production code
never imports the SDK module directly, the RecursionError→typed-error boundary
stays ours, and the RFC 8785 §3.2.2 sample plus nine Appendix B vectors pin
the BEHAVIOR against RFC-published bytes independently of the engine — an SDK
bump that changes hashing fails our vectors. The spec's closed exclusion list
is pinned as literals. The direct `rfc8785` pin is dropped (the SDK requires
it).

## Coordination

- This PR merges first. #1389 (property_list targeting) follows and absorbs
  the reconciliation with this cache before its own merge: its new
  `submitted` response variant becomes cacheable/replayable (cache-guard
  widening, variant-aware replay).
- Follow-up under consideration: coupling the cache write into the booking
  transaction (the reference implementation's single-transaction shape) — a
  path restructure tracked separately; the fail-closed transient covers the
  window at this head.

## Tests

- Wire matrix (IMPL/A2A/MCP/REST): identical-retry replay (adapter not
  re-invoked, byte-level REST body check), same-key/different-payload conflict
  on real wire envelopes, fresh-key independence, missing-key rejection.
- Junction recipes through the production entrypoint (no concurrency needed):
  race-loser fail-closed after a lost cache row (transient, adapter re-invoked
  once, single booking survives); degraded-path conflict on a mutated payload;
  post-TTL `IDEMPOTENCY_EXPIRED`; cross-account same-key independence.
- Rate limit: insert-rate window semantics (burst rejection, trailing-window
  expiry, out-of-window rows not counted), storage-cap `retry_after` exact
  derivation + [1, 3600] clamp, `RATE_LIMITED` on the real REST wire,
  replay-never-rate-limited.
- Capture uniformity: cross-transport identical-retry replay (REST→MCP) and
  deprecated-field-spelling identical-retry replay on REST.
- Wire-hash input pin: two encodings of the same instant conflict instead of
  replaying (a wrapper that dropped the raw payload would replay here).
- DB-level race, eviction (physical-row proof), repository CRUD/TTL suites,
  typed-error unit pins, migration up/down verified on a migrated scratch DB.


KonstantinMirin (changes requested):
The repository layer here is solid — proper UoW integration, tenant isolation, TTL expiry, race condition handling. The architecture is clean. But there are a few gaps that need closing.

**Migration creates `json`, model expects `jsonb`.**

`097b909c7b5f` uses `sa.JSON()` for `response_envelope` (line 42). The ORM model uses `JSONType`, which resolves to `JSONB`. We already have a migration (`1aa2f5893a4d`) that fixed this exact mismatch for other tables. Use `postgresql.JSONB()` in the migration so the column is created correctly from the start.

**The feature contract has no test.**

The PR's stated purpose is "retry with the same idempotency_key returns the cached rejection." No test exercises this. The repository tests prove `record_rejection` and `find_by_key` work at the DB level. But `_build_idempotency_rejection_replay` and `_cache_rejection_envelope` have zero coverage, and no test calls `_create_media_buy_impl` with a key that has a cached rejection. If the replay lookup at lines 1477-1487 were deleted, every test still passes green. The existing `test_idempotency_new_key_proceeds` mocks `find_by_key` to return `None` but never asserts it was called.

**Reject-cache-return is the same 3 times.**

Lines 2007-2019, 2573-2583, and 2955-2967 all do: construct `CreateMediaBuyError`, call `_cache_rejection_envelope` with the same 4 args, return `CreateMediaBuyResult(failed)`. Extract a helper — if the caching protocol changes, three sites need updating in lockstep.

**Adapter rejections aren't cached — intentional?**

Line 3038 returns a failed result without caching. The three validation-error paths cache. If adapter rejections are deliberately excluded (transient errors, should be re-evaluated), add a comment saying so. Four of five review agents flagged this independently, which means the next person reading the code will have the same question.

**What needs to happen:**
- Fix the migration: `sa.JSON()` → `postgresql.JSONB()`
- Add tests: replay path through `_impl`, `_cache_rejection_envelope` called on failure, `_build_idempotency_rejection_replay` produces correct result
- Extract `_reject_and_cache()` helper for the 3 duplicate sites
- Document the adapter caching decision (either cache it or comment why not)

The architecture and repository work are good — this is about filling the gaps.

ChrisHuie:
— addressed all 4 items from your 2026-05-18 review, plus pre-applied 4 pattern-catalog nits from your recent reviews on adjacent PRs.

  **Your 4 items (commit `2951cd65d`):**
  - Migration uses `postgresql.JSONB()` — `alembic/versions/097b909c7b5f_add_idempotency_attempts_table.py:43`
  - `_cache_and_return_rejection` helper extracted from the 3 duplicated validation-reject blocks — defined at `src/core/tools/media_buy_create.py:1531`, called from lines 2261, 2824, 3203
  - Adapter rejection caching wired (excludes `recovery == "transient"`) with inline rationale — `src/core/tools/media_buy_create.py:3279-3285`
  - Replay-through-`_impl` test that breaks if the lookup is deleted — `tests/integration/test_idempotency_replay.py::TestImplReplaysCachedRejection::test_cached_rejection_returned_on_replay` seeds an
  `IdempotencyAttempt` row directly via UoW, calls `_create_media_buy_impl`, and asserts the cached envelope is returned

  **Wire-path completion (commit `1fb34dbc2`):** 
  The `idempotency_key` parameter wasn't declared on the MCP/A2A/REST wrappers, so the field was being stripped by TypeAdapter before reaching `_impl` — the replay feature was dead end-to-end via the
  SDK. Wired it through:
  - MCP `create_media_buy` — `media_buy_create.py:3998`
  - `create_media_buy_raw` (A2A entry) — `media_buy_create.py:4093`
  - A2A `_handle_create_media_buy_skill` — `adcp_a2a_server.py:1543`
  - REST `CreateMediaBuyBody` — `routes/api_v1.py:75,237`

  **Pre-emptive nits from your cross-PR pattern history (commit `1585933f5`):**
  - Hoisted redundant inline `IntegrityError` import (it's already at `media_buy_create.py:18`)
  - Enriched the race-info and exception logs in `_cache_rejection_envelope` to include `tenant_id` and `principal_id`, not just the key
  - Switched the two `_impl`-replay integration tests to `PrincipalFactory.make_identity()` instead of inline `ResolvedIdentity(...)` (per tests/CLAUDE.md single-source-of-truth)
  - Documented that `IdempotencyAttemptRepository.expire_old()` has no production caller yet — replay correctness is unaffected (`find_by_key` filters on `expires_at`), but a periodic cleanup task is a
   follow-up needed to bound storage growth

  **Substrate framing for #1075:** Once this lands, "add idempotency_key support on update_media_buy and sync_creatives" reduces to parameter-swap work — the `IdempotencyAttempt` repo, helpers, and
  `tool_name` column already support both. Added a note on #1075 with the substrate map.

bokelley:
gentle reminder that you get this for free in python sdk 4.4+

KonstantinMirin:
Went through this against the current branch — ran the idempotency suite, didn't just read it.

**Verified good:**
- All 4 items from the 2026-05-18 review are addressed in code (migration `postgresql.JSONB()`, the `_cache_and_return_rejection` helper replacing the 3 duplicate blocks, adapter rejections cached with the `transient` exclusion + rationale, and the replay-through-`_impl` test).
- The three integration files pass against real Postgres — 21/21 (`test_idempotency_replay.py`, `test_idempotency_attempt_repository.py`, `test_idempotency_race.py`). Good coverage on TTL/expiry, tenant/principal/tool isolation, the race paths, and key-scoped miss.
- The core replay test is load-bearing: I disabled the lookup at `media_buy_create.py:1731` and `test_cached_rejection_returned_on_replay` failed (the impl re-evaluates instead of replaying), so it would catch a regression. That's exactly what the earlier review asked for.

**One gap worth closing before merge — the wire path.** Every replay test calls `_create_media_buy_impl` directly. Nothing exercises idempotency through the transport wrappers (MCP `create_media_buy`, A2A `create_media_buy_raw` / `_handle_create_media_buy_skill`, REST `CreateMediaBuyBody`).

That's the one layer that actually broke: per the commit notes, the feature was dead end-to-end until `1fb34dbc` because `idempotency_key` wasn't declared on the wrappers and TypeAdapter stripped it before it reached `_impl`. The wrappers are fixed now, but no test guards that the key survives the boundary. If someone drops `idempotency_key` from the MCP wrapper signature again, all 21 tests stay green and the feature silently dies end-to-end — the same regression, undetected.

Concretely: one test that sends a repeat `idempotency_key` through the MCP wrapper (via the harness `call_mcp` / REST client) and asserts the cached rejection comes back. MCP alone covers the regression; A2A + REST would be better since all three wrappers were touched in the same commit.

**Two scope checks, not blockers:**
- `expire_old()` is tested but has no production caller yet — storage growth is unbounded until a periodic cleanup runs. Fine to defer, just flagging it's still open.
- This is rejection-replay only — no success caching, no `IDEMPOTENCY_CONFLICT` on same-key-different-payload. If that matches the intended scope for #1303 contract 7, all good; if conflict detection was expected here, it isn't built.

The repository and `_impl` work is solid and properly tested. The wire-path test is the thing I'd want in before this merges.

KonstantinMirin:
Chris — glad this closes the wire-path gap, and the three guards point at exactly the right lines. I'd like your read on the *form* before it merges, because I think we can do it with a lot less machinery and a stronger test.

**Team default: integration over mocks, every time.** These run against a live DB already, so we'd rather drive the real path than stand up `MagicMock`/`patch` scaffolding. Two specific places that applies here:

**1. The auth mocking can mostly go away.** `_run_mcp_client` has an integration mode (see its docstring): when the identity carries a real `auth_token` — and `PrincipalFactory` persists one — it patches only `get_http_headers` and runs the *real* chain: header → tenant detection → token→principal DB lookup → `ResolvedIdentity`. Since we have a live DB, seed a real principal and exercise that, instead of `patch(resolve_identity_from_context)` + `MagicMock(_resolve_a2a_identity)` + `MagicMock(_get_auth_token)`. The only seam we genuinely can't avoid in-process is the header getter (no socket), and the harness already isolates exactly that one.

**2. The dispatch + envelope handling is already in the harness — let's not re-implement it.** `BaseTestEnv.call_via(Transport.MCP | A2A | REST, **kwargs)` runs the full FastMCP / A2A-handler / FastAPI-TestClient pipeline and returns a normalized `TransportResult`. Rejection envelopes are normalized for us — `parse_rest_error` reconstructs the `AdCPError` the same way the MCP/A2A unwrappers do, and `assert_rejected(result, code=..., message_contains=...)` checks a rejection across all transports. So no hand-parsing `structured_content` / artifacts / 200-with-status-failed in each test.

Put together, each wire test becomes about this:

```python
def test_mcp_wire_replays_cached_rejection(self, integration_db):
    with MediaBuyCreateEnv(use_real_db=True) as env:
        env.seed_rejection(env.identity.principal_id, idem_key, CACHED_MSG)
        result = env.call_via(Transport.MCP, idempotency_key=idem_key, **valid_create_kwargs)
    assert_rejected(result, code="VALIDATION_ERROR", message_contains=CACHED_MSG)
```

Same coverage and the same "drop `idempotency_key` from a wrapper and the test breaks" guarantee — but the mock seam lives in the harness, the auth chain is real, and it's ~3 lines per transport instead of ~60. The only new thing needed is a `MediaBuyCreateEnv` (we have `media_buy_update.py` and `media_buy_account.py`, just not create yet). Rough shape:

```python
# tests/harness/media_buy_create.py
class MediaBuyCreateEnv(BaseTestEnv):
    MODULE = "src.core.tools.media_buy_create"
    REST_ENDPOINT = "/api/v1/media-buys"
    EXTERNAL_PATCHES = {}   # integration mode: real UoW + real idempotency repo run

    def call_impl(self, **kw):
        req = CreateMediaBuyRequest(**self._defaults(kw))
        return _create_media_buy_impl(req=req, identity=self.identity)

    def call_a2a(self, **kw):
        return self._run_a2a_handler("create_media_buy", CreateMediaBuyResponse, **self._defaults(kw))

    def call_mcp(self, **kw):
        return self._run_mcp_client("create_media_buy", CreateMediaBuyResponse, **self._defaults(kw))

    def parse_rest_response(self, body):  # success path
        return CreateMediaBuyResponse(**body)
    # parse_rest_error is inherited — already maps the rejection envelope

    def seed_rejection(self, principal_id, idempotency_key, message):
        rejection = CreateMediaBuyError(errors=[Error(code="VALIDATION_ERROR", message=message)])
        with MediaBuyUoW(self.identity.tenant_id) as uow:
            uow.idempotency_attempts.record_rejection(
                principal_id=principal_id, tool_name="create_media_buy",
                idempotency_key=idempotency_key,
                response_envelope=rejection.model_dump(mode="json"),
            )

    def _defaults(self, kw):
        return {"brand": {"domain": "harness.example.com"}, "packages": [],
                "start_time": "...", "end_time": "...", "po_number": "HARNESS-1", **kw}
```

I might be missing a reason `call_via` isn't already wired for `create_media_buy`, so tell me if there's a snag — but I think this is the form we want, and it's reusable for every future create_media_buy transport test, not just idempotency. Happy to pair on the env if it helps.

KonstantinMirin:
# Review (round 3)

## Reconciliation with prior rounds + in-flight PR context

The 05-18 blockers (migration `JSONB`, dedup helper, adapter caching decision, replay-through-`_impl` test) are all resolved. The 05-24 wire-path gap is resolved (`TestWirePathReplay` MCP/A2A/REST). The 05-25 harness rework landed (`MediaBuyCreateEnv` + `call_via` — wire tests now use it). New surface since: `payload_hash` + the `IDEMPOTENCY_CONFLICT` raise (`28aab1bb7`), the local canonical hasher, the typed `AdCPIdempotency*Error` classes, the `replayed` flag plumbing.

**In-flight PRs that overlap heavily with the findings below — review intentionally short because most items are eased or partly closed by these landings:**

| PR | What it does | Findings it touches |
|---|---|---|
| **#1306** (merged, this PR builds on it) | Error-emission architecture base, PR 1 of 3 | ARCH-3 (success-replay `replayed` plumbing) and ARCH-4 (`_impl` ↔ envelope coupling) belong properly in PR 2/3 of that series. |
| **#1307** | Error-emission PR 2 | Same as above; likely the natural home for the `replayed`-on-success symmetry fix. |
| **#1374** / **#1384** | Test surface / harness | TEST-2 (wire-assertion strictness) and TEST-4 (`_RepoEnv` consolidation) probably move with these — flagging here so the duplicate `_RepoEnv` copies don't ratchet further while they're in flight. |
| **#1399** | Brings adcp SDK to v5 | Direct bearing on ARCH-1. The SDK landscape this PR was written against (4.3) is two majors out of date by merge time. |

Given that, the structural items below are listed for visibility; the only items I'd block merge on are the ones whose contracted behaviour is unverified by *any* test in *this PR's* own surface.

---

## Architecture problems

### ARCH-1 — Local reimplementation of an SDK primitive: revisit under the new rule
`src/core/idempotency_canonical.py` ships its own `EXCLUDED_FIELDS`, `_NESTED_EXCLUSIONS`, `strip_excluded_fields`, `canonical_payload_hash` — byte-equivalent to `adcp.server.idempotency.{EXCLUDED_FIELDS, strip_excluded_fields, canonical_json_sha256}` in the installed SDK (verified via `inspect.getsource`). The module docstring justifies the fork as "deliberately not adopting the SDK's return-based replay store" — that rationale was right when the SDK was 4.x and the framework was trying to mandate end-to-end flow we can't conform to.

That stance needs to be re-examined under the rule we should hold to from here:

> **The SDK is for primitives, not for orchestration.** Salesagent owns the data flow because it owns data the SDK doesn't see. If the SDK ships a **public pure function** we can use, we use it — the signature is the contract. If it ships a **pure function we'd want but it's private**, file a ticket against the SDK to make it public, then use it. We don't reimplement primitives we already pin; we don't take frameworks that try to own our flow.

Applied here: `strip_excluded_fields` and `canonical_json_sha256` are public, pure, and live in a submodule we can import without touching the store backends we deliberately reject. **#1399 brings the SDK to v5**, and v5.7 likely ships more such primitives — so the right move is to delete the local fork in favour of the SDK helpers, then track v5.7's additional primitives as adoption opportunities rather than reasons to reimplement.

`pyproject.toml`'s explicit `rfc8785>=0.1.4` pin only exists because the local copy bypasses the SDK; drop it when the local helpers go away.

The closed exclusion list is normative spec data. Two copies will drift the moment the spec gains a fourth excluded field — that's the concrete risk that justifies acting now rather than waiting another round.

### ARCH-2 — Symbols ship before their seams are wired (same pattern, third round)
A recurring shape across reviews of this PR: an artifact lands without the path that activates it.

- `AdCPIdempotencyExpiredError` (`exceptions.py:575-584`) — class defined, class-shape test added, no raise site in `src/`. `find_by_key` filters expired rows silently. Dead code.
- `payload_hash` + the `IDEMPOTENCY_CONFLICT` raise at `media_buy_create.py:1772` — built, no behavioural test (see TEST-1).
- `expire_old()` — still has no production caller (open since 05-24).

The 05-18 round was "repo works, contract has no test." The 05-24 round was "wrapper doesn't declare the param, feature dead end-to-end." This round is the third instance on different surfaces. Worth a personal note rather than another finding: **the PR's response pattern to "ship + don't wire" findings is to add the wiring without adding the test that would have caught the original gap**. Each round closes its specific instance and the pattern repeats on the next surface.

### ARCH-3 — `replayed=true` asymmetry between success and rejection envelopes
Rejection-replay sets `exc.replayed = True` (`media_buy_create.py:1511`) → `build_two_layer_error_envelope` emits `replayed: true`. Success-replay (`_build_idempotency_hit_result`, `media_buy_create.py:1594-1644`) does not — `CreateMediaBuySuccess` (`schemas/_base.py:196`) has no `replayed` field. Spec `ProtocolEnvelope.replayed` applies to both.

This properly belongs in **#1307** (error-emission PR 2) or a follow-up in the same series. Acceptable to defer if a docstring on `_build_idempotency_hit_result` acknowledges the gap and points at the tracking bead. Asymmetry without acknowledgement is the worst option.

### ARCH-4 — `_impl` calls `build_two_layer_error_envelope` to derive the cache projection
`media_buy_create.py:1539` — only `_impl` call site of a function whose other five callers are transport-boundary translators. Guard's letter satisfied (cache stores a projection), spirit bent. Cheap fix exists (cache typed-exception fields directly, drop the envelope intermediate). Likely lands cleanly under the **#1306 / #1307** error-emission series — defer there if the series has bandwidth.

### Pattern
ARCH-1 and ARCH-2 are the same shape on different surfaces — *we have the upstream that solves this, we built our own anyway, and the test that would have caught the divergence is also missing*. ARCH-3 and ARCH-4 are Pattern A (error emission) seam issues that the in-flight series exists to clean up.

---

## Testing problems

### TEST-1 — Contracted behaviours added without behavioural tests
- **`IDEMPOTENCY_CONFLICT` branch** (`media_buy_create.py:1770-1775`): reachable only when (a) cached row has non-`NULL` `payload_hash` and (b) `_impl` is called with a request whose `canonical_request_hash` differs. No test satisfies both. Storage-level tests set non-`NULL` hashes for read-back assertions; `_impl`-level tests seed with `payload_hash=None`. The PR's docstring claim at line 1526 is unverified end-to-end.
- **RFC 8785 compliance**: tests cover salesagent contract (exclusion list, key order, non-mutation, hex output) — none cover spec vectors. Combined with ARCH-1, the missing test is the exact safety net that would catch local-vs-SDK drift. Adopting the SDK helpers (per ARCH-1) makes this less acute but `canonical_request_hash` itself — the Pydantic wrapper called from `_impl:1734` — has no dedicated test either way.

### TEST-2 — Wire-assertion the 05-25 round contracted for is silently weakened
`tests/harness/assertions.py:128` — `envelope = result.wire_error_envelope or result.synthesized_error_envelope`. For the three `TestWirePathReplay` tests, the `or` fallback re-introduces the "feature dead end-to-end" failure mode the wire tests exist to catch.

This likely moves under **#1374 / #1384** (harness work). If those PRs land first, the fix lives there; if this PR lands first, the two-line strictness fix belongs here so it doesn't drift further.

### TEST-3 — Symbols defined but no contract test pins them
- `AdCPIdempotencyExpiredError`: class shape covered; the wire/silent-reeval contract is not.
- `test_no_key_is_noop` (`test_idempotency_replay.py:119-130`): asserts nothing observable. Two-line repository-read fix.
- `test_unrelated_key_does_not_replay` (`:284-287`): `try/except AdCPError` masks the success-branch bug case. Convert to `pytest.raises`.

### TEST-4 — DRY: this PR grows the `_RepoEnv` duplicate count
Six copies of `_RepoEnv(IntegrationEnv)` across `tests/integration/`; **two added by this PR**. CLAUDE.md treats DRY as correctness — "do not grow" is the rule. Lift to one `BareIntegrationEnv` in `tests/harness/_base.py`. Pre-existing four copies migrate as follow-up; the two new ones should not ratchet further.

Likely overlaps **#1374 / #1384** — coordinate so the shared class lands once.

### Pattern
TEST-1 / TEST-3 are the testing half of ARCH-2 — each contracted behaviour has a shape test or a storage test, not a behaviour test. The 05-18 review summarised this as "repo works, contract has no test"; the same form is now operating on the new conflict + expiry surface.

---

## Targeted file comments

### Block merge on
1. **`src/core/idempotency_canonical.py` + `pyproject.toml`** — ARCH-1. Replace local primitives with `adcp.server.idempotency` re-exports; keep `canonical_request_hash` as the 2-line Pydantic adapter; drop the explicit `rfc8785` pin. Confirm v4.3 currently in use exposes the public symbols (it does — verified). If the team prefers to wait for #1399's v5 bump before touching this, note the deferral here and file a tracking bead pinned to #1399.
2. **`tests/integration/test_idempotency_replay.py`** — TEST-1. Add `TestImplRaisesConflictOnPayloadHashMismatch`. Seed `record_rejection(..., payload_hash="known-hash")`, call `_impl` with a request whose `canonical_request_hash` differs, assert `AdCPIdempotencyConflictError`. Mirror at one wire transport.
3. **`tests/unit/test_idempotency_canonical.py`** — TEST-1. Parametrized `test_rfc8785_appendix_vectors` (3-5 vectors); `test_canonical_request_hash_pydantic_field_order_stable`.

### Fix in PR (small) OR explicitly defer to a named in-flight PR
4. **`src/core/exceptions.py:575-584`** — ARCH-2 / TEST-3. Pin the expiry contract: either wire `AdCPIdempotencyExpiredError` as a distinct raise from `_impl`, or delete it and document silent re-eval on `find_by_key`. Either is fine; current dead-class state isn't.
5. **`src/core/tools/media_buy_create.py:1594-1644`** — ARCH-3. Defer to #1306/#1307 series; add docstring acknowledgement + bead reference here so the asymmetry isn't silent.
6. **`src/core/tools/media_buy_create.py:1539`** — ARCH-4. Defer to #1306/#1307 series, or cache typed-exception fields directly to drop the `build_two_layer_error_envelope` import from `_impl`.
7. **`tests/harness/assertions.py:128`** — TEST-2. Defer to #1374/#1384, or apply the 2-line strict-wire-envelope fix here.
8. **`tests/integration/test_idempotency_replay.py:30`, `tests/integration/test_idempotency_attempt_repository.py:21`** — TEST-4. Lift `_RepoEnv` to `tests/harness/_base.py`. Coordinate with #1374/#1384 so the shared class lands once.
9. **`src/core/idempotency_canonical.py`** — defence in depth. `try/except RecursionError: raise AdCPValidationError(...)`. Two lines.

### Track / batch (no individual file comments)
- `test_no_key_is_noop` trivial-assertion, `test_unrelated_key_does_not_replay` `try/except` → `pytest.raises` (TEST-3).
- New mock-only `test_idempotency_cached_rejection_replayed` (`tests/unit/test_media_buy.py:1301-1361`) duplicates real-DB coverage.
- `tests/unit/test_create_media_buy_behavioral.py` pre-existing 131-mock budget.
- `_default_recovery="terminal"` parent-override comment, `AdCPGoneError` parent, `Mapped[dict[str, Any]]`, redundant tenant FK, `DEFAULT_REPLAY_TTL` magic constant, `LibraryAccountReference` import-in-function, `Union[...]` migration syntax, three-copy seed block.
- `expire_old()` still has no production caller (open since 05-24).
- Non-idempotency `IntegrityError` rejection-cache hole — worth a test or a comment.

---

## Disposition

| Bucket | Items | Notes |
|---|---|---|
| **Blocking** | ARCH-1, TEST-1 (both halves) | Same-PR fixes. ARCH-1 may legitimately wait for #1399 if the v5 bump is days away — call that out, don't leave it silent. |
| **Fix here OR defer to named PR** | ARCH-2 (expiry contract), ARCH-3 (success `replayed`), ARCH-4 (`_impl` envelope coupling), TEST-2, TEST-4, `RecursionError` defence | Each item names the in-flight PR it would naturally land under. Defer is fine; silent defer is not. |
| **Track / batch** | Everything in the "track / batch" section above | One bead, not nine. |

ChrisHuie:
All review rounds are addressed at head `0e10c4431` — CI 30/30 green on this head (unit, all integration shards including the pinned creative-agent stack, E2E, admin, BDD, migration roundtrip).

## Round-3 items

- **ARCH-1 (hasher fork / `rfc8785` pin):** done. The local RFC 8785 implementation is deleted; `idempotency_canonical` delegates to the SDK canonicalizer (`adcp.server.idempotency`) behind a single production seam — production code never imports the SDK module directly, and the `RecursionError` → typed-error boundary stays local. The direct `rfc8785` pin is dropped (the SDK requires it transitively). Behavior is pinned engine-independently by RFC-published vectors (§3.2.2 sample with its expected canonical output, plus nine Appendix B number-serialization vectors hashed through the production entrypoint); the SDK-parity corpus was removed as tautological once the wrapper delegates to the function it compared against, replaced by a literal pin of the spec's closed exclusion list.
- **TEST-1 (conflict branch):** pinned at the probe, the post-race path, the degraded path, and on all four transports against real wire envelopes.
- **TEST-1 (RFC 8785 appendix vectors + `canonical_request_hash` unit test):** added, as above, plus field-order stability, exclusion invariance, and real-difference tests on the model wrapper.
- **TEST-2 (wire-assert `or`-fallback):** the MCP dispatcher's `wire_error_envelope` is real-wire-or-stash only; synthesized envelopes ride the separate `synthesized_error_envelope` field; the tests/CLAUDE.md authenticity table documents the rule. A dead MCP wire path now fails `assert_envelope_shape`.
- **"Delete the lookup, everything stays green":** re-run empirically at this head — killing the cache lookup reddens 10 tests, killing the cache write reddens 11, killing the conflict check reddens 8 (replay/race/wire-matrix/rate-limit battery, tree restored after each run).

## Spec-grounding corrections beyond the review items

Each verified against the 3.0.1 prose and the installed SDK before changing:

- **Scope tuple is exactly `(agent, account, key)`.** `tool_name` left the unique index and all lookups (it remains as an observability column). A key reused by a different mutating tool now hits the same row and conflicts on its differing payload hash per rule 5, instead of getting an independent per-tool cache. The former isolation test is inverted to pin the spec behavior.
- **`IDEMPOTENCY_CONFLICT` recovery is `correctable`** (the prose example envelope and the storyboard expectation; the buyer resends the original bytes or mints a fresh key). The previous `terminal` rested on an SDK "validator" that does not exist — `STANDARD_ERROR_CODES` is a default applied only when no recovery is supplied, never enforced. `IDEMPOTENCY_EXPIRED` stays `terminal` (no resend of the same key can succeed); both docstrings now state the actual rationale.
- **Insert-rate limit (rule 8 MUST):** a trailing 10s/300 window check on cache inserts now precedes the active-row storage cap; `retry_after` points at when the oldest in-window insert leaves the window, and every `retry_after` clamps to the Error model's [1, 3600].
- **Hash input captured uniformly AS SENT on every transport:** A2A deep-copies the DataPart params before compat normalization and handler mutations; REST hashes the pre-rewrite body bytes stashed ahead of the compat middleware's body replacement; MCP already captured pre-normalization. Pinned by a cross-transport identical-retry replay (REST→MCP) and a deprecated-field-spelling identical-retry replay — seller-side compat-table changes can never flip an honest retry into a conflict.
- **The degraded path fails closed.** Verbatim replay is byte-for-byte or nothing: when the backstop fires and no usable cache row exists, the conflict and TTL-expiry checks run (`IDEMPOTENCY_CONFLICT` via the new `media_buys.payload_hash` column — `raw_request` is not canonicalizable due to injected package_ids and alias-dependent names; `IDEMPOTENCY_EXPIRED` per rule 6), then a transient `SERVICE_UNAVAILABLE` with `retry_after` — never a reconstructed body. The buyer's retry replays verbatim once the winner's cache write lands.
- **Migrations:** `media_buys.payload_hash` is nullable with no backfill; the dup-booking backstop index gains `account_id` via a CONCURRENTLY build-swap-rename (no coverage gap on the hot table); both idempotency-table downgrades are executable over feature data (verified on a migrated scratch DB).

## Known follow-up

Coupling the cache write into the booking transaction (the JS reference's single-transaction shape) is under consideration as a dedicated change: the pending-approval path constructs its response from artifacts created across four sequential transactions, so this is a path restructure rather than a call move. At this head the fail-closed transient covers the window.

KonstantinMirin:
Reviewed via multi-agent pass (DRY / layering / consistency / Python practices / testing). **This is a genuinely clean PR** — `_impl` stays transport-agnostic, all DB access goes through repositories, the tests call real production code with mutation-sensitive assertions (no skip/xfail, and the `schema_inheritance` allowlist even shrank). Three patterns worth addressing, none blocking correctness. Cited sites are illustrations, not an exhaustive checklist.

### 1. The two `create_media_buy` wrappers are edited in lockstep

`create_media_buy` (MCP) and `create_media_buy_raw` (A2A/REST) carry byte-identical request construction — same field list, same `ValidationError → AdCPValidationError` translation, same brand coercion, same `enrich_identity_with_account`. This PR added `account` + the `idempotency_key` omit-when-absent splat (including a verbatim 4-line comment) to **both** copies:
- `src/core/tools/media_buy_create.py:4131-4148` (MCP)
- `src/core/tools/media_buy_create.py:4222-4240` (A2A/REST) — identical except the `to_reporting_webhook()` / `to_context_object()` coercions

The next field added here will land in one and silently miss the other. **Close it at the pattern level:** extract `_build_create_media_buy_request()` plus a shared boundary tail (brand-coerce → validate → account-enrich), with two small adapters for the A2A coercions — then a future field is added once.

### 2. The new `retry_after` / rate-limit feature is where small divergences concentrate

Four of five review lenses flagged something in this region:
- **`retry_after` isn't serialized everywhere it should be.** `AdCPError.retry_after` is emitted only by `build_two_layer_error_envelope()` (`exceptions.py:919`); `to_dict()` and `to_adcp_error()` faithfully emit the sibling `field`/`suggestion` attributes but drop `retry_after` — contradicting the class docstring's sync claim (`exceptions.py:320-378`). Latent (active transports use the envelope builder), but a future flat-serializer caller loses it. Fix: emit it conditionally in all three, or drop the sync claim.
- **Column width disagrees with the canonical schema.** `idempotency_attempts.account_id` is `String(255)` (`models.py:1036` + migration `7a8c3e1170a5`), while every other `account_id` is `String(100)`. Same logical value, joined on the fallback path. A follow-up narrowing migration aligns it.
- **Rate-limit policy lives in the data layer.** The window/ceiling thresholds and `retry_after` clamp sit inside `IdempotencyAttemptRepository.enforce_insert_ceiling` (`idempotency_attempt.py:158-185`). The COUNT/MIN queries belong there; the policy math is borderline-misplaced.

### 3. Eviction runs on the create hot path

`expire_old()` — whose own docstring says *"Designed to be called by a periodic cleanup job"* — runs a tenant-wide `DELETE` in the **same transaction** as `record_success()` on every successful keyed create (`media_buy_create.py:1656-1669`). The best-effort `try/except` keeps the buy itself safe, but a DELETE deadlock under per-tenant concurrency would roll back the just-cached success. Since read-path TTL filtering already guarantees replay correctness, eviction is pure storage reclamation and doesn't need to be synchronous — move it off the hot path (probabilistic ~1%, a separate short transaction, or the scheduled job the docstring anticipates).

Nice touch: the "race" tests are deterministic single-thread TOCTOU recipes (no threads), which hit the real partial-unique-index `IntegrityError` + `_replay_after_race` without scheduler flakiness — disclosed in the docstrings and the right call.

ChrisHuie:
All three patterns addressed at the new head:

- **Lockstep wrappers:** extracted `_build_create_media_buy_request()` — one
  home for the field list, the brand string-shorthand coercion, the
  idempotency omit-when-absent splat, and the ValidationError translation.
  The MCP and A2A/REST wrappers now diverge only in the two A2A input
  coercions (`to_reporting_webhook` / `to_context_object`), applied at the
  call site; a future request field is added once.
- **`retry_after` serializer sync:** `to_dict()` and `to_adcp_error()` now
  emit `retry_after` (conditionally / via the SDK helper's kwarg), matching
  the envelope builder — pinned by a four-way serializer-parity test.
- **Column width:** `idempotency_attempts.account_id` narrowed to
  `String(100)` in the model and in migration `7a8c3e1170a5` itself (this
  branch's migrations are unmerged, so the edit lands in place rather than as
  a follow-up narrowing migration) — verified by a fresh-DB
  `alembic upgrade head` (information_schema reports 100).
- **Policy out of the data layer:** the thresholds, window math, and
  `retry_after` clamps moved to `src/services/idempotency_policy.py`;
  the repository now exposes two pure scope queries
  (`count_inserts_since`, `count_active` — COUNT + MIN each) and carries no
  rejection logic.
- **Eviction off the hot path:** `expire_old()` no longer shares the cache-
  write transaction. It runs in its own short transaction after the write
  commits, on ~1% of keyed successes (`_EVICTION_PROBABILITY`, patchable),
  so a DELETE deadlock can never roll back a just-cached success and the
  hot path almost never carries the DELETE. Pinned both ways: forced
  eviction deletes the expired row; suppressed eviction leaves it with the
  create untouched.

CI is 30/30 green on this head; the narrowed migration is additionally verified by a local fresh-DB `alembic upgrade head`.

KonstantinMirin:
Solid implementation — the architecture holds against the transport-boundary and repository invariants, the shared boundary builder and hash seam collapse the duplication well, and the tests assert on the wire envelope as they should. A set of issues to fix before merge:

**1. Same invariant materialized in two places — give each one home.**
- Replay-window expiry is answered two ways: the cache path reads the stored `expires_at`, the degraded path recomputes `now − MediaBuy.created_at > TTL` (`idempotency_attempt.py:81` vs `media_buy_create.py:1534`). Two rows, two `created_at`s, so the boundaries can diverge for the same key. Anchor both on the stored `expires_at`.
- Advertised `replay_ttl_seconds = 86400` and enforced `DEFAULT_REPLAY_TTL` are independent literals — they can drift. Make the capability read the TTL constant. Move the hardcoded limits to the `int(os.getenv(...))` convention used elsewhere.
- The `retry_after [1,3600]` clamp is written out twice in `idempotency_policy.py:74-94`. Extract `_clamp_retry_after()`.

**2. `find_by_idempotency_key` default re-scopes its other caller.** Adding `account_id=None` renders `IS NULL`, so the unchanged `get_by_id_or_idempotency_key` now matches NULL-account rows only (`media_buy.py:84,108`). Thread `account_id` through it.

**3. Boundary-completeness guard omits `idempotency_key`.** The field this PR turns on is absent from `CREATE_SPEC_FIELDS`, and it's forwarded via a `**`-splat the guard can't see — a wrapper that dropped it would pass. Add a forwarding test that drops the key and expects failure.

**4. `_seed_principal` is copy-pasted in three test files.** Move it to `tests/helpers/idempotency_seeds.py`, which already exists for this.

**5. Migration `1d9b1402` diverges from the other four.** It uses legacy `typing.Union` and omits the column `comment=`. Bring it in line with `collections.abc.Sequence` + PEP 604 + the descriptive comment.

**6. Three names for one concept.** The wire payload is stashed as `raw_arguments` / `raw_wire_parameters` / `raw_wire_body` across MCP/A2A/REST. Pick one.

Once these are in, this is in good shape.

ChrisHuie:
All six addressed at e1659d7f5.

1. Single source per invariant.
   - Replay-window expiry: the degraded path now reads the cache row's stored
     `expires_at` (new `find_including_expired`) instead of recomputing from
     `MediaBuy.created_at`. The MediaBuy.created_at proxy is the fallback only
     when no cache row survives (evicted-after-expiry / in-flight winner) —
     there's no stored boundary to read there. `find_by_key` and
     `find_including_expired` share one `_scope_filter`. New test pins that a
     present-but-expired row drives `IDEMPOTENCY_EXPIRED` even with a freshly
     created MediaBuy backstop.
   - TTL: both capability sites read `int(DEFAULT_REPLAY_TTL.total_seconds())`.
     Rate-limit thresholds moved to the `int(os.getenv(...))` convention;
     `_RETRY_AFTER_MAX` stays a constant (the spec Error-model bound, not an ops
     knob).
   - `retry_after` clamp: extracted `_clamp_retry_after()`; both branches call it.

2. `find_by_idempotency_key` re-scope: threaded `account_id` through
   `get_by_id_or_idempotency_key`. New repo test pins that an account-scoped buy
   resolves under its account and the default (IS NULL) does not match it.
   (That method has no `src/` caller yet — flagging in case it should be wired or
   belongs to a later PR.)

3. Boundary guard: the matcher now models the `**({...} if ... else {})` splat
   (it was structurally blind to it), and `CREATE_SPEC_FIELDS` is now derived
   from `CreateMediaBuyRequest.model_fields` minus a documented exclusion set
   rather than hand-listed — so a new field is required-until-justified. This
   surfaced that `account` was never checked (now is) plus 9 unwired optional
   fields (`proposal_id`, `io_acceptance`, …), now tracked as explicit
   exclusions instead of silent gaps. Negative self-tests included.

4. `_seed_principal` (and a new `seed_media_buy`) now live in
   `tests/helpers/idempotency_seeds.py`; the copies are removed.

5. Migration `1d9b1402` aligned to `collections.abc.Sequence` + PEP 604 + the
   column `comment=`, matching its siblings.

6. One name — `raw_wire_payload` — for the stashed wire payload across
   MCP/A2A/REST (and the convergence param).

Beyond the six, two regression guards: capability/constant parity
(`replay_ttl_seconds` must derive from `DEFAULT_REPLAY_TTL`, not a literal) and
the schema-derived boundary allowlist above.

Verified locally: `make quality`; the idempotency integration suite (repo/flow/
policy), the MCP/A2A/REST wire matrix, and a fresh-DB migration apply; BDD. e2e
and admin run in CI.

KonstantinMirin (commented):
Really solid work on this — the layering is clean, the degraded path genuinely fails closed, and pinning the hash with RFC-8785 vectors independent of the SDK engine is exactly the right instinct. A few shape-level things stood out that I think are worth talking through. Flagging them as patterns rather than line fixes, partly because the feature inverted to success-cache mid-PR and this is really the first full pass over the new shape.

**1. The race recovery is duplicated — and the duplication is hiding a fragile coupling.**
Both booking paths carry the same "attempt the write → on IntegrityError decide whether it's the idempotency backstop → log → replay the winner" envelope. The ordinary DRY cost aside, the part that would actually bite is *what's* duplicated: the "is this the idempotency race?" decision is a substring match against the driver's exception text, living in two places that have to move in lockstep. That ties correctness to the wording of a Postgres message in two spots. Worth knowing our DRY guard (pylint R0801) will never catch this — the two blocks aren't contiguous-identical, they're semantically identical with different `create()` args in the middle, which is precisely the case R0801 is blind to. So a green duplication guard isn't evidence of anything here. I'd want that race-or-not decision to have one home; the exact shape is yours to pick.

**2. "Errors are never cached" is real, but nothing fails if it breaks.**
The end-to-end test drives an adapter error and asserts no row is cached — good. But what's actually enforcing that is the early return on the error path; the `isinstance(result.response, CreateMediaBuySuccess)` clause in the cache-write helper is unreachable, because every caller already hands it a success. So that defensive clause has no path and no oracle — you could delete it and nothing goes red, and the test that "covers" the guarantee is green for an adjacent reason. Not a bug today, but the invariant is one refactor away from regressing silently (the day an error result gets routed through that helper). The principle I care about more than the fix: a claimed invariant wants a test that fails when the invariant breaks, not one that happens to pass. Either the clause earns that test or it goes and a comment names the early return as the real guard.

**3. The "single SDK seam" is true by convention, not enforced.**
Routing all SDK-canonicalizer access through `idempotency_canonical` is the right call, and the rationale (swap the engine in one file, keep the RecursionError→typed-error boundary ours) holds. But nothing stops the next change from importing `adcp.server.idempotency` straight into `_impl` — and the first time that happens, the seam's whole reason for existing is quietly gone. This feels like exactly the architectural intent that belongs in an AST guard next to the others, rather than in a docstring that the tooling can't enforce.

**4. Minor — EXPIRED is the only error code not checked on the wire.**
CONFLICT / RATE_LIMITED / VALIDATION_ERROR are all asserted on the real wire envelope; IDEMPOTENCY_EXPIRED is only verified through the reconstructed exception. Just an asymmetry in the verification surface — one wire assertion makes the set uniform.

None of these are "it's broken" blockers — the feature works and the design is sound. They're the shape-level things that, at the speed this codebase moves, tend to be what bites later. Happy to dig into any of them.

KonstantinMirin (approved):
## Approving — one DRY follow-up requested

Reviewed across five dimensions (layering, DRY, consistency, Python/Pydantic/SQLAlchemy, testing). This is a clean, well-architected change: layer separation is textbook (canonicalization in `core`, policy in `services`, persistence in the repository, orchestration in `_impl`, transport capture in wrappers), the SDK-idempotency seam guard genuinely enforces the boundary, the canonical hash is RFC-8785-deterministic and pinned to spec byte vectors, race recovery anchors on the stored `expires_at`, and the wire-matrix tests drive real REST/MCP/A2A pipelines (verified by mutating the replay/conflict logic — the tests fail on all four transports). No correctness issues found.

**Approving.** One DRY fix to land before merge, since this codebase treats DRY as a correctness invariant and the duplication carries a tenant-isolation hazard.

### Pattern: scope-windowed aggregate + tenant prefix hand-rolled per method

In `src/core/database/repositories/idempotency_attempt.py`, the two aggregate methods each rebuild the `(tenant_id, principal_id, account_id)` isolation prefix inline and re-implement an identical `(COUNT(*), MIN(column))` scope query — differing only in one windowing predicate and the MIN column. This is the same logical operation with parameter substitution, which `R0801` doesn't catch because the method bodies diverge syntactically.

The hazard is concrete: a tenant/principal/account scoping fix now requires three parallel edits, and missing one silently breaks isolation for that query path — the exact failure mode the DRY-as-correctness rule guards against. You already extracted `_scope_filter` for the sibling *keyed* lookups (with the explicit "one home so the two paths cannot scope the same key differently" comment), so the helper pattern is established — the aggregates just didn't get folded in.

Illustrations (not the full queue — the fix is one extraction that covers all of them):
- `idempotency_attempt.py:184-190` — `count_inserts_since` scope tuple (`… + created_at > since`)
- `idempotency_attempt.py:207-212` — `count_active` scope tuple (`… + expires_at > now`); the prefix is byte-identical
- `idempotency_attempt.py:191` vs `:213` — the `count = self._session.scalar(select(func.count())…) or 0` line is byte-identical; only the MIN column (`created_at` vs `expires_at`) differs at `:192`/`:214`

**Requested change:** extract a `_scope_prefix(principal_id, account_id)` returning the `(tenant, principal, account)` terms (shared with `_scope_filter`, which appends the key) and a `_count_and_oldest(predicate, min_column)` helper. Both aggregate methods then become a predicate + column pass-through, and the isolation prefix is single-homed across all five queries in the file.

### Non-blocking nits (optional)
- `canonical_payload_hash` (`idempotency_canonical.py:59-64`) catches only `RecursionError`, but `rfc8785.dumps` can also raise `FloatDomainError`/`CanonicalizationError`; the docstring claims it translates engine errors to typed buyer errors. Not reachable via the documented JSON-safe call paths — robustness gap only.
- 3 rate-limit allow-branch tests (`test_idempotency_rate_limit.py:69,88,140`) assert "no exception raised" rather than an explicit count; would strengthen if `enforce_insert_ceiling` returned its computed count.
- Minor naming/style: new error messages start lowercase vs the capitalized-sentence convention; `find_by_key` abbreviates where `MediaBuyRepository.find_by_idempotency_key` spells it out; `_raise_degraded_replay_outcome` passes `idempotency_key`/`principal_id` positionally while siblings force keyword-only.

Issue #1302, opened 2026-05-14
feat: complete property_list and collection_list end-to-end — adapter targeting, capability honesty, no-match handling per AdCP 3.0.6


ChrisHuie:
## Context

  Parent: #1247 gaps #6 + #7. Storyboard scenarios: `media_buy_seller/inventory_list_targeting` and `media_buy_seller/inventory_list_no_match` from
  `@adcp/sdk@6.11.0/compliance/cache/3.0.6/protocols/media-buy/scenarios/`.

  PR #1276 (issue #1275) lands the **schema, persistence, round-trip, and `property_targeting_allowed=false` enforcement** for both list types. This follow-up tracks every remaining piece required to
  make `property_list` and `collection_list` actually filter inventory end-to-end against AdCP spec 3.0.6.

  The scope split is deliberate: PR #1276 makes the buyer's targeting request *visible and persisted*; this issue makes it *actually deliver*.

  ## Out of scope (PR #1276 baseline)

  - `Targeting.collection_list` / `collection_list_exclude` / `property_list` typed against `adcp.types.generated_poc.core.TargetingOverlay`
  - `AdCPPackageUpdate.targeting_overlay` override for typed coercion
  - `GetMediaBuysPackage.targeting_overlay` round-trip (with legacy `targeting` key fallback)
  - `validate_property_targeting_allowed()` rejecting `property_list` against products with `property_targeting_allowed=false`
  - Obligations UC-002-MAIN-14a/b, UC-003-MAIN-13/14

  ## Gap

  Verified against installed `adcp==4.3.0` (wire spec 3.0.6) and codebase HEAD:

  1. **Adapter targeting passthrough is silently dropped across every adapter.** `grep -rn "property_list\|collection_list\|PropertyListReference\|CollectionListReference" src/adapters/` returns zero
  hits. `mock`, `gam` (and all GAM managers), `kevel`, `triton_digital`, `xandr`, `broadstreet` — none consume the references. The buyer sends the field, we accept it, we persist it, we return it via
  `get_media_buys` — and the ad server sees nothing. Even after PR #1276 merges, buyers using `property_list` / `collection_list` get **no inventory filtering**. The 3.0.6 storyboard narrative calls
  this out by name: *"silently drop the targeting and deliver against unintended inventory ... a real failure mode we've seen in adapter integrations."*

[...]

PR #1313, opened 2026-05-17
feat(adapters): raise UNSUPPORTED_FEATURE on property_list

ChrisHuie:
Important: set the PR base to feature/property-collection-list-targeting, NOT main. This is a stacked PR on #1276 — GitHub will auto-rebase to main when #1276 lands.

  Stacked on #1276 — base is `feature/property-collection-list-targeting`. When #1276 merges, GitHub auto-changes this PR's base to `main`.

  Closes part of #1302 (contract item 5 — each enabled adapter MUST translate OR raise UNSUPPORTED_FEATURE).
  B4 of the inventory-targeting plan.

  AdCP #1302 contract 5 requires that each enabled adapter either compiles `targeting_overlay.property_list` to native ad-server targeting OR returns a clean `UNSUPPORTED_FEATURE` envelope. Per
  RESEARCH.md Finding 5, 4 of the 6 adapters in this codebase have no native property_list path today and were silently dropping the field. This PR makes that explicit.

  ## Architecture

  **New `AdServerAdapter.supports_property_list_filtering: ClassVar[bool] = False`** — adapters set `True` only when they have a working compilation path (e.g. Kevel after B3's resolver lands).

  **New `AdServerAdapter._check_property_list_supported(packages)` helper** — returns `CreateMediaBuyError(code="UNSUPPORTED_FEATURE")` when any package has `property_list` AND the adapter doesn't
  advertise support. Returns `None` otherwise. Centralizes the check so adapters share one implementation.

  ## Adapter wiring

  | Adapter | Finding 5 classification | This PR |
  |---|---|---|
  | **Mock** | NONE (symbolic only) | Early-return UNSUPPORTED in `create_media_buy` |
  | **Broadstreet** | NONE (single-network) | Same pattern |
  | **Xandr** | NONE (adapter is stubbed) | Preventive raise — future implementers can't accidentally let property_list silently drop |
  | **Triton** | NATIVE for audio identifier types only (station_id, podcast_guid, etc.) | Reject all property_list for now; audio-type accept path requires B3's resolver. Documented in code comment. |

  ## Not in this PR (intentional)

  - **GAM** — Finding 5: PROXY (could potentially compile with a future mapping infrastructure). Left for the planned PROXY workstream; tracked separately so this PR stays focused.
  - **Kevel** — Finding 5: NATIVE. Covered by B3 (new `kevel_site_resolver.py` + `supports_property_list_filtering=True` flip for Kevel-using tenants).
  - **Adapter-side error code migration** — adjacent rejection paths in these 4 adapters still use the `return CreateMediaBuyError(...)` pattern (Pattern A). PR #1307 sub-batch 4 already plans to
  migrate all adapter Pattern A sites to typed `AdCPError` raises in one sweep; this PR adopts the same return-envelope shape so #1307's sweep absorbs it cleanly.

  ## Tests

  `tests/unit/test_adapter_property_list_unsupported.py` (10 cases):
  - Base helper: returns None on clean packages; returns UNSUPPORTED_FEATURE on violating packages; native-support adapters bypass the check; error message names the adapter; iteration finds a violator
   even when earlier packages are clean.
  - Per-adapter: class attribute defaults to False for Mock/Broadstreet/Xandr/Triton.
  - Mock end-to-end: `create_media_buy(packages=[with property_list])` returns `CreateMediaBuyError` with code `UNSUPPORTED_FEATURE`.

  ## Test plan

  - [ ] `make quality` clean locally (verified: 4483 unit pass; 1 skipped; 19 xfailed; no new failures)
  - [ ] Manual: send a `create_media_buy` request with `targeting_overlay.property_list` set against a tenant configured for each of Mock/Broadstreet/Xandr/Triton. Confirm the response is a clean
  `UNSUPPORTED_FEATURE` envelope naming the adapter (not a silent drop or generic 500).
  - [ ] Manual: same request against a tenant configured for an adapter that overrides `supports_property_list_filtering=True` (test fixture). Confirm the request is accepted (and downstream targeting
  compilation is exercised — relevant once B3 lands for Kevel).

  ## Behavior change

  Previously: a buyer sending `property_list` against these 4 adapters got a 200 success response with the field silently ignored. After this PR: a buyer gets a 422-equivalent error envelope with code
  `UNSUPPORTED_FEATURE` and a message naming the adapter. This aligns with the `property_list_filtering=False` capability declaration shipped in #1276 (this PR's base).


ChrisHuie:
**Known limitation:** `XandrAdapter` is currently abstract (4 unimplemented methods), so the helper call in Xandr's `create_media_buy` is dead code until a concrete subclass lands. The test uses
  `patch.multiple(..., __abstractmethods__=set())` to verify the wiring (matches the pattern in `tests/unit/test_adapter_packages_fix.py:288`). Documented inline at the test site.

KonstantinMirin (changes requested):
**The goal here is right and spec-mandated** — #1302 contract item 5 requires each enabled adapter to either compile `targeting_overlay.property_list` or reject it, and catching the 4 adapters that were silently dropping it is exactly the honest-declaration behavior AdCP wants. The per-adapter classification table in the description is genuinely useful. The notes below are about *converging the implementation onto one path*, not reversing the PR.

While reviewing, I found the codebase now has **three different answers to the same question** ("what happens when a buyer sends `property_list` to an adapter that can't compile it?"), and they disagree:

1. **A pre-existing soft-advisory path** (tool layer) lets the request *succeed* with `property_list` persisted-but-ignored plus a non-fatal `UNSUPPORTED_FEATURE` warning on the success envelope.
2. **This PR's adapter-layer reject** fires first and hard-rejects — which makes the advisory dead code, and means GAM/Kevel (excluded from the reject) still produce the *opposite* buyer outcome for the identical field.
3. **#1276's product-flag gate** (`property_targeting_allowed=false → VALIDATION_ERROR`), which is correct and separate.

**Three specific issues:**

**(a) Capability name is the wrong feature.** `supports_property_list_filtering` is the *get_products result-filter* capability per `media-buy-features.json:12`. The field this PR gates is `targeting_overlay.property_list`, whose capability is `supports_property_list_targeting`. These are different features that happen to share a word. Suggest renaming the ClassVar/reader.

**(b) The two error classes are both correct but distinct — keep them separate.** #1276's product gate is `VALIDATION_ERROR` (the buyer asked for something the *product* forbids). The adapter-incapacity case should be `UNSUPPORTED_FEATURE` with `recovery: correctable` (`error-code.json:189`, `error-handling.mdx:467`) — the buyer asked for something this *seller* can't compile, and the canonical recovery is "check `get_adcp_capabilities`, drop the field, retry, or use a capable seller." Emitting it with `field` + `suggestion` makes it machine-actionable rather than a dead end. `targeting.json:179` and `update-media-buy-request.json:64` are the "MUST reject rather than silently drop" basis.

**(c) The adapter-layer check never runs end-to-end, and isn't uniform.** Unit tests used `MagicMock`; integration ran `dry_run=True`, which short-circuits in `_create_media_buy_impl` *before* the adapter executes — so the reject path had illusory coverage. And because the check lives on 4 specific adapters, GAM/Kevel keep silently dropping. Moving the check into the `_impl` validation phase (right after adapter resolution, before any dry_run/approval branch) makes it fire on every response path, for every adapter, and the soft-advisory can then be deleted entirely.

I've got a refactor that collapses all three into this single path (correctable `UNSUPPORTED_FEATURE`, capability-gated, advisory deleted) with `_impl`-level integration tests that actually exercise the reject (not `dry_run`, not mocks) — happy to share it if useful. **Out of scope / follow-up:** the wire flag is still named `property_list_filtering` in `get_adcp_capabilities`; splitting the get_products capability from the targeting capability is worth tracking separately so this PR stays focused.

KonstantinMirin:
## Round 2 — collapse landed, verified; small honesty fixes before approval

This is exactly the convergence I was hoping for. I traced it against source rather than the commit message, and the three-paths-into-one collapse is real:

- Enforcement lives in one function — `raise_if_property_list_unsupported()` (`targeting_capabilities.py:237`) — called once each from `_create_media_buy_impl` (`media_buy_create.py:2217`) and `_update_media_buy_impl` (`media_buy_update.py:368`), both after adapter resolution and before the dry_run/approval branch. dry_run no longer short-circuits it.
- The soft-advisory is gone from production, not just its test — `build_property_list_unsupported_advisories` and the per-adapter `_reject_property_list_if_unsupported` helper are both removed; the adapters now carry comments only.
- It fires for every adapter, GAM/Kevel included (no `=True` overrides; base default `False`), so the silent-drop asymmetry is closed.
- `supports_property_list_targeting` rename done; `AdCPUnsupportedFeatureError` is `UNSUPPORTED_FEATURE` / `recovery=correctable` with `field` + `suggestion`, and #1276's `VALIDATION_ERROR` product gate stays distinct and runs first on update.
- Coverage is real: I neutered the production check and the reject tests went red against a real adapter on real Postgres, asserting on the wire envelope. That's the standard the previous round was missing.

Four things to close before this is landable — all small, all about the tests matching what they claim:

1. **The integration-test docstring claims `dry_run=False` coverage that doesn't exist.** `test_property_list_unsupported_capability.py:16,61,65` says it exercises the boundary with `dry_run=False` and names a `test_create_dry_run_false_path_also_rejects` test — but all four tests use the default and that test isn't there. The check is genuinely correct, but this is a claim about the exact gap I flagged last round. Add the ~6-line `dry_run=False` test so the file proves what it says.

2. **The per-adapter `ClassVar=False` pins were deleted with the helper and not replaced.** Nothing now fails if someone sets `supports_property_list_targeting = True` on an adapter that can't actually compile it. Add a parametrized test asserting each concrete adapter's default.

3. **`XandrAdapter` has no `adapter_name` ClassVar** (`xandr.py`), so its reject message reads "XandrAdapter does not support…" — the raw class name — where every other adapter reads "mock does not support…" etc. Add the ClassVar.

4. **Stale comments describing the deleted advisory** at `capabilities.py:168-171` and `media_buy_create.py:1621-1627`, a docstring at `targeting_capabilities.py:252` naming a method that doesn't exist, and hardcoded source line numbers in the test docstring at `test_property_list_unsupported_capability.py:56-60` (already drifting — `get_adapter` is at 2207, not 2210). Clean these up so the comments match the code.

The wire-flag split (`property_list_filtering` → a separate targeting capability) stays a tracked follow-up as we agreed — no objection to it riding separately. Nice work converging this.

KonstantinMirin (commented):
# Round 3 review — pattern-organized

Production layering is clean. The headline architectural claim verifies: single `_impl` boundary (`raise_if_property_list_unsupported` called once each from `media_buy_create.py:2240` and `media_buy_update.py:369`), no transport leaks under `src/`, zero new guard violations, allowlist count unchanged. The deletion of `test_property_list_unsupported_advisory.py` (347 LOC) is justified — the advisory-on-success pattern was retired in production and replacement coverage (capability + wire + adapter-defaults pin-tests) is broader, not narrower.

**The problem is concentrated in `tests/integration/test_property_list_unsupported_wire.py`**, where three disease patterns interact. Pattern 1 contains a test that is **currently red on this branch** (Comment 1 below).

## Pattern 1 — Tests trust intent over behavior

The wire test was written against the contract it *wanted* to verify, not against the merged code path. Docstrings describe behavior the merge has since changed; assertions are loose enough to pass without exercising what they claim to verify.

- **Comment 1** — A2A test will fail with `DID NOT RAISE` (merge deleted `_adcp_to_a2a_error`; `on_message_send` returns a failed Task now)
- **Comment 2** — local envelope helper sniffs both shapes, defeats the cross-transport pin
- **Comment 3** — MCP test asserts on `str(exc)` substrings that the request body itself contains
- **Comment 5** — `try / except / pytest.fail` with no positive assertion on the awaited result

## Pattern 2 — Tests bypass the canonical harness

`MediaBuyCreateEnv` + `assert_envelope_shape()` exist for exactly this case; this file ignores them and hand-rolls transport scaffolding, which is the *mechanism* that lets Pattern 1's loose assertions exist.

- **Comment 4** — rewrite the file on `MediaBuyCreateEnv.call_via(transport)`; rolls up Comments 1–3
- **Comment 6** — two parallel setup-checklist bypass mechanisms (monkeypatch + `x-test-session-id`); pick one

## Pattern 3 — Docstring drift from the merged code (defer)

- Wire test module docstring claims `UC-003` (update) coverage; only `create_media_buy` is tested
- Wire test line 154 references `_adcp_to_a2a_error` which the merge deleted — same drift that misled my local review
- `_build_idempotency_hit_result` docstring (`src/core/tools/media_buy_create.py:1493-1496`) claims the replay is "always against a capable adapter"; `raise_if_property_list_unsupported` short-circuits per-package when `overlay.property_list is None` (`targeting_capabilities.py:261`), so the claim is false when the request carries no `property_list`. Conclusion (`errors=None`) is correct, stated reasoning is not.

## Clean signals (positive)

- **Layering**: zero findings from the layering pass. Adapter declares `ClassVar[bool] = False`; service owns policy; `_impl` is the single call site; wrappers translate via existing plumbing. No `ToolError`/`fastmcp`/`a2a`/`starlette` imports added under `src/`.
- **Adapter-side DRY claim verifies**: 4 adapters carry only `ClassVar` + pointer comment, no per-adapter raise code that could drift.
- **`test_adapter_property_list_targeting_defaults.py`** is a clean parametrized pin-test (good positive example).
- **`test_property_targeting_allowed_enforcement.py`** fixed a real vacuous-compound-assertion bug.
- **True-path coverage added** (`test_adapter_with_property_list_support_advertises_capability`) — closes the previously missing flip on the capability declaration.

## Defer to follow-up (not blocking)

- Pattern 3 docstring drift (the three items above)
- Push `adapter_name: ClassVar[str]` onto the base class; drop the `getattr(adapter.__class__, "adapter_name", __name__)` fallback at `src/services/targeting_capabilities.py:263`
- Style cleanup: `errors=None` repeated 7× where it defaults, scalar-vs-list `pytestmark` inconsistency between the two new integration files, naming prefix (`test_property_list_unsupported_*` vs `test_adapter_property_list_targeting_defaults`), Xandr honest-declaration reminder in docstring vs 3-others' free-standing `#` comment

## Honest miss callout

My local 5-agent code review certified Comment 1 as a stylistic "tighten the `pytest.raises` catch" nit. It read this test's own docstring + the recent `51ada1363` commit history that mentions `_adcp_to_a2a_error` and assumed production still matched — it never `grep`'d `src/` to confirm the symbol survived the merge. Externally-run `ultrareview` traced through `on_message_send` and caught the live failure. Process gap; the testing-review agent should verify symbols its target docstrings name still exist in `src/` before triaging.

tests/integration/test_property_list_unsupported_wire.py
```diff
@@ -0,0 +1,252 @@
...
+        "x-adcp-tenant": TENANT_ID,
+        "x-test-session-id": "prop-list-wire-a2a-session",
+    }
+    server_context = ServerCallContext(
+        state={AUTH_CONTEXT_STATE_KEY: AuthContext(auth_token=ACCESS_TOKEN, headers=headers)}
+    )
+
+    handler = AdCPRequestHandler()
+    with pytest.raises((A2AError, InternalError)) as excinfo:
+        await handler.on_message_send(params, server_context)
```
KonstantinMirin, line 187: **Critical — this test is currently red on this branch (`DID NOT RAISE`).**

`_adcp_to_a2a_error` (referenced in this test's own docstring at line 154) was deleted in the merge. `on_message_send` at `src/a2a_server/adcp_a2a_server.py:693` now catches `AdCPError`, builds a failed skill result via `_build_failed_skill_result` (line 702), appends it to `results`, and returns a `TASK_STATE_FAILED` Task at line 769 — it does not raise. The inline comment on line 694 makes the contract explicit: *"AdCP-level errors are async-task failures, not JSON-RPC errors."*

The envelope lives in `task.artifacts[0].parts[0].data` as `error_envelope`. `tests/harness/dispatchers.py:A2ADispatcher` already reconstructs `AdCPError` from the artifact `DataPart` and exposes `wire_error_envelope` — that's what this test should consume (and Comment 4's rewrite picks up automatically).

tests/integration/test_property_list_unsupported_wire.py
```diff
@@ -0,0 +1,252 @@
...
+    assert err.get(code_key) == "UNSUPPORTED_FEATURE", f"expected UNSUPPORTED_FEATURE; got {err.get(code_key)!r}"
+    assert err.get("recovery") == "correctable", f"expected recovery=correctable; got {err.get('recovery')!r}"
+    assert err.get("field") == "packages[0].targeting_overlay.property_list", (
+        f"field must identify the offending package; got {err.get('field')!r}"
+    )
+    assert err.get("suggestion"), "suggestion must be present so the buyer agent can act"
+    assert "property_list_filtering" in err["suggestion"], (
+        "suggestion must reference the canonical capability flag so the buyer "
+        "agent can locate a capable seller via get_adcp_capabilities"
+    )
```
KonstantinMirin, line 105: This local helper sniffs `"errors" in envelope` and accepts both the flat (REST) shape and the nested (A2A/MCP) shape from any transport. That defeats the pin's purpose — a regression where REST starts emitting the nested two-layer shape, or A2A/MCP flattens, slips through silently.

`tests/CLAUDE.md` "Error Verification Policy" requires `assert_envelope_shape(result.wire_error_envelope, code=..., recovery=..., field=...)` from `tests/helpers/envelope_assertions.py:22`. Per-transport shape variance, if real, should be visible as an explicit failure — that's the whole point of pinning the wire envelope.

tests/integration/test_property_list_unsupported_wire.py
```diff
@@ -0,0 +1,252 @@
...
+    # parenthesized tuple: ``('CODE', 'message', 'recovery', '{json_details}')``.
+    # Parse the components out and verify each one matches the contract.
+    exc = excinfo.value
+    msg = str(exc)
+    assert "UNSUPPORTED_FEATURE" in msg, f"MCP ToolError must surface error_code=UNSUPPORTED_FEATURE; got: {msg!r}"
+    assert "correctable" in msg, f"MCP ToolError must surface recovery=correctable; got: {msg!r}"
+    assert "packages[0].targeting_overlay.property_list" in msg, (
+        f"MCP ToolError must surface the offending field; got: {msg!r}"
+    )
+    assert "property_list_filtering" in msg, f"MCP ToolError must surface the capability-flag suggestion; got: {msg!r}"
```
KonstantinMirin, line 252: Three of the four substring assertions are satisfied by the request body itself: `"packages[0].targeting_overlay.property_list"` is exactly the field path the buyer sent in `_build_packages()`, and `"UNSUPPORTED_FEATURE"` / `"correctable"` could appear in any error string mentioning the contract. A FastMCP regression that echoes the request body back in the error message would still green these checks.

The docstring above already describes the structured tuple FastMCP serializes (`('CODE', 'message', 'recovery', '{json_details}')`) — parse that into a dict and assert structurally via `assert_envelope_shape()`, same as REST and A2A. Only the `"property_list_filtering"` assertion is genuinely tight, because that string only appears in the suggestion.

tests/integration/test_property_list_unsupported_wire.py
```diff
@@ -0,0 +1,252 @@
...
+The ``_impl``-level tests in
+``tests/integration/test_property_list_unsupported_capability.py`` prove the
+exception is raised with the correct fields. These tests prove the exception
+actually traverses each transport boundary (REST, A2A, MCP) and reaches the
+buyer as a spec-compliant wire envelope with ``code=UNSUPPORTED_FEATURE``,
+``recovery=correctable``, ``field``, and ``suggestion``.
+
+Covers: UC-002 honest-declaration property_list reject — wire shape
+Covers: UC-003 honest-declaration property_list reject — wire shape
+"""
```
KonstantinMirin, line 12: **Structural ask: rewrite this file on top of `MediaBuyCreateEnv`.**

The harness for exactly this use case already exists. `tests/harness/media_buy_create.py:41` defines `MediaBuyCreateEnv(IntegrationEnv)` with `REST_ENDPOINT = "/api/v1/media-buys"` (line 56) and inherits `call_via(transport: Transport) -> TransportResult` from `tests/harness/_base.py:422`. `TransportResult.wire_error_envelope` is the canonical two-layer envelope per `tests/CLAUDE.md`. The A2A dispatcher (`tests/harness/dispatchers.py:A2ADispatcher`) already reconstructs `AdCPError` from the failed Task's artifact `DataPart` — which means the post-merge A2A contract Comment 1 calls out is supported out of the box.

A single parametrized test:

```python
@pytest.mark.parametrize("transport", [Transport.REST, Transport.A2A, Transport.MCP])
async def test_create_media_buy_property_list_unsupported_envelope(env, transport):
    result = await env.call_via(transport, ...)
    assert_envelope_shape(
        result.wire_error_envelope,
        code="UNSUPPORTED_FEATURE",
        recovery="correctable",
        field="packages[0].targeting_overlay.property_list",
    )
    assert "property_list_filtering" in result.wire_error_envelope["errors"][0]["suggestion"]
```

collapses ~250 LOC of hand-rolled 3-transport setup and fixes Comments 1, 2, 3 in one move. Happy to pair on the rewrite.

tests/integration/test_property_list_unsupported_capability.py
```diff
@@ -0,0 +1,306 @@
...
+    # real media buy.
+    try:
+        await _create_media_buy_impl(req=request, identity=_make_identity())
+    except AdCPCapabilityNotSupportedError as e:
+        pytest.fail(
+            "Adapter declared supports_property_list_targeting=True but the "
+            f"boundary check rejected anyway: field={e.field!r} suggestion={e.suggestion!r}. "
+            "This indicates raise_if_property_list_unsupported is not honoring the "
+            "ClassVar override."
+        )
```
KonstantinMirin, line 225: `try / except AdCPCapabilityNotSupportedError / pytest.fail` with no positive assertion on the awaited result. A regression where `_create_media_buy_impl` returns `None` silently, or where the coroutine never materializes (e.g. a missing `await` somewhere upstream), would green this test.

Bind the return and assert at minimum `result.media_buy_id` (or truthiness on the canonical response field). Same pattern repeats at lines 257-260 — fix both.

tests/integration/test_property_list_unsupported_wire.py
```diff
@@ -0,0 +1,252 @@
...
+    tests need to reach the property_list check on a minimal test tenant, so
+    we patch the setup-checklist validator to a no-op for the duration of
+    the fixture. This is the same shape ``test_session_id`` uses internally
+    in ``_create_media_buy_impl``, but the patch keeps the test transport-
+    agnostic (every transport surface honors it identically without needing
+    each route to wire ``x-test-session-id`` plumbing).
+    """
+    monkeypatch.setattr(
+        "src.core.tools.media_buy_create.validate_setup_complete",
+        lambda tenant_id: None,
```
KonstantinMirin, line 46: Two parallel mechanisms to bypass the setup-checklist gate: this fixture monkeypatches `validate_setup_complete`, and each per-transport test (lines 129, 174, 222) also sends `x-test-session-id: prop-list-wire-*-session`. Either is sufficient on its own; running both confuses the next reader about which path actually disables the gate, and one of the two becomes dead code without anyone noticing if `validate_setup_complete` moves or `x-test-session-id` plumbing changes.

Pick one — `x-test-session-id` is the production-shaped path the harness already speaks, so drop the monkeypatch.

PR #1314, opened 2026-05-17
feat(kevel): compile property_list to native siteIds

ChrisHuie:
Important: set the PR base to feature/b4-adapter-unsupported-property-list, NOT main. This is a stacked PR on B4 (which stacks on #1276). When B4 + #1276 land, GitHub auto-rebases this PR's base 
  through the chain.

  Stacked on B4 (#TBD-B4) — base is `feature/b4-adapter-unsupported-property-list`. B4 stacks on #1276. When this chain lands, GitHub auto-rebases through to `main`.

  Closes part of #1302 (contract item 5 — Kevel becomes the first salesagent adapter to satisfy the "translate" side of the contract).
  B3 of the inventory-targeting plan.

  Buyers can specify `targeting_overlay.property_list` to constrain a media buy to a curated set of publisher properties. Until now no salesagent adapter compiled that field — every adapter silently
  dropped it (B4 makes that explicit by raising `UNSUPPORTED_FEATURE`). This PR adds the first compilation path: Kevel's `Site.Id` primitive maps cleanly to AdCP's `domain`/`subdomain` identifiers, so
  a Kevel-backed campaign with a `property_list` now flows the resolved set of `siteIds` into the native targeting payload.

  ## Architecture

  **New `src/services/kevel_site_resolver.py`**
  `KevelSiteResolver.resolve(ref: PropertyListReference) -> ResolvedSiteIds(site_ids, unsupported_types, unresolvable_values)` — the resolver:
  1. Fetches the property list (via `resolve_property_list_typed_sync`, shared with the discovery resolver).
  2. Routes each identifier by type — `domain`/`subdomain` go to siteId lookup, other types (`ios_bundle`, `podcast_guid`, etc.) accumulate in `unsupported_types`.
  3. Looks up Kevel `siteId` by domain via paginated `GET /v1/site`. Site index is cached per `(base_url, network_id)` with a 5-min TTL.
  4. Domain normalization strips scheme/path/`www.` and lowercases — `"https://www.espn.com/sports"`, `"www.espn.com"`, and `"ESPN.COM"` all match Kevel's `Site.Url`.

  **Shared resolver refactor (`src/core/property_list_resolver.py`)**
  The existing async resolver cached `list[str]` (just identifier values). The new sync path needs `list[Identifier]` (with type). Refactored to cache typed identifiers; the existing async
  `resolve_property_list(ref) -> list[str]` keeps its signature for back-compat with the discovery caller via a thin wrapper. Extracted shared helpers (`_build_request`, `_check_cache`,
  `_store_in_cache`, `_http_error_message`) so the sync and async fetches don't duplicate logic.

  **Kevel adapter (`src/adapters/kevel.py`)**
  - `supports_property_list_filtering: ClassVar[bool] = True` at class level — opt in to the capability declaration.
  - Override `_check_property_list_supported` to validate identifier types via the resolver. Lists containing `ios_bundle`/`podcast_guid`/etc. emit `UNSUPPORTED_FEATURE` with the full type list in the
  message. Zero-match (all types supported but no sites resolved) is **accepted**, per the inventory-targeting plan SD2 — downstream `inventory_list_no_match` storyboard handles surfacing this to the
  buyer.
  - `_build_targeting` writes the resolved `siteIds` into the Kevel native targeting payload, unioned with any `custom.kevel.site_ids` the buyer already passed, and sorted for deterministic output.

  **Capabilities (`src/core/tools/capabilities.py`)**
  `property_list_filtering` is now adapter-aware: `True` iff the tenant's configured adapter sets `supports_property_list_filtering = True` at class level. This reverses B1's blanket downgrade for
  Kevel-using tenants while leaving Mock/Broadstreet/Xandr/Triton honestly declared as not supporting the feature.

  ## Tests (28 new)

  - `tests/unit/test_kevel_site_resolver.py` — 18 cases covering domain normalization (URL parsing, www-stripping, case-folding), identifier-type routing (domain compiles, ios_bundle goes to
  unsupported, mixed splits correctly), site-index caching (single fetch across repeated resolves, separate networks have separate caches, expired entry is refetched), Kevel pagination (multi-page
  response aggregation), HTTP error translation (`AdCPAdapterError` on 5xx/timeout/connection failure).
  - `tests/unit/test_kevel_property_list_compilation.py` — 10 cases pinning the three observable behaviors: class attribute (`supports_property_list_filtering=True`), identifier-type rejection in
  `_check_property_list_supported` including zero-match accept (SD2), and siteId output in `_build_targeting` including union with `custom.kevel.site_ids` and dry-run no-op.
  - `tests/unit/test_get_adcp_capabilities.py` — 2 new cases for the adapter-aware `property_list_filtering` flag (True for `supports=True` adapter, False for `supports=False` adapter).

  ## Out of scope (intentional follow-ups)

  - **App-level Kevel inventory** (`ios_bundle`, `android_package`, etc.) — Kevel API supports app-level inventory entities but they don't map to `Site.Id`. Identifying the right Kevel primitive (and
  whether to handle them in this resolver or a sibling) is a future expansion per actual buyer demand.
  - **Triton audio-type filter** — the inventory-targeting plan notes Triton can compile audio identifier types (`station_id`, `podcast_guid`, etc.). That's a separate adapter-side resolver and isn't
  part of B3. Triton continues to raise `UNSUPPORTED_FEATURE` for all `property_list` (per B4) until that work lands.
  - **B5 — faithful structural intersection in discovery** — the products-side `filter_products_by_property_list` still uses the legacy by_tag shortcut that drops some products. Tracked separately.

  ## Test plan

  - [ ] `make quality` clean (verified locally: 4491 unit pass; lint clean; mypy clean; only the pre-existing `media_buy_create.py:2789` ruff/black format mismatch remains and is not introduced by this
   PR)
  - [ ] `tox -e integration` exercises the resolver against real Postgres (no integration test added since the resolver is HTTP-based and tested at unit level with mocked Kevel responses)
  - [ ] Manual against a Kevel-configured tenant: send a `create_media_buy` request with `targeting_overlay.property_list` referencing a list of domain identifiers; confirm the Kevel campaign receives
  a `siteIds` list matching the resolved set. Confirm `get_adcp_capabilities` declares `property_list_filtering=true`.
  - [ ] Manual: send the same request with a property list containing `ios_bundle` identifiers; confirm `UNSUPPORTED_FEATURE` envelope with the type list in the message.
  - [ ] Manual: send the request with a property list whose domains aren't onboarded to Kevel; confirm the buy is created with empty/missing `siteIds` (zero-match accept-with-context, per plan SD2).

  ## Behavior change vs B4

  For Kevel-configured tenants only:
  - `get_adcp_capabilities.media_buy.features.property_list_filtering` flips from `false` (B1) → `true`.
  - `targeting_overlay.property_list` is now compiled to native `siteIds` instead of rejected with `UNSUPPORTED_FEATURE` (B4).
  - Lists containing identifier types Kevel can't compile (`ios_bundle`, etc.) still emit `UNSUPPORTED_FEATURE` — but the error message names the specific types instead of rejecting the field
  categorically.

  For all other tenants, B4's behavior holds.


KonstantinMirin (commented):
# PR #1314 — First review (B3 Kevel property_list compilation)

**Verdict**: production layering is clean (no transport leaks, no `get_db_session()` in `_impl`, adapter→service direction correct). The PR ships a non-trivial new caching/locking surface (`KevelSiteResolver`) with reasonable concurrency design, but five interconnected issues should be addressed before merge — the dominant one being a 4-agent consensus on **allowlist growth driven by real duplication** that's resolvable with a single helper extraction.

A code-review symbol-grep pass confirmed: no `_adcp_to_a2a_error`-style docstring drift this time. Every production symbol the new tests reference exists in `src/`.

## Pattern 1 — Allowlist grew because the same error literal is in 11 places (must-fix, 4-agent consensus)

Cross-flagged by testing, dry, layering, and python-practices passes — same root cause, four dimensions.

`tests/unit/test_architecture_no_error_construction_in_impl.py` `PATTERN_A_PER_FILE_CAP` raised: `src/adapters/base.py: 0 → 1` (new entry) and `src/adapters/kevel.py: 5 → 6`. The PR self-justifies in a comment ("established return-based error convention"), which is the "existing debt is not a template" anti-pattern.

Grep'd `src/adapters/` + `src/services/`: **11 sites** construct `Error(code="UNSUPPORTED_FEATURE", ...)` literally. Two are new in this PR (`base.py:416`, `kevel.py:151`). One helper on `AdServerAdapter` collapses ≥4 sites in `kevel.py` alone, *shrinks* the allowlist, kills the duplication, and prevents error-shape drift across adapters. See Comment 1.

## Pattern 2 — The Kevel cache hides real failures (must-fix on 2a; should-fix on 2b/2c)

Four findings on the same surface, dominantly from the deep-trace pass.

- **2a (must-fix)** — `kevel_site_resolver.py:215-218` uses `payload.get("items") or []` and `payload.get("totalPages") or 1`. A malformed Kevel response (`{}`, `{"items": null}`) looks like a successful empty fetch, gets cached for the full 5-minute TTL, and every subsequent `create_media_buy` silently writes `siteIds=[]` — indistinguishable from "publisher not onboarded". Violates "No Quiet Failures". See Comment 2.
- **2b (should-fix)** — sibling `src/core/property_list_resolver.py:72` still uses `del _cache[cache_key]`, the race the new resolver explicitly fixed with `pop(key, None)`. The PR adds a sync entry point that shares this `_cache`, making the race more reachable. See Comment 6.
- **2c (should-fix)** — `test_concurrent_cold_cache_resolves_share_same_lookup` (test_kevel_site_resolver.py:242) only asserts post-condition coherence, not `fetch_mock.call_count`. The docstring's "stampede is acceptable" contract isn't actually testable as written. See Comment 7.
- **2d (defer)** — `_site_cache` keys on `(base_url, network_id)` only. Moot today (per-tenant `network_id`), but if two tenants ever share one, the second tenant's API key never gates cache hits. Add `api_key` to the key or document the assumption.

## Pattern 3 — Tests assert ClassVar values, not adapter behavior (must-fix)

`tests/unit/test_adapter_property_list_unsupported.py:172-196` — three of the four reject-adapter test classes contain exactly one assertion: `assert <Adapter>.supports_property_list_filtering is False`. Removing `_reject_property_list_if_unsupported` from Broadstreet/Xandr/Triton's `create_media_buy` path leaves these tests green. Only Mock has a real end-to-end test. Parameterize the existing Mock body over the four adapters. See Comment 3.

## Pattern 4 — Identifier-type classification duplicated between adapter and resolver (must-fix)

Same loop in two places:
- `src/adapters/kevel.py:101-105` (dry-run branch of `_resolve_property_list`)
- `src/services/kevel_site_resolver.py:121-125` (live `resolve`)

Adapter even imports `SUPPORTED_IDENTIFIER_TYPES` as a module constant from the resolver to support the duplicate loop — a sign that the helper belongs on the resolver. See Comment 5. Fixes a layering finding too (resolver no longer needs to expose the constant publicly).

## Pattern 5 — Two resolvers with inconsistent concurrency models (medium, document or unify)

This PR ships two cache-bearing modules side-by-side:

| | `src/core/property_list_resolver.py` | `src/services/kevel_site_resolver.py` |
|--|---|---|
| State | module-level `_cache: dict` | ClassVar `_site_cache: dict` |
| Lock | none | `threading.Lock` |
| Removal | `del _cache[key]` (race) | `pop(key, None)` (safe) |
| Surface | functional helpers | class with `clear_cache()` |

The core resolver predates this PR; the PR's new sync entry point inherits the inconsistency rather than fixing it. Pick one cache idiom (probably the class-with-lock one) or document why they diverge.

## Pattern 6 — Python/connection-pool details (mixed)

- **PP-04 (must-fix)** — `with httpx.Client(...)` inside the `while page <= total_pages:` loop at `kevel_site_resolver.py:206`. New connection per page; defeats keep-alive. The cache lock that was added to amortize this exact work is fighting against per-page reconnect. See Comment 4.
- **PP-06 (should-fix)** — `self._site_resolver: KevelSiteResolver | None = None` annotation lives only on the dry-run branch at `kevel.py:57`. Live branch assigns without annotation. The `if self._site_resolver is None` guard at line 99 is a disguised `if self.dry_run`. Move the annotation up, or drop the optional and the dead guard. See Comment 5.

## What's clean (positive signal)

- **No docstring drift**: every production symbol referenced in test docstrings (`_reject_property_list_if_unsupported`, `_site_resolver`, `_resolve_property_list`, `_property_list_cache`, `KevelSiteResolver._cache_lock`, `clear_cache`) was grep-verified in `src/`. Explicit fix from the PR #1313 process gap.
- **Layering**: no transport imports under `src/`, no `ToolError`, no `get_db_session()` in new code.
- **`property_list_resolver.py` refactor**: extracts clean helpers (`_build_request`, `_check_cache`, `_store_in_cache`, `_http_error_message`).
- **`_normalize_domain` + `TestNormalizeDomain` table**: textbook good test.
- **Dry-run identifier-type validation**: pinned by `test_dry_run_still_rejects_unsupported_identifier_types`. `ios_bundle`-only lists surface as `UNSUPPORTED_FEATURE` in dry-run instead of silent accept.
- **Per-request resolver memoization**: pinned by `test_resolver_called_once_per_request_across_check_and_build`.
- **Kevel capability declaration**: uses the same `ClassVar` shape as the 4 rejecting adapters; `test_adapter_with_property_list_support_advertises_capability` closes the TRUE-path coverage.

## Defer to follow-up

- Pattern 2d (multi-tenant cache key)
- Integration coverage of the wire path (defensible if B5 #1315 adds it; risky if it stays unit-only)
- Test-fixture extraction across the 3 new test files
- Naming + style consistency nits (`details=None` vs `details={...}`, `Error` import source, missing cache-hit DEBUG log)

tests/unit/test_architecture_no_error_construction_in_impl.py
```diff
@@ -42,9 +42,15 @@
 # When cleanup sub-batches land, drop the relevant entry below to zero rather
 # than gradually lowering it — keep the cap honest.
 PATTERN_A_PER_FILE_CAP: dict[str, int] = {
+    # property_list-unsupported rejection: adapters return
+    # CreateMediaBuyError(errors=[Error(code="UNSUPPORTED_FEATURE")]) following the
+    # established return-based error convention (callers isinstance-check the
+    # returned *Error result — adapters cannot raise). Same convention as the
+    # sibling kevel/broadstreet/gam sites below.
+    "src/adapters/base.py": 1,
     "src/adapters/broadstreet/adapter.py": 13,
```
KonstantinMirin, line 51: **Allowlist must shrink, never grow.** `base.py: 0 → 1` is a new entry and `kevel.py: 5 → 6`; the justification comment is exactly the "existing debt is not a template" pattern that `feedback_no_allowlist_growth_ever.md` blocks on.

The root cause is real duplication — `grep -rn 'code="UNSUPPORTED_FEATURE"' src/adapters/ src/services/` returns **11 sites** constructing `Error(code="UNSUPPORTED_FEATURE", ...)` literally (4 in `kevel.py`, 3 in `google_ad_manager.py`, 2 in `triton_digital.py`, 1 each in `broadstreet/adapter.py`, `base.py:416` — new in this PR — and `services/targeting_capabilities.py`).

Proposed: extract on `AdServerAdapter`
```python
def _unsupported_feature_error(self, message: str) -> CreateMediaBuyError:
    return CreateMediaBuyError(
        errors=[Error(code="UNSUPPORTED_FEATURE", message=message, details=None)],
    )
```
Calling sites collapse from a 3-line literal to one helper call. The allowlist count *decreases* (`kevel.py: 5 → 1`; the new `base.py: 0 → 1` stays). Error-shape drift across adapters becomes impossible. Cross-flagged by testing TQ-06, dry DRY-03, layering LR-01, python PP-08.

src/services/kevel_site_resolver.py
```diff
@@ -0,0 +1,251 @@
...
+            except (httpx.HTTPStatusError, httpx.TimeoutException, httpx.RequestError) as exc:
+                raise AdCPAdapterError(
+                    f"Failed to fetch Kevel site list (network {self.network_id}, page {page}): {exc}"
+                ) from exc
+
+            payload = response.json()
+            items = payload.get("items") or []
+            sites.extend(items)
+            total_pages = payload.get("totalPages") or 1
+            if page >= total_pages:
```
KonstantinMirin, line 218: `payload.get("items") or []` and `payload.get("totalPages") or 1` silently absorb malformed responses. A `{}` or `{"items": null, "totalPages": null}` from Kevel looks like a successful single-page fetch returning zero sites. That zero-site lookup is cached for the full 5-minute TTL. Every subsequent `create_media_buy` against this network silently compiles `siteIds=[]` — indistinguishable from "publisher not onboarded".

`CLAUDE.md` "No Quiet Failures". Fix:
```python
if "items" not in payload or "totalPages" not in payload:
    raise AdCPAdapterError(
        f"Kevel /v1/site response missing required keys "
        f"(network {self.network_id}, page {page}): {sorted(payload)}"
    )
items = payload["items"]
total_pages = payload["totalPages"]
```
An actual empty list (`items: []`) is fine to cache. A missing key is not.

tests/unit/test_adapter_property_list_unsupported.py
```diff
@@ -0,0 +1,210 @@
...
+    """Triton: audio-type accept path is gated on B3's resolver.
+
+    Until the property_list → identifier resolver lands, Triton rejects all
+    property_list targeting with UNSUPPORTED_FEATURE. The class attribute will
+    flip to True (with an override of _check_property_list_supported that
+    inspects identifier types) once B3 wires the resolver.
+    """
+
+    def test_class_default_unsupported(self):
+        assert TritonDigital.supports_property_list_filtering is False
```
KonstantinMirin, line 196: These three classes each do exactly one assertion:
```python
class TestBroadstreetAdapterRejectsPropertyList:
    def test_supports_property_list_filtering_is_false(self):
        assert BroadstreetAdapter.supports_property_list_filtering is False
```
Same shape for Xandr and Triton. If someone removes `_reject_property_list_if_unsupported` from those three adapters' `create_media_buy` path, these tests still pass — they verify the *declaration*, not the *behavior*. Only `TestMockAdapterRejectsPropertyList` (line 158) actually calls `create_media_buy()` and asserts on the returned envelope.

Parameterize:
```python
@pytest.mark.parametrize("adapter_cls", [MockAdServer, BroadstreetAdapter, XandrAdapter, TritonDigital])
def test_create_media_buy_returns_unsupported_envelope(self, adapter_cls):
    # the body that exists for Mock today, generalized
```
One ~15-line parametrization replaces three trivial classes and gives real coverage. Flagged by testing TQ-01.

src/services/kevel_site_resolver.py
```diff
@@ -0,0 +1,251 @@
...
+    def _fetch_all_sites(self) -> list[dict]:
+        """Page through Kevel ``GET /v1/site`` and return the aggregated list."""
+        sites: list[dict] = []
+        page = 1
+        while True:
+            url = f"{self.base_url}/site?page={page}&pageSize={_KEVEL_SITE_PAGE_SIZE}"
+            try:
+                with httpx.Client(timeout=self.timeout_seconds) as client:
+                    response = client.get(url, headers={"X-Adzerk-ApiKey": self.api_key})
+                    response.raise_for_status()
```
KonstantinMirin, line 208: `httpx.Client(...)` is constructed inside the pagination loop — a new TCP connection per page, no keep-alive. The cache lock that was added to amortize this exact fetch ends up fighting against per-page reconnect.

Hoist the client outside the loop:
```python
with httpx.Client(
    timeout=self.timeout_seconds,
    headers={"X-Adzerk-ApiKey": self.api_key},
) as client:
    while True:
        url = f"{self.base_url}/site?page={page}&pageSize={_KEVEL_SITE_PAGE_SIZE}"
        try:
            response = client.get(url)
            response.raise_for_status()
        except (httpx.HTTPStatusError, httpx.TimeoutException, httpx.RequestError) as exc:
            ...
```
Flagged by python PP-04.

src/adapters/kevel.py
```diff
@@ -47,17 +54,104 @@ def __init__(
...
+            identifiers = resolve_property_list_typed_sync(ref)
+            unsupported_types: set[str] = set()
+            for ident in identifiers:
+                ident_type = ident.type.value if hasattr(ident.type, "value") else str(ident.type)
+                if ident_type not in SUPPORTED_IDENTIFIER_TYPES:
+                    unsupported_types.add(ident_type)
+            resolved = ResolvedSiteIds(site_ids=set(), unsupported_types=unsupported_types, unresolvable_values=[])
+        else:
+            resolved = self._site_resolver.resolve(ref)
+
```
KonstantinMirin, line 110: **Two issues stacked on the same branch.**

**(a) Identifier-type classification duplicated** — the loop at lines 101-105 reproduces `KevelSiteResolver.resolve()` at `src/services/kevel_site_resolver.py:121-125`: same `ident.type.value if hasattr(...) else str(...)` extraction, same `unsupported_types` set construction. The adapter has to import `SUPPORTED_IDENTIFIER_TYPES` from the resolver module (line 13) just to do this — a strong signal the helper belongs on the resolver.

Proposed:
```python
# on KevelSiteResolver
@classmethod
def classify_identifier_types(cls, identifiers) -> set[str]:
    return {
        (i.type.value if hasattr(i.type, "value") else str(i.type))
        for i in identifiers
        if (i.type.value if hasattr(i.type, "value") else str(i.type)) not in cls.SUPPORTED_IDENTIFIER_TYPES
    }
```
Both the live `resolve()` and the adapter's dry-run branch call this; `SUPPORTED_IDENTIFIER_TYPES` becomes a private class attribute (drop the module-level export).

**(b) `_site_resolver` annotation is dry-run-only** — `self._site_resolver: KevelSiteResolver | None = None` lives only on the dry-run branch at line 57. The live branch (line 62) assigns the same attribute *without* annotation. The `if self._site_resolver is None:` guard at line 99 is effectively `if self.dry_run:` — type-narrowing made invisible. Move the annotation above the branch so both arms share it:
```python
self._site_resolver: KevelSiteResolver | None = None
if self.dry_run:
    self.log(...)
elif not self.network_id or not self.api_key:
    raise ValueError(...)
else:
    self.headers = {...}
    self._site_resolver = KevelSiteResolver(...)
```
Flagged by dry DRY-01, python PP-06 + PP-11, layering LR-05.

src/core/property_list_resolver.py
```diff
@@ -39,79 +51,113 @@ def _validate_agent_url(agent_url: str) -> None:
...
+
+
+def _check_cache(agent_url: str, list_id: str) -> list[Identifier] | None:
+    """Return cached identifiers if present and not expired; drop the entry on expiry."""
+    cache_key = (agent_url, list_id)
+    if cache_key not in _cache:
+        return None
+    identifiers, expires_at = _cache[cache_key]
+    if datetime.now(UTC) >= expires_at:
+        del _cache[cache_key]
```
KonstantinMirin, line 72: `del _cache[cache_key]` is the race the new `KevelSiteResolver` explicitly fixed with `pop(key, None)`. Two callers, one finds the key expired, both try `del`, second raises `KeyError`.

This PR adds a new sync entry point that shares this `_cache`, so the latent race is now more reachable. Mirror the safe pattern:
```python
_cache.pop(cache_key, None)
return None
```
Flagged by python PP-02, layering LR-02.

tests/unit/test_kevel_site_resolver.py
```diff
@@ -0,0 +1,375 @@
...
+            resolver.resolve(_ref())
+            # Manually expire the cache entry
+            cache_key = (resolver.base_url, resolver.network_id)
+            lookup, _expires = KevelSiteResolver._site_cache[cache_key]
+            KevelSiteResolver._site_cache[cache_key] = (lookup, datetime.now(UTC) - timedelta(seconds=1))
+            resolver.resolve(_ref())
+
+        assert fetch_mock.call_count == 2
+
+    def test_concurrent_cold_cache_resolves_share_same_lookup(self):
```
KonstantinMirin, line 242: `test_concurrent_cold_cache_resolves_share_same_lookup` asserts only post-condition coherence (both threads got the same result), not `fetch_mock.call_count == 1`. The docstring claims "stampede is acceptable" but that's not actually pinned by the test — and the lock that exists in production *should* prevent the stampede.

Either:
- assert `fetch_mock.call_count == 1` (and rename → `_lock_prevents_stampede`), or
- accept the stampede semantics, rename the test + docstring to match, and add a separate test that pins the lock-held-across-call invariant.

The companion `test_concurrent_expired_cache_drop_does_not_raise` admits in its own docstring "Two sequential calls" — it isn't concurrent. Either make it concurrent or rename. Flagged by testing TQ-02, deep-trace M2.

Issue #1038, opened 2026-02-12
update_media_buy with creative_ids does not sync to GAM


estofko-scope3:
When `update_media_buy` is called with `creative_ids` to add/remove creatives from a package, the changes are saved to the database but never uploaded to GAM. Creatives assigned after initial approval will not appear in the GAM line items.

**Symptoms:**
1. Call update_media_buy with creative_ids to assign creatives to a package
2. Database shows the assignment exists in creative_assignments table
3. GAM order/line item has no creatives associated

**Impact**
- Affected: Any creative assignment changes after initial media buy approval
- User expectation: update_media_buy with creative_ids should sync to GAM automatically

**Note**
This fix depends on `platform_line_item_id` being present (see [related issue](https://github.com/prebid/salesagent/issues/1037)). The GAM creatives manager uses the order ID to query line items, so it may work independently, but full functionality requires both fixes.

ChrisHuie:
Closing in favor of #1529.

- The `creative_ids` field this issue is framed around was removed from the AdCP package-update object at spec 3.0.0; the current (3.1.0-beta.3) mechanisms are `creative_assignments` and `creatives`. The replacement issue is reframed accordingly.
- Both blockers cited here are resolved: #1037 (platform_line_item_id) and #1046 (clickthrough URL).
- PR #1337 addressed only the `approve_creative` retroactive-push path and explicitly did not cover the `update_media_buy` path. The underlying defect — assignment changes via `update_media_buy` never reaching the ad server — remains, and is tracked in #1529 along with its pitfalls and pre-implementation decisions.

PR #1315, opened 2026-05-17
feat(products): faithful property_list intersection (Stream B5)

ChrisHuie:
Replaces the legacy `filter_products_by_property_list` shortcut with a faithful
  intersection service that handles all three AdCP selector variants (`all`,
  `by_id`, `by_tag`) correctly. Fixes the silent drop of `by_tag`-only products
  flagged in AdCP RESEARCH.md Finding 3 (#1247).

  This is Stream B5 of the inventory-targeting plan, the final piece — with this
  landed, Stream B (B1–B6) is complete.

  ## Stacked on

  `feature/b3-kevel-property-list-compilation` (PR #1314).
  Auto-rebases when #1314 merges.

  ## What changes

  - `AuthorizedPropertyRepository` — tenant-scoped repo with `list_by_domain()`
    and `list_by_tags()` for resolving `by_tag` selectors. JSONB-overlap done
    client-side after a single domain-scoped fetch.
  - `PropertyIntersection.filter_products` — routes selectors by `selection_type`:
    - `all`    → unbounded, always include
    - `by_id`  → direct intersection with explicit property_ids
    - `by_tag` → resolve via repo, intersect with allowed set
  - `IntersectionResult(kept_products, dropped_products)` carries structured
    `DropReason` (`NO_RESOLVABLE_PROPERTIES`, `NO_PROPERTY_OVERLAP`,
    `STRICT_MODE_VIOLATION`). The `zero_match` property drives advisory logging.
  - UoW wiring: `ProductUoW.authorized_properties` and
    `MediaBuyUoW.authorized_properties` attached — discovery + create_media_buy
    reach the repo via existing UoW.
  - Discovery (`_get_products_impl`): swaps `filter_products_by_property_list`
    for `PropertyIntersection`. By-tag products survive when resolved IDs
    overlap the buyer's allowed set.
  - Create-media-buy (`_emit_property_list_advisories`): WARN-level advisory
    on zero-overlap, never raises (plan SD2 accept-with-context). Strict
    `property_targeting_allowed` validation above continues to reject the
    genuinely-illegal case. 
  
  ## Test plan

  - [ ] `tox -e unit` — new suite `tests/unit/test_property_intersection.py` (14 cases) passes
  - [ ] All three selector variants covered: `all`, `by_id`, `by_tag`
  - [ ] Mixed-selector aggregation case asserts kept/dropped accounting
  - [ ] Strict vs permissive `property_targeting_allowed` preserved
  - [ ] Zero-match advisory flag asserted
  - [ ] `tox -e integration` — discovery path still green under real Postgres
  - [ ] Architecture guards pass (line-number shift in `products.py` allowlist already adjusted)


KonstantinMirin:
## Review summary

Thanks for this — the concurrency consolidation in particular is genuinely strong work: `ThreadRegistry`, the `lifecycle.register_shutdown` inversion, the `pin_task` weak-ref fix, the dual-dict ABBA avoidance, and the running-loop detection before `asyncio.run`/`create_task` are all correct and well-documented.

I'd like to hold merge on a few things, though. Nothing here is unfixable, but there's a cluster of real issues plus a testing-approach gap I think we should close first. Grouped by theme below; I've tried not to drop anything.

### Testing approach (the thing I'd most like to discuss)

The branch *does* add integration, BDD, e2e, factory and harness tests — credit for that. But the **newest and riskiest components** (`kevel_site_resolver`, `thread_registry`, the reaper consolidation, and the new cache layers) are covered **only by mock-based unit tests**, with no integration/harness/BDD exercise. That matters here specifically because:

- **HIGH-02 — vacuous-pass test.** `test_execute_approved_pending_review_filter.py::test_pending_review_creative_not_uploaded` mocks only two `.first()` results, but `execute_approved_media_buy` makes a third (`media_buy_create.py:695`, the Product load). The third call raises `StopIteration`, the per-package `try/except` swallows it, and the function returns *before* the creative filter runs. `add_creative_assets.assert_not_called()` then passes because the code crashed — not because pending creatives were filtered. The test would still pass if the filter were deleted. This is the core #1038 hold-back guarantee, currently untested while looking covered. Could we drive this through the existing harness rather than a hand-built session mock?
- The three concurrency findings below (HIGH-01, MED-02, MED-03) are precisely the class of bug a mock can't see. A concurrent integration test against the real path would.
- **TQ-02** — `test_class_default_unsupported` asserts only a class-attribute literal; behaviorally weak in isolation.

I'd suggest the new external-integration and concurrency components get the same integration/harness treatment the property-intersection enforcement already got (`test_property_targeting_allowed_enforcement.py` is a good model).

### Concurrency / thread-safety

- **HIGH-01** — `kevel_site_resolver.py:80,135-141`: `_site_cache` is a class-level dict with an unlocked check-then-`del` on expiry, and there's no lock in the file. Kevel `create_media_buy` runs synchronously in request threads, so concurrent calls for the same network race the `del` into a `KeyError` (plus duplicate paginated fetches). Fix: class-level `Lock`, and `pop(key, None)` instead of `del`.
- **MED-02** — `property_list_resolver.py` `_check_cache`: same unguarded `del`-on-expiry, now reachable from the multithreaded sync path. Same one-line fix; worth doing together with HIGH-01.
- **MED-03** — `background_approval_service.py:46-58`: TOCTOU between `contains()` and `add()` — two callers with the same `thread_id` can both spawn a polling thread. Suggest adding `ThreadRegistry.add_if_absent(key, thread) -> bool` (check+add under the lock) and branching on it. (The delivery simulator and GAM reporting manager already guard this way.)
- **PP-04** — `thread_registry.py:40-46`: `add_reap_callback` mutates `_reap_callbacks` outside the lock. Benign today (constructor-time registration) but contradicts the docstring's "every method locks." Either lock it or document the constraint.
- **PP-05** — eager f-strings in logging across the new services; notably `protocol_webhook_service.py:158-160` logs the full webhook `payload` at debug. Prefer `%s` lazy form so it's never built when debug is off (and to avoid logging buyer payloads).

### Duplication (DRY)

- **HIGH-03** — the per-package targeting-overlay validation loop is copy-pasted between `media_buy_create.py:2144-2155` and `media_buy_update.py:327-335`, and it's **already diverged** (create raises `ValueError`, update raises `AdCPValidationError`). Suggest extracting `validate_package_overlay(overlay) -> list[str]` and letting each caller choose the exception.
- **HIGH-04** — `products.py:70-147` (`extract_product_property_ids`, `should_include_product_for_property_list`, `filter_products_by_property_list`) is dead in production (no callers; `PropertyIntersection` replaced it) but still pinned by tests — and per the new docstring it's the *buggy* version that "silently dropped products." Suggest deleting it and migrating those tests onto `PropertyIntersection`.
- **DRY-03** — `targeting_overlay` rehydration with the legacy-key fallback is duplicated (`media_buy_create.py:769`, `media_buy_list.py:205`); the lookup contract should live in one helper even though the error policy differs.
- **DRY-04** — two `_normalize_domain` implementations with subtly different rules (`kevel_site_resolver.py:194`, `property_discovery_service.py:35`).

### Layering / placement

- **LR-01** — `KevelSiteResolver` does live `httpx` calls (Layer 4 / adapter work) but lives under `src/services/`, and is only consumed by the Kevel adapter. Suggest relocating under `src/adapters/`.
- **MED-04 / LR-02** — `media_buy_delivery.py:302-312`: inline `AuditLog(...)` + `uow.session.add()` in `_impl` (Pattern #3). Tracked as `FIXME(salesagent-9f2)`; fine to keep as tracked debt, flagging since the file was rewritten here.
- **LR-03** — `media_buy_delivery.py:211-212` hand-builds `ProductRepository(uow.session, ...)`. This branch *added* `MediaBuyUoW.products` (`uow.py:133`), so the fix is now available — suggest `uow.products` here.
- **MED-01 / CON-01** — the new `supports_property_list_filtering` flag (`base.py:197`, `kevel.py:23`) is a bare `ClassVar` read via `getattr(adapter.__class__, ...)`, while every other `supports_*` flag is a field on the `AdapterCapabilities` dataclass (`base.py:124-148`). Suggest moving it there.

### Minor / consistency

- **CON-02** — `kevel_site_resolver.py:170-177` hand-rolls an HTTP error message; the sibling `property_list_resolver` (same branch) extracted a shared `_http_error_message()` formatter — reuse it.
- **CON-03** — `kevel_site_resolver.py` uses `httpx` while the owning `kevel.py` adapter uses `requests`. Defensible (shares the httpx cache), but worth a comment noting why.
- **PP-06** — `schemas/delivery.py:299-349`: `model_dump` override + the nested-serializer mixin double-serialize. Correct, just a perf smell; no action unless responses get large.
- **PP-07** — `schemas/_base.py:1028-1033`: the Pydantic v1 `Targeting.dict()` override is retained (pre-existing). Migrate callers to `model_dump()` and drop it in a follow-up.

One meta-note: `kevel_site_resolver.py` shows up in five of these (HIGH-01, CON-02, CON-03, DRY-04, LR-01) — it's the least-settled new file, so it might be worth a focused second pass there.

Happy to pair on any of these — the two cache fixes (HIGH-01 + MED-02) are a couple of lines, and HIGH-04 is a clean deletion.

Issue #1317, opened 2026-05-18
bug(impl): _get_media_buy_delivery_impl does not resolve AccountReference [P1 correctness]


KonstantinMirin:
## Summary

`src/core/tools/media_buy_delivery.py` — IMPL accepts the `account` kwarg but never resolves it against the DB. Account resolution only happens at the REST wrapper boundary (`src/routes/api_v1.py:265-289`); A2A, MCP, and direct-IMPL calls all skip it.

## Impact

Cross-transport contract is broken. Same buyer with same token gets different behavior depending on which transport they use. `account_not_found` rows pass through IMPL silently.

## Fix

Move `enrich_identity_with_account` from the REST wrapper into `_impl` (or into a shared helper that every transport calls before `_impl`). Paired with the A2A fix (C1).

## Acceptance

- [ ] `_get_media_buy_delivery_impl` resolves `AccountReference` against DB, raising appropriately on `account_not_found`
- [ ] All four transports (IMPL, A2A, MCP, REST) return identical results for the same buyer+token+account input
- [ ] REST wrapper no longer needs to call `enrich_identity_with_account` (moved into _impl) — or shared helper applies uniformly
- [ ] BDD `T-UC-004-partition-account` and `-boundary-account` error-code rows flip from xfail to pass on all transports
- [ ] `c2` marker block in `tests/bdd/conftest.py` can flip to `strict=True` (FIXME removed, doc entry removed)

## References

- Tracked: `docs/test-debt-bdd-strict-markers.md#c2` (item C2)
- Pair: item C1 (A2A skill drops account param)
- Audit: Batch 3, 2026-05-08

ChrisHuie:
Two related notes — and folding a duplicate of ours into this issue.

1. PR #1438 forwards the validated `req.account` in the A2A `get_media_buy_delivery` skill (closes #1434, the current-symptom restatement of #1316). It does not move resolution into `_get_media_buy_delivery_impl`, so this issue's core ask — resolve `AccountReference` inside `_impl` so IMPL/A2A/MCP/REST share one path and REST drops its boundary-only `enrich` — remains open.

[...]

Issue #1316, opened 2026-05-18
bug(a2a): silently discards account parameter on get_media_buy_delivery [P1 security]


KonstantinMirin:
## Summary

`src/a2a_server/adcp_a2a_server.py:1937-1980` — the A2A skill handler for `get_media_buy_delivery` does not forward the `account` parameter to `_raw()` and does not call `enrich_identity_with_account`.

## Impact

**Security gap.** A buyer with a token scoped to one tenant could request delivery against another account, and the account scope is silently ignored at deserialize time. The REST transport correctly resolves account at `src/routes/api_v1.py:265-289`; A2A and IMPL skip the step.

## Reproduction

BDD scenarios `T-UC-004-partition-account` and `T-UC-004-boundary-account`, error-code rows on the `[a2a]` transport, currently xpass on REST/MCP and xfail on IMPL/A2A.

## Threat model

- A2A endpoint is exposed to untrusted callers in production — that is the protocol's purpose
- A tenant's token authorizes them to query only accounts they own — but the parameter that scopes that is being dropped
- No known abuse; logs may not surface the discard since the parameter is silently ignored at deserialize time

## Fix

Forward `account` to `_raw()` in `_handle_get_media_buy_delivery_skill`; call `enrich_identity_with_account` on the A2A path. Paired with C2 (move resolution into `_impl` so all transports share the step).

## Acceptance

- [ ] A2A handler forwards `account` to `_raw()`
- [ ] `enrich_identity_with_account` runs on the A2A path
- [ ] BDD `T-UC-004-partition-account` and `-boundary-account` error-code rows on `[a2a]` transport flip from xfail to pass
- [ ] When fixed, the `c1` marker block in `tests/bdd/conftest.py` can flip to `strict=True` (FIXME removed, doc entry removed)

## References

- Tracked: `docs/test-debt-bdd-strict-markers.md#c1` (item C1)
- Pair: item C2 (IMPL `_get_media_buy_delivery_impl` does not resolve `AccountReference`)
- Audit: Batch 3, 2026-05-08

ChrisHuie:
The symptom here has drifted since this was filed. As of #1334, the A2A handler does forward `account` — but as the raw params dict, so `get_media_buy_delivery` now raises `AttributeError: 'dict' object has no attribute 'root'` in `resolve_account` (account_helpers.py) rather than silently discarding the scope. That current symptom is tracked in #1434.

[...]

PR #1320, opened 2026-05-19
fix: resolve delivery account references across transports

### LINKED ISSUE #1318: bug(repo): cross-principal media_buy access returns 200+empty instead of 403 [P1 security]
## Summary

`src/core/database/repositories/media_buy.py:99-107`, `MediaBuyRepository.get_by_principal`, filters silently by `principal_id`. A request from a principal who does not own any of the requested media_buys receives an empty deliveries list, not an authorization error.

## Impact

**Security gap.** A 200 OK with empty body is indistinguishable from "everything ran, nothing matched." A 403 (or 404 with suggestion) would correctly indicate the access was denied.

The BDD ownership scenarios were the only place this could be observed — and they do not actually swap identity (see B3 — separate test bug), so the gap is invisible to the suite today.

## Threat model

- A buyer cannot enumerate other principals' media_buy_ids directly, but a guess-and-check attacker could
- A caching layer might cache the empty response — worth checking that it doesn't leak

## Fix

Raise `AdCPAuthorizationError` (or `AdCPNotFoundError` with suggestion text "media_buy not found or not owned by you") when the requesting principal does not own any of the requested media_buys.

## Acceptance

- [ ] `MediaBuyRepository.get_by_principal` (or its caller) raises `AdCPAuthorizationError` / `AdCPNotFoundError` when none of the requested media_buy_ids belong to the requesting principal
- [ ] A partial-overlap case (some owned, some not) is decided and tested — either 403 for the entire request or return only the owned subset with a warning, per AdCP spec interpretation
- [ ] BDD `T-UC-004-partition-ownership` mismatch rows flip from xfail to pass (after B3 also lands so identity actually swaps)
- [ ] `c3` marker block in `tests/bdd/conftest.py` can flip to `strict=True` (FIXME removed, doc entry removed)
- [ ] Pair with B3: BDD test fix to actually swap identity, otherwise the security fix is untested

## References

- Tracked: `docs/test-debt-bdd-strict-markers.md#c3` (item C3)
- Pair: item B3 (resolution/ownership scenarios use symbolic names that never swap identity)
- Audit: Batch 3, 2026-05-08

sangilish:
## Summary
- Resolve `GetMediaBuyDeliveryRequest.account` in the shared delivery `_impl` path after identity/principal-id validation and before principal lookup.
- Pass account references through A2A, REST, MCP, and raw delivery callers so transport behavior stays consistent.
- Add account-resolution regression coverage for explicit account IDs, natural-key accounts, missing accounts, and A2A forwarding.
- Seed UC-004 account rows with real account/access fixtures without editing generated feature files.

## Validation
- `make quality`
- `uv run pytest tests/unit/test_delivery.py tests/unit/test_rest_api_endpoints.py tests/unit/test_mcp_tool_type_alignment.py tests/harness/test_harness_delivery_poll.py -q`
- `uv run pytest tests/unit/test_delivery_account_resolution_flow.py -q`
- `uv run pytest tests/unit/test_auth_consistency.py::TestMissingTokenConsistency::test_get_media_buy_delivery_missing_auth_returns_error_response -q`
- `uv run pytest tests/unit/test_pydantic_schema_alignment.py::TestPydanticSchemaAlignment::test_model_accepts_all_schema_fields[/schemas/latest/media-buy/get-media-buy-delivery-request.json-GetMediaBuyDeliveryRequest] -q`

## Not fully verified locally
- `uv run pytest tests/harness/test_delivery_account_resolution.py -q -rs` skipped 13 DB-backed cases because local PostgreSQL `DATABASE_URL` is unavailable.
- `uv run pytest tests/bdd/test_uc004_deliver_media_buy_metrics.py -k "account" -q -rs` skipped 60 selected UC-004 BDD cases for the same reason.

## Scope note
This PR fixes account reference resolution/authorization path consistency for #1316 and #1317. It intentionally does not claim to fix #1318 cross-principal ownership semantics or add delivery-result filtering by `account_id`.

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1967,6 +1967,7 @@ class AdCPRequestHandler(RequestHandler):
                 status_filter=params.get("status_filter"),
                 start_date=params.get("start_date"),
                 end_date=params.get("end_date"),
+                account=req.account,
                 context=params.get("context"),
                 identity=identity,
             )
--- a/src/core/tools/media_buy_delivery.py
+++ b/src/core/tools/media_buy_delivery.py
@@ -105,6 +105,13 @@ def _get_media_buy_delivery_impl(
             context=context_val,
         )
 
+    account_ref = getattr(req, "account", None)
+    if account_ref is not None:
+        from src.core.transport_helpers import enrich_identity_with_account
+
+        identity = enrich_identity_with_account(identity, account_ref)
+        assert identity is not None  # identity was validated above
+
     # Get the Principal object
     principal = get_principal_object(principal_id, tenant_id=identity.tenant_id)
     if not principal:
@@ -614,12 +621,6 @@ async def get_media_buy_delivery(
     """
     identity = (await ctx.get_state("identity")) if isinstance(ctx, Context) else None
 
-    # Handle account resolution at boundary (same as sync_creatives pattern)
-    if account is not None and identity is not None:
-        from src.core.transport_helpers import enrich_identity_with_account
-
-        identity = enrich_identity_with_account(identity, account)
-
     # Create AdCP-compliant request object
     try:
         req = GetMediaBuyDeliveryRequest(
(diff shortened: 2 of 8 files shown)


sangilish:
I have read the IPR Policy

ᐧ

2026년 5월 19일 (화) 오전 10:52, github-actions[bot] ***@***.***>님이
작성:

sangilish:
recheck

sangilish:
Superseded by #1321.

PR #1321, opened 2026-05-19
fix: resolve delivery account references across transports

sangilish:
## Summary
- Resolve `GetMediaBuyDeliveryRequest.account` in the shared delivery `_impl` path after identity/principal-id validation and before principal lookup.
- Pass account references through A2A, REST, MCP, and raw delivery callers so transport behavior stays consistent.
- Add account-resolution regression coverage for explicit account IDs, natural-key accounts, missing accounts, and A2A forwarding.
- Seed UC-004 account rows with real account/access fixtures without editing generated feature files.

## Validation
- `make quality`
- `uv run ruff check tests/integration/test_delivery_account_resolution.py tests/helpers/account_resolution.py`
- `uv run pytest tests/unit/test_delivery.py tests/unit/test_rest_api_endpoints.py tests/unit/test_mcp_tool_type_alignment.py tests/harness/test_harness_delivery_poll.py -q`
- `uv run pytest tests/unit/test_delivery_account_resolution_flow.py -q`
- `uv run pytest tests/unit/test_auth_consistency.py::TestMissingTokenConsistency::test_get_media_buy_delivery_missing_auth_returns_error_response -q`
- `uv run pytest tests/unit/test_pydantic_schema_alignment.py::TestPydanticSchemaAlignment::test_model_accepts_all_schema_fields[/schemas/latest/media-buy/get-media-buy-delivery-request.json-GetMediaBuyDeliveryRequest] -q`
- `uv run pytest tests/integration/test_delivery_account_resolution.py --collect-only -q -m "delivery"`

## Not fully verified locally
- `uv run pytest tests/integration/test_delivery_account_resolution.py -q -rs` skipped 13 DB-backed cases because local PostgreSQL `DATABASE_URL` is unavailable.
- `scripts/run-test.sh --db tests/integration/test_delivery_account_resolution.py -q -rs` could not start the local test database because the Docker daemon is unavailable on this machine.
- `uv run pytest tests/bdd/test_uc004_deliver_media_buy_metrics.py -k "account" -q -rs` skipped 60 selected UC-004 BDD cases for the same local DB reason.

## CI coverage note
The DB-backed delivery account regression now lives under `tests/integration` and is marked `integration`, `requires_db`, and `delivery`, matching the repo GitHub Actions integration-test path for the media-buy/delivery matrix.

## Scope note
[...]

PR #1330, opened 2026-05-19
fix(ci): restore security audit and schema alignment

sangilish:
## Summary
- Add an explicit `idna>=3.15` security floor and refresh `uv.lock` from `idna 3.11` to `3.15`.
- Update the delivery request schema-alignment allowlist for current upstream AdCP `latest` fields: `time_granularity` and `include_window_breakdown`.
- Keep this as a small CI-unblock PR separate from the delivery account-resolution behavior change in #1321.

## Why
Two independent main-branch CI gates are currently blocking PR verification:

1. `Security Audit` fails on `GHSA-65pc-fj4g-8rjx` for transitive `idna==3.11`; the advisory reports `3.15` as the fixed version.
2. `Unit Tests` fail in `test_pydantic_schema_alignment.py` because AdCP `latest` added delivery pull-window fields that are not yet in the Python library/model. This is a known schema-vs-library mismatch, not a runtime behavior change.

## Validation
- `uv lock --check`
- `uvx uv-secure --no-check-uv-tool --ignore-vulns GHSA-7gcm-g887-7qv7,GHSA-5239-wwwm-4pmq`
- `uv run --locked python -c "import idna; print(idna.__version__)"` -> `3.15`
- `uv run pytest tests/unit/test_pydantic_schema_alignment.py -q --tb=short`
- `uv run pytest tests/unit/ -q --tb=short`
- `git diff --check`

Note: local `uvx uv-secure` without `--no-check-uv-tool` also scans my local global `uv 0.9.30` and reports `GHSA-pjjw-68hj-v9mw`; GitHub Actions installs `uv 0.11.6`, so this local tool finding is not expected to reproduce in CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dependencies = [
     "python-multipart>=0.0.27",  # GHSA-wp53-j4wj-2cfg + GHSA-mj87-hwqh-73pj + GHSA-pp6c-gr5w-3c5g
     "pyasn1>=0.6.3",  # GHSA-jr27-m4p2-rc6r
     "mako>=1.3.12",  # GHSA-2h4p-vjrc-8xpq (transitive via alembic)
+    "idna>=3.15",  # GHSA-65pc-fj4g-8rjx
 ]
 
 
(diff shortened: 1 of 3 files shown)


ChrisHuie:
@sangilish thanks for this :) Have this on another pr in-flight so good to just pull out here

sangilish:
Thanks, makes sense. I’ll close this duplicate PR and leave the security-audit fix to the in-flight PR. I’ll keep the delivery account-resolution PR separate.

Issue #1319, opened 2026-05-18
tracking: BDD strict-marker test debt — 16 items pending resolution


KonstantinMirin:
## Purpose

Tracking issue for the **16 non-security items** in `docs/test-debt-bdd-strict-markers.md`. The doc is the canonical source of truth — please update the doc when items resolve, then check the corresponding box here.

Security-tier items (C1, C2, C3) are filed separately as #1316, #1317, #1318.

## Lifecycle (per item)

1. Engineer picks an item from the doc
2. Fix the gap (production change, test rewrite, or both)
3. Flip the `tests/bdd/conftest.py` marker to `strict=True` (or remove the marker block entirely if no rows remain xfail)
4. Remove the FIXME comment that references the doc item
5. Remove the entry from `docs/test-debt-bdd-strict-markers.md`
6. Check the box below

Step 3 forces step 5 because a stale `strict=True` causes the suite to fail on next xpass drift.

---

### [ ] C4 — Pydantic `ValidationError` → `AdCPError(INVALID_REQUEST, suggestion)` translation **(P2, broad payoff)**

- **Scope:** ~32 partition rows across UC-004 (`reporting_dimensions`, `attribution_window`, possibly more), currently `strict=True` xfail
- **Where:** transport boundary in `_get_media_buy_delivery_impl` and other `_impl` functions
- **Fix:** add a transport-boundary translator that wraps Pydantic `ValidationError` in `AdCPError(INVALID_REQUEST, suggestion=…)`
- **Unblocks:** ~32 currently-xfailed rows across UC-004 plus probably more across UC-005

### [ ] C5 — `include_package_daily_breakdown` is a no-op **(P2 feature gap)**

- **Scope:** `T-UC-004-partition-daily-breakdown` and `-boundary-daily-breakdown` valid rows (currently `strict=False`)
- **Where:** `src/core/tools/media_buy_delivery.py:480` hardcodes `daily_breakdown=None` regardless of `req.include_package_daily_breakdown`
- **Fix:** populate `daily_breakdown` per-package when the flag is True; strengthen the Then-step to verify shape differential
- **Note:** All 6 valid rows currently pass vacuously — illusory coverage

### [ ] C6 — Date-range validation returned in success envelope, not raised **(P3, cosmetic — already strict)**

- **Scope:** `T-UC-004-daterange-invalid` and `T-UC-004-daterange-equal` (currently `strict=True` with stale reason)
[...]

PR #1331, opened 2026-05-19
fix(mcp): accept creative format relationship filters

sangilish:
## Summary

Part of #1319, item C8.

This updates the MCP `list_creative_formats` wrapper to accept `output_format_ids` and `input_format_ids`, then forwards both fields into `ListCreativeFormatsRequest` so MCP uses the existing shared `_list_creative_formats_impl` filtering logic.

It also removes the C8 BDD debt entry and the MCP xfail markers for output/input format-id filters. REST filter xfails are intentionally left untouched because REST filter forwarding is a separate gap.

## Validation

- `uv run pytest tests/unit/test_list_creative_formats_new_filters.py tests/unit/test_mcp_tool_schemas.py -q --tb=short`
- `uv run ruff check src/core/tools/creative_formats.py tests/unit/test_list_creative_formats_new_filters.py tests/unit/test_mcp_tool_schemas.py tests/bdd/conftest.py`
- `uv run pytest tests/unit/test_architecture_wrapper_typed_params.py tests/integration/test_list_creative_formats_params.py -q --tb=short`
- `uv run pytest tests/integration/test_mcp_tools_audit.py -q --tb=short`
- `git diff --check`

## Local BDD Note

I also tried:

`uv run pytest tests/bdd/test_uc005_discover_creative_formats.py -q -k "(output_format_ids or input_format_ids or brrule049_inv9 or brrule049_inv10) and mcp" -rs --tb=short`

That selected 154 UC-005 C8-adjacent BDD cases, but all were skipped locally because `DATABASE_URL` is not configured. The draft PR CI should be the DB-backed confirmation path.

### DIFF
--- a/docs/test-debt-bdd-strict-markers.md
+++ b/docs/test-debt-bdd-strict-markers.md
@@ -379,7 +379,7 @@ that scenario causes a hard suite failure rather than silent xpass.
 - **Test-helper improvements:** 2 (H1, H2)
 - **Total:** ~18 items
 
-Severity distribution: 3× P1 (security/correctness), 7× P2, 7× P3.
+Severity distribution: 3× P1 (security/correctness), 6× P2, 9× P3.
 
 If filed as one umbrella GH issue with a checklist, this fits comfortably
 in a single tracking issue. If filed individually, only the 3 P1s warrant
--- a/docs/test-debt-bdd-strict-markers.md
+++ b/docs/test-debt-bdd-strict-markers.md
@@ -28,7 +28,7 @@ only record). Reconciled items were fixed by the Phase-2 wave (run
 | C5 (`include_package_daily_breakdown` no-op) | OPEN P2 | `salesagent-kzk0` |
 | C6 (date-range validation in success envelope) | OPEN P3 | `salesagent-t6y9` |
 | C7 (end-only date_range default) | OPEN P3 | `salesagent-losz` |
-| C8 (MCP list_creative_formats missing fmt-id params) | OPEN P2 | `salesagent-95q3` |
+| C8 (MCP list_creative_formats missing fmt-id params) | RECONCILED (PR #1331) | — |
 | C10 (description-only spec constraints) | OPEN P3 | `salesagent-o9w4` |
 | C11 (reporting_period echo) | RECONCILED (salesagent-18h.1) | — |
 | B1 (Gherkin `pending_activation`) | OPEN P2 | `salesagent-8c78` |
@@ -154,21 +154,6 @@ only record). Reconciled items were fixed by the Phase-2 wave (run
   media buy creation date (look up `MediaBuy.created_at`).
 - **Severity:** P3
 - **Origin:** Batch 2 audit
-
-### C8 — MCP `list_creative_formats` wrapper missing `output_format_ids`, `input_format_ids`
-- **Scope:** `T-UC-005-inv-049-9-violated`, `-9-nofield`, `-10-violated`,
-  `-10-nofield` on the `[mcp]` transport (currently strict=False, labeled
-  "vacuous pass")
-- **Where:** `src/core/tools/creative_formats.py:440-494` — MCP wrapper
-  signature does not accept these filter params; `mcp_compat_middleware`
-  strips them
-- **Impact:** Tests pass on MCP for the wrong reason (the param disappears
-  before reaching the filter). 4 vacuous xpasses.
-- **Unblocks:** add both params to the MCP wrapper signature so MCP becomes
-  a real test surface. Then flip strict=True (the underlying filter is
-  already implemented in `_impl`).
-- **Severity:** P2
-- **Origin:** Batch 5 audit
 
 ### C10 — Description-only spec constraints in adcp library
 - **Scope:** Two specific Examples rows already xfailed strict=True:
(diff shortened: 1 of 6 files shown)


sangilish:
I have read the IPR Policy

ChrisHuie:
https://github.com/prebid/salesagent/issues/1352 - will rerun infra test that is failing.  Saw this test being flaky earlier

KonstantinMirin:
Thanks for this — and the fix itself is correct: the MCP wrapper genuinely wasn't forwarding `output_format_ids` / `input_format_ids`, so this closes a real gap. Removing the matching `xfail` markers in `tests/bdd/conftest.py` was exactly the right move, and those scenarios now run through the real MCP path (I confirmed they fail if the wrapper drops the params again, so the coverage has teeth).

I want to flag one thing about the *added* tests, because it's the same pattern from your last PR and it's worth getting straight once so it saves you time going forward.

The two new unit tests (`test_list_creative_formats_new_filters.py`, `test_mcp_tool_schemas.py`) mock out `_list_creative_formats_impl` and inspect the function signature. They'll pass even if the filter logic never runs — they prove the param is *wired*, not that it *works*. In this repo, the thing that proves a transport-level filter actually works end-to-end is the BDD harness, and that coverage already exists:

- `tests/bdd/features/BR-UC-005-discover-creative-formats.feature` — the `inv-049-9/10` scenarios (`only "universal-builder" should be returned`, `"audio-builder" should not be returned`) assert the real filter results against a seeded registry, across all four transports including MCP.
- `tests/integration/test_creative_formats_behavioral.py` already tests this filter at the impl level — that's the in-repo pattern to follow if you want non-BDD coverage, rather than mocking `_impl` out entirely.
- The `xfail` ledger in `docs/test-debt-bdd-strict-markers.md` (item C8) literally described this gap and pointed at the fix.

That ledger + those scenarios *are* the spec for this change. The workflow that would've saved you the round-trip:

1. Find the xfailed/failing scenario that describes the behavior (here: C8 → the `*-fmtids` and `inv-049-9/10` scenarios).
2. Treat it as the contract — make it pass by fixing production code and removing the `xfail`. You did this part, and it's the right instinct.
3. Don't add mock-based unit tests on top to "prove" it — they assert wiring, the harness asserts behavior, and the harness is what we trust for "done." If you want extra coverage, write it *through* the harness (`tests/harness/creative_formats.py` → `CreativeFormatsEnv.call_mcp`, with `FormatIdFactory` from `tests/factories/format.py`) so it executes the path instead of mocking it away.

More broadly — and this is the important part — please read through `CLAUDE.md` and the `docs/development/` folder before the next PR. It documents almost everything you'll touch: `architecture.md` and `patterns-reference.md` (the schema-inheritance and transport-boundary `_impl`-vs-wrapper conventions), `structural-guards.md` (the automated checks that fail the build on violations), `admin-bdd-pattern.md` (how the harness is meant to be used), `a2a-mcp-agent-flows.md` (which is directly relevant here — it explains how the MCP and A2A wrappers relate), and `contributing.md`. These already encode the answers, including the rule that a fix isn't "tested" without coverage that executes the real path. If you hit something they *don't* cover — a real gap in the harness, a missing fixture, an undocumented pattern — the move is to **address that gap** (extend the harness, add the factory, update the doc) so the next person inherits it, not to reinvent a one-off or copy-paste around it. Reinventing is what produces tests that look green but don't run the code, which is most of the back-and-forth here.

Concretely for this PR: drop the two new mock unit tests (the BDD scenarios already cover the behavior for real), keep the wrapper fix and the `xfail` removals, and fix the arithmetic slip in the `docs/test-debt-bdd-strict-markers.md` severity histogram (C8 was a P2, but the edit reduced the P3 count instead — should read 6× P2, 9× P3).

And please fix the same bug fully while you're in here, rather than leaving it half-done. The root cause isn't "two params were missing" — it's that the MCP and A2A wrappers each rebuild `ListCreativeFormatsRequest` by hand, so any field can silently fall out of one transport. This PR patches two of those fields; `wcag_level` is still in exactly the same broken state (`_impl` filters on it, neither wrapper forwards it), and its BDD scenarios are still xfailed for that reason. If we only fix the symptom we noticed, the next field drifts the same way — that's how this got here. So:

1. Extract a single shared request-builder both wrappers call, so a field can't be present in one transport and missing in the other (`a2a-mcp-agent-flows.md` covers how these two relate).
2. Route `wcag_level` through it too and remove its now-obsolete `xfail` markers, confirming the scenarios pass on MCP for real (same as you did for the fmt-id filters).

That turns this from "the one filter I noticed" into "this whole class of drift is closed," which is the bar. Happy to point you at the exact lines if it helps.

Appreciate you picking this up — the fix is right. This is purely about which layer earns the "tested" checkmark, and about leaning on the patterns the repo already has so you're not doing more work than the change needs.

sangilish:
Thanks, this is addressed in `3f7fd41a`.

Changes made:
- Dropped the added mock/signature assertions for the relationship filter params.
- Added a shared `build_list_creative_formats_request(...)` helper and routed both MCP and A2A through it.
- Added `wcag_level` to the same shared wrapper path and removed the now-obsolete MCP `wcag_level` xfail rows.
- Fixed the BDD debt severity histogram to `6× P2, 9× P3`.

Validation run after the update:
- `uv run ruff check src/core/tools/creative_formats.py src/a2a_server/adcp_a2a_server.py tests/bdd/conftest.py tests/unit/test_list_creative_formats_new_filters.py tests/unit/test_mcp_tool_schemas.py`
- `uv run pytest tests/unit/test_mcp_tool_schemas.py tests/unit/test_list_creative_formats_new_filters.py -q` — 22 passed
- `scripts/run-test.sh --db tests/bdd/test_uc005_discover_creative_formats.py -k "fmtids or wcag or inv_049_9 or inv_049_10" -rA --tb=short` — 164 passed, 519 deselected
- `scripts/run-test.sh --db tests/bdd/test_uc005_discover_creative_formats.py -q --tb=line` — 490 passed, 193 xfailed
- `scripts/run-test.sh --db tests/integration/test_creative_formats_mcp_filters.py tests/integration/test_creative_formats_behavioral.py -q --tb=short` — 38 passed
- `uv run mypy src/core/tools/creative_formats.py src/a2a_server/adcp_a2a_server.py`
- `git diff --check`

KonstantinMirin (approved):
LGTM. All four asks from the last round addressed and verified at source — `_schedule_tmp_sync`-style shared builder (`build_list_creative_formats_request`) called from both MCP and A2A wrappers closing the drift class, `wcag_level` routed through it, the two mock unit tests dropped, and the histogram reconciled.

Re-ran the BDD harness on the de-xfailed scenarios to confirm they pass on MCP for real rather than trusting the marker removal: the wcag partition/boundary, output/input `format_ids` partition/boundary, and `inv-049` holds scenarios are all green on `[mcp]`. Guards + ruff clean, no allowlist growth. Thanks for the clean turnaround.

Issue #1333, opened 2026-05-20
fix(creative-registry): per-format resilient ingestion — one non-conforming format must not invalidate the whole list_creative_formats response


KonstantinMirin:
`_validate_formats_tolerant` in `src/core/creative_agent_registry.py` currently nukes the entire creative-format batch when any single format fails validation for a non-additive-asset_type reason. A predecessor refactor (commit `327676f8` on the branch behind the open PR adding the regression test below) introduced the additive-asset_type tolerance path; that was an incomplete generalization of the *"one non-conforming format must not invalidate the whole `list_creative_formats` response"* invariant.

## Symptom

When the live public creative agent (`https://creative.adcontextprotocol.org`) drops or renames formats that buyers expect (e.g. `display_image`, `display_html`, `display_js`, `video_standard`), or includes a format whose shape diverges in a way the pinned adcp library can't validate, our ingestion path discards every conforming format alongside the offending one. Buyers see zero formats from the agent instead of the conforming remainder.

## Where (two nuke-all code paths)

1. **`_validate_formats_tolerant`** (`src/core/creative_agent_registry.py`, around L117–143).
   The per-format loop tolerates only the additive-`asset_type` `ValidationError` (skip + continue around L135–137). Any other `ValidationError` hits L138 `raise` → the whole batch is lost.

2. **SDK `FAILED`-status branch** (around L447–465).
   When the adcp client returns `TaskStatus.FAILED` because wholesale schema validation rejected one bad format in a multi-format response, the code currently `raise AdCPAdapterError(...)`. It should fall through to the raw-MCP path and validate per-format there instead.

## Acceptance

- Replace L138 `raise` with `skipped_count += 1` + structured warning (identical pattern to the additive-asset_type branch). Always return the conforming remainder.
- In the SDK-FAILED branch, detect wholesale-validation FAILED responses and fall through to `_fetch_formats_raw_mcp` for per-format ingestion instead of raising.
- The regression test landing in the same PR as this issue (`tests/unit/test_creative_agent_registry_tolerance.py::test_non_asset_type_malformed_format_must_not_nuke_batch`) is currently `xfail(strict=True)` pinning this gap; when the fix lands, it xpasses → strict-fail → marker is removed.
- Add additional unit tests for the SDK-FAILED-status fall-through path.

## Priority

P2. Real robustness issue against an external service that demonstrably drifts. The gating PR adds a regression test (xfail-strict) so the gap is now visible.

[...]

Issue #1308, opened 2026-05-16
feat: accept adcp_major_version on get_products + sync_accounts + other MCP tools (#1247 gap #1)


ChrisHuie:
#1247 master tracker gap #1. Unblocks all 3 storyboard scenarios currently
  gated as advisory in `.github/workflows/test.yml` (`storyboard-compliance` job).

  ## Problem
  
  The @adcp/sdk storyboard runner sends `adcp_major_version: 3` as a request
  parameter on every MCP tool call (per AdCP 3.x spec). Salesagent's Pydantic
  schemas reject it as an unexpected keyword argument:

  mcp_error: 1 validation error for call[get_products]
  adcp_major_version
    Unexpected keyword argument [type=unexpected_keyword_argument, input_value=3, input_type=int]

  Evidence: CI run 25967283035 on PR #1274 — all 3 matrix scenarios
  (`refine_products`, `inventory_list_targeting`, `inventory_list_no_match`)
  fail in their `discover` / `sync_accounts` phase before ever reaching the
  scenario-specific assertions.

  ## Scope

  - `get_products` MCP wrapper: accept `adcp_major_version: int | None = None`
  - `sync_accounts` MCP wrapper: same
  - Audit every MCP tool wrapper for parameter acceptance — the runner sends
    this on every call
  - Decide: pass through to `_impl` (ignore) vs validate vs persist for audit
  - Update existing tests to cover the parameter

  ## Definition of Done

  - [ ] All 3 storyboard scenarios reach their scenario-specific phases (not
        blocked in `discover`/`sync_accounts` on this validation error)
  - [ ] When this lands + auth token issue is resolved + PR #1276 is merged,
        remove `continue-on-error: true` from `.github/workflows/test.yml`
        `storyboard-compliance` job (~line 387) and re-add
        `storyboard-compliance.result` to `test-summary`'s failure check

  ## Related

  - #1247 master tracker (gap #1)
  - PR #1274 (added the storyboard CI scaffolding; matrix gates these scenarios)
  - The auth token mismatch (separate gap; same blocker for storyboard)

KonstantinMirin:
## Additional missing get-products fields surfaced by the same alignment test

The Pydantic schema-alignment suite catches three missing fields on `GetProductsRequest` against `/schemas/latest/media-buy/get-products-request.json`, not just one:

- `adcp_major_version` — the envelope value `'1.0.0'` rejected as `extra_forbidden` (the case this issue already calls out).
[...]

PR #1334, opened 2026-05-20
fix: gh-#1264 memory-leak follow-ups and test-suite integrity

KonstantinMirin:
## Summary

Originally scoped as memory-leak follow-ups to #1264 (fire-and-forget asyncio task pinning, dead-thread reaper coverage, lifespan handle cleanup), the branch absorbed the BDD strict-marker reconciliation tracked under #1319 and the test-suite integrity work that surfaced while running the gates. The authoritative `./run_all_tests.sh` still has three known failures — the `get-products` schema drift already tracked in #1308, independent of any surface this PR touches. Every other suite passes; every non-strict xfail marker remaining in the BDD suite is the subject of an open tracking issue.

### Memory-leak follow-ups (gh-#1264)
- Pin fire-and-forget asyncio tasks in `activity_feed`.
- Extract `pin_task()` helper to `src/core/async_utils.py`.
- Extract `ThreadRegistry`, consolidating five reaper sites behind a single primitive.
- Lifecycle shutdown-callback registry.
- Replace the private-global webhook import with `get_webhook_service_or_none()`.
- Regression suites: dual-dict reaping (delivery-simulator, GAM-reporting, context-manager webhook pinning), FastAPI lifespan webhook close, and a single DRY parametrized `ThreadRegistry` reaper test file across all reaper sites (mutation-verified).

### Delivery / A2A / schema fixes
- `list_media_buys` honors persisted `MediaBuy.status` (`status_filter` no longer derived from dates).
- Keep `push_notification_config` out of `create_media_buy` request validation.
- `GetMediaBuyDelivery` now populates and echoes `attribution_window`; derives the geo classification system; applies placement `sort_by` ordering with spend fallback.
- `GetMediaBuyDeliveryRequest` extended with `time_granularity` + `include_window_breakdown`.
- A2A webhook camelCase wire-format alignment + matching test fixes.
- Tolerant per-format ingestion in the creative registry for additive `asset_type` growth.
- Bound Prometheus label cardinality.

### BDD strict-marker reconciliation (#1319)
- The three generic partition/boundary/status `Then` steps are strengthened with `AdCPError` / `ValidationError` type guards and degenerate-shell rejection — vacuous "any response is good enough" passes are gone.
- The eight UC-004 partition/boundary fields (status_filter, daily_breakdown, account, sampling, resolution, ownership, credentials, date_range) are wired to real request kwargs / setup instead of dispatching the axis name as a literal `extra=forbid` kwarg. The blanket `strict=False` sets covering these are now empty.
- UC-005 `disclosure_positions` filter + `inv-049-9/10` format-id markers reconciled: 39 vacuous xpasses eliminated, six vacuous `strict=False` entries removed.
- UC-006 `account_resolution` rewired to genuinely dispatch absent / both-keys payloads to production (no more synthesized validation errors in the step layer).
- The reporting-dims/attribution boundary rows that lost coverage when the blanket sets were emptied are restored with transport-scoped `strict=True` markers (a2a/mcp/rest only — impl genuinely passes).
- Net picture: every remaining BDD non-strict marker is either tracked under #1319 (the test-debt umbrella) or under one of the security issues below.

### Test-suite integrity (local gate fidelity)
- **Creative-agent test isolation.** `scripts/creative-agent-stack.sh` is the single source of truth for the pinned reference creative agent (commit `ca70dd1e2a6c`). `run_all_tests.sh`, `tox.ini`, and `.github/workflows/test.yml` all consume it; the inline CI snippet is gone, so CI and local cannot drift. A guard fixture in `tests/integration/test_creative_agent_live.py` hard-fails when `CREATIVE_AGENT_URL` is unset or points at the live public agent — no silent live fallback (an explicit `ALLOW_LIVE_CREATIVE_AGENT=1` opt-in remains for manual prod testing).
- **Test-stack DB connect hardening.** A shared `_connect_with_retry` (bounded exponential backoff) in `tests/conftest_db.py` replaces both bare `psycopg2.connect` sites and eliminates the parallel-`tox` SSL-negotiation race that was flipping bdd red non-deterministically.

### Test-first regression gates for the deferred production gaps
Three new failing tests now gate items tracked under #1319 and #1333:

- `tests/unit/test_creative_agent_registry_tolerance.py` — proves one non-conforming format currently nukes the whole `list_creative_formats` batch (#1333).
- `tests/integration/test_kzk0_daily_breakdown_shape.py` — proves `include_package_daily_breakdown` is a no-op (the response field is hardcoded `None`) (#1319 item).
- `T-UC-004-daterange-end-only` upgraded from `strict=False` to `strict=True` (retires the last vacuous-tolerance state on daterange).

Each is `xfail(strict=True)`: today the suite stays green; the moment the underlying gap closes the marker will force-fail, prompting its removal.

### Build / CI
- Run the BDD suite in CI and gate it in the test-summary job.
- Architecture guard: every suite in `./run_all_tests.sh` must have a gating CI job.
- Deterministic non-colliding port allocation for parallel e2e stacks.
- E2E assertions strengthened (tenant isolation; vacuous-pass guards in creative_assignment removed).

## Verification

Authoritative `./run_all_tests.sh`:

| Suite | Result |
|---|---|
| `unit` | 4668 passed / 2 failed → upstream schema drift (see Follow-ups) |
| `admin` | 79 passed / 0 failed |
| `e2e` | 91 passed / 1 failed → same upstream drift |
| `ui` | 5 passed / 0 failed |
| `bdd` | **1490 passed / 0 failed / 0 error**; 4 `xpassed` rows tracked under #1316/#1317 |
| `integration` | **1887 passed / 0 failed / 0 error** — the nine `test_creative_agent_live.py` tests pass against the pinned reference agent |

## Follow-ups (out of scope of this PR)

Tracked in `prebid/salesagent`:

- **Account-resolution security at the `_impl` boundary:** #1316 (A2A) and #1317 (impl).
- **Cross-principal `media_buy` access returns 200+empty instead of 403:** #1318.
- **BDD test-debt umbrella (16 non-security items):** #1319 — covers the `ValidationError → AdCPError(INVALID_REQUEST)` transport translator, `include_package_daily_breakdown` no-op, MCP `list_creative_formats` missing `output_format_ids` / `input_format_ids`, `disclosure_positions` filter implementation, and `reporting_dimensions` breakdowns (`by_geo` / `by_device_type` / `by_audience`).
- **`get-products` schema drift** (the three reds in the authoritative run): all three missing fields — `adcp_major_version` envelope, `if_catalog_version`, and `if_pricing_version` — tracked under #1308 (audit-every-MCP-tool-wrapper scope).
- **Creative-agent registry per-format resilience** (gated by the new unit test landing here): #1333.

## Notable for reviewers

- The BDD reconciliation half touches zero `src/` code — strict-marker reconciliation, test wiring, and step-definition fixes are test-only. Production fixes (delivery / A2A / schemas / metrics / memory) are bundled into this single commit but visible in the diff.
- No vacuous `strict=False` xfail markers remain outside the surface tracked by #1316/#1317/#1318.
- Branch base is `c70cbb02` (the prior #1299 head); `origin/main` has not advanced, so no merge conflicts.


ChrisHuie:
Three new inline imports were added in this PR, none with circular-dep justification. Consistent with the hoist policy applied across recent PRs. Can you move these guys up please


- src/services/protocol_webhook_service.py:545 — from src.core.lifecycle import register_shutdown inside get_protocol_webhook_service(). lifecycle doesn't import this module, no circularity.

- src/core/metrics.py:53 — from src.core.exceptions import (AdCPRateLimitError, AdCPServiceUnavailableError, AdCPValidationError) inside categorize_error(). exceptions doesn't import metrics.

- src/core/tools/media_buy_delivery.py — 8 sites of from typing import cast as type_cast inside function bodies (lines 233, 245, 339, 459, 522, 537, 574, 797). cast is already imported at module top (line 14: from typing import Annotated, Any, cast); the as type_cast rename serves no purpose. Line 456 uses bare cast — inconsistent with the rename. Suggest dropping the rename and using the top-level cast everywhere.

ChrisHuie (approved):

PR #1338, opened 2026-05-21
ci: supply-chain hardening (PR 1 of issue #1234)

mkostromin-sigma:
Closes #1234.

## Summary

Part of the [#1234](https://github.com/prebid/salesagent/issues/1234) rollout — 6-PR supply-chain hardening.

**This PR is pure defense, additive only. Zero behavior change to production code or tests.**

Closes drift items: PD3, PD4, PD5, PD6, PD7, PD13, PD14, PD15, PD23, PD24.

### What's included

| Commit | Change | Drift item |
|--------|--------|------------|
| `e9b6534f` | `SECURITY.md` + `[project.urls]` + description in `pyproject.toml` | PD5, PD23 |
| `38ae0ff5` | `CONTRIBUTING.md` rewritten as thin pointer to `docs/development/contributing.md` | PD7 |
| `f94e380e` | `.github/CODEOWNERS` — routes review to `@chrishuie` for CI, security, test infra, ratchet baselines | PD4 |
| `ad32a354` | `.github/dependabot.yml` — weekly updates for pip/actions/pre-commit, monthly for docker, no auto-merge | PD6 |
| `1bbca90b` | `.github/workflows/security.yml` — zizmor + pip-audit + gitleaks + pinact + actionlint; gitleaks pre-commit hook added | PD13 |
| `e52d21f8` | `.github/workflows/codeql.yml` + config — CodeQL advisory mode (2-week ramp, then flip to gating per D10) | PD14, PD15 |
| `7922e02d` | `docs/decisions/adr-001` (uv.lock single source) + `adr-002` (solo-maintainer bypass) | — |
| `245f710b` | SHA-pin all GitHub Actions via `pinact`; `permissions: {}` on `test.yml`; `persist-credentials: false` on all checkouts | PD3 |
| `41fa5fff` | `.github/zizmor.yml` allowlist + `adr-003` (pull_request_target trust boundary) + `security` PR title type | PD13 |

### Pre-commit hook changes

- `pre-commit autoupdate --freeze` applied — all `rev:` values are now commit SHAs with `# frozen: <tag>` comments
- `gitleaks` hook added (staged-file scan at commit time)

### Decisions referenced

D1, D2, D5, D10, D12, D15, D16, D17, D21, D35

### Reviewer notes

- **CodeQL is advisory** (`continue-on-error: true`) — flip to gating after ~2 weeks by setting `continue-on-error: false` in `.github/workflows/codeql.yml`
- **zizmor** gates from day 1 on `medium+` severity findings
[...]

PR #1339, opened 2026-05-21
fix(deps): bump pydantic-ai 1.56 -> 1.99 for GHSA-cqp8-fcvh-x7r3

ChrisHuie:
uv-secure flagged pydantic-ai 1.56.0 as vulnerable (GHSA-cqp8-fcvh-x7r3); fix version is 1.99.0. Resolved to 1.100.0 by uv lock refresh; transitive deps (openai, pydantic-evals, pydantic-graph, pydantic-ai-slim, temporalio, s3transfer, nexus-rpc) also bumped as part of the resolution.

Unblocks Security Audit on every open PR (same pattern as #1298 and #1336).


mkostromin-sigma:
Hey @ChrisHuie please check this [PR](https://github.com/prebid/salesagent/pull/1338)
Its already added in pr1 and in pr2 dependecy should be upgraded

src/services/ai/factory.py
```diff
@@ -149,7 +149,11 @@ def _create_provider_model(self, provider: str, model_name: str, api_key: str |
 
             if api_key:
                 return GoogleModel(model_name, provider=GoogleProvider(api_key=api_key))
-            return GoogleModel(model_name, provider="google-gla")
+            # pydantic-ai 1.99 renamed the "google-gla" provider literal to "google"
+            # (Google AI Studio / Generative Language API). The internal "google-gla"
+            # name we use above is our own normalization layer (e.g. "gemini" → "google-gla");
+            # the rename only applies at the SDK boundary.
+            return GoogleModel(model_name, provider="google")
```
KonstantinMirin, line 156: The rename was applied here only, which leaves the codebase's duplication worse than before this PR rather than fixing it.

`gemini -> google-gla` is already normalized in **two** places — `factory.py:122` and `config.py:90`. This PR adds a **third** hop (`google-gla -> "google"`) here in `_create_provider_model` only. `build_model_string` (`config.py:93`) still emits the deprecated `"google-gla:<model>"`.

Verified against the resolved `pydantic-ai==1.100.0`:
- `"google"` is accepted in **both** the `provider=` kwarg (`Literal['google','google-cloud','gateway']`) **and** the `"provider:model"` string form, with no warning.
- `"google-gla"` is deprecated in both forms: *"The 'google-gla:' prefix is deprecated and will be removed in v2.0. Use 'google:' instead."*

So the fix should remove `google-gla` from our code entirely via a single `normalize_provider()` (mapping `gemini`/`google-gla` -> `google`) used by both the factory dispatch and `build_model_string` — not a second translation layered on top. As written, the next person updating the provider mapping updates one copy and misses the other, and the untouched `build_model_string` path becomes a hard break at pydantic-ai v2.0.

src/services/ai/factory.py
```diff
@@ -149,7 +149,11 @@ def _create_provider_model(self, provider: str, model_name: str, api_key: str |
 
             if api_key:
                 return GoogleModel(model_name, provider=GoogleProvider(api_key=api_key))
-            return GoogleModel(model_name, provider="google-gla")
+            # pydantic-ai 1.99 renamed the "google-gla" provider literal to "google"
```
KonstantinMirin, line 152: Wording: `google-gla` was **deprecated, not renamed/removed** — it still resolves in 1.100.0, just with a deprecation warning (removal is in v2.0). Since this comment is the stated justification for the change, it should be precise. It also cites 1.99 while the lockfile resolves 1.100.0.

scripts/security-audit.sh
```diff
@@ -37,6 +47,6 @@
 # Extra arguments are forwarded to uv-secure (e.g. ``--no-check-uv-tool``).
 set -euo pipefail
 
-IGNORED_VULNS="PYSEC-2026-89,PYSEC-2025-183"
+IGNORED_VULNS="PYSEC-2026-89,PYSEC-2025-183,GHSA-cqp8-fcvh-x7r3"
```
KonstantinMirin, line 50: This adds a CVE to the ignore list that this same PR fixes.

`pyproject.toml` here bumps pydantic-ai to `>=1.99.0` and `uv.lock` resolves `pydantic-ai 1.100.0` / `fastmcp 3.3.1` — which is exactly the work the comment above (lines 28-37) describes as future work "that warrants its own dedicated migration PR." That comment now contradicts the diff.

This file's own convention (lines 38-41: protobuf, Pygments) is that once a bump resolves an advisory, it's **removed** from `IGNORED_VULNS` and demoted to the "Previously ignored, now resolved" note. Adding it instead permanently silences a now-fixed CVE in the scanner, masking any future regression if the pin is loosened below the fixed version.

Please remove `GHSA-cqp8-fcvh-x7r3` from `IGNORED_VULNS` and move its note to the resolved section. (The "inapplicable / no `force_download`" rationale is accurate and worth keeping in that note — `grep -rn 'force_download\|allow-local' src/ tests/` is 0 hits — but that justifies not panicking, not permanently ignoring a fixed CVE.)

KonstantinMirin (commented):
The dependency bump and the `mcp_server_enhanced.py` deletion look correct (verified the deleted module is dead code with no references, and that it imported the now-removed `fastmcp.server.context`). Two changes need rework before merge — inline comments below.

The theme across both: the change reads as correct line-by-line but wasn't traced through the rest of the file. The provider rename was added next to existing duplication instead of replacing it, and the security-audit edit suppresses a CVE that this same PR fixes.

**One thing not inline (the test file isn't in this diff):** nothing anchors the change. The edited line (`provider="google"`, no-API-key branch at `factory.py:156`) has no test — every factory test sets `GEMINI_API_KEY` and takes the `if api_key:` branch, so line 156 never runs. And `tests/unit/test_ai_service_factory.py::TestBuildModelString::test_gemini_provider` still asserts the deprecated `"google-gla:..."` output, which now disagrees with the intent of this PR. One assertion on the provider literal would anchor the rename.

KonstantinMirin (changes requested):
**Requesting changes** — two blocking issues (see inline comments on `src/services/ai/factory.py:156` and `scripts/security-audit.sh:50`).

1. **The provider rename was copied into one spot, not consolidated.** `gemini -> google-gla` is already normalized in two places (`factory.py:122`, `config.py:90`), and this PR adds a third hop (`google-gla -> "google"`) in `_create_provider_model` only — `build_model_string` still emits the deprecated `"google-gla:"`. Verified against `pydantic-ai==1.100.0`: `"google"` works in both the kwarg and string forms with no warning; `"google-gla"` is deprecated in both (removed in v2.0). The fix should route everything through one `normalize_provider()` and drop `google-gla` from our code, leaving the file with less duplication than before — not more.

2. **`security-audit.sh` suppresses a CVE this PR fixes.** `GHSA-cqp8-fcvh-x7r3` was added to `IGNORED_VULNS` even though the bump (pydantic-ai 1.100.0 / fastmcp 3.3.1 in `uv.lock`) resolves it — and the comment still describes that bump as future work "warranting its own PR." Per the file's own convention, a resolved advisory is removed from the ignore list and demoted to the history note. As written it permanently masks the fixed CVE.

The dependency bump and the `mcp_server_enhanced.py` deletion are correct. The common thread in the two issues above is that each edit reads fine in isolation but wasn't traced through the rest of the file — and nothing in the test suite anchors the change (the edited no-API-key branch at `factory.py:156` is never exercised; `test_gemini_provider` still asserts the deprecated output).

Issue #1234, opened 2026-04-24
refactor: CI and pre-commit flow


ChrisHuie:
> **Status:** P0 · 6-PR rollout · ~19.5-23.5 engineer-days · solo-admin model · 12 audit rounds applied (planning corpus on `docs/ci-refactor-planning` branch)
> **Read order for agents:** TL;DR → Locked decisions → 6-PR table → Acceptance criteria → expand `<details>` for deep context. Cold-start: `.claude/notes/ci-refactor/RESUME-HERE.md` (orientation, ~3k tokens).
> **Research dates:** 2026-04-23 → 2026-04-26. Re-verify `.pre-commit-config.yaml` line numbers and `uv.lock` versions if you pick this up >2 weeks later. PR #1217 (adcp 3.10 → 3.12) and PR #1221 (Flask → FastAPI v2.0) may have moved.

## TL;DR

`.pre-commit-config.yaml` has drifted in three compounding ways:

1. **Stale `adcp==3.2.0`** in mypy's `additional_dependencies` while the project runs `adcp>=3.10.0` — 8 minor versions of API drift hitting the type-check surface.
2. **Six other entries** in the same block are stale-pinned or completely unpinned, including production-runtime libraries (`fastmcp`, `alembic`).
3. **`psf/black` pinned to `25.1.0`** while `uv.lock` resolves `26.3.1` — `pre-commit run black` and `uv run black` produce different output **today**. Active formatting bug, not theoretical.

**Root cause:** `additional_dependencies` creates an isolated pre-commit venv with no mechanism to stay synchronized with `uv.lock`. Compounding governance gaps: hook `rev:` accepts mutable tags, no CODEOWNERS, no SECURITY.md, no Dependabot, no hash verification.

**This is a systemic fix, not a version bump.** It establishes:

- `uv.lock` as single source of truth for Python deps in pre-commit
- Full-SHA pinning on every external hook + workflow action (CPython pattern)
- CODEOWNERS + SECURITY.md + Dependabot + zizmor + CodeQL + gitleaks governance overlay
- Layered hook architecture: pre-commit = formatters + hygiene only; **CI is authoritative**
- Single-source anchors for Python, Postgres, and uv versions
- Image supply-chain: cosign keyless signing + SBOM + harden-runner + Trivy OS-layer scan + reproducible builds (`SOURCE_DATE_EPOCH`)
- Container hardening: base-image `@sha256:` digest pin + `USER` non-root

[...]

PR #1344, opened 2026-05-22
fix(ci): pass GITHUB_TOKEN to pinact to avoid anonymous rate limit

mkostromin-sigma:
Closes #1234.

Follow-up to #1338.

`pinact run --check` resolves action tag SHAs via the GitHub API. Without authentication it hits the 60 req/h anonymous limit (403 rate limit exceeded). Pass `GITHUB_TOKEN` so requests are authenticated (5000 req/h).

The `actions/checkout` version bump is handled separately by Dependabot PR #1342.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -71,6 +71,8 @@ jobs:
           grep "  ${PINACT_TARBALL}$" pinact_checksums.txt | sha256sum -c -
           tar -xzf "${PINACT_TARBALL}" -C /usr/local/bin/ pinact
           pinact run --check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   actionlint:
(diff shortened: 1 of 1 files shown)


numarasSigmaSoftware (approved):

KonstantinMirin (approved):
Approved. Correct, minimal fix for the pinact job introduced in #1338.

- Safe: trigger is `pull_request` (not `pull_request_target`), so fork tokens are read-only; job permission is `contents: read`; `env:` is scoped to the single step; `pinact` only reads workflow files + queries the GitHub API (no PR code execution); checkout uses `persist-credentials: false`.
- Necessary: `pinact run --check` resolves action tag→SHA via the GitHub API; anonymous = 60 req/h (403 rate-limit), authenticated = 5000 req/h. Without this the new SHA-pin enforcement job flakes/fails on real PRs.

Textbook fix, well scoped.

PR #1351, opened 2026-05-24
fix(delivery): resolve account references in shared path

sangilish:
## Summary

Refs #1316 and #1317.

This is the smaller current-main replacement for the stale #1321 draft after #1334 changed the delivery/A2A/BDD surfaces. It focuses on AccountReference propagation and shared resolution for get_media_buy_delivery; it does not claim #1318 or media-buy ownership filtering.

## What changed

- A2A now passes the validated `req.account` value to the raw delivery call instead of the original raw params dict.
- MCP/raw and REST pass account into `GetMediaBuyDeliveryRequest` instead of resolving account at each wrapper boundary.
- REST validates `body.account` into a typed `AccountReference` before passing it to the raw path.
- `_get_media_buy_delivery_impl` resolves `req.account` once through `enrich_identity_with_account` before principal lookup.
- Added unit regression coverage for raw forwarding, shared impl resolution, A2A forwarding, and REST forwarding.

## Out of scope

This does not claim #1318. Cross-principal media-buy ownership remains separate.

## Testing

- `uv run pytest tests/unit/test_delivery_account_resolution_flow.py tests/unit/test_auth_consistency.py::TestMissingTokenConsistency::test_get_media_buy_delivery_missing_auth_returns_error_response -q --tb=short`
- `uv run pytest tests/unit/test_delivery.py tests/unit/test_rest_api_endpoints.py tests/unit/test_delivery_account_resolution_flow.py -q --tb=short`
- `uv run pytest tests/unit/test_pydantic_schema_alignment.py::TestPydanticSchemaAlignment::test_model_accepts_all_schema_fields[/schemas/latest/media-buy/get-media-buy-delivery-request.json-GetMediaBuyDeliveryRequest] -q --tb=short`
- `uv run ruff check src/core/tools/media_buy_delivery.py src/a2a_server/adcp_a2a_server.py src/routes/api_v1.py tests/unit/test_delivery_account_resolution_flow.py`
- `uv run mypy src/core/tools/media_buy_delivery.py src/a2a_server/adcp_a2a_server.py src/routes/api_v1.py`
- `git diff --check`

Keeping this draft until the full GitHub Actions matrix runs or maintainers signal that this smaller current-main split is wanted.

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1986,7 +1986,7 @@ class AdCPRequestHandler(RequestHandler):
                 reporting_dimensions=req.reporting_dimensions,
                 attribution_window=req.attribution_window,
                 include_package_daily_breakdown=req.include_package_daily_breakdown,
-                account=params.get("account"),
+                account=req.account,
                 context=params.get("context"),
                 identity=identity,
             )
--- a/src/core/tools/media_buy_delivery.py
+++ b/src/core/tools/media_buy_delivery.py
@@ -650,7 +650,7 @@ def get_media_buy_delivery_raw(
     reporting_dimensions: ReportingDimensions | None = None,
     attribution_window: AttributionWindow | None = None,
     include_package_daily_breakdown: bool | None = None,
-    account: LibraryAccountReference | dict[str, Any] | None = None,
+    account: LibraryAccountReference | None = None,
     context: ContextObject | None = None,
     ctx: Context | ToolContext | None = None,
     identity: ResolvedIdentity | None = None,
--- a/src/core/tools/media_buy_delivery.py
+++ b/src/core/tools/media_buy_delivery.py
@@ -93,6 +93,12 @@ def _get_media_buy_delivery_impl(
     if identity is None:
         raise AdCPValidationError("Context is required", recovery="correctable")
 
+    if isinstance(req, GetMediaBuyDeliveryRequest) and req.account is not None:
+        from src.core.transport_helpers import enrich_identity_with_account
+
+        identity = enrich_identity_with_account(identity, req.account)
+        assert identity is not None
+
     # Extract testing context for time simulation and event jumping
     testing_ctx = identity.testing_context or AdCPTestContext()
 
@@ -615,12 +621,6 @@ async def get_media_buy_delivery(
     """
     identity = (await ctx.get_state("identity")) if isinstance(ctx, Context) else None
 
-    # Handle account resolution at boundary (same as sync_creatives pattern)
-    if account is not None and identity is not None:
-        from src.core.transport_helpers import enrich_identity_with_account
-
-        identity = enrich_identity_with_account(identity, account)
-
     # Create AdCP-compliant request object
     try:
         req = GetMediaBuyDeliveryRequest(
@@ -631,6 +631,7 @@ async def get_media_buy_delivery(
             reporting_dimensions=reporting_dimensions,
             attribution_window=attribution_window,
             include_package_daily_breakdown=include_package_daily_breakdown,
+            account=account,
             context=cast(ContextObject | None, context),
         )
 
@@ -649,7 +650,7 @@ def get_media_buy_delivery_raw(
     reporting_dimensions: ReportingDimensions | None = None,
     attribution_window: AttributionWindow | None = None,
     include_package_daily_breakdown: bool | None = None,
-    account: LibraryAccountReference | None = None,
+    account: LibraryAccountReference | dict[str, Any] | None = None,
     context: ContextObject | None = None,
     ctx: Context | ToolContext | None = None,
     identity: ResolvedIdentity | None = None,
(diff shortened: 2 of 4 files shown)


KonstantinMirin (changes requested):
**Verdict: this PR is a net regression and meets none of its own acceptance criteria. Recommend closing and rewriting against the existing harness.**

I ran the designed scenarios that already exist for this exact contract — `tests/bdd/test_uc004_deliver_media_buy_metrics.py -k account`, generated from `BR-UC-004-deliver-media-buy-metrics.feature:777,796` — against a real PostgreSQL, changing only the 3 production files between `main` and this branch.

## Evidence

| Production code | account scenarios |
|---|---|
| `main` | **23 passed, 37 xfailed, 0 failed** (green) |
| this PR | **19 passed, 34 xfailed, 7 failed** (red) |

Reproduction:
```bash
# isolated DB, then:
uv run pytest tests/bdd/test_uc004_deliver_media_buy_metrics.py -k account -rA --tb=line
# swap src/core/tools/media_buy_delivery.py, src/a2a_server/adcp_a2a_server.py,
# src/routes/api_v1.py between origin/main and HEAD to compare
```

## 1. Existing passing tests were broken (the headline)

Four scenarios that are **green on `main`** now hard-fail on this branch:

- `partition[impl-explicit_account_id-{"account_id":"acc_acme_001"}-valid]`
- `partition[impl-natural_key-{"brand":{"domain":"acme-corp.com"},"operator":"acme-corp.com"}-valid]`
- `boundary[impl-account_id present + account exists-valid]`
- `boundary[impl-brand + operator present + single match-valid]`

All four fail with:
```
AssertionError: Expected valid account result but got error:
  AdCPAccountNotFoundError("Account 'acc_acme_001' not found.")
```

The PR added account resolution into `_impl` but **dropped the UC-004 account fixtures** that the superseded #1321 had seeded ("Seed UC-004 account rows with real account/access fixtures"). With no `acc_acme_001` in the DB, every valid-account path now raises not-found. The feature didn't get safer — the happy path got deleted.

## 2. The BDD gate goes red on unreconciled strict markers

Three rows flip to `XPASS` under `xfail(strict=True)` and therefore **fail the suite** (`tests/bdd/conftest.py:715-761`, `_UC002_VALIDATION_XFAIL`): `partition[impl-account_not_found]`, `boundary[impl-account_id present + not found]`, `boundary[a2a-account_id present + not found]`. The PR never touches `conftest.py`, so the markers that pin these as "still broken" were left in place. `git diff origin/main...HEAD --stat -- tests/bdd/conftest.py` is empty.

## 3. Zero deliverables met

The #1316/#1317 acceptance criterion is "`T-UC-004-partition/boundary-account` `[a2a]` rows flip xfail→pass." **Not one `a2a-*`/`mcp-*`/`rest-*` valid row flips** — they all remain xfailed (the missing fixtures mask whether the new wiring even resolves). The PR's stated goal is unachieved by every row it names.

## 4. The new unit tests are vanity

`tests/unit/test_delivery_account_resolution_flow.py` mocks `enrich_identity_with_account` — the exact function this PR relocates — and asserts a mock received a value. No DB, no resolution, no rejection path. That is why they stay green while the real behavior regresses 7 ways. There is also no DB-backed integration test; #1321 had one (`tests/integration/test_delivery_account_resolution.py`), and it was not carried over.

## 5. Wrong architectural layer (documented invariant)

Resolution was moved **into** `_get_media_buy_delivery_impl` (`src/core/tools/media_buy_delivery.py:96-100`). This contradicts written rules:

- **CLAUDE.md Pattern #5**: wrappers resolve identity; `_impl` receives a resolved `ResolvedIdentity` and does business logic only, with zero transport imports.
- The helper's own docstring (`src/core/transport_helpers.py:112`): "Called at the transport boundary after resolve_identity()."
- The sibling tools agree — `src/core/tools/media_buy_create.py:3956` literally comments "Resolve account at transport boundary (before _impl)"; `sync_creatives` does the same.

Delivery is now the lone tool resolving in `_impl`. The transport-agnostic guard stays green only because it doesn't follow transitive imports (`transport_helpers` imports `fastmcp`), which is also why the new import is function-local — it dodges the guard rather than satisfying it.

## Recommendation

**If this branch was produced by an AI agent: close it and start over.** The defects are structural — wrong layer, deleted fixtures, ignored harness, mocked-away tests — and span every file. Patching them one by one costs more than a clean redo and tends to leave the original wrong scaffolding in place.

**If this was hand-written: it's fine to keep going**, but in the correct paradigm:

1. Resolve `account` at the **transport boundary** in the 3 delivery wrappers (mirror `create_media_buy`/`sync_creatives`). Keep the typed `account` field + forwarding — that part is the right idea.
2. **Restore the UC-004 account fixtures** so valid accounts resolve.
3. As rows go green, **shrink the `a2a-*`/`mcp-*`/`rest-*` substrings out of `_UC002_VALIDATION_XFAIL`** in `tests/bdd/conftest.py` and remove the matching FIXME/doc entries.
4. **Delete the mocked unit file**; let the BDD harness + a DB-backed integration test be the evidence (the harness already exists and is wired to all four transports — use it, don't reinvent it).
5. Gate on `tox -e bdd -k account` → 0 failed, 0 unexpected xpass.

What's salvageable: the A2A diagnosis (`params.get("account")` raw dict → `req.account` typed) is real, and REST's `model_validate` is the right idiom. Everything else needs to be rebuilt at the boundary.

sangilish:
Thanks for the detailed review. I agree with the architectural point: delivery account resolution should stay at the transport boundary, consistent with create_media_buy and sync_creatives, rather than moving into _impl.

I am going to close this draft and rebuild from current main only if I can validate it against the existing UC-004 account BDD harness: boundary-level account resolution, real account/access fixtures, xfail marker cleanup only after rows pass, and DB-backed evidence rather than mocked resolver-only tests.

PR #1355, opened 2026-05-24
chore(adcp): pin SDK exactly + document spec version

ChrisHuie:
Pins adcp Python SDK to exact version 4.3.0 (was >=4.3.0) and adds infrastructure to make the targeted AdCP spec version explicit.

- pyproject.toml: adcp>=4.3.0 → adcp==4.3.0
- New tests/unit/test_adcp_spec_version.py: CI guard asserting adcp.get_adcp_spec_version() == "3.0.1"
- New docs/adcp-spec-version.md: SDK-to-spec mapping + bump procedure
- README.md: adds AdCP Compatibility section
- CLAUDE.md: adds reference to the doc

adcp 4.3.0 ships Pydantic models code-generated from AdCP spec 3.0.1 (per adcp.get_adcp_spec_version()). The previous open-upper-bound pin allowed silent dependency-resolution shifts to a 4.x release targeting a different spec version. The exact pin + CI guard prevents accidental drift; the doc makes the current target discoverable.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,6 +114,15 @@ AST-scanning tests enforce architecture invariants on every `make quality` run. 
 
 ---
 
+## AdCP Spec Version
+
+This project targets AdCP spec **3.0.1** via the `adcp==4.3.0` Python SDK. See
+[docs/adcp-spec-version.md](docs/adcp-spec-version.md) for the version mapping
+and bump procedure. The CI guard at `tests/unit/test_adcp_spec_version.py`
+fails on pin drift.
+
+---
+
 ## 🚨 Critical Architecture Patterns
 
 ### 1. AdCP Schema: Extend Library Schemas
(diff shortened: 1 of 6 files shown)


KonstantinMirin (approved):
**Scoped and does what the title says.** One commit, 6 files (+105/−3). Verified locally: installed SDK is `4.3.0`, `adcp.get_adcp_spec_version()` returns `3.0.1`, and the new guard test passes. The pin change, CI guard, and docs all match the stated intent — no scope creep.

**One thing to fix before merge:** `uv.lock` doesn't only update the adcp specifier — it also downgrades the lockfile format (`revision = 3` → `revision = 2`). That happens when the lock is regenerated with an older `uv` than the one that produced main's lockfile. It'll cause merge churn and anyone on a newer `uv` will bump it back to 3 on their next `uv lock`. Please regenerate with current `uv` so only the adcp line changes and `revision` stays at 3.

**Two non-blocking notes:**
- The SDK→spec mapping table documents future rows (`4.4.x → 3.0.5`, `5.x → 3.0.7`, `6.x → 3.1.0-beta`) that aren't verifiable from this repo and can go stale. Only the `4.3.x → 3.0.1` row is enforced by the guard. Fine to keep, just flagging it as maintenance surface.
- Branch is `-v2` — confirm any earlier superseded branch/PR gets closed.

PR #1364, opened 2026-05-26
feat: add /pre-review skill

pmezzich:
Adds a \`/pre-review\` skill that catches the patterns @ChrisHuie and @KonstantinMirin consistently flag in code review before they have to flag them.

## How it was built

Pulled 102 review comments across 14 merged PRs filtered to Chris and Konstantin only (skipping bokelley). Frequency-ranked the recurring patterns and built a checklist around the top 10.

## What it checks

- **Hoisting** — inline imports without circular-dep justification (Chris's named pattern)
- **Type refinement** — `list[Any]` / `product: Any` where a concrete type is known (Konstantin's named pattern)
- **Vacuous assertions** — OR short-circuits, always-true `hasattr`, zero-assertion happy paths
- **Create/update parity** — validation present in create path but missing from update (exploitable bypass)
- **Raw dicts to typed fields** — `{"code": ...}` where `Error(...)` is expected
- **Test DRY** — helper extracted but old callsites not updated
- **Dead/unreachable code** — exception branches the type system proves can't fire
- **Redundant DB queries** — double SELECT / N+1 on hot paths
- **Weakened BDD assertions** — OR replacing AND in Then steps
- **Project conventions** — subprocess returncode, no PR numbers in comments, etc.

Uses `curl` for all GitHub API calls per project convention.

## Validation

Ran against PR #1107 as a blind test — caught 4/4 of Chris's actual findings before looking at his comments.

PR #1365, opened 2026-05-27
feat(skill): add error-emission checks to /pre-review

pmezzich:
## Summary

- Adds checks **K** and **L** to the `/pre-review` skill, aligning with the error-emission architecture conventions from PR #1306
- **K — Wire envelope assertions**: New error-path tests should assert on `result.wire_error_envelope` via `assert_envelope_shape()`, not on reconstructed exceptions (`isinstance`/`.error_code`), since reconstruction is lossy (e.g. `AdCPAuthenticationError` and `AdCPAuthorizationError` both map to `AUTH_REQUIRED`)
- **L — Typed exception subclasses in `_impl`**: `_impl` functions should raise typed `AdCPError` subclasses instead of building wire-shape dicts, raising raw `ValueError`, or using inline `Error(code=...)` construction

## Context

Built on top of #1364 (the base `/pre-review` skill with checks A–J). These two checks are derived from reading the error-emission architecture PR #1306 — if that lands, the skill is already ahead of it.

## Test plan

- [ ] Run `/pre-review` against a PR that touches error paths and verify checks K/L fire correctly
- [ ] Run `/pre-review` against a PR with no error changes and verify K/L report N/A
- [ ] Confirm checks A–J still work unchanged

### DIFF
--- a/.claude/skills/pre-review/SKILL.md
+++ b/.claude/skills/pre-review/SKILL.md
@@ -1,0 +1,404 @@
+---
+name: pre-review
+description: >
+  Pre-review a salesagent PR before human review. Checks for the recurring
+  patterns that ChrisHuie and KonstantinMirin consistently flag across
+  merged PRs: hoisting, type refinement, vacuous assertions, create/update
+  parity, raw dicts to typed fields, dead code, test DRY violations,
+  redundant DB queries, error envelope compliance, and untyped exception
+  usage. Saves review rounds by catching these before submission.
+args: <pr-number>
+---
+
+# Pre-Review Skill
+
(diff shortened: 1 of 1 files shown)


pmezzich:
I have read the IPR Policy

KonstantinMirin:
I think that it makes more sense to ground the skill overall in the docs/development rather than "Chris and Konstantin point out", making it more procedural.
Also, I believe we should not have two PRs about one skill. This supercedes the previous one.

pmezzich:
Superseded by #1371 — reworked from skill/command into rule files per review feedback.

PR #1371, opened 2026-05-27
feat(rules): add review-derived code and test pattern rules

pmezzich:
## Summary

- Adds two rule files to `.claude/rules/patterns/` so agents internalize review patterns upfront rather than catching them in a post-hoc review pass
- **review-code-patterns.md**: hoisting, type refinement, create/update parity, raw dicts → typed constructors, dead code, redundant DB queries, typed exceptions in `_impl`, project conventions
- **review-test-patterns.md**: vacuous assertions, test DRY, BDD assertion strength, wire envelope assertions over reconstructed exceptions

## Context

Reworked from PRs #1364 and #1365 (which used a skill/command approach) into rule files per review feedback. These load into agent context at session start so the patterns are followed one-shot.

Derived from recurring patterns in code review history, plus error-emission architecture conventions from PR #1306.

Follows the existing format in `.claude/rules/patterns/` (`code-patterns.md`, `testing-patterns.md`, `mcp-patterns.md`).

Supersedes #1364 and #1365.

## Test plan

- [ ] Start a new Claude session in the salesagent repo and verify the rules load
- [ ] Write code that violates a pattern (e.g., inline import) and confirm the agent avoids or self-corrects it
- [ ] Confirm no overlap/contradiction with existing pattern files

### DIFF
--- a/.claude/rules/patterns/review-code-patterns.md
+++ b/.claude/rules/patterns/review-code-patterns.md
@@ -1,7 +1,7 @@
-# Review-Derived Code Patterns
+# Code Review Patterns
 
-Patterns consistently flagged by ChrisHuie and KonstantinMirin across 102
-review comments on 14 merged PRs. Follow these when writing or modifying code.
+Recurring patterns extracted from code review history. Follow these when
+writing or modifying code.
 
 ## Hoisting — No Inline Imports Without Circular-Dep Justification
 
--- a/.claude/rules/patterns/review-code-patterns.md
+++ b/.claude/rules/patterns/review-code-patterns.md
@@ -148,7 +148,9 @@ with an inline code string. The structural guard tests cap these sites.
 
 ## Project Conventions
 
-- No issue/PR numbers in code comments — reference in commit messages only
+- FIXME comments must carry a GitHub issue reference — `FIXME(#1234)`, not
+  bare `FIXME` and not internal tool IDs (e.g. beads). Reviewers ask for the
+  issue-tagged form so the revisit condition is trackable.
 - No `# noqa: F401` on imports that are actually used
 - No `--link` in Docker commands — use named networks
 - Check `subprocess.run()` result `returncode` when the result drives a gate
--- a/.claude/rules/patterns/review-code-patterns.md
+++ b/.claude/rules/patterns/review-code-patterns.md
@@ -1,0 +1,154 @@
+# Review-Derived Code Patterns
+
+Patterns consistently flagged by ChrisHuie and KonstantinMirin across 102
+review comments on 14 merged PRs. Follow these when writing or modifying code.
+
+## Hoisting — No Inline Imports Without Circular-Dep Justification
+
+Move imports to module level unless a circular dependency requires deferral.
+
+```python
+# WRONG — inline import without circular dep
+def process_order(order_id: str) -> None:
+    from src.core.schemas import Order  # no circular dep exists
+    ...
+
+# CORRECT — module-level import
+from src.core.schemas import Order
+
+def process_order(order_id: str) -> None:
+    ...
+
+# ACCEPTABLE — circular dep requires deferral
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from src.core.schemas import Order
+```
+
+## Type Refinement — No `Any` Where a Concrete Type Is Known
+
+If a function only ever receives or returns one specific type, annotate it.
+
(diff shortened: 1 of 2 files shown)


pmezzich:
I have read the IPR Policy

KonstantinMirin (changes requested):
The patterns are solid and the writing is concrete — `# WRONG / # CORRECT` pairs, an actual *why* line under each rule (e.g. "Asymmetric validation lets buyers bypass checks via the update path"), no metaphors or filler. Create/Update parity in particular is a real review pattern that isn't documented anywhere else in the repo and catches a real security class. Vacuous-assertions has four well-chosen subtypes.

Two structural issues and a coverage gap before this is ready to land. Fixing is small.

### Required before merge

**1. Cross-reference the authoritative sources where the rule already exists.**

Two sections restate rules that already live in canonical project docs. Restating creates drift risk — the two copies will eventually disagree. Keep your worked examples (they add value), but anchor each section to the source of truth.

- `review-code-patterns.md` § "_impl Functions — Raise Typed Exceptions, Don't Build Wire Shape" overlaps with **CLAUDE.md Critical Pattern #5 — Transport Boundary: Layer Separation**. Add a one-liner at the top: *"Reinforces CLAUDE.md Critical Pattern #5 with concrete examples."*
- `review-test-patterns.md` § "Error Tests — Assert on Wire Envelope, Not Reconstructed Exceptions" overlaps with **`tests/CLAUDE.md` § Error Verification Policy**. Add: *"See `tests/CLAUDE.md` § Error Verification Policy for the full policy, helpers, and migration path."*

**2. Add the missing review-derived patterns.** This file documents patterns extracted from review history. Four patterns that show up in nearly every review are missing:

a. **No allowlist growth — hard block.** Structural guard allowlists in this repo are ratchets — they only shrink. PRs that add a new FIXME-deferred entry to any allowlist (model_dump, weak-mock-assertions, raw-select, type-ignore baseline, duplication baseline, etc.) are blockers, not Mediums. Existing entries are debt, not templates for new code to follow. Pre-commit + the structural-guard tests both enforce this; reviewers should never approve "match the existing convention" rationale when the convention is allowlisted.

b. **Tests must fail when production is mutated.** A test that survives the production line being deleted or inverted is not coverage. Verify before approving a test — read the assertion, then mentally delete or invert the line under test, then ask whether the assertion would fire. Common failure modes:
- Mock that bypasses the predicate being tested (e.g. `scalars(...).all()` stubbed to a static list — tenant-scoping fix can be reverted with no test failure).
- `patch(...)` of the function under test that only ever raises (e.g. `_finalize_approval` patched to raise — the real function's branches are never exercised).
- Assertion on a reconstructed value the test built itself, rather than on the real call site's output.

c. **Done means harness/integration coverage, not manual-mock unit tests.** A fix isn't done unless an integration test (`@pytest.mark.requires_db`) or e2e/BDD test using `tests/harness/` + `tests/factories/` (a) drives the real entrypoint, (b) exercises the code path the fix touches, and (c) fails when the fix is reverted. Manual `patch()` / `MagicMock` stacks let wire-correctness, JSONB `flag_modified`, transaction ordering, and race-condition behaviors slip through.

d. **No `pytest.mark.skip` / `xfail` to bypass a failure.** The only acceptable `xfail` is a tracked stub for unimplemented work that the test-surfacing skill (`/surface`) manages. Never `xfail` to mute a real regression, "infrastructure issue," or "pre-existing failure." Project policy is in CLAUDE.md "Test Integrity Policy" — cross-reference it.

**3. Three smaller fixes:**

- `review-code-patterns.md` § "Raw Dicts → Typed Constructors" mixes two patterns: (1) typed constructors at construction time, (2) `.model_dump(mode="json")` on both sides of `==` comparisons. Split into two sections — they apply at different points and the comparison rule is easy to miss when buried inside a constructor-shape rule.
- `review-code-patterns.md` § "Dead / Unreachable Code" #1 (the `ToolError around TypeAdapter` case) is one instance of a broader rule. State the rule first ("Only catch exception types the `try` block can actually raise — read what the called code emits") and use the TypeAdapter case as the example.
- Add a short section on the xfail-as-bug pattern: verify a "production gap" xfail against the AdCP `error_code` enum and the source-of-truth hierarchy (generated `BR-*.feature` traces to upstream AdCP requirements repo; `docs/test-obligations/` is bootstrap only, not authoritative) **before** recommending implementation. Recurring failure mode where a wiring-PR author would otherwise be asked to implement non-spec behavior.

Issue #1233, opened 2026-04-23
ci: divergences between GitHub Actions and local test tooling


ChrisHuie:
> **Status:** Mostly **subsumed by #1234** (12 of 15 D-items absorbed by #1234 PRs 3 + 4). Three items remain in this issue's scope: D9 (informational), D11 (`requires_server` decision), D13 (GAM live tests → nightly-cron follow-up). Close this issue when #1234 merges + the 3 remaining items are resolved.
> **Read order:** Status table → Remaining scope → expand `<details>` for the original 15-divergence catalog and historical context.

## TL;DR

Two entire test suites — `tests/admin/` and `tests/bdd/` — were never run in CI. The audit found 15 CI-vs-local divergences (D1-D15). **#1234's CI/pre-commit refactor absorbed 12 of them** (PR 3 adds the missing jobs as part of the new 14-frozen-check-name structure; PR 4 deletes D15's dead hook). 3 items are decisions/follow-ups that need explicit resolution.

The original framing — "CI green doesn't mean all tests pass" — is fixed by #1234. This issue closes when #1234 lands AND the 3 remaining items are decided.

## Status of each divergence (post-#1234 planning)

| ID | Divergence | Status | Closed by |
|---|---|---|---|
| **D1** | Admin + BDD suites never run in CI | ✅ Absorbed | #1234 PR 3 commit 3 — `Admin UI Tests` + `BDD Tests` are 2 of the 14 frozen check names |
| **D2** | Integration marker matrix fragile; `architecture` not routed | ✅ Absorbed | #1234 PR 3 commit 3 — matrix collapsed to single xdist job (`--dist=loadscope` per D33) |
| **D3** | `tests/harness/` collected locally, ignored by CI | ✅ Absorbed | #1234 PR 3 — `tox -e unit` runs `tests/unit/ tests/harness/` (verified at tox.ini:40) |
| **D4** | Coverage collected from unit only; never combined | ✅ Absorbed | #1234 PR 3 commit 6 — `.coverage-baseline=53.5` hard-gate + combine job (D11 of #1234) |
| **D5** | `make quality` pre-commit hooks don't re-run in CI | ✅ Absorbed | #1234 PR 3 commit 3 — `Quality Gate` job runs `pre-commit run --all-files` |
| **D6** | Ruff linting advisory-only in CI (`\|\| true`) | ✅ Absorbed | #1234 PR 3 commit 7 — removes `\|\| true` and `continue-on-error: true` |
[...]

PR #1372, opened 2026-05-28
ci: authoritative pipeline with frozen required checks (PR 3 of #1234)

mkostromin-sigma:
Closes #1234.

## Summary

Part of **#1234 — PR 3 of 6**. Replaces legacy `.github/workflows/test.yml` with authoritative **`.github/workflows/ci.yml`**, establishes frozen branch-protection check names, and makes CI the enforcement layer for quality gates (building on PR #1370 / PR2).

## Stacking & review scope

**Depends on PR #1370 (PR2)** — merged. Branch rebased onto `upstream/main`.

| View | Files changed | When |
|------|---------------|------|
| **This PR vs `main` (rebased)** | ~27 files | PR3-only delta — review scope |

**For reviewers:** focus on the PR3-only delta (workflow, composite actions, guards, `make quality-ci`, integration shards). PR2 changes belong in #1370.

See [Branch protection migration](#branch-protection-migration) — a **maintainer pre-merge step** on `prebid/salesagent`.

## Why (issue #1234 + #1233)

Issue #1234 PR3 scope: CI becomes authoritative; reusable setup; frozen rendered check names; coverage hard-gate; creative-agent isolation; absorb #1233 CI/local divergences.

PR2 (#1370) fixed pre-commit drift. PR3 completes the CI half: one workflow, stable check names, no duplicate unit-test runs, integration parallelism restored from legacy `test.yml`.

### Authoritative CI workflow
- **New:** `.github/workflows/ci.yml` — single `CI` workflow with `permissions: {}` at top level
- **Removed:** `.github/workflows/test.yml`
- **Composite actions:** `_setup-env`, `_pytest` (paths/flags/coverage), `_postgres` (wait + optional migrate)
- **Workflow-level `env:`** — shared test credentials, `DATABASE_URL`, `GEMINI_API_KEY`, `ADCP_TESTING`
- **`Summary` job** — final aggregation gate

**Dedup note (review follow-up):** composites + workflow `env:` deduplicate pytest flags and credentials. Postgres `services:` blocks remain **inline per job** — GHA does not support YAML anchors in `services:` (actionlint/runtime reject). E2E overrides `ADCP_TESTING` on its pytest step so `conftest.py` starts `docker-compose.e2e.yml` (legacy `test.yml` parity).

### Frozen required checks (18 rendered names)

| Rendered check name | Job |
|---------------------|-----|
| `CI / Quality Gate` | Static analysis (`make quality-ci`) — no pytest |
| `CI / Type Check` | mypy |
| `CI / Schema Contract` | AdCP contract tests |
| `CI / Security Audit` | uv-secure via `scripts/security-audit.sh` |
| `CI / Quickstart` | docker compose health |
| `CI / Smoke Tests` | import smoke + skip-decorator guard |
| `CI / Unit Tests` | unit + harness, coverage artifact |
| `CI / Integration (creative)` | integration shard |
| `CI / Integration (product)` | integration shard |
| `CI / Integration (media-buy)` | integration shard |
| `CI / Integration (infra)` | integration shard |
| `CI / Integration (other)` | integration shard |
| `CI / E2E Tests` | full stack (self-managed docker) |
| `CI / Admin UI Tests` | admin suite |
| `CI / BDD Tests` | BDD suite (sharding is #1379) |
| `CI / Migration Roundtrip` | alembic upgrade |
| `CI / Coverage` | `--fail-under` from `.coverage-baseline` |
| `CI / Summary` | aggregation gate |

Enforced by `tests/unit/test_architecture_required_ci_checks_frozen.py`. Capture names with `./scripts/capture-rendered-names.sh`.

### Quality Gate split (no duplicate unit runs)
- **`make quality-ci`** — ruff + mypy + duplication guard (**no pytest**)
- **`make quality`** — `quality-ci` + unit pytest (local full gate)
- **Unit Tests** job runs unit + harness once; **Coverage** job enforces `.coverage-baseline`
- **tox** and CI share `scripts/ci/coverage_report.sh` for `.coverage-baseline` gate (no hardcoded 30% drift)
- Unit job no longer passes `-p no:randomly` (aligned with local `make quality`)

### Integration (5-way parallel shards, strict partition)

Measured overlap **before** strict partition (`pytest --collect-only`): 14 duplicate file assignments across shards 1–4 (182 total files). **After** `and not` on shards 2–4: all pairwise overlaps **0**, union/total still 182/182.

| Shard | Markers |
|-------|---------|
| creative | `creative` |
| product | `product and not creative` |
| media-buy | `(media_buy or delivery) and not creative and not product` |
| infra | `(transport or auth or … or agent) and not creative and not product and not media_buy and not delivery` |
| other | negation of all 15 markers above |

Each shard: Postgres service container, `-p no:randomly`, 15m timeout. Creative shard uses pinned adcp commit on `creative-net`.

### Other jobs
- **E2E Tests** — clears workflow `ADCP_TESTING` so conftest starts docker stack; docker cleanup before/after
- **Type Check**, **Schema Contract**, **Security Audit**, **Quickstart**, **Smoke Tests**
- **Admin UI Tests**, **BDD Tests** (`-n auto --dist=loadfile`)
- **Migration Roundtrip** — alembic upgrade + current

### Tooling & hygiene
- CI **uv 0.11.15** (GHSA-4gg8-gxpx-9rph)
- Removed dead `[tool.black]` from `pyproject.toml` (ruff canonical)
- **`docs/development/ci-pipeline.md`** — pipeline docs + branch protection checklist
- Guard tests updated for strict partition + `_pytest` routing

## Relationship to PR2 (#1370)

PR3 assumes PR2 is merged: `uv sync --group dev`, lockfile ruff/mypy, `-p no:randomly` in integration, DB pool env overrides.

## Deferrals (later PRs in #1234)

| Item | PR |
|------|-----|
| BDD parallel shards | #1379 |
| Hook relocation | PR4 |
| Version anchors / Dockerfile digest pin | PR5 |
| cosign / SBOM / Trivy | PR6 |
| Coverage ratchet to 53.5% | follow-up |

## Branch protection migration

**Who:** repo admin on `prebid/salesagent` — not automated by this PR.

**When:** after PR2 merges and this PR is rebased onto `main`, **before** merging PR3.

**How:** `./scripts/capture-rendered-names.sh` → replace `Test Suite / …` with 18 `CI / …` names. Do **not** add #1379 BDD shard names yet.

Full checklist: `docs/development/ci-pipeline.md`.

## Test plan

- [x] Architecture CI guard tests pass
- [x] `make quality` passes locally
- [x] Pre-commit hooks pass
- [x] CI green on PR (all 18 `CI / …` checks + Summary + Security actionlint)
- [x] **Maintainer:** branch protection rules updated — required before merge

## Related

- Part of #1234 (PR 3 of 6)
- Depends on #1370 (PR2)
- BDD sharding follow-up: #1379


mkostromin-sigma:
@ChrisHuie Ready for review — PR 3 of #1234 (authoritative CI pipeline).

**Merge order:** please review/merge **#1370 (PR2)** first. After PR2 lands on `main`, I will rebase this branch — the diff will shrink from ~113 files (stacked) to ~**88 files** (PR3-only: `ci.yml`, composite actions, guards, integration shards).

**What to review here (PR3-only delta):**
- Legacy `test.yml` → authoritative `ci.yml` with **18 frozen `CI / …` check names** (integration split into 5 parallel shards)
- Composite actions: `_setup-env`, `_pytest`, `_postgres`
- `make quality-ci` for Quality Gate (static analysis only — no duplicate unit test run)
- Coverage hard-gate via `.coverage-baseline` (46.5%)
- Pinned creative-agent stack (`scripts/creative-agent-stack.sh`)
- Structural guards + `docs/development/ci-pipeline.md`

**Before merge (maintainer step on `prebid/salesagent`):** branch protection rules must be migrated from legacy `Test Suite / …` to the 18 `CI / …` names — see [Branch protection migration](https://github.com/prebid/salesagent/pull/1372#branch-protection-migration) in the description. Not testable from a fork PR alone; I'll coordinate after PR2 merge + rebase.

**Follow-up:** BDD parallel shards are in draft #1379 — separate branch-protection update, not part of this PR.

CI is green on the stacked PR. Happy to walk through anything.

KonstantinMirin (commented):
PR delivers what the description claims at the surface — the CI workflow is restructured, the 18 rendered check names are pinned by a guard, and the quality-gate split is correct. The deeper concerns are around how much of the "deduplication" framing the code actually backs up, and around two new instances of the CI/local drift this PR was meant to eliminate. Specific instances are inline.

1. **The "composite actions deduplicate" framing oversells what's there.** `_setup-env` does its job — 12 callsites, no inline Python+uv setup anywhere else. But `_pytest/action.yml` is a literal no-op wrapper (`run: ${{ inputs.command }}`) used at 1 of 7 pytest sites; `_postgres` is reused at 2 of 3 postgres-consuming jobs (`migration-roundtrip` re-implements the wait inline because it needs DB before alembic, not migrate.py); the `services:` block is copy-pasted verbatim three times; `DATABASE_URL` repeats five times; the same five test env vars repeat five times. The single-source ambition is the right target — but the current state is closer to "one composite action and three near-misses" than to the description's framing.

2. **The PR is meant to kill CI/local drift; it introduces two new instances of it.** `.coverage-baseline = 46.5` is read at runtime by the CI Coverage job, but `tox.ini:114` hardcodes `coverage report --fail-under=30`. Same metric, two thresholds; `make test-all` passes at 30% while CI fails at the same coverage. Separately, CI Unit Tests passes `-p no:randomly` while `tox -e unit` does not — unit tests run randomized locally and deterministic in CI, exactly the inversion of PR2's tox/integration pattern. A flake at random ordering hits the developer, not the gate.

3. **One env-var value disagrees with the other five copies.** `smoke-tests` uses `GEMINI_API_KEY: test_key`; the other five jobs use `test_key_for_mocking`. Either intentional (then a comment) or drift from the verbatim copy-paste (then unify). Without a comment, the next person editing one of these copies will pick the wrong canonical and the divergence becomes permanent.

4. **Integration shard partition is exhaustive but not strictly mutually exclusive.** "other" is correctly the negation of the 15 markers used by the first four shards, so every test runs somewhere. But the first four shards' markers themselves overlap: a file tagged `product` + `delivery` runs in both `product` and `media-buy`; a file tagged `creative` + `targeting` runs in both `creative` and `infra`. The frozen-shards guard checks set membership, not partition exclusivity. The cost is CI wall time and confusing cross-shard failure signal for the same test. Either tighten the marker expressions with `and not (<earlier-shard-markers>)`, or document the overlap as intentional.

5. **Same fact in three places, one of them dead.** `alembic/versions/` is excluded by `pyproject.toml [tool.ruff] exclude` (canonical), `.pre-commit-config.yaml`'s ruff hook (parallel), and `pyproject.toml [tool.black] extend-exclude` (third copy; black isn't the formatter anymore per PR2's ADR-001). The dead black entry presents as live config to a reader and is a footgun the next exclusion will trip over.

For closure on earlier review threads: 18 frozen rendered check names match the guard exactly; `make quality-ci` is correctly composed and CI Unit Tests does not re-run static checks; `.type-ignore-baseline` is correctly ratcheted to 60 (closes the raise I flagged on #1370); workflow `permissions: {}` at top level with per-job `contents: read` minimum; uv 0.11.15 single-sourced and forwarded to all 12 setup invocations; adcp pin `ca70dd1e2a6c` lives in `scripts/creative-agent-stack.sh:20` only; integration shards use `strategy.matrix.include` rather than five hand-rolled jobs; the 27 `BR-UC-*.feature` files lost only a trailing blank line (mechanical end-of-file-fixer pass); the `src/` ruff format sweep is purely cosmetic. One BDD nuance worth confirming: the trailing-newline removal is on auto-generated files that are locally-overwritten on regeneration — confirm the upstream generator now outputs no trailing newline, otherwise the edit gets wiped on the next regen.

.github/actions/_pytest/action.yml
```diff
@@ -0,0 +1,12 @@
...
+description: Execute a pytest command through uv.
+inputs:
+  command:
+    description: Full pytest command to execute.
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Run pytest
+      shell: bash
```
KonstantinMirin, line 31: 6. This is `run: ${{ inputs.command }}`. It runs at 1 of 7 pytest sites in `ci.yml`; the other six inline `uv run pytest …`. As-is it adds an indirection without consolidating anything — either delete it and let all sites inline, or extend it to actually wrap the shared `uv run pytest ... -p no:randomly --cov=src ...` flag-set and route every site through it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,417 @@
...
+          POSTGRES_USER: adcp_user
+          POSTGRES_PASSWORD: test_password
+          POSTGRES_DB: adcp_test
+        options: >-
+          --health-cmd "pg_isready -U adcp_user"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+        ports:
+          - 5432:5432
```
KonstantinMirin, line 209: 7. Same `services:` block also at `:284-297` and `:327-340` (e2e, admin-ui, migration-roundtrip). GHA forbids `services:` inside composite actions, but a YAML anchor at the top of the file, or a top-level `env:` carrying the connection string + a single Postgres setup helper script, would single-source the three identical copies and the five `DATABASE_URL` repeats.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,417 @@
...
+          persist-credentials: false
+      - uses: ./.github/actions/_setup-env
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          uv-version: ${{ env.UV_VERSION }}
+      - uses: ./.github/actions/_pytest
+        with:
+          command: uv run pytest tests/smoke/ -v --tb=short
+        env:
+          GEMINI_API_KEY: test_key
```
KonstantinMirin: 8. Outlier: this value is `test_key`, the five other jobs use `test_key_for_mocking` (`:142, :208, :245, :271, :312`). Either intentional (add a one-line comment naming why smoke-tests needs the different fixture) or drift (unify). Without a comment, the next env-var edit picks the wrong canonical.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,417 @@
...
+          - group: product
+            marker: product
+          - group: media-buy
+            marker: media_buy or delivery
+          - group: infra
+            marker: transport or auth or adapter or schema or admin or infra or inventory or targeting or workflow or policy or agent
+          - group: other
+            marker: not (creative or product or media_buy or delivery or transport or auth or adapter or schema or admin or infra or inventory or targeting or workflow or policy or agent)
+    services:
+      postgres:
```
KonstantinMirin, line 197: 9. `other` is correctly exclusive (negates all 15 markers from the other four shards). The first four shards aren't mutually exclusive among themselves: `product` matches files also tagged `delivery`; `media-buy` matches the same `delivery`. Same for `creative` ∩ `infra` markers (`targeting`, etc.). The cost is duplicate execution per overlapping file. `and not (<earlier-shard-markers>)` on shards 2-4 makes the partition strict; or accept the overlap and document it next to the matrix.

.github/actions/_pytest/action.yml
mkostromin-sigma: Agreed — will extend `_pytest` to wrap the shared flag-set (paths, coverage, `-p no:randomly` where needed) and route all pytest jobs through it, rather than keeping a one-site no-op wrapper.

.github/workflows/ci.yml
mkostromin-sigma: Agreed — will add a YAML anchor for the Postgres service block and a workflow-level `env:` for `DATABASE_URL` and the repeated test credentials. Will also extend `_postgres` with a wait-only mode for migration-roundtrip.

.github/workflows/ci.yml
mkostromin-sigma: Agreed — drift, not intentional. Will unify on `test_key_for_mocking` via workflow-level `env:` (smoke tests are import-only and do not depend on the key value).

mkostromin-sigma:
@KonstantinMirin — thank you for the thorough review. All points are actionable; I have replied on each inline thread with the intended fix.

**Context:** PR2 (#1370) was upgraded with your and @ChrisHuie’s feedback (CI green on latest). I will rebase this branch onto PR2 before pushing further changes, so the diff shrinks to PR3-only.

**Steps to get this PR merge-ready:**

1. **Rebase** `ci/pr3-ci-authoritative-pipeline` onto latest PR2 (`ci/pr2-uv-lock-single-source`), then onto `main` after PR2 lands.
2. **Measure integration shard overlap** — `pytest --collect-only` per current CI markers (same filters as CI); post file-level intersection counts on the shard thread before changing markers.
3. **Integration shards** — apply strict `and not` partition if overlap is confirmed; otherwise reply with data and pick the appropriate fix.
4. **Workflow dedup** — Postgres service YAML anchor, workflow-level test `env:`, extend `_pytest` / `_postgres` (wait-only for migration-roundtrip).
5. **CI/local alignment** — `tox` coverage reads `.coverage-baseline` (not hardcoded 30); remove `-p no:randomly` from CI unit job to match local/tox; unify `GEMINI_API_KEY`.
6. **Config hygiene** — remove dead `[tool.black]` alembic exclude; keep ruff canonical + pre-commit exclude with comment.
7. **Verify** — architecture guards, `./scripts/capture-rendered-names.sh` (18 checks), CI green; update PR description.
8. **Maintainer (pre-merge)** — branch protection migration from `Test Suite / …` to 18 `CI / …` names (documented in `docs/development/ci-pipeline.md`).

I will follow up here with commits and overlap measurements once pushed.

.github/workflows/ci.yml
mkostromin-sigma: Good catch on the theoretical overlap. Plan: run `pytest --collect-only` per current shard markers first, post file-level intersection counts here, then apply strict `and not (<earlier-shard-markers>)` partition if overlap is confirmed (or document + decide if zero). Will reply with measured before/after numbers.

mkostromin-sigma: **Overlap measurement** (rebased branch, `pytest --collect-only`, CI filters):

| Shard | Files |
|-------|------:|
| creative | 32 |
| product | 36 |
| media-buy | 24 |
| infra | 76 |
| other | 28 |
| **Union / total** | **182 / 182** (no coverage gap) |

**Pairwise overlaps (shards 1–4):**
- creative ∩ product: **2** (`test_get_products_format_id_filter.py`, `test_product_multiple_format_ids.py`)
- creative ∩ media-buy: **1** (`test_update_media_buy_creative_assignment.py`)
- creative ∩ infra: **4** (creative format/schema/auth files)
- product ∩ media-buy: **0**
- product ∩ infra: **5** (GAM pricing, get_products policy/auth, inventory profile)
- media-buy ∩ infra: **2** (format conversion approval, inventory profile media buy)

Overlap confirmed — proceeding with strict `and not` partition. Will post after counts once pushed.

mkostromin-sigma: **After strict partition** (`and not` on shards 2–4): all pairwise overlaps among shards 1–4 are **0**; union/total remains **182/182**. Pushing code shortly.

.github/workflows/ci.yml
mkostromin-sigma: Update: YAML anchors are **not supported** in GHA workflow \`services:\` (actionlint + runtime reject). Kept inline \`services: postgres\` on integration/BDD/migration-roundtrip jobs + workflow-level \`env:\` for shared credentials + \`_postgres\` composite (wait-only on migration-roundtrip via \`run-migrations: false\`). Dedup is at composite/env layer, not YAML anchor layer.

mkostromin-sigma:
## Ready for re-review — Jun 4 review feedback addressed + CI green

@KonstantinMirin — all four inline threads from your Jun 4 review are addressed. CI is green on the latest push (`82cdaea8`).

### Review checklist (your items)

| # | Item | Status |
|---|------|--------|
| 1 | Route all pytest jobs through extended `_pytest` composite | Done |
| 1 | Workflow-level `env:` (credentials, `DATABASE_URL`, `GEMINI_API_KEY`) | Done |
| 1 | Postgres dedup | **Partial by GHA constraint** — YAML anchors rejected by actionlint/runtime; kept inline `services: postgres` on integration/BDD/admin/migration jobs + `_postgres` composite (`run-migrations: false` on migration-roundtrip). Replied on services thread. |
| 2a | tox coverage reads `.coverage-baseline` | Done |
| 2b | Unit job aligned with local (no `-p no:randomly` drift in CI) | Done |
| 3 | `GEMINI_API_KEY` unified via workflow `env:` | Done |
| 4 | Integration strict partition | Done — measured overlap posted (14 dupes before → 0 after `and not` on shards 2–4); guard test added |
| 5 | Remove dead `[tool.black]` | Done |

### Post-review CI fixes (3 commits after your review)

1. **`2699703d`** — `_pytest` injects `extra_args` at compile time (bash word-split broke marker expressions with parentheses)
2. **`71b673f8`** — E2E overrides workflow `ADCP_TESTING` so `conftest.py` starts `docker-compose.e2e.yml` (legacy `test.yml` parity)
3. **`82cdaea8`** — Admin UI job: Postgres service + `_postgres` migrate (admin tests use `integration_db`; was missing, caused 20m connection-retry timeout)

### CI status (run `26986925894`)

All **18** frozen `CI / …` checks + **Summary** + Security (**actionlint**, pip-audit, zizmor) — **pass**.

PR description updated with strict-partition measurements, anchor limitation, and E2E override note.

### Remaining (not in this PR's code)

- **Stacking:** still depends on **#1370 (PR2)** — will rebase onto `upstream/main` after PR2 merges
- **Maintainer pre-merge:** branch protection migration to 18 `CI / …` names (`./scripts/capture-rendered-names.sh`) — see `docs/development/ci-pipeline.md`

From our side this is merge-ready once PR2 lands and branch protection is updated. Happy to walk through any thread.

KonstantinMirin:
**Re-review status — pulled latest (`82cdaea85`)**

|---|---|---|
| Pattern 1 (composite/services dedup oversold; `_pytest` no-op) | ✅ Fully addressed | `_pytest` action rewritten as a real wrapper (paths / extra_args / disable_randomly / with_coverage / tb). Used 7× across jobs; 0 inline `uv run pytest`. `DATABASE_URL`, `GEMINI_API_KEY`, `GOOGLE_*`, `SUPER_ADMIN_EMAILS`, `ADCP_TESTING` lifted to top-level `env:`. `_postgres` got a `run-migrations` toggle; now reused at 4 sites (was 2). The 4× inline `services:` block is acknowledged in commit `3b5e43b19` as a GHA limitation (YAML anchors rejected inside `services:`) — accepted |
| Pattern 2 (CI/local coverage threshold drift; CI/local `-p no:randomly` drift) | ✅ Fully addressed | `tox.ini:114` now `coverage report --fail-under=$(tr -d '[:space:]' < {toxinidir}/.coverage-baseline)`. `-p no:randomly` parity: off in both `tox -e unit` and CI Unit Tests; on in both `tox -e integration` and CI Integration (via `_pytest disable_randomly: "true"`) |
| Pattern 3 (smoke-tests `test_key` outlier) | ✅ Fully addressed | Single top-level `GEMINI_API_KEY: test_key_for_mocking`. No outlier left |
| Pattern 4 (integration shard non-mutex) | ✅ Fully addressed | Each shard now carries `and not <earlier-shard-markers>` — partition is strictly exclusive |
| Pattern 5 (alembic excluded in 3 places, dead `[tool.black]`) | ✅ Fully addressed | `[tool.black]` removed. Pre-commit ruff hooks still carry explicit `exclude:` but with an inline comment explaining the constraint (ruff ignores `pyproject.toml` excludes when pre-commit passes paths explicitly). That's a tool limitation, not duplicated truth |

**Scanned for new issues**: the three follow-up commits after the main fix are real downstream bugs caught and repaired:
- `2699703d0` — `extra_args` injection switched to GHA compile-time substitution so marker-expression parens aren't bash-word-split.
- `71b673f88` — E2E job clears `ADCP_TESTING` / `DATABASE_URL` at step level because the workflow-level defaults would make conftest wait for an existing server instead of starting `docker-compose.e2e.yml`. Documented.
- `82cdaea85` — admin-ui-tests was missing the Postgres service after the dedup; added it AND added a regression guard in `test_architecture_ci_suite_coverage.py`.

The workflow-level `ADCP_TESTING: true` default does mean any future test job that doesn't want it must override explicitly — but the E2E fix shows that pattern is in place and a future regression in E2E behavior is caught by the new guard tests.

All 28 required checks green. From my side: ready.

mkostromin-sigma:
@ChrisHuie @KonstantinMirin — PR3 rebased onto latest PR2 (#1370) stack.

**Stacking**
- Base: latest `ci/pr2-uv-lock-single-source` (`6ec1ff7f`, includes Jun 8 PR2 follow-up).
- PR3-only: **13 commits**, **~88 files** (`git diff ci/pr2-uv-lock-single-source..HEAD`).
- Merge order unchanged: **#1370 first**, then rebase this branch onto `main` and migrate branch protection.

**Rebase**
- Replayed PR3 commits with `git rebase --onto ci/pr2-uv-lock-single-source feec0015`.
- One conflict in `tests/admin/test_creatives_blueprint.py` — kept upstream `#1395` patch-target comment + `_PUSH_PATCH` at use site.

**Review status (unchanged from Jun 6)**
All four Jun 4 threads + three post-review CI fixes remain in place on the new base:
- `_pytest` composite routes all pytest jobs
- Workflow-level `env:` + `_postgres` (`run-migrations: false` on migration-roundtrip)
- Strict integration shard partition (0 pairwise overlap)
- tox + CI coverage both read `.coverage-baseline`
- Unit job aligned with local/tox on `pytest-randomly`
- E2E clears workflow `ADCP_TESTING`; admin-ui-tests has Postgres service + guard

**After PR2 merges**
1. Rebase onto `upstream/main` (expect ~88-file PR3-only diff).
2. Maintainer: branch protection → 18 `CI / …` names (`docs/development/ci-pipeline.md`).

CI should be running on the latest push. Ready for re-review when you have a moment.

mkostromin-sigma:
Rebased onto latest PR2 (`d6c84dd4` — round-2 review fixes + green CI).

**Method:** cherry-picked 11 substantive PR3 commits onto `origin/ci/pr2-uv-lock-single-source`; skipped format-sweep commits (`83d97994`, `d5ecf5a4`) that conflicted with post-#1307 main. Resolved conflicts in `ci.yml`, frozen-check guard, and BDD assertion allowlist line numbers (kept current HEAD line refs).

**Stack:** PR2 (10 commits) + PR3-only (11 commits) → force-pushed to `ci/pr3-ci-authoritative-pipeline`.

CI should run on this push.

mkostromin-sigma:
@ChrisHuie — **PR3 review / merge note (#1372)**

CI is green. KonstantinMirin review threads are addressed. Merge **after #1370**.

### Your action for this PR
- Review + approve + merge when #1370 is on `main`.

### After **you** merge this PR — **Chris maintainer tasks**
1. Run `./scripts/capture-rendered-names.sh` on `main` (or the rebased PR branch) and copy the **18** `CI / …` names.
2. GitHub → Settings → Rules → `main` branch protection:
   - **Remove** legacy required checks prefixed with `Test Suite /`.
   - **Add** all **18** `CI / …` checks from the script output.
   - Keep existing non-CI required checks (`check-pr-title`, CodeQL, `security.yml` jobs, etc.).
3. Confirm a test PR shows all 18 `CI / …` checks as **required** and green before treating migration complete.

| Who | Other post-merge task |
|-----|----------------------|
| **Author (me)** | Rebase #1379 (PR3-2) onto `main` |

Details: `docs/development/ci-pipeline.md` § **Branch protection migration**.

### Not yet (later PRs in chain)
- **#1379:** +3 BDD shard required checks (→ 21 total) — separate BP update after PR3-2 merges
- **#1401:** dependency graph, `Security / Dependency Review`, secret scanning

Thanks!

ChrisHuie:
# Review — findings with fixes (verified at head `4692d0249`)

The core of this PR holds up: the `test.yml` → `ci.yml` replacement drops nothing (all 9 legacy jobs map across, every pytest flag carried), the strict shard partition measures exactly exclusive and exhaustive at this head (185 integration files, shard union 185/185, all pairwise overlaps 0), the frozen-names and suite-coverage guards fail correctly under mutation testing, and all 28 checks are green on the post-#1370 rebase.

The items below are what's left. **Items 1–7 should land before merge** (several get more expensive to change after branch protection freezes the 18 check names); items 8–11 are hardening worth folding into the same commit. Every diff shown was applied and verified together as one change set: both workflow guards pass (12/12), `./scripts/capture-rendered-names.sh` output is byte-identical to the frozen 18 (branch protection unaffected), ruff/mypy clean on the touched Python file, and the full quality gate passes (4763 passed / 12 skipped / 26 xfailed).

---

## 1. No `concurrency` group — every superseded push runs the full pipeline to completion

**Where:** `.github/workflows/ci.yml`, workflow top level (legacy `test.yml` lacked one too, but this PR owns the file now).

**Problem:** One push costs ~65 runner-minutes across 18 jobs (~23 min wall, measured from the latest run). Without a concurrency group, pushing a fix-up while CI is running lets the stale run burn to completion. This PR alone saw 4+ pushes in a single review day, and four more stacked PRs (#1379 → #1401) will repeat that pattern.

**Fix:**

````diff
 permissions: {}

+concurrency:
+  # One run per PR head ref; a superseded push cancels the in-flight run.
+  # push/dispatch events get per-sha groups so runs on main never cancel or
+  # queue behind each other.
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
````

**Why the conditional instead of the common `group: ${{ github.ref }}` + `cancel-in-progress: true`:** GitHub cancels *queued* runs in a shared group. With the naive form, two merges to `main` in quick succession leave the middle merge commit with a cancelled run — a permanently non-green check in main's history. The conditional gives PR runs a shared per-ref group (`refs/pull/N/merge`) with cancellation, while `push`/`workflow_dispatch` runs get per-SHA groups that can never collide.

---

## 2. The BDD job bypasses the `_pytest` composite again — reopens the open review thread

**Where:** `.github/workflows/ci.yml:332-342` and `.github/actions/_pytest/action.yml`.

**Problem:** The review thread on `_pytest/action.yml:27` asked for the composite to either be deleted or to wrap the shared flag-set and route **every** pytest site through it. That was achieved (7/7 sites) — then the Coverage-job commit reverted BDD to an inline `uv run pytest`, because the composite cannot express three things BDD needs: a custom `COVERAGE_FILE`, a custom XML report path (`coverage-bdd.xml`), and a suppressed terminal report. Current state is 6/7 routed, with the inline command re-duplicating `-q --tb=line` — exactly the flag drift the composite exists to prevent.

**Fix, part 1 — teach the composite the xdist-coverage shape** (one new input; existing callers unaffected):

````diff
   with_coverage:
     description: When true, pass --cov=src with xml/term reports.
     required: false
     default: "false"
+  coverage_xml:
+    description: Custom XML coverage report path; enables --cov=src with no terminal report (xdist-friendly).
+    required: false
+    default: ""
   tb:
````

````diff
         if [ "${{ inputs.with_coverage }}" = "true" ]; then
           args+=(--cov=src --cov-report=xml --cov-report=term)
         fi
+        if [ -n "${{ inputs.coverage_xml }}" ]; then
+          args+=(--cov=src "--cov-report=xml:${{ inputs.coverage_xml }}" --cov-report=)
+        fi
         # Inject extra_args at compile time — do not bash-word-split (marker exprs contain parentheses).
````

(`with_coverage` and `coverage_xml` are intended to be mutually exclusive; no job sets both.)

**Fix, part 2 — route the BDD job through it.** `COVERAGE_FILE` stays as step-level `env:` on the `uses:` step (step env propagates into composite `run:` steps — same pattern as `CREATIVE_AGENT_URL` on the creative shard). The fragment-combine becomes its own step because the composite `exec`s pytest:

````diff
-      - name: Run BDD tests
+      - uses: ./.github/actions/_pytest
         env:
+          # Separate data file so the artifact cannot collide with unit's .coverage.
           COVERAGE_FILE: .coverage.bdd
+        with:
+          paths: tests/bdd/
+          coverage_xml: coverage-bdd.xml
+          extra_args: -n auto --dist=loadfile --timeout=60
+      - name: Combine xdist coverage fragments
         run: |
-          uv run pytest tests/bdd/ -n auto --dist=loadfile -q --tb=line \
-            --cov=src --cov-report=xml:coverage-bdd.xml --cov-report= --timeout=60
           shopt -s nullglob
           fragments=(.coverage.bdd.*)
           if [ "${#fragments[@]}" -gt 0 ]; then
             uv run coverage combine --data-file=.coverage.bdd .coverage.bdd "${fragments[@]}"
           fi
````

Verified: `test_bdd_job_runs_the_bdd_suite` still passes (the guard reads `_pytest` `paths:` inputs as well as `run:` text); produced artifacts are identical (`.coverage.bdd` + `coverage-bdd.xml`); restores 7/7 routing. Note for #1379: the shard rewrite of this job inherits this shape — each shard sets its own `COVERAGE_FILE`.

---

## 3. mypy executes twice per pipeline — decide now, while the check names are still fluid

**Where:** `Makefile:11` (`make quality-ci` includes `uv run mypy src/ --config-file=mypy.ini`, run by the Quality Gate job) and `.github/workflows/ci.yml:54-55` (the Type Check job runs the same invocation).

**Problem:** The legacy `lint` job ran mypy once. The Quality Gate / Type Check split runs the identical check in two required jobs — duplicate execution in a PR whose stated principle is "no duplicate runs." Options:

- **(a) Keep both, document the decision** — costs ~1.5 runner-minutes, zero wall-clock (parallel jobs), keeps a dedicated mypy failure signal, and keeps `make quality-ci` as the single local/CI target. **Recommended.**
- (b) Drop the Type Check job — changes the frozen 18 names during the branch-protection migration; worst possible timing.
- (c) Remove mypy from `quality-ci` — strips type-checking from the local gate (`make quality` calls `quality-ci`), recreating the local/CI drift this stack exists to eliminate.

**Fix (option a):**

````diff
+  # mypy also runs inside Quality Gate via `make quality-ci` (the unified local
+  # target — removing it there would drift local from CI). The dedicated job is
+  # kept as a separate required-check signal; jobs run in parallel, so the
+  # overlap costs runner minutes, not wall-clock time.
   type-check:
     name: Type Check
````

---

## 4. Comment references a "CI storyboard job" that has never existed

**Where:** `src/core/tools/capabilities.py:256-260`.

**Problem:** The comment says "The CI storyboard job is advisory (see `.github/workflows/ci.yml`)" — but no storyboard job exists in `ci.yml`, in the deleted `test.yml`, or anywhere in workflow history (`git log -S 'storyboard' -- .github/workflows/` over main is empty; the comment arrived in #1276 already pointing at a job that wasn't there). This PR's `test.yml` → `ci.yml` pointer update re-validates a dangling reference: a reader auditing why two declared-but-failing scenarios don't block merge is sent to a file containing no such job.

**Fix** (keeps the true policy claim, drops the false mechanism claim):

````diff
     # We declare the specialism even though `pending_creatives_to_start` and
-    # `invalid_transitions` are not yet fully green. The CI storyboard job is
-    # advisory (see `.github/workflows/ci.yml`), so those scenario failures
-    # don't block merge — and the public declaration forces prioritization of
-    # the remaining gaps instead of hiding them.
+    # `invalid_transitions` are not yet fully green. Storyboard compliance runs
+    # are advisory — no required CI job executes them — so those scenario
+    # failures don't block merge, and the public declaration forces
+    # prioritization of the remaining gaps instead of hiding them.
````

---

## 5. Quality Gate step is named "Run pre-commit" but runs `make quality-ci`

**Where:** `.github/workflows/ci.yml:38-39`. The legacy lint job never ran pre-commit either, so this isn't a leftover — it's just wrong, and it's the first thing someone debugging a Quality Gate failure reads. Step names are not rendered check names, so branch protection is unaffected.

````diff
-      - name: Run pre-commit
+      - name: Run static analysis (make quality-ci)
         run: make quality-ci
````

---

## 6. E2E and Admin collect coverage that nothing consumes

**Where:** `.github/workflows/ci.yml:265` (e2e) and `:314` (admin), both `with_coverage: "true"`.

**Problem:** Only the unit and BDD jobs upload coverage artifacts; the Coverage gate downloads and combines exactly those two. E2E/Admin coverage data is generated and discarded every run. E2E's is misleading on top of wasteful: the server under test runs inside Docker, so host-side coverage measures the client harness, not `src/` (the legacy e2e job collected no coverage — this restores parity). Admin's coverage *is* meaningful (in-process tests), but wiring it into the gate changes the measured percentage and belongs with the deferred coverage-ratchet work, not the branch-protection migration window.

````diff
       - uses: ./.github/actions/_pytest
         with:
           paths: tests/e2e/
-          with_coverage: "true"
+          # No coverage: the server under test runs inside Docker, so host-side
+          # coverage would measure the client harness, not src/.
           extra_args: --timeout=300
````

````diff
         with:
           paths: tests/admin/
-          with_coverage: "true"
+          # The Coverage gate combines unit + BDD only; add an artifact upload
+          # before re-enabling coverage here, or the data is silently dropped.
           extra_args: -m "not requires_server and not slow"
````

---

## 7. Documentation states the wrong tool for Security Audit, plus two other doc drifts

**Where:** `docs/development/ci-pipeline.md:32`, `:45`, `:115-117`; `.github/LINK_VALIDATION_CI.md:18-27`.

**Problem (a):** Both the check table ("pip-audit + uv-secure") and the Security Audit section ("runs `uv run pip-audit`") misstate what the job does. `scripts/security-audit.sh` ends in `exec uvx uv-secure --ignore-vulns … uv.lock` — **uv-secure only**. pip-audit is a separate required check in `security.yml`; the two tools share only the suppression list (`scripts/security-ignored-vulns.sh`). The same wrong claim is in the PR description's table.

````diff
-| `CI / Security Audit` | pip-audit + uv-secure |
+| `CI / Security Audit` | uv-secure lockfile audit (pip-audit is a separate `security.yml` check) |
````

````diff
-The `Security Audit` job runs `uv run pip-audit` against pinned dependencies.
-Any known vulnerability in a direct dependency fails the build. Fix by bumping
-the affected package in `pyproject.toml` and running `uv lock`.
+The `Security Audit` job runs `./scripts/security-audit.sh`, which executes
+**uv-secure** against `uv.lock`. **pip-audit** runs as a separate required
+check in `security.yml`; both tools read the shared suppression list in
+`scripts/security-ignored-vulns.sh` (single source of truth for ignored
+advisories). Any known vulnerability in a direct dependency fails the build.
+Fix by bumping the affected package in `pyproject.toml` and running `uv lock`.
````

**Problem (b):** The Coverage row predates the combine rework — the gate now uses unit **and** BDD data (and the suite-coverage guard enforces the combine):

````diff
-| `CI / Coverage` | `--fail-under` from `.coverage-baseline` using unit artifact |
+| `CI / Coverage` | `--fail-under` from `.coverage-baseline` using combined unit + BDD artifacts |
````

**Problem (c):** `.github/LINK_VALIDATION_CI.md` shows a config snippet headed `# .github/workflows/test.yml` — a file this PR deletes — in the pre-composite invocation style:

````diff
 ## CI Configuration
 ```yaml
-# .github/workflows/test.yml
+# .github/workflows/ci.yml — integration-tests job (excerpt; the real job adds
+# entity-shard markers via strategy.matrix and shared composite actions)
 integration-tests:
   runs-on: ubuntu-latest
   services:
     postgres:
       image: postgres:15
   steps:
-    - name: Run integration tests
-      run: |
-        uv run pytest tests/integration/ -v --cov=. -m "not requires_server and not skip_ci"
+    - uses: ./.github/actions/_pytest
+      with:
+        paths: tests/integration/
+        disable_randomly: "true"
+        extra_args: --timeout=60 -m "(${{ matrix.marker }}) and not requires_server and not skip_ci"
 ```
````

---

## 8. Eight jobs have no `timeout-minutes` — a hang costs the 360-minute default

**Where:** `quality-gate`, `type-check`, `schema-contract`, `security-audit`, `smoke-tests`, `unit-tests`, `coverage`, `summary` (integration/e2e/admin/bdd/quickstart already have bounds). With item 1's concurrency in place, a hung job would also pin its group slot for six hours.

**Fix** — one `timeout-minutes:` line per job, sized from measured durations. For the first seven, insert immediately after the job's `contents: read` line; `summary` has no `permissions:` block, so insert after its `runs-on:` line:

| Job | Insert after | Value | Observed |
|---|---|---|---|
| `quality-gate` | `contents: read` | `timeout-minutes: 15` | 8m36s |
| `unit-tests` | `contents: read` | `timeout-minutes: 15` | 5m24s |
| `type-check` | `contents: read` | `timeout-minutes: 10` | 1m25s |
| `schema-contract` | `contents: read` | `timeout-minutes: 10` | 33s |
| `security-audit` | `contents: read` | `timeout-minutes: 10` | 31s |
| `smoke-tests` | `contents: read` | `timeout-minutes: 10` | 36s |
| `coverage` | `contents: read` | `timeout-minutes: 10` | 19s |
| `summary` | `runs-on: ubuntu-latest` | `timeout-minutes: 5` | 2s |

Example (quality-gate; the other seven are identical in shape):

````diff
     permissions:
       contents: read
+    timeout-minutes: 15
     steps:
````

---

## 9. Cache key omits the Python version — and PR5 is "version consolidation"

**Where:** `.github/actions/_setup-env/action.yml:35`. The cache stores `.venv` keyed only on `uv.lock`; a Python bump that doesn't touch the lockfile would restore a stale 3.12 venv. The next PR in this stack changes version anchors, so land the guard first:

````diff
-        key: ${{ runner.os }}-uv-v3-${{ hashFiles('uv.lock') }}
+        key: ${{ runner.os }}-uv-v4-py${{ inputs.python-version }}-${{ hashFiles('uv.lock') }}
````

The `v3` → `v4` bump invalidates existing caches once (~1–2 min cold sync per job on the next run, then warm again).

---

## 10. Smoke's skip-decorator grep also matches `skipif`

**Where:** `.github/workflows/ci.yml:143` (carried verbatim from legacy `test.yml:76`). `grep "@pytest.mark.skip"` matches `@pytest.mark.skipif` as a substring. The repo has zero `skipif` uses today, so this is latent — but the first legitimate platform-conditional skip will fail CI with a misleading "Found skip decorators!" message.

````diff
-          if grep -r "@pytest.mark.skip" tests/ --include="test_*.py" | grep -v "skip_ci"; then
+          # \b is unsupported in POSIX grep -E; the group below matches `skip` and
+          # `skip(...)` but not `skipif` (platform-conditional skips are legitimate).
+          if grep -rE "@pytest\.mark\.skip(\(|[[:space:]]|$)" tests/ --include="test_*.py" | grep -v "skip_ci"; then
````

Verified against a synthetic file: matches `@pytest.mark.skip` and `@pytest.mark.skip(reason=…)`, ignores `@pytest.mark.skipif(…)`.

---

## 11. `PYTEST_CURRENT_TEST: true` in workflow env is dead configuration

**Where:** `.github/workflows/ci.yml:21` (carried from legacy `test.yml:70`). The repo has no reader of this variable; the only references are `tests/conftest_db.py:13` (which **sets** it at every pytest session start, so test behavior cannot depend on the workflow copy) and a `tox.ini` passenv pass-through. pytest manages this variable itself during runs.

````diff
   ADCP_TESTING: true
-  PYTEST_CURRENT_TEST: true
   ENCRYPTION_KEY: MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA=
````

---

## PR description edits (no commit needed)

1. The stacking table promises "~88–89 files" after rebase; the actual PR3-only diff is **25 files** (the restack dropped the two format-sweep commits).
2. The check table's Security Audit row says "pip-audit + uv-secure" — same correction as item 7a.

## Two facts worth recording before the coverage ratchet

- **The gate's headroom is 1.5 points:** the combined unit+BDD run measures 48.0% against the 46.5 floor — roughly 600 of 40,010 statements of slack. A single sizeable untested module added to `src/` could trip it; the deferred "ratchet to 53.5%" implies adding coverage inputs or tests, not just raising the number.
- **The baseline means different things locally vs CI:** `tox -e coverage` combines all five suites against the same floor (a much higher number — integration dominates), so 46.5 effectively binds only CI. Worth a doc note now; deciding which measurement defines the number belongs to the ratchet discussion.

## Verification performed on the full change set

All diffs above applied together at `4692d0249`: `test_architecture_required_ci_checks_frozen.py` + `test_architecture_ci_suite_coverage.py` pass 12/12 (which also proves the modified YAML parses); `./scripts/capture-rendered-names.sh` output byte-identical to the frozen 18 names; `ruff format --check`, `ruff check`, and `mypy` clean on `capabilities.py`; full `make quality` green (4763 passed / 12 skipped / 26 xfailed). Not run locally: actionlint/zizmor — both run as checks on push, and the only constructs added (conditional `concurrency`, step-level `env:` on a `uses:` step, `timeout-minutes`) are standard GHA.

**Comment by mkostromin-sigma** [comment]
@ChrisHuie — all review items addressed in separate commits (pushed to `ci/pr3-ci-authoritative-pipeline`):

| # | Item | Commit |
|---|------|--------|
| 1 | PR concurrency group | `61f26372` |
| 2 | BDD via `_pytest` + `coverage_xml` | `281c26a6` |
| 3 | Document duplicate mypy (Quality Gate + Type Check) | `e3933309` |
| 4 | Fix misleading storyboard CI comment | `06601c0b` |
| 5 | Rename step → "Run static analysis (make quality-ci)" | `69be9bb0` |
| 6 | Drop unused E2E/Admin coverage collection | `d1c8b2a2` |
| 7 | Docs: Security Audit = uv-secure; Coverage = unit+BDD; LINK_VALIDATION → ci.yml | `67360298` |
| 8 | `timeout-minutes` on 8 jobs | `95397d94` |
| 9 | uv cache key includes Python version (`uv-v4`) | `d9d016e7` |
| 10 | Smoke skip grep excludes `skipif` | `d9f0ab89` |
| 11 | Remove dead `PYTEST_CURRENT_TEST` workflow env | `0f2ec2f4` |

@KonstantinMirin — prior threads also addressed on this branch: all 7 pytest sites use `_pytest`; GEMINI unified via workflow `env:`; strict integration shard partition (0 overlap among shards 1–4); Postgres blocks stay inline (GHA rejects YAML anchors in `services:`).

CI re-running on latest push. Ready for re-review.

**Comment by ChrisHuie** [comment]
Verified all 11 items from my review landed at head 0f2ec2f4 — CI green, guards pass, shard partition still measures 0 overlap. Two fixes before I approve:

**1. The tox coverage gate doesn't execute (`tox.ini:114`).** tox runs `commands` as argv without a shell, so `$(tr ...)` is passed literally:

    $ tox -e coverage
    coverage: commands[1]> coverage report '--fail-under=$(tr' -d '[:space:]' '<' '…/.coverage-baseline)'
    option --fail-under: invalid floating-point value: '$(tr'
    coverage: FAIL code 1

CI's copy is fine (real shell block), but locally `run_all_tests.sh` wraps this env in `|| echo "(non-fatal)"`, so the combined-coverage gate, htmlcov, and coverage.json are silently never produced — the exact CI/local drift this PR is meant to kill. Fix: wrap in an explicit shell (`bash -c '...'` + `allowlist_externals = bash`) or move the gate into a small script shared with the ci.yml Coverage job. Please verify with an actual `tox -e coverage` run.

**2. `test_merge_head_downgrade_target_uses_first_parent` asserts the *current* head is a merge migration** (`tests/unit/test_architecture_single_migration_head.py:60`). Reproduced: add any ordinary single-parent migration and the test fails (`assert 1 > 1`) in a file that PR never touched. Restructure it like the sibling single-parent test: search the graph for any merge revision and assert `resolve_roundtrip_downgrade_target(rev) == downs[0]`, skipping only if the graph has no merge revision.

**3. PR description:** stacking table still says "~88–89 files (expected)" (rebased diff is 26 files) and the Security Audit row still says "pip-audit + uv-secure" (the job runs uv-secure only).

After those: approve + branch-protection migration per the runbook.

**Comment by mkostromin-sigma** [comment]
@ChrisHuie — round-2 review items addressed:

| # | Item | Commit |
|---|------|--------|
| 1 | tox coverage gate via shared `scripts/ci/coverage_report.sh` (also wired into CI Coverage job) | `14c638cf` |
| 2 | Graph-agnostic merge-head downgrade test | `b9581dc4` |
| 3 | PR description updated (~27 files vs main; Security Audit = uv-secure only) | (this comment) |

`tox config -e coverage` now resolves `bash …/coverage_report.sh` instead of literal `$(tr …)`.
Ready for re-review + approve.

**Review by ChrisHuie — APPROVED** [review]

File: .github/workflows/ci.yml [hunk 5]
```diff
@@ -0,0 +1,484 @@
...
+        if: always()
+        with:
+          name: bdd-coverage
+          path: |
+            .coverage.bdd
+            coverage-bdd.xml
+          include-hidden-files: true
+          if-no-files-found: error
+
+  migration-roundtrip:
```
KonstantinMirin, line 375: **`migration-roundtrip` is the only job without `timeout-minutes`.** All 13 other jobs set 5–25 min. A hung alembic step pins the runner for GitHub's 6-hour default; because this job is in `summary.needs`, the required check stops reporting for hours and blocks the PR. Add `timeout-minutes: 15` (matches `integration-tests`).

scripts/ci/migration_roundtrip.py
```diff
@@ -0,0 +1,44 @@
...
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from alembic.config import Config
+
+from alembic import command
+from tests.unit._migration_helpers import resolve_roundtrip_downgrade_target
```
KonstantinMirin, line 16: **PAT-01 — inverted dependency: a CI-invoked script importing from the test tree.** This script runs at `ci.yml:407`, yet imports `resolve_roundtrip_downgrade_target` from `tests/unit/_migration_helpers.py`. The migration-graph resolver is operational logic with no test-framework dependency — move it to `scripts/ci/` (or `src/core/database/`) and have the guard tests import *down* from there. Operational code must not depend on `tests/`.

scripts/capture-rendered-names.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Print rendered CI check names ("CI / <job name>") from ci.yml, including matrix expansion.
+uv run python - <<'PY'
+from tests.unit.workflow_helpers import rendered_ci_check_names
```
KonstantinMirin, line 6: **PAT-01 — same inverted dependency.** This branch-protection-setup script imports `rendered_ci_check_names` from `tests/unit/workflow_helpers.py`. CI-YAML parsing is operational; it belongs under `scripts/ci/`. As written, production tooling can't run without `tests/` on the path.

tests/unit/_migration_helpers.py
```diff
@@ -109,6 +109,38 @@ def get_migration_heads() -> set[str]:
...
+            return down_revisions[0]
+        # At a merge head, Alembic downgrade to any parent revision undoes the
+        # merge and restores all branch tips (see alembic branches docs).
+        return down_revisions[0]
+
+    msg = f"Migration file not found for head revision {head_revision}"
+    raise ValueError(msg)
+
+
 def is_merge_migration(tree: ast.Module) -> bool:
```
KonstantinMirin, line 144: **PAT-01 — duplicated correctness predicate.** `is_merge_migration` / `is_empty_body` here mirror `.pre-commit-hooks/check_migration_completeness.py:16-39`, kept in sync by hand because the hook can't import from `tests/`. Once these helpers move to an importable home, the hook imports the single definition and this copy goes away. Two copies of a correctness predicate is a latent drift bug.

tests/unit/test_architecture_ci_suite_coverage.py
```diff
@@ -72,29 +69,94 @@ def test_bdd_job_has_postgres_service(self):
...
+        )
+
+        assert "unit-tests" in needs, "coverage job must depend on unit-tests."
+        assert "bdd-tests" in needs, "coverage job must depend on bdd-tests."
+        assert steps_text.count("download-artifact") >= 2, "coverage job must download unit and BDD coverage artifacts."
+        run_steps = " ".join(str(step.get("run", "")) for step in coverage_job.get("steps", []))
+        assert "pytest" not in run_steps, "coverage job must not re-run tests."
+        assert "coverage combine" in run_steps, "coverage job must combine unit and BDD coverage data."
+
     def test_summary_gates_bdd_and_e2e(self):
```
KonstantinMirin, line 138: **PAT-02 — this guard protects 2 of 13 gated jobs.** `test_summary_gates_bdd_and_e2e` iterates only `("bdd-tests", "e2e-tests")`. I removed `migration-roundtrip` from both `summary.needs` and the validate step and this test stayed green — i.e. this PR's own new gate can be silently un-gated. Iterate the full `summary.needs` set, cross-checked against the job list, the way `test_architecture_required_ci_checks_frozen.py` does set-equality.

scripts/ci/migration_roundtrip.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def main() -> int:
```
KonstantinMirin, line 19: **PAT-02 — the roundtrip this PR adds as a gate has no test.** Only the pure AST helper is covered, and its test asserts `== downs[0]` — restating the helper's own "first parent" choice, not that alembic accepts the resolved target. Exercise `main()` end-to-end against a real (or fixture) alembic graph so the gate is verified, not assumed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,484 @@
...
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/_setup-env
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          uv-version: ${{ env.UV_VERSION }}
+      - name: Run mypy
+        run: uv run python -m mypy src --config-file=mypy.ini
```
KonstantinMirin, line 67: **PAT-03 — CI/local mypy drift.** CI runs `uv run python -m mypy src`; `make typecheck` runs `uv run mypy src/`. Two invocation forms for one check, in a PR whose point is parity. Have the job call `make typecheck` (or align both to one form).

.github/actions/_pytest/action.yml
```diff
@@ -0,0 +1,47 @@
+name: Run pytest
+description: Run pytest via uv with shared CI defaults.
+inputs:
+  paths:
+    description: Space-separated pytest paths or files.
+    required: true
+  extra_args:
```
KonstantinMirin, line 7: **PAT-03 — input-naming inconsistency across sibling actions.** `_pytest` uses snake_case (`extra_args`, `with_coverage`, `coverage_xml`); `_setup-env`/`_postgres`, added in this same PR, use kebab-case (`python-version`, `database-url`). Pick one convention now, while there are three actions, not thirty.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,484 @@
...
+          if [ ! -f .coverage ]; then
+            echo "Unit coverage data file missing."
+            exit 1
+          fi
+          if [ ! -f .coverage.bdd ]; then
+            echo "BDD coverage data file missing."
+            exit 1
+          fi
+          uv run coverage combine --data-file=.coverage.combined .coverage .coverage.bdd
+          ./scripts/ci/coverage_report.sh --data-file .coverage.combined
```
KonstantinMirin, line 442: **PAT-03 — coverage gate measures a different population than local.** This combines unit+BDD only; `tox -e coverage` combines all five suites. The `--fail-under` baseline therefore means two different things. Either combine the same suites, or add a comment/guard pinning `.coverage-baseline` to the unit+BDD subset it was calibrated against.

tests/e2e/conftest.py
```diff
@@ -42,10 +42,16 @@ def port_scan_start(start_port: int, end_port: int, pid: int | None = None) -> i
     span = end_port - start_port
     if span <= 1:
         return start_port
+    # Test suites may temporarily monkeypatch process metadata. Keep allocator
+    # robust by normalizing PID-like values while preserving scan scatter.
+    try:
+        normalized_pid = int(pid)
+    except (TypeError, ValueError):
+        normalized_pid = id(pid)
```
KonstantinMirin, line 50: **PAT-04 — production loosened for a test.** This `int(pid)` → `id(pid)` fallback is dead for real PIDs (`os.getpid()` is always int) and exists only so a `MagicMock` test input won't raise. `id()` is a memory address — non-deterministic — which breaks the determinism this function's docstring and `TestPortScanStartScattersByProcess` promise. Pass a real int in the test, or raise on a non-int here. Don't let the test pin the loosened contract.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,484 @@
...
+          python-version: ${{ env.PYTHON_VERSION }}
+          uv-version: ${{ env.UV_VERSION }}
+      - uses: ./.github/actions/_pytest
+        with:
+          paths: tests/smoke/
+          tb: short
+          extra_args: -v
+      - name: Check for skipped tests
+        run: |
+          if grep -rE '@pytest\.mark\.skip(\(|[[:space:]]|$)' tests/ --include="test_*.py" | grep -v "skip_ci"; then
```
KonstantinMirin, line 143: **PAT-02 — skip-decorator check duplicated with divergent exclusions.** This inline `grep` re-implements `TestNoSkippedTests` from the smoke suite, but the two have different exclusion lists, so they can disagree on the same input. Drop the shell step; the smoke test is the single source of truth (DRY invariant).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,484 @@
...
+            coverage-bdd.xml
+          include-hidden-files: true
+          if-no-files-found: error
+
+  migration-roundtrip:
+    name: Migration Roundtrip
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    services:
```
KonstantinMirin, line 380: **`migration-roundtrip` also missing `permissions:`.** Every other job declares `permissions: contents: read`; this one inherits the workflow-level `{}`. Add it so a future step that needs a scope fails loudly at review, not at runtime with a 403.

tests/unit/workflow_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def load_ci_workflow() -> dict:
+    """Load and parse the canonical CI workflow file."""
+    return yaml.safe_load(CI_WORKFLOW_PATH.read_text(encoding="utf-8"))
+
+
+def rendered_ci_check_names(workflow: dict | None = None) -> set[str]:
+    """Return rendered ``CI / …`` check names, expanding strategy.matrix jobs."""
+    jobs = (workflow or load_ci_workflow())["jobs"]
```
KonstantinMirin, line 19: **PAT-02 — opaque failure on the input it's meant to guard.** `(workflow or load_ci_workflow())["jobs"]` and the per-job indexing raise a bare `KeyError` on a renamed/empty `ci.yml`, throwing away the helpful assertion messages the guards otherwise invest in. `load_ci_workflow()` also returns `None` on an empty file → `TypeError`. Guard the lookups and assert with a message naming the missing job.

tests/unit/_migration_helpers.py
```diff
@@ -109,6 +109,38 @@ def get_migration_heads() -> set[str]:
...
+    """
+    if head_revision is None:
+        heads = get_migration_heads()
+        if len(heads) != 1:
+            msg = f"Expected exactly 1 migration head, found {sorted(heads)}"
+            raise ValueError(msg)
+        head_revision = next(iter(heads))
+
+    for path in get_migration_files():
+        revision, down_revisions = extract_revision_info(path)
```
KonstantinMirin, line 128: **PAT-02 — this new resolver depends on a helper that silently drops broken migrations.** `extract_revision_info` (pre-existing, ~line 43) swallows `SyntaxError` → `(None, [])`, so a migration that doesn't parse vanishes from the graph this roundtrip/single-head guard protects — the invariant could pass *because* a file is broken. Since this PR makes the guard depend on that behavior, surface the parse failure instead of absorbing it. (`parse_function` also misses `async def` — latent today, as alembic funcs are sync.)

.github/actions/_setup-env/action.yml
```diff
@@ -0,0 +1,39 @@
...
+inputs:
+  python-version:
+    description: Python version to use.
+    required: false
+    default: "3.12"
+  uv-version:
+    description: uv version to install.
+    required: false
+    default: "0.11.15"
+  dependency-group:
```
KonstantinMirin, line 12: **PAT-03 — dead input.** `dependency-group` is declared and consumed at line 39, but no caller passes it. Either wire the callers that should use it or drop the input — an unused knob invites a future caller to set it expecting an effect the defaults already decided.

KonstantinMirin (changes requested):
## Review: Request changes

Nothing here blocks the pipeline as written — it works today. But several items will quietly degrade future CI quality if they land as-is, and they are far cheaper to fix now, while the code is fresh, than after the next ten PRs build on top of them. This is an *authoritative pipeline* PR; its whole job is to make the next change unable to silently regress CI. So the bar isn't "does it work today" — it's "can a future change break this without a guard noticing." Everything below is a vector for exactly that.

### What's right (the load-bearing part)

The headline claim holds. The 18 frozen required-check names match exactly across three sources — the job `name:` fields in `ci.yml`, `REQUIRED_RENDERED_CHECKS` in `test_architecture_required_ci_checks_frozen.py`, and the table in `docs/development/ci-pipeline.md`. The `summary` job's validate step inspects all 13 upstream jobs, so the pipeline gates correctly today. The composite-action extraction (`_setup-env`, `_pytest`, `_postgres`) is real — no inline `uv run pytest` survives. That contract is the thing this PR most needed to get right, and it's right.

The findings group into four patterns. The inline comments are **illustrations** of each — not the whole work queue. Fix each at the pattern level, not just the cited site.

### PAT-01 — Operational logic lives in the test tree, and CI scripts import *up* into it

The new CI scripts pull their logic out of `tests/`:
- `scripts/ci/migration_roundtrip.py` (run in CI at `ci.yml:407`) imports the migration-graph resolver from `tests/unit/_migration_helpers.py`
- `scripts/capture-rendered-names.sh` imports CI-YAML parsing from `tests/unit/workflow_helpers.py`
- and because that test-tree home isn't importable from the pre-commit hook, the migration-completeness AST predicate is duplicated into `.pre-commit-hooks/check_migration_completeness.py` and kept in sync by hand.

The dependency arrow points backwards: operational code must never depend on the test tree. Both helpers are operational logic with no test-framework dependency.

**Close it at the pattern level:** move both helper modules to an operational home (`scripts/ci/` or `src/core/database/`). Tests, CI scripts, and the hook then all import *down* from one source, and the hand-synced duplication collapses to a single definition.

### PAT-02 — Guards that assert over a slice of the invariant they name

The frozen-names guard does this right: exact set-equality over all 18 checks. It's the template. The others don't follow it:
- `test_summary_gates_bdd_and_e2e` checks 2 of the 13 jobs in `summary.needs`. I removed `migration-roundtrip` (this PR's own new gate) from both `summary.needs` and the validate step, and the test stayed green.
- `migration_roundtrip.py::main()` — the actual upgrade→downgrade→upgrade — has no test. Only the pure AST helper is covered, and that test asserts the helper returns its own arbitrary "first parent" choice, not that alembic accepts the target.

A guard that protects 2 of 13 sites is worse than no guard: it reads as coverage while the thing it names breaks at the other 11.

**One question that decides severity:** does branch protection require all 18 named checks individually, or only `summary`? If all 18, the summary-gate gap is defense-in-depth. If only `summary`, dropping a job from `summary.needs` is a real open door. Please confirm the branch-protection config.

**Close it at the pattern level:** every guard asserts over the full population it names — iterate all of `summary.needs` cross-checked against the job list; exercise `main()` end-to-end against a real alembic graph.

### PAT-03 — CI/local parity is asserted in this PR and drifts inside the same PR

The stated goal is "CI runs what a developer runs." Inside the same change:
- Type Check runs `uv run python -m mypy src` in CI vs `uv run mypy src/` in `make typecheck`
- `_pytest` uses snake_case inputs (`extra_args`, `with_coverage`) while sibling actions `_setup-env`/`_postgres` — added in this same PR — use kebab-case (`python-version`, `database-url`)
- the coverage gate combines unit+BDD in CI vs all five suites locally, so the `--fail-under` baseline means something different in each place

Each is a seam where CI and local diverge under maintenance — the exact failure this PR sets out to close.

**Close it at the pattern level:** CI jobs shell out to the same Make targets developers run; pick one input-naming convention for composite actions; tie `.coverage-baseline` to the subset it was calibrated against.

### PAT-04 — Production loosened to make a test pass

`port_scan_start` gained an `int(pid)` → `id(pid)` fallback "in case test suites monkeypatch process metadata." Real PIDs from `os.getpid()` are always int, so the fallback is dead in production and exists only so a `MagicMock` test input won't raise. Worse, `id()` returns a memory address — non-deterministic — contradicting the determinism the docstring and the test class name (`TestPortScanStartScattersByProcess`) promise. The test now pins the loosened contract.

**Close it at the pattern level:** the test passes a real `int`, or production raises on a non-int. Production shape is driven by the contract, never by test convenience.

### One near-term fix worth its own line

`migration-roundtrip` is the only job of 14 with no `timeout-minutes`. A hung alembic step pins the runner for GitHub's 6-hour default, and since the job is in `summary.needs`, the required check just stops reporting for hours. One line: `timeout-minutes: 15`.

---

Inline comments mark concrete instances. They're evidence for the patterns above, not a checklist to close one-by-one — please fix at the pattern level.

mkostromin-sigma:
@KonstantinMirin — Post-merge review items PAT-01…PAT-04 from PR3 are addressed in **#1424** (PR 3-3 of #1234):

- **PAT-01:** migration/workflow helpers moved to `scripts/ci/` (operational layer, not `tests/unit/`)
- **PAT-02:** full `summary.needs` guard; duplicate skip grep removed from smoke-tests job
- **PAT-03:** Type Check uses `make typecheck`; `dependency-group` input removed from `_setup-env`
- **PAT-04:** strict non-int pid contract in `port_scan_start`; `_GETPID` bound at import so unit-suite `os.getpid` mocks do not break `find_free_port`

CI green on latest push (`4ae7ce43`): all 20 `CI / …` checks + security workflows pass.

Please re-review in that context or close the outstanding thread.

mkostromin-sigma:
Follow-up for E2E setup/timeout layering: #1666

mkostromin-sigma:
E2E setup/timeout follow-up implementing the #1666 design: https://github.com/prebid/salesagent/pull/1669 (closes #1667).

PR #1379, opened 2026-06-02
feat(ci): parallel BDD CI shards with merged coverage (stacked on PR3)

mkostromin-sigma:
Closes #1234.

## Summary

Stacked follow-up to **PR #1372** (merged as `68fd9066`). Parallelizes **BDD tests only** via **2 greedy scenario-count shards** while keeping branch protection stable.

| Suite | Before (PR3 on main) | After (this PR) |
|-------|----------------------|-----------------|
| BDD Tests | 1 job, all `tests/bdd/test_*.py` sequential | **2 parallel shards** + aggregate **`CI / BDD Tests`** (status proxy) |

**Branch protection:** grows from **18 → 20** rendered names (+2 `CI / BDD Tests (Shard N/2)`). Aggregate `CI / BDD Tests` gates shard pass/fail only; **Coverage** job combines unit + BDD shard artifacts and enforces `.coverage-baseline`.

**Head:** `2ff10d1c` · **CI:** [green — run 27373922220](https://github.com/prebid/salesagent/actions/runs/27373922220) · **Diff:** 13 files / ~510 insertions vs `main`

## How it works

- **`bdd-tests-shard`** — matrix `BDD Tests (Shard 1/2)` … `(Shard 2/2)` via `${{ strategy.job-total }}`; disjoint subsets via `scripts/ci/shard_split.py` greedy assignment.
- **`bdd-tests`** — aggregate status proxy only (fails if any shard fails).
- **`coverage`** — per-shard artifact gate (`SHARD_COUNTS['bdd']`), then `coverage combine`, `--fail-under`.
- **Structural guards** — partition via pytest collect-only; frozen check guard (**20 names**).

## Shard layout

13 BDD test files, greedy-balanced by scenario count:

| Shard | Files | Scenarios |
|-------|-------|-----------|
| 1/2 | 5 | ~479 |
| 2/2 | 8 | ~441 |

### ChrisHuie (items 1–5) — `956027ba`, `f35ce7b4`

- [x] Fail shard job on `shard_paths.py` error / empty output
- [x] Guard `shard_count > len(files)`
- [x] Partition guard anchored to pytest collect-only
- [x] Scenario assert `>= 1`; remove dead round-robin path
- [x] Local shard repro + load-metric note in `docs/development/ci-pipeline.md`

### KonstantinMirin (items 1–6 + follow-up) — `f7b06163`, `2ff10d1c`

- [x] Coverage gate requires **all** BDD shard artifacts (no partial combine)
- [x] Shard denominator bound to `strategy.job-total`; coverage reads `SHARD_COUNTS`
- [x] Migration roundtrip asserts DB revision **heads** after upgrade / downgrade / upgrade (`2ff10d1c` fixes merge-head multi-tip downgrade)
- [x] Every job declares `permissions` + `timeout-minutes`
- [x] DRY `iter_migration_trees()`; `parse_function` supports `AsyncFunctionDef`
- [x] Stack-aware collect-only partition guard; separate `Scenario` vs `Scenario Outline` regexes

## Test plan

- [x] Arch guards pass locally (`test_architecture_required_ci_checks_frozen`, `test_architecture_ci_bdd_shard_manifest`, `test_architecture_ci_suite_coverage`, `test_architecture_single_migration_head`)
- [x] All **20** branch-protection checks green on [CI run 27373922220 (tip `2ff10d1c`; re-run if needed after stack updates)](https://github.com/prebid/salesagent/actions/runs/27373922220)

Depends on: ~~#1372~~ (merged)
Part of: #1234
Stacked: #1424 (PR3-3, `4c4b2913`) → #1390 (PR4, `ce20bdbd`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,8 +316,12 @@ jobs:
           paths: tests/admin/
           extra_args: -m "not requires_server and not slow"
 
-  bdd-tests:
-    name: BDD Tests
+  bdd-tests-shard:
+    name: BDD Tests (Shard ${{ matrix.shard }}/2)
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2]
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -347,30 +351,53 @@ jobs:
       - uses: ./.github/actions/_postgres
         with:
           database-url: ${{ env.DATABASE_URL }}
+      - name: Resolve BDD shard paths
+        id: bdd_shard
+        run: |
+          mapfile -t TEST_FILES < <(uv run python scripts/ci/shard_paths.py bdd ${{ matrix.shard }})
+          joined="${TEST_FILES[*]}"
+          echo "paths=${joined}" >> "$GITHUB_OUTPUT"
       - uses: ./.github/actions/_pytest
         env:
-          # Separate data file so the artifact cannot collide with unit's .coverage.
-          COVERAGE_FILE: .coverage.bdd
+          # Separate data file per shard so artifacts cannot collide with unit's .coverage.
+          COVERAGE_FILE: .coverage.bdd-${{ matrix.shard }}
         with:
-          paths: tests/bdd/
-          coverage_xml: coverage-bdd.xml
+          paths: ${{ steps.bdd_shard.outputs.paths }}
+          coverage_xml: coverage-bdd-${{ matrix.shard }}.xml
           extra_args: -n auto --dist=loadfile --timeout=60
       - name: Combine xdist coverage fragments
         run: |
           shopt -s nullglob
-          fragments=(.coverage.bdd.*)
+          fragments=(.coverage.bdd-${{ matrix.shard }}.*)
           if [ "${#fragments[@]}" -gt 0 ]; then
-            uv run coverage combine --data-file=.coverage.bdd .coverage.bdd "${fragments[@]}"
+            uv run coverage combine --data-file=.coverage.bdd-${{ matrix.shard }} \
+              .coverage.bdd-${{ matrix.shard }} "${fragments[@]}"
           fi
       - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         if: always()
         with:
-          name: bdd-coverage
+          name: bdd-shard-${{ matrix.shard }}-coverage
           path: |
-            .coverage.bdd
-            coverage-bdd.xml
+            .coverage.bdd-${{ matrix.shard }}
+            coverage-bdd-${{ matrix.shard }}.xml
           include-hidden-files: true
           if-no-files-found: error
+
+  bdd-tests:
+    name: BDD Tests
+    runs-on: ubuntu-latest
+    needs: [bdd-tests-shard]
+    if: always()
+    permissions:
+      contents: read
+    steps:
+      - name: Aggregate BDD shards
+        run: |
+          if [ "${{ needs.bdd-tests-shard.result }}" != "success" ]; then
+            echo "One or more BDD shards failed."
+            exit 1
+          fi
+          echo "BDD shards passed."
 
   migration-roundtrip:
     name: Migration Roundtrip
@@ -409,8 +436,8 @@ jobs:
   coverage:
     name: Coverage
     runs-on: ubuntu-latest
-    needs: [unit-tests, bdd-tests]
-    if: ${{ needs.unit-tests.result == 'success' && needs.bdd-tests.result == 'success' }}
+    needs: [unit-tests, bdd-tests-shard]
+    if: ${{ needs.unit-tests.result == 'success' && needs.bdd-tests-shard.result == 'success' }}
     permissions:
       contents: read
     timeout-minutes: 10
@@ -427,18 +454,22 @@ jobs:
           name: unit-coverage
       - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
         with:
-          name: bdd-coverage
+          pattern: bdd-shard-*-coverage
+          merge-multiple: true
+          path: bdd-coverage-shards
       - name: Coverage gate
         run: |
+          shopt -s nullglob
+          mapfile -t bdd_data_files < <(find bdd-coverage-shards -name '.coverage.bdd-*' -type f | sort)
+          if [ "${#bdd_data_files[@]}" -eq 0 ]; then
+            echo "No BDD coverage data files downloaded."
+            exit 1
+          fi
           if [ ! -f .coverage ]; then
             echo "Unit coverage data file missing."
             exit 1
           fi
-          if [ ! -f .coverage.bdd ]; then
-            echo "BDD coverage data file missing."
-            exit 1
-          fi
-          uv run coverage combine --data-file=.coverage.combined .coverage .coverage.bdd
+          uv run coverage combine --data-file=.coverage.combined .coverage "${bdd_data_files[@]}"
           ./scripts/ci/coverage_report.sh --data-file .coverage.combined
 
   summary:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -354,7 +354,15 @@ jobs:
       - name: Resolve BDD shard paths
         id: bdd_shard
         run: |
-          mapfile -t TEST_FILES < <(uv run python scripts/ci/shard_paths.py bdd ${{ matrix.shard }})
+          set -euo pipefail
+          shard_list=$(mktemp)
+          trap 'rm -f "$shard_list"' EXIT
+          uv run python scripts/ci/shard_paths.py bdd ${{ matrix.shard }} > "$shard_list"
+          mapfile -t TEST_FILES < "$shard_list"
+          if [ "${#TEST_FILES[@]}" -eq 0 ]; then
+            echo "::error::shard_paths.py produced no files for bdd shard ${{ matrix.shard }}"
+            exit 1
+          fi
           joined="${TEST_FILES[*]}"
           echo "paths=${joined}" >> "$GITHUB_OUTPUT"
       - uses: ./.github/actions/_pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -317,7 +317,7 @@ jobs:
           extra_args: -m "not requires_server and not slow"
 
   bdd-tests-shard:
-    name: BDD Tests (Shard ${{ matrix.shard }}/2)
+    name: BDD Tests (Shard ${{ matrix.shard }}/${{ strategy.job-total }})
     strategy:
       fail-fast: false
       matrix:
@@ -398,6 +398,7 @@ jobs:
     if: always()
     permissions:
       contents: read
+    timeout-minutes: 5
     steps:
       - name: Aggregate BDD shards
         run: |
@@ -412,6 +413,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    timeout-minutes: 15
     services:
       postgres:
         image: postgres:15
@@ -467,10 +469,19 @@ jobs:
           path: bdd-coverage-shards
       - name: Coverage gate
         run: |
-          shopt -s nullglob
-          mapfile -t bdd_data_files < <(find bdd-coverage-shards -name '.coverage.bdd-*' -type f | sort)
-          if [ "${#bdd_data_files[@]}" -eq 0 ]; then
-            echo "No BDD coverage data files downloaded."
+          set -euo pipefail
+          expected_bdd_shards="$(uv run python -c "from scripts.ci.shard_split import SHARD_COUNTS; print(SHARD_COUNTS['bdd'])")"
+          bdd_data_files=()
+          for shard in $(seq 1 "${expected_bdd_shards}"); do
+            found="$(find bdd-coverage-shards -name ".coverage.bdd-${shard}" -type f | head -1)"
+            if [ -z "${found}" ]; then
+              echo "::error::Missing BDD coverage artifact for shard ${shard} (.coverage.bdd-${shard})"
+              exit 1
+            fi
+            bdd_data_files+=("${found}")
+          done
+          if [ "${#bdd_data_files[@]}" -ne "${expected_bdd_shards}" ]; then
+            echo "::error::Expected ${expected_bdd_shards} BDD coverage files, found ${#bdd_data_files[@]}"
             exit 1
           fi
           if [ ! -f .coverage ]; then
@@ -484,6 +495,8 @@ jobs:
     name: Summary
     runs-on: ubuntu-latest
     if: always()
+    permissions:
+      contents: read
     timeout-minutes: 5
     needs:
       [
--- a/docs/development/ci-pipeline.md
+++ b/docs/development/ci-pipeline.md
@@ -63,6 +63,16 @@ status-only. Shard jobs route pytest through the shared `_pytest` composite.
 | 1/2 | 5 | 479 |
 | 2/2 | 8 | 441 |
 
+Reproduce a CI shard locally:
+
+```bash
+uv run pytest $(uv run python scripts/ci/shard_paths.py bdd 1)
+```
+
+`make test-bdd` (full `tests/bdd/`) runs the same scenarios as both shards combined.
+
+**Load metric note:** greedy assignment balances Gherkin `Scenario` line counts (~920 total). Collected pytest items are much higher (~9×) because each scenario expands by `Examples:` rows and transport parametrization. Wall-clock is dominated by slow DB-heavy scenarios, not test volume — the reason 2 shards was chosen over 3–4.
+
 Structural guard: `tests/unit/test_architecture_ci_bdd_shard_manifest.py`.
 
 ## Integration Test Shards
--- a/docs/development/ci-pipeline.md
+++ b/docs/development/ci-pipeline.md
@@ -18,11 +18,14 @@ git rebase upstream/main
 git diff upstream/main --stat   # target: ~20-30 files, no alembic/versions/ changes
 ```
 
-## Required checks (18 rendered names)
+## Required checks (20 rendered names after PR #1379)
 
 Run `./scripts/capture-rendered-names.sh` after any job rename. The frozen guard
 [`tests/unit/test_architecture_required_ci_checks_frozen.py`](../../tests/unit/test_architecture_required_ci_checks_frozen.py)
 must stay in sync.
+
+PR #1372 (PR3) establishes **18** checks. PR #1379 adds **2 BDD shard** checks
+plus the aggregate `CI / BDD Tests` status proxy (replacing PR3's single BDD job).
 
 | Rendered check name | Job |
 |---------------------|-----|
@@ -40,10 +43,27 @@ must stay in sync.
 | `CI / Integration (other)` | integration shard |
 | `CI / E2E Tests` | full stack |
 | `CI / Admin UI Tests` | admin suite |
-| `CI / BDD Tests` | BDD suite (single job in PR3; sharding is PR #1379) |
+| `CI / BDD Tests (Shard 1/2)` | BDD greedy scenario shard |
+| `CI / BDD Tests (Shard 2/2)` | BDD greedy scenario shard |
+| `CI / BDD Tests` | aggregate status proxy (shard pass/fail) |
 | `CI / Migration Roundtrip` | alembic upgrade |
 | `CI / Coverage` | `--fail-under` from `.coverage-baseline` using unit + BDD artifacts |
 | `CI / Summary` | aggregation gate |
+
+## BDD Test Shards
+
+BDD tests (`tests/bdd/test_*.py`, 13 files) run in **2 parallel shards** via
+greedy min-load assignment by Gherkin scenario count (`scripts/ci/shard_split.py`).
+Each shard uploads coverage; the **Coverage** job combines unit + BDD shard
+artifacts and enforces `.coverage-baseline`. The aggregate `BDD Tests` job is
+status-only. Shard jobs route pytest through the shared `_pytest` composite.
+
+| Shard | Files | Scenarios (approx.) |
+|-------|------:|--------------------:|
+| 1/2 | 5 | 479 |
+| 2/2 | 8 | 441 |
+
+Structural guard: `tests/unit/test_architecture_ci_bdd_shard_manifest.py`.
 
 ## Integration Test Shards
 
@@ -139,7 +159,7 @@ required (or stay blocked on stale `Test Suite / …` names).
    - Add all 18 checks prefixed with `CI /` from the script output.
 4. Keep non-CI required checks unchanged (`check-pr-title`, CodeQL, `security.yml` jobs, etc.).
 5. Merge PR #1372 only after a test PR shows all 18 `CI / …` checks as required and green.
-6. For BDD parallel sharding (PR #1379): update branch protection again when shard check names are added — do **not** combine with PR3.
+6. After PR #1379 (BDD sharding): add the 2 `CI / BDD Tests (Shard …/2)` required checks and keep `CI / BDD Tests` as the aggregate proxy — **separate** branch-protection update from PR3.
 
 ### Local vs CI quality targets
 
--- a/scripts/ci/migration_roundtrip.py
+++ b/scripts/ci/migration_roundtrip.py
@@ -15,16 +15,20 @@ from alembic.runtime.migration import MigrationContext
 from sqlalchemy import create_engine
 
 from alembic import command
-from tests.unit._migration_helpers import get_migration_heads, resolve_roundtrip_downgrade_target
+from tests.unit._migration_helpers import (
+    expected_heads_after_roundtrip_downgrade,
+    get_migration_heads,
+    resolve_roundtrip_downgrade_target,
+)
 
 
-def _assert_revision(cfg: Config, expected: str, label: str) -> None:
+def _assert_heads(cfg: Config, expected: set[str], label: str) -> None:
     engine = create_engine(cfg.get_main_option("sqlalchemy.url"))
     with engine.connect() as conn:
         context = MigrationContext.configure(conn)
-        current = context.get_current_revision()
+        current = set(context.get_current_heads())
     if current != expected:
-        msg = f"{label}: expected revision {expected}, got {current}"
+        msg = f"{label}: expected heads {sorted(expected)}, got {sorted(current)}"
         raise RuntimeError(msg)
 
 
@@ -34,16 +38,17 @@ def main() -> int:
 
     print("Running alembic upgrade head...")
     command.upgrade(cfg, "head")
-    _assert_revision(cfg, head_revision, "after initial upgrade")
+    _assert_heads(cfg, {head_revision}, "after initial upgrade")
 
     target = resolve_roundtrip_downgrade_target(head_revision)
+    expected_after_downgrade = expected_heads_after_roundtrip_downgrade(head_revision)
     print(f"Running alembic downgrade {target}...")
     command.downgrade(cfg, target)
-    _assert_revision(cfg, target, "after downgrade")
+    _assert_heads(cfg, expected_after_downgrade, "after downgrade")
 
     print("Running alembic upgrade head...")
     command.upgrade(cfg, "head")
-    _assert_revision(cfg, head_revision, "after final upgrade")
+    _assert_heads(cfg, {head_revision}, "after final upgrade")
 
     print("Migration roundtrip completed successfully.")
     return 0
--- a/scripts/ci/migration_roundtrip.py
+++ b/scripts/ci/migration_roundtrip.py
@@ -11,26 +11,39 @@ if str(_REPO_ROOT) not in sys.path:
     sys.path.insert(0, str(_REPO_ROOT))
 
 from alembic.config import Config
+from alembic.runtime.migration import MigrationContext
+from sqlalchemy import create_engine
 
 from alembic import command
-from tests.unit._migration_helpers import resolve_roundtrip_downgrade_target
+from tests.unit._migration_helpers import get_migration_heads, resolve_roundtrip_downgrade_target
+
+
+def _assert_revision(cfg: Config, expected: str, label: str) -> None:
+    engine = create_engine(cfg.get_main_option("sqlalchemy.url"))
+    with engine.connect() as conn:
+        context = MigrationContext.configure(conn)
+        current = context.get_current_revision()
+    if current != expected:
+        msg = f"{label}: expected revision {expected}, got {current}"
+        raise RuntimeError(msg)
 
 
 def main() -> int:
     cfg = Config(str(_REPO_ROOT / "alembic.ini"))
+    head_revision = next(iter(get_migration_heads()))
 
     print("Running alembic upgrade head...")
     command.upgrade(cfg, "head")
-    command.current(cfg)
+    _assert_revision(cfg, head_revision, "after initial upgrade")
 
-    target = resolve_roundtrip_downgrade_target()
+    target = resolve_roundtrip_downgrade_target(head_revision)
     print(f"Running alembic downgrade {target}...")
     command.downgrade(cfg, target)
-    command.current(cfg)
+    _assert_revision(cfg, target, "after downgrade")
 
     print("Running alembic upgrade head...")
     command.upgrade(cfg, "head")
-    command.current(cfg)
+    _assert_revision(cfg, head_revision, "after final upgrade")
 
     print("Migration roundtrip completed successfully.")
     return 0
--- a/scripts/ci/shard_paths.py
+++ b/scripts/ci/shard_paths.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+"""Print test paths for a CI BDD shard (one path per line)."""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from scripts.ci.shard_split import SHARD_COUNTS, paths_for_shard  # noqa: E402
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("suite", choices=sorted(SHARD_COUNTS))
+    parser.add_argument("shard", type=int, help="1-based shard index")
+    args = parser.parse_args()
+    try:
+        paths = paths_for_shard(args.suite, args.shard)
+    except ValueError as exc:
+        print(exc, file=sys.stderr)
+        return 1
+    if not paths:
+        print(f"No tests assigned to {args.suite} shard {args.shard}", file=sys.stderr)
+        return 1
+    print("\n".join(paths))
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/scripts/ci/shard_split.py
+++ b/scripts/ci/shard_split.py
@@ -1,0 +1,82 @@
+"""Deterministic CI test sharding for BDD parallel jobs.
+
+BDD files are assigned one at a time (sorted path order) to the shard with the
+lowest current scenario count. Ties go to the lowest shard index. Scenario counts
+are read from each test file's ``scenarios("features/....feature")`` binding.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+
+_SCENARIO_LINE = re.compile(r"^\s*Scenario", re.MULTILINE)
+_SCENARIOS_CALL = re.compile(r"""scenarios\s*\(\s*["'](features/[^"']+)["']""")
+
+# Keep in sync with strategy.matrix.shard in .github/workflows/ci.yml (bdd-tests-shard).
+SHARD_COUNTS: dict[str, int] = {
+    "bdd": 2,
+}
+
+SUITE_GLOBS: dict[str, str] = {
+    "bdd": "tests/bdd/test_*.py",
+}
+
+
+def list_suite_files(suite: str, repo_root: Path | None = None) -> list[str]:
+    if suite not in SUITE_GLOBS:
+        raise KeyError(f"Unknown suite {suite!r}")
+    root = repo_root or _REPO_ROOT
+    return sorted(str(p.relative_to(root)) for p in root.glob(SUITE_GLOBS[suite]))
+
+
+def bdd_scenario_count(test_path: str, repo_root: Path | None = None) -> int:
+    """Return the number of Gherkin scenarios bound by a BDD test module."""
+    root = repo_root or _REPO_ROOT
+    text = (root / test_path).read_text(encoding="utf-8")
+    match = _SCENARIOS_CALL.search(text)
+    if match is None:
+        raise ValueError(f"No scenarios() feature binding found in {test_path}")
+    feature_path = root / "tests/bdd" / match.group(1)
+    if not feature_path.is_file():
+        raise ValueError(f"Feature file not found for {test_path}: {feature_path}")
+    return len(_SCENARIO_LINE.findall(feature_path.read_text(encoding="utf-8")))
+
+
+def _assign_round_robin(files: list[str], shard_count: int) -> dict[int, list[str]]:
+    buckets: dict[int, list[str]] = {i: [] for i in range(1, shard_count + 1)}
+    for index, path in enumerate(files):
+        buckets[(index % shard_count) + 1].append(path)
+    return buckets
+
+
+def _assign_greedy_by_scenario_count(
+    files: list[str],
+    shard_count: int,
+    repo_root: Path,
+) -> dict[int, list[str]]:
+    loads: dict[int, int] = dict.fromkeys(range(1, shard_count + 1), 0)
+    buckets: dict[int, list[str]] = {i: [] for i in range(1, shard_count + 1)}
+    for path in files:
+        shard = min(range(1, shard_count + 1), key=lambda index: (loads[index], index))
+        buckets[shard].append(path)
+        loads[shard] += bdd_scenario_count(path, repo_root=repo_root)
+    return buckets
+
+
+def assign_files_to_shards(suite: str, repo_root: Path | None = None) -> dict[int, list[str]]:
+    root = repo_root or _REPO_ROOT
+    files = list_suite_files(suite, repo_root=root)
+    shard_count = SHARD_COUNTS[suite]
+    if suite == "bdd":
+        return _assign_greedy_by_scenario_count(files, shard_count, root)
+    return _assign_round_robin(files, shard_count)
+
+
+def paths_for_shard(suite: str, shard: int, repo_root: Path | None = None) -> list[str]:
+    buckets = assign_files_to_shards(suite, repo_root=repo_root)
+    if shard not in buckets:
+        raise ValueError(f"Shard {shard} out of range 1..{SHARD_COUNTS[suite]} for {suite}")
+    return buckets[shard]
--- a/scripts/ci/shard_split.py
+++ b/scripts/ci/shard_split.py
@@ -45,18 +45,13 @@ def bdd_scenario_count(test_path: str, repo_root: Path | None = None) -> int:
     return len(_SCENARIO_LINE.findall(feature_path.read_text(encoding="utf-8")))
 
 
-def _assign_round_robin(files: list[str], shard_count: int) -> dict[int, list[str]]:
-    buckets: dict[int, list[str]] = {i: [] for i in range(1, shard_count + 1)}
-    for index, path in enumerate(files):
-        buckets[(index % shard_count) + 1].append(path)
-    return buckets
-
-
 def _assign_greedy_by_scenario_count(
     files: list[str],
     shard_count: int,
     repo_root: Path,
 ) -> dict[int, list[str]]:
+    if shard_count > len(files):
+        raise ValueError(f"shard_count={shard_count} exceeds {len(files)} bdd files; a shard would be empty")
     loads: dict[int, int] = dict.fromkeys(range(1, shard_count + 1), 0)
     buckets: dict[int, list[str]] = {i: [] for i in range(1, shard_count + 1)}
     for path in files:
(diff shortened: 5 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,483 @@
...
+  bdd-tests-shard:
+    name: BDD Tests (Shard ${{ matrix.shard }}/3)
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write
```
KonstantinMirin: 5. `unit-tests` does the same `actions/upload-artifact@v4.6.2` with only `contents: read` and works. No step in this job (checkout / `_setup-env` / `_postgres` / pytest / upload-artifact) is documented to require `actions: write`. Drop it.

KonstantinMirin (commented):
PR delivers the parallelism — three matrix shards, deterministic round-robin assignment, a coverage-merging aggregate, and a structural guard on the partition. The shard mechanism itself is well-built. The concerns are around how the framing in the description lines up with what the guard and the docs actually say, plus a permissions discipline regression vs PR3.

1. **Three sources of truth now disagree about how many BDD checks branch protection needs.**
   - The PR description says **one** (the aggregate `CI / BDD Tests`).
   - `test_architecture_required_ci_checks_frozen.py:26-28` says **four** (the aggregate plus three shards).
   - `docs/development/ci-pipeline.md:146` says **four** and instructs the maintainer to add the three shard checks to branch protection.

   The guard and the docs agree; the description is the outlier. The mechanism behind it: `tests/unit/workflow_helpers.py rendered_ci_check_names()` expands `matrix.shard: [1, 2, 3]` into three separate rendered check names, so the workflow now exposes four BDD check names total. The guard asserts set equality against the rendered set, which is why the +3 lines had to land — they're forced by matrix expansion, not a policy choice. Either branch protection grows from 18 to 21 required checks and the PR description should say so, or `rendered_ci_check_names()` should stop expanding matrices and the +3 lines should be reverted.

2. **Permissions discipline regressed against PR3.** PR3's convention was `permissions: {}` at the workflow top, `contents: read` per-job minimum. The new shard job adds `actions: write`. The `unit-tests` job uses the same `actions/upload-artifact@v4.6.2` with only `contents: read` and works fine; the GHA docs for upload-artifact v4 don't require `actions: write`. None of the steps in the shard job (checkout, `_setup-env`, `_postgres`, pytest, upload-artifact) need the elevation. Drop it back to the default; if some future step does need it, scope it then.

3. **The aggregate job has two responsibilities that fail differently.** `bdd-tests` is both (a) a status proxy for the matrix shards (the `if: always()` + `needs.bdd-tests-shard.result` check) AND (b) a coverage merger (checkout + setup + download + `coverage combine` + `coverage xml` + upload). The two failure modes are independent: a green shard run with a corrupted artifact will fail the aggregate via the combine step, turning a BP-required check red on a passing test run. Split them — the status proxy stays the BP-required name; the coverage merge is best-effort.

4. **BDD coverage is collected and discarded, same pattern as PR3.** Shards run `--cov=src`, aggregate merges into `coverage-bdd.xml` and uploads it as `bdd-coverage`. The `Coverage` gate job from PR3 still consumes only `unit-coverage` and gates against `.coverage-baseline`. The merged BDD coverage feeds nothing. Either gate it (`Coverage` job downloads `bdd-coverage` too) or drop `--cov=src` from the shards and remove the aggregate's merge step — pick one.

Smaller note: pre-PR `scripts/` worked as a PEP 420 namespace package (no `__init__.py` anywhere). Only `scripts/ci/__init__.py` is needed to make `from scripts.ci.shard_split import ...` work. The new top-level `scripts/__init__.py` turns `scripts/` into a regular package whose siblings (`scripts/ops`, `scripts/setup`) remain namespace dirs — mixed convention. Drop it; keep `scripts/ci/__init__.py`.

For closure on what's verified clean: the round-robin assignment is deterministic and correct (5/4/4 split alphabetically); the `mapfile -t TEST_FILES < <(...)` pattern verified safe (no trailing empty argument leaks into pytest); `scripts/__init__.py` does not pull `scripts/` into pytest collection (`testpaths = tests`, `norecursedirs` covers it); workflow `name: CI` unchanged; `uv.lock` churn is entirely the two dep bumps with no hidden wholesale regen; coverage artifact names match PR3's former single-job convention; the new manifest guard correctly catches duplicate assignments via `len(assigned) > len(expected)`; CVE versions in pyproject are minimum-viable for the cited advisories.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,483 @@
...
+        with:
+          name: bdd-shard-${{ matrix.shard }}-coverage
+          path: |
+            .coverage.bdd-${{ matrix.shard }}
+            coverage-bdd-${{ matrix.shard }}.xml
+          include-hidden-files: true
+          if-no-files-found: error
+
+  bdd-tests:
+    name: BDD Tests
```
KonstantinMirin, line 395: 6. The "preserves frozen check name" framing in the PR description only works if `rendered_ci_check_names()` does NOT expand matrices. It does — verified by running it — so the workflow's rendered set has four BDD names, and the frozen guard at `test_architecture_required_ci_checks_frozen.py:26-28` was mechanically forced to add the 3 shard names. `ci-pipeline.md:146` then correctly tells the maintainer to add all 3 shard checks to BP. So the description, the guard, and the docs aren't aligned on whether BP grows from 18 to 21 or stays at 18. Resolve the contradiction one way or the other.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,483 @@
...
+      - name: Merge BDD coverage reports
+        run: |
+          shopt -s nullglob
+          mapfile -t data_files < <(find bdd-coverage-shards -name '.coverage.bdd-*' -type f)
+          if [ "${#data_files[@]}" -eq 0 ]; then
+            echo "No BDD coverage data files downloaded."
+            exit 1
+          fi
+          export COVERAGE_FILE="${{ github.workspace }}/.coverage.bdd"
+          uv run coverage combine "${data_files[@]}"
```
KonstantinMirin: 7. This is the second responsibility this job carries. A shard with a malformed `.coverage.bdd-N` will fail combine and red the BP-required `CI / BDD Tests` check on an otherwise green test run. The status-proxy contract (early `exit 1` if `needs.bdd-tests-shard.result != "success"`) doesn't need the coverage merge — split into two jobs so the BP gate only reflects test status.

.github/workflows/ci.yml
mkostromin-sigma: Agreed — will drop `actions: write`; shard job will use `contents: read` only (same as `unit-tests` + `upload-artifact`).

.github/workflows/ci.yml
mkostromin-sigma: Agreed — will align PR description with guard + docs: branch protection grows to **21** rendered names (18 from PR3 + 3 BDD shard checks). Aggregate `CI / BDD Tests` stays the status proxy; shard checks are required the same way integration shards are.

.github/workflows/ci.yml
mkostromin-sigma: Agreed — will split responsibilities: `bdd-tests` becomes status-proxy only (fail if any shard failed). Coverage merge moves to the `Coverage` job so a bad artifact cannot red the BDD BP gate on a green test run.

mkostromin-sigma:
@KonstantinMirin — thanks for the thorough review; all points are actionable. Replying inline on each thread; summary of the fix plan:

1. **BP / frozen names (18→21):** Update PR description + keep guard/docs in sync; same pattern as integration matrix expansion.
2. **Permissions:** Drop `actions: write` on `bdd-tests-shard`.
3. **Job split:** `bdd-tests` = shard status proxy only; no coverage merge there.
4. **BDD coverage:** Gate it — `Coverage` job will `coverage combine` unit + BDD shard artifacts, then `--fail-under` from `.coverage-baseline`.
5. **`scripts/__init__.py`:** Remove; keep only `scripts/ci/__init__.py`.

Rebasing onto latest PR3 (#1372) first — stacked diff had drifted. Will follow up once rebased + CI green.

mkostromin-sigma:
## Next steps to make this PR merge-ready

Stack context: PR2 (#1370) and PR3 (#1372) are rebased and review-addressed. Rebasing this branch onto latest `ci/pr3-ci-authoritative-pipeline` before code changes land.

**Phase 1 — Rebase & conflicts**
- Rebase `ci/pr3-2-ci-parallel-shards` onto `ci/pr3-ci-authoritative-pipeline`
- Resolve ~9 expected conflicts (`uv.lock`, `pyproject.toml`, `ci.yml`, AI factory files, etc.)
- Rule: PR3 wins for non-BDD files; port only BDD sharding delta

**Phase 2 — Review fixes**
- Drop `actions: write` on BDD shard job
- Split `bdd-tests`: status proxy only; move coverage merge to `Coverage` job (unit + BDD shard artifacts)
- Align frozen guard + docs to **21** rendered check names; update PR description
- Remove `scripts/__init__.py`

**Phase 3 — Local verify**
- `make quality`
- Targeted arch guard tests + `./scripts/capture-rendered-names.sh` (expect 21 names)

**Phase 4 — Push (me)**
- Commit + `git push --force-with-lease` to `origin/ci/pr3-2-ci-parallel-shards`

**Phase 5 — Post-push hygiene (agent)**
- Poll CI green → resolve review threads → undraft → follow-up comment

**After PR3 merges to main (agent):** rebase onto `upstream/main`, confirm PR-only diff is BDD sharding (~8–10 files).

**Maintainer (post-merge):** separate branch-protection update to add 3 `CI / BDD Tests (Shard …)` required checks (documented in `docs/development/ci-pipeline.md`).

mkostromin-sigma:
Ready for re-review — rebased onto PR3, review feedback addressed, CI green on `968c9d96`.

KonstantinMirin:
**Re-review status — pulled latest (`968c9d968`)**

|---|---|---|
| Pattern 1 (three sources disagree on BDD checks in BP) | ✅ Fully addressed | All three now agree on **21 rendered names**. `test_architecture_required_ci_checks_frozen.py` counts 21 `"CI / …"` entries; `docs/development/ci-pipeline.md` instructs adding the 3 shard checks; PR description rewritten ("Branch protection: grows from **18 → 21** rendered names (+3 `CI / BDD Tests (Shard N/3)`)"). The architectural choice was made explicit (option B: BP enumerates all matrix-rendered names) rather than papered over |
| Pattern 2 (`actions: write` on shard job) | ✅ Fully addressed | `grep -rn "actions: write" .github/` returns only the unrelated `ipr-agreement.yml`. Shard job is back to `contents: read` |
| Pattern 3 (aggregate dual responsibility — status proxy + coverage merge) | ✅ Fully addressed | `bdd-tests` (`ci.yml:361-375`) is now pure status proxy: `if: always()`, checks `needs.bdd-tests-shard.result`, exits 1 on failure. No checkout, no setup, no coverage work. A corrupt shard artifact can no longer red the BP-required check on a green test run |
| Pattern 4 (BDD coverage collected and discarded) | ✅ Fully addressed | `coverage` job (`ci.yml:413-450`) now `needs: [unit-tests, bdd-tests-shard]`, downloads `unit-coverage` + `bdd-shard-*-coverage`, combines into `.coverage.combined`, and gates against `.coverage-baseline`. The follow-up `968c9d968` correctly uses `--data-file=.coverage.combined` so combine doesn't read-then-overwrite `.coverage` |
| Body nit (top-level `scripts/__init__.py` mixes conventions) | ✅ Fully addressed | Removed. Only `scripts/ci/__init__.py` remains. The three existing `from scripts.X import …` callers (manifest guard, BDD traceability test, GAM tenant integration test) continue to resolve via PEP 420 namespace packages |

**Scanned for new issues**: the follow-up commit `968c9d968` is a real xdist coverage bug it caught and fixed — BDD shards run `-n auto --dist=loadfile`, which writes coverage fragments like `.coverage.bdd-N.M`. The fix combines fragments into `.coverage.bdd-${shard}` before artifact upload, and the downstream Coverage job's `combine --data-file=.coverage.combined` avoids a read/write collision on `.coverage`. No new debt.

All 31 required checks green. From my side: ready.

mkostromin-sigma:
@ChrisHuie @KonstantinMirin — PR3-2 rebased onto latest PR3 (#1372) stack, which is now stacked on latest PR2 (#1370).

**Stacking**
- Base: `ci/pr3-ci-authoritative-pipeline` at `30d876c9` (PR3 rebased onto PR2 `6ec1ff7f` today).
- PR3-2-only: **2 commits**, **8 files** (`git diff ci/pr3-ci-authoritative-pipeline..HEAD`).
- Merge order unchanged: **#1370 → #1372 → this PR**.

**Rebase**
- Replayed PR3-2 commits with `git rebase --onto ci/pr3-ci-authoritative-pipeline 82cdaea8`.
- Clean replay — no conflicts (BDD delta only touches `ci.yml`, shard scripts, guards, and `ci-pipeline.md`).

**Review status (unchanged from Jun 6 re-review on `968c9d96`)**
- 21 frozen rendered check names (18 PR3 + 3 BDD shards + aggregate `CI / BDD Tests`)
- BDD shard job: `contents: read` only (no `actions: write`)
- `bdd-tests` = status proxy only; `Coverage` job combines unit + BDD shard artifacts with `--data-file=.coverage.combined`
- xdist fragment merge in shard jobs before artifact upload
- `scripts/__init__.py` removed; `scripts/ci/__init__.py` only

**After upstream merges**
1. After PR2 → rebase PR3 onto `main`
2. After PR3 → rebase this branch onto `main` (expect ~8-file PR-only diff)
3. Maintainer: branch protection grows **18 → 21** (+3 `CI / BDD Tests (Shard N/3)`)

Will push rebased branch shortly; CI needs a fresh run. Ready for re-review when you have a moment.

mkostromin-sigma:
Rebased onto updated PR3 (stacked on PR2 `d6c84dd4`).

**Method:** cherry-picked 2 PR3-2 commits onto the new PR3 tip; resolved `test_architecture_ci_suite_coverage.py` conflict (kept BDD shard guards + summary/E2E gating tests). Local CI guard tests: 15/15 passed.

**Stack:** PR2 + PR3 + PR3-2 (2 commits) → force-pushed to `ci/pr3-2-ci-parallel-shards`.

CI should run on this push. Branch protection still needs 21 rendered check names when this merges (per prior note).

mkostromin-sigma:
@ChrisHuie — **PR3-2 review / merge note (#1379)**

CI is green. KonstantinMirin re-review (Jun 6): all findings addressed. Merge **after #1372**.

### Your action for this PR
- Review + approve + merge when #1372 is on `main`.

### After **you** merge this PR — **Chris maintainer tasks**
**Separate branch-protection update** (documented in `docs/development/ci-pipeline.md` step 6):

1. Run `./scripts/capture-rendered-names.sh` — expect **21** `CI / …` names (18 from PR3 + 3 BDD shards + aggregate `CI / BDD Tests` already in the set).
2. GitHub → Settings → Rules → `main`:
   - **Add** these **3** new required checks:
     - `CI / BDD Tests (Shard 1/3)`
     - `CI / BDD Tests (Shard 2/3)`
     - `CI / BDD Tests (Shard 3/3)`
   - Keep the existing 18 `CI / …` checks from the PR3 migration.

| Who | Other post-merge task |
|-----|----------------------|
| **Author (me)** | Rebase #1390 (PR4) onto `main` |

### Not yet (later PRs)
- **#1401 (PR6):** dependency graph + `Security / Dependency Review` required check; secret scanning + push protection; OpenSSF badge; harden-runner `block` mode follow-up (≥2 weeks post-merge)

Thanks!

mkostromin-sigma:
Rebased onto latest PR3 review fixes (`ci/pr3-ci-authoritative-pipeline`, all 11 ChrisHuie items).

**What changed in this rebase:**
- Inherits PR3: concurrency, timeouts, uv cache key, smoke skip grep, E2E/Admin coverage removal, docs, `_setup-env` / `_pytest` updates
- BDD shards (4×) now route through `_pytest` composite + `coverage_xml` (same pattern as PR3 single BDD job)
- Greedy 4-shard scenario assignment unchanged; frozen checks = **22** names (4 shards + aggregate proxy)
- Guard + docs updated for 4-shard layout

Ready for re-review alongside PR #1372.

mkostromin-sigma:
Rebased onto latest PR3 (`b9581dc4` — includes `coverage_report.sh` + graph-agnostic merge-head test).

**Coverage job:** combines unit + 4 BDD shard artifacts, gates via shared `./scripts/ci/coverage_report.sh` (same as PR3/tox).

Guard tests: 17/17 passed locally. CI re-running on force-push.

mkostromin-sigma:
**Experiment: 2 BDD shards** (`a37b5de1`) — comparing CI wall-clock vs prior 4-shard runs.

| | 4 shards (previous) | 2 shards (this push) |
|--|---------------------|----------------------|
| Jobs | Shard 1–4/4 | Shard 1–2/2 |
| Frozen checks | 22 | **20** |
| Max scenario load | ~256 | **~479** |
| CI wall-clock proxy | max(shard times) | max(~479s work, ~441s work) in **2** jobs |

Greedy assignment unchanged; only `SHARD_COUNTS["bdd"]` 4→2.

**What to compare in Actions:** slowest of `CI / BDD Tests (Shard 1/2)` vs `Shard 2/2` vs previous max among Shard 1–4/4. Fewer parallel jobs but heavier shards — net effect TBD.

Revert to 4 if wall-clock regresses.

mkostromin-sigma:
@ChrisHuie — PR3-2 is rebased onto merged #1372 and ready for review + merge.

**What changed since last push**
- Rebased onto latest `main` (#1372 merged as `68fd9066`) — clean single commit (`d8e27629`), **8 files** only
- Conflicts resolved (stacked PR3 history dropped; delta reapplied on top of main)
- **2 parallel BDD shards** (greedy scenario-count assignment) after wall-clock experiment vs 4 shards
- Updated PR description to match (20 frozen check names, not 21)

**Architecture (unchanged intent from prior review)**
- `bdd-tests-shard` matrix: `BDD Tests (Shard 1/2)` + `(Shard 2/2)` via `scripts/ci/shard_split.py`
- `bdd-tests` = aggregate status proxy only
- `coverage` = combines unit + BDD shard artifacts, `--fail-under` from `.coverage-baseline`
- Guards: `test_architecture_ci_bdd_shard_manifest.py` + frozen checks (20 names)

**Shard balance**
| Shard | Files | Scenarios |
|-------|-------|-----------|
| 1/2 | 5 | ~479 |
| 2/2 | 8 | ~441 |

**CI:** all 20 `CI / …` checks + Summary green on `d8e27629`.

**Branch protection note:** required checks grow 18 → 20 (+2 shard names; aggregate `CI / BDD Tests` retained).

Please review and merge when convenient — next in #1234 chain is PR4 (#1390).

ChrisHuie:
# Review — #1379 BDD shards (verified at head `d8e27629`)

The mechanism holds up. The `test.yml` → `ci.yml` BDD job becomes a 2-way matrix whose split is exhaustive and mutually exclusive (5 + 8 = 13 files, union = all, zero overlap) and — on true collected test count — almost perfectly balanced (4,457 vs 4,448, ratio 1.002). Each shard writes a distinct `.coverage.bdd-N` that the Coverage job downloads by matching artifact name and combines into `.coverage.combined` (no overwrite, no gap); the aggregate `CI / BDD Tests` is a pure `if: always()` status proxy (no checkout/coverage work, 3s); and the three earlier review fixes still hold at 2 shards (`actions: write` dropped, proxy split out, coverage merge in the Coverage job). Guards pass (17/17) and all required checks are green.

Nothing below blocks merge. Item 1 is a small hardening worth folding in; 2–5 are optional nits; the last section is informational (and explains why 2 shards beats 3–4).

## 1. Surface a `shard_paths.py` failure instead of silently running the whole suite

**Where:** `.github/workflows/ci.yml`, the `Resolve BDD shard paths` step.

**Problem:** the step has no `set -euo pipefail` and no empty-output guard. `shard_paths.py` exits 1 on an empty/out-of-range shard, or on a `tests/bdd/test_*.py` with no `scenarios()` binding — but `mapfile -t TEST_FILES < <(…)` does not propagate that exit, so `paths=` is written empty. `_pytest` then runs `args=(uv run pytest ${{ inputs.paths }})` → bare `uv run pytest`, and `pytest.ini`'s `testpaths = tests` makes the shard collect the **entire** `tests/` tree (locally: 7,234 items) under `-n auto` — a green-but-wrong run. Not reachable at the current 2-shard/13-file config; reachable the moment a shard goes empty or a non-binding BDD module is added.

**Fix:**

```diff
       - name: Resolve BDD shard paths
         id: bdd_shard
         run: |
+          set -euo pipefail
           mapfile -t TEST_FILES < <(uv run python scripts/ci/shard_paths.py bdd ${{ matrix.shard }})
+          if [ "${#TEST_FILES[@]}" -eq 0 ]; then
+            echo "::error::shard_paths.py produced no files for bdd shard ${{ matrix.shard }}"
+            exit 1
+          fi
           joined="${TEST_FILES[*]}"
           echo "paths=${joined}" >> "$GITHUB_OUTPUT"
```

(`mapfile < <(…)` won't propagate the script's exit even under `pipefail` — process substitution — so the array-length check is what actually catches it.)

## 2. Guard the greedy split against a shard count above the file count

**Where:** `scripts/ci/shard_split.py`, `_assign_greedy_by_scenario_count`.

**Problem:** if `SHARD_COUNTS["bdd"]` is ever raised above the number of BDD files, a shard gets zero files and the balance guard's `max(loads) / min(loads)` raises an opaque `ZeroDivisionError`. Can't happen at 2 shards; cheap to prevent.

```diff
 def _assign_greedy_by_scenario_count(
     files: list[str],
     shard_count: int,
     repo_root: Path,
 ) -> dict[int, list[str]]:
+    if shard_count > len(files):
+        raise ValueError(
+            f"shard_count={shard_count} exceeds {len(files)} bdd files; a shard would be empty"
+        )
     loads: dict[int, int] = dict.fromkeys(range(1, shard_count + 1), 0)
```

## 3. The partition guard is self-referential

**Where:** `tests/unit/test_architecture_ci_bdd_shard_manifest.py::test_bdd_shards_partition_suite`.

**Problem:** `expected` (`list_suite_files`) and `assigned` (`assign_files_to_shards`) derive from the **same** glob, so `set(assigned) == set(expected)` is true by construction — it proves the function permutes its own input, not that the sharded set equals what pytest actually collects. It passes today because the glob and pytest's collection coincide. To catch a future glob-vs-collection drift, anchor one side to an independent source (parse `uv run pytest tests/bdd/ --collect-only` — note pytest 9 prints a `<Module …>` tree, not `path::test` lines, so parse accordingly). Low urgency.

## 4. Two small tightenings

- **`assert … >= 0` proves nothing** (`test_bdd_shards_have_discoverable_scenario_counts`): the count is `len(...)`, always ≥ 0; the real effect is that `bdd_scenario_count` doesn't raise. Assert the invariant instead — `>= 1` (every BDD module should bind at least one scenario).
- **`_assign_round_robin` is unreachable** (`scripts/ci/shard_split.py`): only the `bdd` suite exists and it uses the greedy path; `shard_paths.py` restricts the CLI to `choices=["bdd"]`. Remove it, or add a test if a second suite is imminent.

(The `(Shard N/2)` denominator in the job name is hardcoded, but it's part of the frozen check name — leave it; the frozen-names guard already catches drift.)

## 5. Document how to run a shard locally

**Where:** `docs/development/ci-pipeline.md`, BDD shard section. `Makefile`/`tox.ini`/`run_all_tests.sh` don't know about shards, so a developer can't reproduce "what Shard 1 ran." One line closes the gap:

```
Reproduce a CI shard locally:
  uv run pytest $(uv run python scripts/ci/shard_paths.py bdd 1)
`make test-bdd` (full tests/bdd/) runs the same scenarios as both shards combined.
```

## Informational — the load metric, and why 2 shards is the right call

`shard_split.py` balances on `^Scenario` line counts (920 total). The real collected count is **8,905** (~9.7× more): each scenario expands by `Examples:` rows and by transport (`rest`/`mcp`). This undercount is currently harmless — the greedy split is near-perfect on true counts too (4,457 vs 4,448, ratio 1.002), because the expansion factor is roughly uniform across files.

The useful takeaway for future tuning: **wall-clock isn't predictable from any test count.** The two shards have essentially equal true counts yet ran 9m57s vs 12m39s (ratio 1.27) — time is dominated by a handful of slow scenarios (DB/fixtures), not test volume. That's exactly why 2 shards beats 3–4: you can't balance time by adding shards (the slow tail is unpredictable), while each shard adds fixed setup overhead. No change requested — recording it so the metric's limits are explicit if the shard count is ever revisited.

## Verification performed

At `d8e27629`: ran the three delta guards → 17 passed; `pytest tests/bdd/ --collect-only` → 8,905 tests, per-shard 4,457 / 4,448; bare `uv run pytest` → 7,234 items (the whole `tests/` tree), confirming the empty-paths consequence; reproduced the `ZeroDivisionError` at 14 shards / 13 files. Not run: the `mapfile` exit-swallow on the runner's bash (inferred from bash semantics; local bash is 3.2 without `mapfile`), and a deliberately-failing-shard run to exercise the status-proxy failure path. Branch protection grows 18 → 20 — add the two `CI / BDD Tests (Shard N/2)` names from `./scripts/capture-rendered-names.sh`.

mkostromin-sigma:
**Documentation sync (2026-06-11)** — canonical state for reviewers:

| Source | BDD shards | Frozen `CI / …` checks | Assignment |
|--------|------------|------------------------|------------|
| PR description (this PR) | **2** | **20** | greedy scenario-count |
| `scripts/ci/shard_split.py` | `SHARD_COUNTS["bdd"] = 2` | — | greedy |
| `.github/workflows/ci.yml` | matrix `[1, 2]` | — | — |
| `docs/development/ci-pipeline.md` | 2 shards, table 479/441 | 20 | greedy |
| frozen guard | Shard 1/2, 2/2 + proxy | 20 | — |

**Older comments on this PR** (3 round-robin shards, 4-shard experiment, 21/22 check names) are **historical** — superseded by the PR description + `d8e27629` rebase onto merged #1372.

Downstream #1390–#1401 rebased and PR bodies updated to **20** checks / **2** shards.

mkostromin-sigma:
@ChrisHuie — Addressed your review at `956027ba` (items 1, 2, 4, 5; item 3 deferred as low urgency).

| # | Item | Status |
|---|------|--------|
| **1** | Fail shard job when `shard_paths.py` errors / empty output | ✅ `set -euo pipefail` + temp-file capture (propagates script exit) + empty-array guard with `::error::` |
| **2** | Guard `shard_count > len(files)` | ✅ `ValueError` in `_assign_greedy_by_scenario_count` + guard test |
| **3** | Partition guard self-referential vs pytest collect | ⏸ deferred (low urgency; glob matches collection today) |
| **4** | `>= 1` scenario assert; remove dead `_assign_round_robin` | ✅ both |
| **5** | Local shard reproduction + load-metric note in `ci-pipeline.md` | ✅ |
| **info** | Wall-clock / 2-shard rationale | acknowledged — no code change |

Guard tests: **18/18** pass locally (`test_architecture_ci_bdd_shard_manifest` now 5 tests). CI re-running on push.

**Downstream:** #1390 rebased onto `956027ba` → tip `b9f3ff2b`.

mkostromin-sigma:
Follow-up: **item 3** now addressed at `f35ce7b4`.

`test_bdd_shards_partition_suite` anchors `expected` to pytest-collected BDD modules (`pytest tests/bdd/ --collect-only`, parse `<Module …>` lines) instead of the same glob as `shard_split`. Catches future glob-vs-collection drift.

All five actionable review items (1–5) are now closed. #1390 rebased → tip pending CI.

mkostromin-sigma:
@ChrisHuie — **Review closure map** (head `f35ce7b4`, CI green). All items addressed **in this PR (#1379)** — nothing deferred to #1390.

| # | Item | Where fixed | PR |
|---|------|-------------|-----|
| **1** | Fail shard job on `shard_paths.py` error / empty output | `.github/workflows/ci.yml` | **#1379** (`956027ba`) |
| **2** | Guard `shard_count > len(files)` | `scripts/ci/shard_split.py` + guard test | **#1379** (`956027ba`) |
| **3** | Partition guard anchored to pytest collect-only | `test_architecture_ci_bdd_shard_manifest.py` | **#1379** (`f35ce7b4`) — *was briefly deferred; now in this PR* |
| **4** | `>= 1` scenario assert; remove dead `_assign_round_robin` | guard + `shard_split.py` | **#1379** (`956027ba`) |
| **5** | Local shard repro + load-metric note | `docs/development/ci-pipeline.md` | **#1379** (`956027ba`) |
| **info** | Wall-clock / 2-shard rationale | acknowledged in PR description | no code change |

**#1390 (PR4):** rebased onto `f35ce7b4` for stack hygiene only (`2bc3fa9c`). It **inherits** the above; no separate Chris review follow-ups planned there for BDD sharding.

Ready for re-review / merge when you have a moment.

KonstantinMirin:
## Review: parallel BDD CI shards

Strong PR. The two ways a sharded test pipeline usually breaks silently are both genuinely closed here, and I verified them rather than taking the guard names at face value:

- **A test that runs in no shard (or two) fails the build.** `test_bdd_shards_partition_suite` runs the real splitter and asserts set-equality against `pytest --collect-only` — an independent authority, not a restated constant.
- **A job rename that orphans a branch-protection check fails the build.** The frozen-checks guard renders names from the workflow YAML with matrix expansion (20 == 20, no drift), not against a copy of itself.
- The shard-split algorithm is complete, disjoint, and deterministic (verified identical across 3 runs). The `capabilities.py` edit is comment-only and correctly in scope (it fixes a comment pointing at the `test.yml` this PR deletes).

Findings below in priority order.

### 1. Coverage-merge gate accepts a partial shard set (blocker)

`ci.yml:471-480` — the Coverage gate is:

```bash
mapfile -t bdd_data_files < <(find bdd-coverage-shards -name '.coverage.bdd-*' -type f | sort)
if [ "${#bdd_data_files[@]}" -eq 0 ]; then exit 1; fi
...
uv run coverage combine --data-file=.coverage.combined .coverage "${bdd_data_files[@]}"
```

With `merge-multiple: true`, this checks "did **any** shard's coverage arrive," not "did **all** arrive." If one of N shard artifacts is missing or misnamed, `find` returns the survivors, the `-eq 0` check passes, and `coverage combine` proceeds with fewer files — green CI with under-reported coverage, which can then pass `--fail-under` falsely. Total loss is caught (it fails loud); the partial set is the gap.

This is one instance of finding 2.

### 2. Shard count is single-sourced and guarded, but two consumers bypass the source

`SHARD_COUNTS` is the authority. The matrix (`shard: [1,2]`) is guarded against it (`test_architecture_ci_bdd_shard_manifest.py:72`) and so is the partition. But two places that *depend on the count* are free literals nothing ties back to it:

- the coverage-merge gate above (`-eq 0`, not `== SHARD_COUNTS["bdd"]`)
- the job name `BDD Tests (Shard ${{ matrix.shard }}/2)` — `ci.yml:320`, hardcoded `/2` (bump to 3 and it renders "Shard 3/2")

These are illustrations, not the work queue — the fix is to bind both consumers to `SHARD_COUNTS` and add one string-shape guard (like the others in `test_architecture_ci_suite_coverage.py`) asserting the gate requires exactly `SHARD_COUNTS["bdd"]` shard files and that the upload-name / download-pattern / find-glob prefixes agree. Change the gate from `-eq 0` to `-ne` the expected count.

### 3. Decision logic is tested; the execution that depends on it isn't

`migration_roundtrip.py`'s target resolver is well covered (`test_architecture_single_migration_head.py:44-63`, both branches), but the actual `upgrade(head) -> downgrade(target) -> upgrade(head)` cycle runs only in the live CI job and never asserts post-state. A valid-AST-but-wrong `downgrade()` passes everything, because the final `upgrade(head)` recreates the schema regardless. Make the script prove the property it claims — assert final revision == head (and ideally reflected-vs-ORM schema parity) — rather than trusting the CI job as its only executor.

### 4. Hygiene convention broken in the same file it's set in

`summary` omits the `permissions: contents: read` block the other 14 jobs carry; `bdd-tests-shard` and `migration-roundtrip` omit `timeout-minutes` (inherit the 360-min default). A one-line guard "every job declares permissions + timeout" would close this.

### 5. DRY: migration-tree parsing copy-pasted across 5 methods

`test_architecture_migration_completeness.py:87-218` copy-pastes the read/`ast.parse`/skip preamble across 5 methods — extract an `iter_migration_trees()` generator into the `_migration_helpers.py` this PR already adds.

### 6. Lower-stakes items

- Partition guard reconstructs paths from basenames (`test_architecture_ci_bdd_shard_manifest.py:23,52`), so a nested BDD file would produce a misleading failure and a basename-collision blind spot — parse the nodeid path instead.
- Scenario-balance metric counts a `Scenario Outline` as 1 (`shard_split.py:35-45`), so the balance it enforces is approximate and an unrelated feature addition can trip the 1.35 threshold.
- `parse_function` ignores `ast.AsyncFunctionDef`; `shard_paths.py` catches only `ValueError` while `migration_roundtrip.py` catches broad `Exception`; `tests/e2e/conftest.py:45-54` `id(pid)` fallback is opaque coercion to satisfy a `MagicMock`.

Net: the guards that matter most are real and the sharding logic is sound. The coverage gate is the one place green CI can hide a real loss; the rest are pattern-level tidy-ups.

mkostromin-sigma:
@KonstantinMirin Thanks for the detailed review — all six items addressed in `f7b06163`:

1. **Coverage partial-shard leak (blocker):** Coverage gate now loops `shard 1..SHARD_COUNTS['bdd']` and fails with `::error::` if any `.coverage.bdd-N` artifact is missing before `coverage combine` (no more `find … | head -1` survivor combine).

2. **Shard denominator binding:** BDD shard job name uses `${{ strategy.job-total }}` (renders as `Shard 1/2`, `Shard 2/2`); coverage gate reads `SHARD_COUNTS` from `shard_split.py`. Guard: `test_ci_bdd_shard_job_name_uses_matrix_total` + frozen-check renderer expands `strategy.job-total`.

3. **Migration roundtrip revision assertions:** `migration_roundtrip.py` asserts DB revision **heads** after initial upgrade, after downgrade, and after final upgrade (`MigrationContext.get_current_heads()` via `_assert_heads()`).

4. **Workflow hygiene:** `summary` now has `permissions: contents: read`; `migration-roundtrip` and `bdd-tests` have `timeout-minutes`; guard `test_ci_jobs_declare_permissions_and_timeout` enforces both on every job.

5. **DRY migration AST parsing:** `iter_migration_trees()` in `_migration_helpers.py`; `test_architecture_migration_completeness` uses it; `parse_function` supports `AsyncFunctionDef`.

6. **Partition guard + scenario counting:** BDD partition guard uses pytest collect-only with stack-aware `tests/bdd/…` paths (not basename-only); `shard_split` splits `Scenario` vs `Scenario Outline` regexes so outlines are not double-counted.

CI green — verified map below ([run 27373922220](https://github.com/prebid/salesagent/actions/runs/27373922220), head `2ff10d1c`).

| # | Item | Fix | Commit |
|---|------|-----|--------|
| 1 | Coverage accepts partial BDD shard set | Loop `1..SHARD_COUNTS['bdd']`; fail if any `.coverage.bdd-N` missing before combine | `f7b06163` |
| 2 | Hardcoded `/2` in shard job name | `${{ strategy.job-total }}` in job name; coverage reads `SHARD_COUNTS`; frozen-check renderer expands job-total | `f7b06163` |
| 3 | Migration roundtrip missing revision assert | `_assert_heads()` after upgrade/downgrade/upgrade | `f7b06163` |
| 4 | `summary` permissions / job timeouts | All jobs declare `permissions` + `timeout-minutes`; arch guard enforces | `f7b06163` |
| 5 | DRY migration AST parsing | `iter_migration_trees()` shared helper | `f7b06163` |
| 6 | Partition guard / Scenario Outline / AsyncFunctionDef | Stack-aware pytest collect-only; separate Scenario vs Outline regexes; `AsyncFunctionDef` in `parse_function` | `f7b06163` |
| — | **Follow-up:** roundtrip failed on merge-head downgrade | `get_current_heads()` + `expected_heads_after_roundtrip_downgrade()` (merge restores all branch tips) | `2ff10d1c` |

All 20 branch-protection checks green on that run, including Migration Roundtrip, Coverage, both BDD shards, and Summary. Ready for re-review.

Let me know if anything else should change before merge.

mkostromin-sigma:
@ChrisHuie — quick stack update (your items unchanged since `f35ce7b4`):

- Konstantin review addressed at `f7b06163` + merge-head roundtrip fix `2ff10d1c`
- **Head:** `2ff10d1c` · **CI:** [green — run 27373922220](https://github.com/prebid/salesagent/actions/runs/27373922220)
- **#1390** rebased onto this tip (`6aa7c8d5`) — inherits all BDD shard + review fixes

PR description and #1234 rollout comment updated to match. Re-review requested on both threads when you have a moment.

ChrisHuie:
## Review — BDD CI shards (verified at head `2ff10d1c`, diff vs merged PR3 `68fd9066`)

Strong PR. I re-verified the full review history across the redesign (4-shard round-robin → 2-shard greedy): **every prior reviewer item is closed in current code** — including KonstantinMirin's original four (permissions, dual-responsibility aggregate, discarded coverage, source-of-truth alignment) and both 2026-06-11 reviews. The sharding mechanism is sound, the coverage gate is now real (not collected-and-discarded), and the migration-roundtrip fix is exercised at both unit and live-DB levels.

**No blockers.** The two `SHOULD-FIX` items below are *new* observations from this pass, not unaddressed feedback — and notably neither is caught by an automated gate, so they're flagged for human judgment.

<details>
<summary><b>Prior feedback — all closed (re-verified against current code)</b></summary>

**KonstantinMirin — 2026-06-04** (re-review 2026-06-06 signed off the then-4-shard design; the 2-shard redesign re-lands all of it):
| # | Item | Closed by |
|---|------|-----------|
| 1 | Description / guard / docs disagree on BDD check count | All three now state **20** (desc "18→20", `ci-pipeline.md` "20 rendered names", `rendered_ci_check_names()` renders 20 — exact match, 0 diff) |
| 2 | `actions: write` on shard job (regresses PR3's least-privilege) | Removed — no `actions:` permission anywhere in `ci.yml`; shard job is `contents: read` |
| 3 | Aggregate `bdd-tests` carries two failure modes (proxy + merge) | Split — `bdd-tests` is status-proxy only; `test_bdd_aggregate_is_status_proxy_only` asserts `coverage combine` is absent |
| 4 | BDD coverage collected then discarded | Gated — `coverage` job downloads `bdd-shard-*-coverage`, combines into `.coverage.combined`, enforces `--fail-under` |
| note | Top-level `scripts/__init__.py` mixes package conventions | Removed — only `scripts/ci/__init__.py` remains (siblings stay PEP-420 namespaces) |

**ChrisHuie — 2026-06-11:**
| # | Item | Closed by |
|---|------|-----------|
| 1 | Surface `shard_paths.py` failure (don't silently run whole suite) | `set -euo pipefail` + temp-file redirect (propagates the script's exit, which `mapfile < <(…)` would have swallowed) + empty-array guard |
| 2 | Guard `shard_count > len(files)` | `ValueError` in `_assign_greedy_by_scenario_count` + `test_bdd_greedy_split_rejects_shard_count_above_file_count` |
| 3 | Partition guard self-referential | `expected` re-anchored to `pytest --collect-only` (confirmed it emits the `<Module>` tree here; 13 paths, independent of the splitter glob) |
| 4 | `assert >= 1`; remove dead round-robin | Both done |
| 5 | Document local shard repro | Added to `ci-pipeline.md` + load-metric note |

**KonstantinMirin — 2026-06-11:**
| # | Item | Closed by |
|---|------|-----------|
| 1 | Coverage gate accepts a partial shard set (blocker) | Gate loops `1..SHARD_COUNTS['bdd']`, fails on any missing `.coverage.bdd-N`, asserts exact count |
| 2 | Shard count single-sourced but two consumers bypass it | Job name uses `strategy.job-total`; gate reads `SHARD_COUNTS`; + `test_ci_bdd_shard_job_name_uses_matrix_total` + `test_coverage_gate_requires_all_bdd_shard_artifacts` |
| 3 | Roundtrip never asserts post-state | `_assert_heads` after each step; merge-head fix returns `set(down_revisions)`. Verified on a live DB — the current head is itself a merge (`597485e1799a` → `b4e2bffdd4f8`, `d12aae262177`), so the CI `Migration Roundtrip` job exercises this path directly |
| 4 | permissions + timeout hygiene | All jobs declare both; `test_ci_jobs_declare_permissions_and_timeout` enforces it |
| 5 | DRY migration-tree parsing | `iter_migration_trees()` extracted; the 3 iterate-all-files methods converted (the 2 allowlist-staleness methods correctly left — single-named-file domain) |
| 6 | AsyncFunctionDef / Outline regex / nodeid-path partition | All addressed |

</details>

---

### SHOULD-FIX 1 — DRY: the new head-resolver duplicates the existing one

**Where:** `tests/unit/_migration_helpers.py` — `expected_heads_after_roundtrip_downgrade` (added here) vs the pre-existing `resolve_roundtrip_downgrade_target`.

**Problem:** Same operation — resolve the head, scan migration files, find the head's `down_revisions` — differing only in return shape (`str` first-parent vs `set` of all parents). The `head_revision is None` preamble is byte-identical and the file-scan loop is identical. This re-introduces the duplication the PR's own helper extraction set out to remove. Per the repo's DRY invariant (`CLAUDE.md` → *DRY — Non-Negotiable Invariant*), this is a correctness concern, not a style preference. It isn't caught by `check_code_duplication.py` (pylint R0801 scores the file 10.00/10 — the two functions are below its similarity threshold), so it's a human-review call.

**Fix (gold standard):** one private resolver returning the parent list; both public functions become thin shapers (semantics preserved — `resolve_…` → first parent; `expected_…` → single-element or full set):

```diff
+def _downgrade_parents_for_head(head_revision: str | None = None) -> list[str]:
+    """Resolve the head revision and return its down_revisions (parent list)."""
+    if head_revision is None:
+        heads = get_migration_heads()
+        if len(heads) != 1:
+            raise ValueError(f"Expected exactly 1 migration head, found {sorted(heads)}")
+        head_revision = next(iter(heads))
+    for path in get_migration_files():
+        revision, down_revisions = extract_revision_info(path)
+        if revision != head_revision:
+            continue
+        if not down_revisions:
+            raise ValueError(f"Cannot downgrade from base revision {head_revision}")
+        return list(down_revisions)
+    raise ValueError(f"Migration file not found for head revision {head_revision}")
+
+
 def resolve_roundtrip_downgrade_target(head_revision: str | None = None) -> str:
     """Return an explicit Alembic downgrade target one step back from head. ..."""
-    if head_revision is None:
-        ...                        # preamble + scan loop (now duplicated)
-    raise ValueError(f"Migration file not found for head revision {head_revision}")
+    # Downgrading a merge head to any one parent undoes the merge and restores all tips.
+    return _downgrade_parents_for_head(head_revision)[0]


 def expected_heads_after_roundtrip_downgrade(head_revision: str | None = None) -> set[str]:
     """Return alembic_version heads expected after CI roundtrip downgrade from head. ..."""
-    if head_revision is None:
-        ...                        # preamble + scan loop (duplicated)
-    raise ValueError(f"Migration file not found for head revision {head_revision}")
+    parents = _downgrade_parents_for_head(head_revision)
+    return {parents[0]} if len(parents) == 1 else set(parents)
```

(Side benefit: `resolve_…` currently has a dead branch — `if len(down_revisions) == 1: return down_revisions[0]` followed by an identical `return down_revisions[0]` — which this collapses.)

---

### SHOULD-FIX 2 — `pytest.skip` self-disables the new merge-head tests, and evades the no-skip hook

**Where:** `tests/unit/test_architecture_single_migration_head.py` — `test_merge_head_downgrade_restores_all_branch_tips` and `test_merge_head_downgrade_target_uses_first_parent`.

**Problem:** Both tests end with `pytest.skip("No merge migration in graph.")` when no merge migration is found, while their sibling single-parent test uses `pytest.fail` for its empty case. The repo's *Test Integrity — ZERO TOLERANCE* policy bans skips, enforced by the `no-skip-tests` pre-commit hook — but that hook greps `@pytest\.mark\.skip` (the **decorator** form) only, so an imperative `pytest.skip(...)` call passes straight through. It's a dead branch today (the head is itself a merge and 45 merge migrations exist — the test PASSES, so the logic is genuinely covered), but if the graph ever lost all merge migrations, the only unit pin on this PR's new logic would silently vanish with no signal.

**Fix:** match the sibling's fail-loud pattern (one line in each test):

```diff
-        pytest.skip("No merge migration in graph.")
+        pytest.fail("No merge migration in graph — merge-head roundtrip logic is unexercised.")
```

*Low practical urgency given the migration graph would have to shed every merge migration to trigger it — but it's a one-line hardening that also closes a real blind spot in the skip guard.*

---

### NITs

- **`scripts/ci/migration_roundtrip.py` — `create_engine(cfg.get_main_option("sqlalchemy.url"))`** passes `str | None` where `str | URL` is expected (`mypy [arg-type]`). Dormant (`alembic.ini` always defines the url) and un-gated (CI runs mypy on `src/` only). Guard: `url = cfg.get_main_option("sqlalchemy.url"); if url is None: raise RuntimeError("sqlalchemy.url not configured")` before the call.
- **`scripts/ci/shard_split.py` — `bdd_scenario_count`** weights a `Scenario Outline` as 1, ignoring `Examples:` rows (the real driver of generated test count). It's a balance heuristic, mitigated by the `<= 1.35` guard (current ratio 1.086) and acknowledged in the docs — a one-line comment marking the approximation would prevent a future reader treating it as an exact count. Separately, `_SCENARIOS_CALL.search()` reads only the *first* `scenarios()` binding; dormant (all 13 modules bind one feature) and would only skew the weight, never lose a test.
- **`tests/unit/test_architecture_ci_bdd_shard_manifest.py` — `_pytest_bdd_module_paths`** rebuilds paths from an append-only stack anchored on the first `bdd`; a BDD test nested in a subpackage would reconstruct to the wrong path. Fails safe today (flat layout — the non-recursive glob and pytest collection would mismatch → loud failure, not a silent pass). Pop on dedent, or use the module's own nodeid path, if BDD tests are ever sub-foldered.
- **`scripts/ci/migration_roundtrip.py` — `next(iter(get_migration_heads()))`** skips the `len(heads) == 1` check its callee helpers enforce; on a multi-head graph it would pick an arbitrary head rather than erroring clearly. Fail-closed either way (the single-head guard runs in `unit-tests`, and `alembic upgrade head` errors on multiple heads), so cosmetic.

### Examined and not flagged
- The coverage-gate and permissions/timeout guards are YAML string-matches (presence, not runtime behavior) — acceptable; they match the existing `test_architecture_ci_*` config-contract family, and the real gate logic is exercised by CI.
- The narrow `except ValueError` in `shard_paths.py` is correct (`KeyError` is unreachable via argparse `choices`; an `OSError` would still surface loudly under `set -euo pipefail`).
- The manifest guard's `pytest --collect-only` subprocess has no import-time DB dependency (runs green offline); ~10s cost is the only consideration.

### Verification performed
- **Ran:** full `test_architecture_*` suite (241 passed); the touched guards (26 affected, all passed); `pytest tests/bdd/ --collect-only -q` (tree format, 8,905 selected, 13 modules); `rendered_ci_check_names()` (20 names, exact match, `strategy.job-total` → 2); `migration_roundtrip.py` against an ephemeral Postgres (exit 0; downgrading the current merge head restored both branch tips); `pylint R0801`, `mypy`, `ruff` on the changed Python; all CI checks confirmed green for head `2ff10d1c`. Re-verified every prior reviewer item against current code (no author filter).
- **Not run:** the live GitHub Actions workflow (shard split / coverage upload-download-combine execute only in a real run — CI is green); the full `run_all_tests.sh ci` gate (e2e/admin/integration); per-shard wall-clock; an audit of individual migration `downgrade()` destructiveness (this PR adds no new migration files).

ChrisHuie (approved):

PR #1380, opened 2026-06-02
fix: detect v3 pricing fields in product summary

### LINKED ISSUE #1246: Bug: pricing_option_has_rate misses v3 fixed_price / floor_price → every authenticated buyer gets the "anonymous" pricing message

sravan27:
Closes #1246.

## Summary
- Treat AdCP v3 `fixed_price` and `floor_price` as buyer-visible pricing in `pricing_option_has_rate`
- Keep the existing anonymous/no-`rate` behavior for price-guidance-only responses
- Add regressions for authenticated v3 fixed-price and auction-floor products so `GetProductsResponse.__str__` does not append the anonymous pricing message

This is intentionally narrower than #1260: it only addresses the buyer-visible #1246 message bug and avoids the broader v2-compat changes from that stale dirty PR.

## Verification
- `uv run pytest tests/unit/test_get_products_response_str.py -q`
- `uv run pytest tests/unit/test_get_products_response_str.py tests/unit/test_anonymous_user_pricing.py -q`
- `uv run ruff check src/core/helpers/pricing_helpers.py tests/unit/test_get_products_response_str.py`

### DIFF
--- a/src/core/helpers/pricing_helpers.py
+++ b/src/core/helpers/pricing_helpers.py
@@ -7,27 +7,33 @@ from typing import Any
 
 
 def pricing_option_has_rate(pricing_option: Any) -> bool:
-    """Check if a pricing option has a rate value.
+    """Check if a pricing option exposes buyer-visible pricing.
 
     Handles multiple formats:
-    - Dict format (from JSON/serialization): checks po["rate"]
-    - Pydantic RootModel wrapper (adcp 2.14.0+): checks po.root.rate
-    - Direct attribute access (SQLAlchemy models): checks po.rate
+    - Dict format (from JSON/serialization)
+    - Pydantic RootModel wrapper (adcp 2.14.0+)
+    - Direct attribute access (SQLAlchemy models)
+
+    AdCP v2 used ``rate``. AdCP v3 renamed that value to ``fixed_price`` for
+    fixed pricing and ``floor_price`` for auction floors, so all three fields
+    indicate that pricing is present.
 
     Args:
         pricing_option: A pricing option in any supported format
 
     Returns:
-        True if the pricing option has a non-None rate value
+        True if the pricing option has a non-None pricing value.
     """
+    price_fields = ("rate", "fixed_price", "floor_price")
+
     # Dict format (JSON/serialization)
     if isinstance(pricing_option, dict):
-        return pricing_option.get("rate") is not None
+        return any(pricing_option.get(field) is not None for field in price_fields)
 
     # Try RootModel wrapper first (adcp 2.14.0+ Pydantic models)
     root = getattr(pricing_option, "root", None)
     if root is not None:
-        return getattr(root, "rate", None) is not None
+        return any(getattr(root, field, None) is not None for field in price_fields)
 
     # Direct attribute (SQLAlchemy models or plain objects)
-    return getattr(pricing_option, "rate", None) is not None
+    return any(getattr(pricing_option, field, None) is not None for field in price_fields)
--- a/tests/unit/test_get_products_response_str.py
+++ b/tests/unit/test_get_products_response_str.py
@@ -110,6 +110,56 @@ def test_get_products_response_str_anonymous_user():
     )
 
 
+def test_get_products_response_str_v3_fixed_price_is_not_anonymous():
+    """Regression for #1246: v3 fixed_price options are buyer-visible pricing."""
+    product = Product(
+        product_id="test-v3-fixed",
+        name="Test V3 Fixed",
+        description="A test",
+        format_ids=[create_test_format_id("banner")],
+        delivery_type="guaranteed",
+        delivery_measurement={"provider": "test_provider", "notes": "Test measurement"},
+        publisher_properties=[create_test_publisher_properties_by_tag(publisher_domain="test.com")],
+        pricing_options=[
+            {
+                "pricing_option_id": "cpm_usd_fixed",
+                "pricing_model": "cpm",
+                "currency": "USD",
+                "fixed_price": 10.0,
+            }
+        ],
+    )
+
+    response = GetProductsResponse(products=[product])
+
+    assert str(response) == "Found 1 product that matches your requirements."
+
+
+def test_get_products_response_str_v3_floor_price_is_not_anonymous():
+    """Regression for #1246: v3 floor_price auction options are buyer-visible pricing."""
+    product = Product(
+        product_id="test-v3-auction",
+        name="Test V3 Auction",
+        description="A test",
+        format_ids=[create_test_format_id("banner")],
+        delivery_type="guaranteed",
+        delivery_measurement={"provider": "test_provider", "notes": "Test measurement"},
+        publisher_properties=[create_test_publisher_properties_by_tag(publisher_domain="test.com")],
+        pricing_options=[
+            {
+                "pricing_option_id": "cpm_usd_auction",
+                "pricing_model": "cpm",
+                "currency": "USD",
+                "floor_price": 5.0,
+            }
+        ],
+    )
+
+    response = GetProductsResponse(products=[product])
+
+    assert str(response) == "Found 1 product that matches your requirements."
+
+
 def test_get_products_response_model_dump_still_has_full_data():
     """Verify that model_dump() still returns full structured data."""
     product = Product(
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.12"
 resolution-markers = [
     "python_full_version >= '3.14'",


sravan27:
I have read the IPR Policy

KonstantinMirin (changes requested):
Thanks for this — the fix is correct as far as it goes, and the two regression tests do fail against the old helper, so they're real guards. I'm requesting changes on two points, though, because one of them is the core premise of the PR and the other is about testing the behavior the way this repo tests behavior. Both are addressable here.

## 1. The bug this PR fixes is still live in `create_media_buy` (blocking)

The change fixes v3 pricing detection in the product summary, but the same v2-only field read is still in the create path:

`src/core/tools/media_buy_create.py:3066-3074`
```python
cpm = 10.0  # Default
...
rate = getattr(inner_option, "rate", None)   # v2 field only
if rate:
    cpm = float(rate)
```

`pkg_product` here comes from `get_product_catalog()` (`:2712`), which produces **AdCP v3** options via `convert_pricing_option_to_adcp` (`product_conversion.py:69-75`: `rate → fixed_price`). The resulting `CpmPricingOption` has no `rate` attribute, so this read is always `None` and `cpm` stays at the `10.0` default for every product. It then feeds `impressions = total_budget / cpm * 1000` (`:3112`) and the stored package CPM (`:3111`).

Confirmed at runtime: a DB option with `rate=25.0` converts to `fixed_price=25.0`, and the read returns `None` → CPM 10.0, impressions overstated 2.5×.

This is the same #1246 disease (reading the v2 `rate` name off a v3 object), one call away. Since "detect v3 pricing fields" is the premise of this PR, I'd like it fixed here rather than deferred.

## 2. Wire the existing scenarios instead of unit-testing a proxy (blocking)

This behavior is already specified in `tests/bdd/features/`, and the systematic way to lock it down is to wire those scenarios through the harness, where each scenario runs across all four transports automatically (`tests/CLAUDE.md`: "Every BDD scenario is automatically parametrized across all 4 transports"). Current state:

- **`BR-UC-001` → `BR-RULE-004 INV-2`** ("authenticated request, full pricing retained", lines 447-451) is the spec-level statement of #1246 — but its step "every product should retain its full pricing_options" **has no step definition.** It's unwired; the invariant never executes.
- The message this PR changes ("connect through an authorized buying agent…", `schemas/product.py:291`) is **never asserted in BDD or integration** — only in the new unit test.
- **`BR-UC-002`** has pricing scenarios, but the assertion (`then_media_buy.py:100-112`) only checks that `pricing_option_id` is present — never the computed CPM/impressions, so item 1 is invisible to the suite. (Note: the XOR boundary uses `fixed_price=10.00`, which equals the buggy default — assert against a different value.)

On the unit test itself: `tests/unit/test_get_products_response_str.py` exercises one layer and a proxy — `__str__` is an internal display convenience (`schemas/product.py:272`, "conversational text without adding non-spec fields"), not a spec obligation. The obligation is `BR-RULE-004 INV-2`. The `ProductEnv` harness already exists, so grounding this in the scenario across transports is available now (full e2e over HTTP arrives with #1430, but isn't required for this).

## Non-blocking

- **Duplication** is why this bug appeared twice: "which field is the price" is written in ≥4 places with different field sets (`pricing_helpers.py:27`, `media_buy_create.py:3072`, `tools/products.py:491-498`, `schemas/_base.py:571-578`). A shared accessor/constant in `pricing_helpers.py` that every site uses would make item 1 a one-liner and stop the next copy from drifting. Optional, but it addresses the root cause.
- **Naming:** `fixed_price`/`floor_price` are the SDK/spec names and are fine; only our own `pricing_option_has_rate` lags now that it checks three fields. Optional rename + fix the stale comment at `schemas/product.py:284`.

## What done looks like

1. Fix `media_buy_create.py:3072` to read the v3 price (`fixed_price`, fall back to legacy `rate`).
2. Wire the unwired `BR-UC-001 INV-2` step and assert that an authenticated buyer with v3 pricing gets pricing presented — covering #1246 at the spec level across all four transports.
3. Strengthen the `BR-UC-002` pricing assertion to check the computed CPM/impressions, using a `fixed_price` other than `10.00`.
4. Optional: the shared-accessor cleanup and the rename.

The unit test can stay as a fast model-level guard — it just shouldn't be the coverage we rely on for this behavior. Happy to pair on wiring the two scenarios.

Issue #1391, opened 2026-06-06
chore: Migrate 91 tests from inline CreativeAsset construction to CreativeAssetFactory


KonstantinMirin:
## Context

The adcp SDK 5.7 upgrade (#1388) changed `CreativeAsset` from a flat Pydantic model to a `RootModel[CreativeAsset1 | CreativeAsset2]` discriminated union. The `CreativeAssetFactory` was updated during the upgrade to produce valid discriminated-union data, but **91 unit tests** still construct `CreativeAsset` with old-style raw dicts that fail validation.

These tests were not broken by the upgrade itself (they were already using the factory-compatible test helpers in production paths), but their test data setup bypasses the factory and constructs assets inline.

## Scope

Files with inline CreativeAsset construction (counts are approximate):

| File | Inline constructions |
|------|---------------------|
| `test_creative.py` | 43 |
| `test_sync_creatives_format_validation.py` | 8 |
| `test_creative_coverage_gaps.py` | 8 |
| `test_extract_url_from_assets.py` | 6 |
| `test_update_media_buy_behavioral.py` | 5 |
| `test_sync_creatives_assignment_reporting.py` | 5 |
| `test_creative_response_serialization.py` | 4 |
| +8 more files | 1-2 each |

## Action

Replace each inline `CreativeAsset(...)` or `CreativeAsset(**{...})` with `CreativeAssetFactory.build(...)` or `CreativeAssetFactory.create_sync(...)`. The factory already handles the discriminated union correctly.

This is mechanical work — no business logic changes.

## Related

- Parent: #1388 (adcp SDK 4.3 -> 5.7 upgrade)
- Beads: salesagent-svt8

PR #1399, opened 2026-06-08
feat: Upgrade adcp SDK 4.3.0 → 5.7.0 (AdCP spec 3.1.0-beta.3)

### LINKED ISSUE #1388: chore: Upgrade adcp SDK 4.3.0 → 5.7.0 (AdCP spec 3.0.1 → 3.1.0-beta.3)
## Goal

Upgrade the `adcp` Python SDK from `4.3.0` (AdCP spec 3.0.1) to `5.7.0` (AdCP spec 3.1.0-beta.3 — the spec our BDD feature files target). Direct upgrade: keep the deep `generated_poc` import paths (they are unambiguous), types-only (do not adopt the adcp server framework), no behaviour change beyond what the new model shapes require.

This is mechanical reconciliation work. The actionable steps are A–E; the full field-level changeset that drives them is in the **Reference** section at the bottom.

---

## A. Moved / renamed symbols — fix the import line

| # | File:line | 4.3 | 5.7 | Note |
|---|---|---|---|---|
| 1 | `src/core/tools/accounts.py:26` | `Account` (from `adcp.types.generated_poc.account.sync_accounts_request`) | `Accounts` | Cascade root — `accounts` is re-exported via `src/core/tools/__init__.py`, so this one break drives ~34 collection errors. `Accounts` and `Accounts3` are structurally identical; use `Accounts`. This is a parenthesized multi-symbol import. Update the import and any `Account`-named usage in the file. |
| 2 | `src/core/tools/signals.py:20` | `SignalId18` (from `core.signal_id`) | `SignalId5` | Field-identical (`{agent_url, id, source}`). Do **not** use `SignalId4` (`{data_provider_domain, id, source}`). |
| 3 | `src/core/tools/accounts.py:358` | `Setup` (from `account.sync_accounts_response`) | `adcp.types.Setup` | Fields `{expires_at, message, url}`. Change to `from adcp.types import Setup` and confirm the call site uses those fields. |

## B. Shape drift — reconcile field usage

4. **`Creative`** (`creative.sync_creatives_response`) — removed `account, assigned_to, assignment_errors, changes, expires_at, platform_id, preview_url, status, warnings`; added `adcp_major_version, adcp_version`. We **write** `sync_result.assigned_to` and `.assignment_errors` at `src/core/tools/creatives/_assignments.py:216,221`. Our `SyncCreativeResult` (`src/core/schemas/creative.py:352`) extends the library type. **Action:** declare every removed field we read or write as our own field on the `SyncCreativeResult` extension so it survives the SDK removal. Grep every `.assigned_to / .assignment_errors / .platform_id / .preview_url / .status / .warnings / .changes / .expires_at` access on sync-creative objects and confirm each is owned or reconciled.
5. **`Creative.action`** changed type `CreativeAction` (enum) → `CreativeAction | str`. **Silent failure risk:** `c.action == CreativeAction.created` returns `False` when the value is the bare string `"created"` — no error, just a wrong count. Normalize to the enum (or compare on `.value`) anywhere we compare `action`. Audit `SyncCreativesResponse.__str__` and any code branching on `action`.
6. **`GetProductsResponse.products`** `list[Product]` → `list[Product] | None`, **`GetSignalsResponse.signals`** `Sequence[Signal]` → `Sequence[Signal] | None`. Guard `len()` / iteration against `None` at our call sites.
7. **`CreativeAsset`** flat model → `RootModel[CreativeAsset1 | CreativeAsset2]` (discriminated by `asset_type`); asset dict values are now `list[AssetVariant]`. Reconcile any code that reads the old flat shape.
8. **`create_mcp_webhook_payload()`** return `dict[str, Any]` → `McpWebhookPayload`, and its 3rd positional arg changed meaning (`result` → `task_type`). Fix `**`-unpacking call sites and the positional arg.
9. **`GovernanceAgent`** (`core.account`) removed `categories`. We do not read `.categories` (verified) — safe; confirm with a grep before closing.
10. **Additive / non-breaking, confirm only:** `Brand` +`trademarks`, `MediaBuyFeatures` +`committed_metrics_supported`, `adcp_version`/`adcp_major_version` envelope fields across request/response types. Confirm no test asserts an exact field set on these. `SignalId` (`core.signal_id`) is unchanged — no action.

## C. Contract drift — `tests/unit/test_adcp_contract.py` (5 tests)

Reconcile our extending schemas (`src/core/schemas/*`) against the 5.7 models for:
- `TestSchemaMatchesLibrary::test_all_request_schemas_match_library`
- `test_sync_creatives_response_adcp_compliance`
- `test_list_creatives_request_adcp_compliance`
- `test_get_products_response_adcp_compliance`
- `test_task_status_mcp_integration`

## D. Schema-redefinition guard

`tests/unit/test_architecture_schema_inheritance.py` flags `GetMediaBuyDeliveryRequest.time_granularity` and `include_window_breakdown` — 5.7 now provides these (alongside `adcp_version`). Delete our redeclarations in `src/core/schemas/delivery.py` and inherit.

## E. Version pin + spec constant + docs

- `pyproject.toml`: `adcp==4.3.0` → `==5.7.0`; `uv lock` (transitive: `mcp` 1.25.0→1.27.2, adds `tldextract` — verify the lock resolves and builds).
- `tests/unit/test_adcp_spec_version.py:5`: `EXPECTED_SPEC_VERSION = "3.0.1"` → `"3.1.0-beta.3"`.
- `docs/adcp-spec-version.md`: version↔spec table is stale (says 5.x = 3.0.7; 5.7 reports 3.1.0-beta.3) — fix the table and the 3.0.1 prose.
- `CLAUDE.md`: update the `adcp==4.3.0` / spec 3.0.1 references.

## Process notes

- **Enumerate imports with an AST walk, not a line regex** — `accounts.py:26` is a parenthesized multi-line import that single-line matching misses. Enumerate every `from adcp... import ...` (including multi-line) before declaring scope complete.
- **Verify field-set identity before swapping numbered variants** — the SDK has several same-named/numbered variants (`SignalId4` vs `SignalId5`, `Accounts` vs `Accounts3`). Confirm `model_fields` match before applying each replacement; let `mypy` catch silent mismatches.
- **Keep the deep import paths.** Do not flatten collision-prone symbols (`Brand`, `Account`, etc.) to `adcp.types.X`; the deep paths are unambiguous.

## Verification

`make quality` clean, then `tests/unit/test_adcp_contract.py`, `test_adcp_spec_version.py`, and `test_architecture_schema_inheritance.py` green. Full BDD suite shows no regressions vs the current pin.

---

# Reference — full 4.3.0 → 5.7.0 field changeset

### Fields removed (per model)

| Model | Fields removed | Was required? |
|---|---|---|
| `sync_creatives_response.Creative` | `account, assigned_to, assignment_errors, changes, expires_at, platform_id, preview_url, status, warnings` | all optional |
| `sync_creatives_response.SyncCreativesResponse1` | `context, dry_run, ext, sandbox` | all optional |
| `sync_accounts_response.Account` | **entire class removed** (16 fields: `account_id, account_scope, action, billing, billing_entity, brand, credit_limit, errors, name, operator, payment_terms, rate_card, sandbox, setup, status, warnings`) | — |
| `sync_accounts_response.SyncCreativesResponse1` | `accounts, context, dry_run, ext` | **`accounts` was required** |
| `sync_accounts_request.Account` | **renamed → `Accounts`** (same fields + 2 new) | — |
| `core.signal_id.SignalId18` | **entire class removed → `SignalId5`** (same fields) | — |
| `core.account.GovernanceAgent` | `categories` | optional |

### Fields added

| Model | Fields added |
|---|---|
| `sync_creatives_response.Creative` | `adcp_major_version, adcp_version` |
| `sync_creatives_response.SyncCreativesResponse1` | `adcp_major_version, adcp_version, status` |
| `sync_accounts_request.Accounts` (was `Account`) | `notification_configs, preferred_reporting_protocol` |
| `GetMediaBuyDeliveryRequest` | `adcp_version, include_window_breakdown, time_granularity` |
| `core.media_buy_features.MediaBuyFeatures` | `committed_metrics_supported` |
| `brand.Brand` | new class with 28 fields |
| most request/response types | `adcp_version` (protocol envelope field) |

### Type changes (same field, different type)

| `Model.field` | 4.3 type | 5.7 type | Impact |
|---|---|---|---|
| `Creative.action` | `CreativeAction` (enum) | `CreativeAction \| str` | equality comparison breaks (`== CreativeAction.created` is `False` for str `"created"`) — silent |
| `GetProductsResponse.products` | `list[Product]` | `list[Product] \| None` | `len()` crashes on `None` |
| `GetSignalsResponse.signals` | `Sequence[Signal]` | `Sequence[Signal] \| None` | iteration crashes on `None` |
| `CreativeAsset` | flat Pydantic model | `RootModel[CreativeAsset1 \| CreativeAsset2]` | asset dict values must be lists with `asset_type` discriminator |
| `create_mcp_webhook_payload()` return | `dict[str, Any]` | `McpWebhookPayload` | `**` unpacking crashes |
| `create_mcp_webhook_payload()` 3rd arg | `result` | `task_type` | positional-arg meaning changed |
| response aliases (`CreateMediaBuySuccessResponse`, …) | direct class | `getattr()` dynamic resolve | mypy "invalid base class" (runtime OK) |

### Structural changes

| Change | Description |
|---|---|
| Response envelope collapse | Sync response types stripped to a minimal protocol envelope; runtime compat preserved only by `extra='allow'` on the models |
| Version metadata injection | `adcp_version` + `adcp_major_version` added to envelope types |
| `CreativeAsset` discriminated union | Asset dict values are now `list[AssetVariant]` tagged by an `asset_type` discriminator, not flat dicts |

### Why a minor spec bump renames Python types

The SDK is generated by `datamodel-code-generator`, which inlines each JSON-Schema `$ref` into a per-message local class rather than sharing one type. In `adcp.types.generated_poc`, 431 class names are defined in ≥2 modules (`Account`×37, `Error`×35, `Brand`×33), and class names carry an ordinal suffix tied to position in the inlined expansion. So a wire-additive spec change can still rename/renumber the class we import (`SignalId18` → `SignalId5`, identical fields). This is why the upgrade needs AST-based import enumeration and per-symbol field-set verification rather than a blind pin bump.

KonstantinMirin:
## Summary

Upgrades the `adcp` Python SDK from 4.3.0 (spec 3.0.1) to 5.7.0 (spec 3.1.0-beta.3 — the spec the BDD feature files target). Direct upgrade: keeps deep `generated_poc` import paths, types-only (does not adopt adcp server framework).

Closes #1388

### What changed

**SDK pin & lockfile**: `adcp==5.7.0`, `mcp` 1.25→1.27, `+tldextract`

**Import fixes (3 moved symbols)**:
- `Account` → `Accounts` (sync_accounts_request)
- `SignalId18` → `SignalId5` (field-identical)
- `Setup` → `adcp.types.Setup`

**Fields removed from SDK, declared locally** (spec 3.1 collapsed response envelopes):
- `SyncCreativeResult`: assigned_to, assignment_errors, platform_id, status, action (as str)
- `SyncCreativesResponse`: dry_run, creatives (typed as SyncCreativeResult)
- `SyncAccountsResponse`: accounts, dry_run, context, ext
- `CreateMediaBuySuccess`: account, sandbox
- `SyncResponseAccount`: new local model (was SDK type, now gone)

**Fields deleted from local schemas** (SDK now provides):
- `GetMediaBuyDeliveryRequest`: time_granularity, include_window_breakdown, include_package_daily_breakdown

**CreativeAction enum→str normalization** (adcontextprotocol/adcp-client-python#913):
SDK's `CreativeAction` is plain `Enum` (not `StrEnum`), so `"created" == CreativeAction.created` is `False`. `SyncCreativeResult.action` normalizes to string via `field_validator`. All comparisons use string literals.

**`create_mcp_webhook_payload` signature change**: 3rd arg is now `task_type`, returns `McpWebhookPayload` (not dict). Fixed 5 production call sites.

**`CreativeAsset` discriminated union**: Asset dict values must be `list[AssetVariant]` with `asset_type` tag. Added `make_image_assets()` / `make_video_assets()` / `make_text_assets()` helpers. Shared test helpers extracted to reduce duplication (95 blocks, down from 101 baseline).

**Production fix**: `_processing.py` — extract message from SDK 5.7 RootModel-wrapped assets via shared `_extract_message_from_assets()` helper.

### Upstream issues filed

- adcontextprotocol/adcp-client-python#913 — 1,410 generated enums are plain `Enum`, not `StrEnum`
- adcontextprotocol/adcp-client-python#914 — `buyer_ref` ghost field on `CreateMediaBuyResponse1`

### Follow-up

- #1391 — Migrate remaining inline `CreativeAsset` construction to `CreativeAssetFactory`

## Test plan

- [x] `make quality` green (4,733 unit tests, 0 failures)
- [x] `./run_all_tests.sh` green (8,432 tests across 6 suites, 0 failures)
- [x] Duplication baseline improved: 101 → 95
- [x] No mypy regressions (remaining are SDK dynamic-type issues, suppressed)
- [x] BDD: 0 regressions (1,477 passed, 5,372 xfailed are unimplemented 3.1 stubs)


ChrisHuie:
## Review — adcp 4.3.0 → 5.7.0 (spec 3.0.1 → 3.1.0-beta.3)

Solid, well-scoped bump. I grounded the behavioral findings in the 3.1.0-beta.3 JSON schemas + the graded conformance storyboards (not SDK codes/prose), and ran the suites locally on the head SHA. **No blockers.** The wire is schema- and storyboard-conformant on every point I checked:

- The locally re-declared "collapsed envelope" fields (`SyncCreativeResult`, `SyncCreativesResponse`, `SyncAccountsResponse`, `CreateMediaBuySuccess`, `SyncResponseAccount`) each match the 3.1 response schemas — they restore the spec shape the 5.7 code-gen dropped.
- `Product` correctly stays on the `anyOf` v1 arm (`format_ids`); the `response_schema` storyboard check passes.
- Auth `recovery: "terminal"` is conformant — the storyboards grade the error *code*, never the recovery class.
- The `get_products` inventory-profile inversion (drop non-spec fields) passes on all four transports against production.

Local runs on this head: unit 4918p/0f, changed integration 437p/0f, BDD inventory-profile 24/24 (4 transports), BDD @account UC-006 52p/0f, admin creatives 14p/0f, e2e 90p/0f; mypy/ruff/format clean; duplication 98→91.

### Suggested fold-ins (this PR)

**1. Auth-error docstrings are stale after the bump** (`src/core/exceptions.py`, ~ll.402-409/429-431; same wording in `tests/integration/test_mcp_error_envelope.py:253` and `tests/integration/test_a2a_error_responses.py:926`). They cite the 4.3 table, call `AUTH_TOKEN_INVALID` "a standard SDK code" (it's absent from 5.7 `STANDARD_ERROR_CODES` and from the 3.1 enum `AUTH_REQUIRED`/`AUTH_MISSING`/`AUTH_INVALID`), and say `terminal` "matches the installed SDK's validator" (5.7 says `correctable`). The wire value is fine — recommend correcting the rationale only:

> Recovery is intentionally `terminal`. The 3.1 conformance storyboards grade the error code, not the recovery class, so this is a free, conformant choice. (5.7's STANDARD_ERROR_CODES lists AUTH_REQUIRED as `correctable`; we deviate deliberately.)

**2. `enum_value` duplicates the existing `resolve_enum_value`.** `src/core/enum_helpers.py` adds `enum_value`, but `src/core/validation_helpers.py:18` already has `resolve_enum_value` (4 callers in `adapters/base.py`, `services/targeting_capabilities.py`), and `tests/bdd/steps/generic/then_success.py:45` now inlines the same pattern a third time. Contracts differ (`-> str` vs `-> str | None`), so worth converging deliberately on one helper rather than three forms.

**3. `task_type` fallback leaks the wire value into the internal label** (`src/core/context_manager.py` ~ll.842-876). The new validate-with-fallback rewrites `task_type_str` to `"update_media_buy"`, and that rewritten value then flows into `metadata["task_type"]` — the label keying the delivery-webhook guards + audit. No live collision today, but suggest keeping them separate:

```python
wire_task_type = task_type_str
try:
    _TaskType(wire_task_type)
except ValueError:
    wire_task_type = "update_media_buy"
# pass wire_task_type to create_mcp_webhook_payload(task_type=...) only;
# leave metadata["task_type"] = the original action label
```

**4. New `no_bare_basemodel` guard is currently uninformative** (`tests/unit/test_architecture_no_bare_basemodel.py`): it matches 0 sites and has no positive/negative self-test, and the matcher only models `ast.Name "BaseModel"` (misses `pydantic.BaseModel` and `RootModel`). A known-positive/known-negative self-test would make its green meaningful.

**5. `local_schema_imports` guard allowlist** (`tests/unit/test_architecture_local_schema_imports.py`, ~ll.48-66): 12 entries without `# FIXME(salesagent-xxxx)` at the source lines, and the in-test refs use `#1360` rather than the repo's `salesagent-xxxx` convention.

**6. `test_async_submitted_response`** (`tests/unit/test_creative.py:4687`) dropped `assert "context"/"ext" in model_fields` when adapting to required `task_id` — those fields still exist under 5.7, worth re-adding alongside the `task_id` assertions.

### Follow-up (separate PR)

**Cross-transport envelope shape**: MCP returns the raw model (`creatives/sync_wrappers.py:72`), so it ships `dry_run/context/ext: null` + wider advisory `errors[]` items than A2A/REST (`model_dump(exclude_none=True)`). Pre-existing mechanism, but the 5.7 `Error` adds `issues/source/sdk_id`, so it's worth a whole-class pass on whether SDK-default envelope fields should ship on the wire.

### NITs

- `creative_deadline` is dropped by 5.7 but still emitted (default `now+2d` from `adapters/base.py`) via `extra="allow"` — declare it typed beside `account`/`sandbox` in `CreateMediaBuySuccess` for parity.
- `Product` RootModel-union serializer warning is cosmetic (wire is correct via the nested-serializer override) but noisy — a targeted `filterwarnings` would keep a real future regression visible.
- `get_adcp_version()` is deprecated in 5.7; 3 remaining call sites (`tests/unit/test_webhook_delivery_service.py:136`, `tests/e2e/test_a2a_endpoints_working.py:80,211`) can move to `get_adcp_spec_version()`.
- `task_type` validation is guarded in `context_manager` but not at the 3 admin webhook sites (`admin/blueprints/creatives.py:254`, `operations.py:497,587`) — unreachable on current paths, and the move off `model_construct` is a net improvement; noting for symmetry.

KonstantinMirin:
@ChrisHuie thanks for the schema/storyboard-grounded review — all six fold-ins plus the four NITs are now pushed (`9f8c04fcf..99b5ac07c`). Point by point:

### Fold-ins

**#1 — Auth-error docstrings stale** ✅ `8a3f99bd3`
Rewrote the `AdCPAuthenticationError`/`AdCPAuthorizationError` rationale to your exact framing (storyboards grade the error *code*, not the recovery class; we deviate from 5.7's `correctable` deliberately). Verified `STANDARD_ERROR_CODES["AUTH_REQUIRED"].recovery == "correctable"` and that the wire `terminal` comes from the hardcoded `_default_recovery` ClassVar, never from the SDK table — so this is docs-only, no wire change. Matching comments in `test_mcp_error_envelope.py:253` and `test_a2a_error_responses.py:926` updated too.

**#2 — `enum_value` duplicates `resolve_enum_value`** ✅ `d87a4013a`
Converged on `enum_value` (the `-> str | None` superset) as the single source of truth. Deleted `resolve_enum_value`, migrated the 4 callers (`adapters/base.py`, `services/targeting_capabilities.py`) and the inlined third form in `then_success.py:45`. Handled the contract gap by guarding the `None` case at the `targeting_capabilities` call site (empty-string keys still pass through). Added `test_enum_helpers.py` locking the superset contract so a narrower copy can't silently reappear.

**#3 — `task_type` fallback leaks into the internal label** ✅ `44a921ab7`
Exactly your suggested split: a validated `wire_task_type` copy goes to the webhook payload only; `metadata["task_type"]` keeps the original action label. Extracted a shared `validate_webhook_task_type()` helper (also covers the NIT below). Regression test drives the real `_send_push_notifications` and is mutation-verified (reusing the coerced copy for the label fails it).

**#4 — `no_bare_basemodel` guard uninformative** ✅ `683cff60a`
Added known-positive/known-negative self-tests, broadened the matcher to `ast.Attribute` (`pydantic.BaseModel`) and `RootModel`, and widened scope to `src/routes/api_v1.py`. That surfaced 11 bare-`*Body` classes there; they're allowlisted with `# FIXME` markers for now, since migrating them to the project base flips REST extra-field handling and wants integration coverage — I'll file that as a follow-up to drain the allowlist to empty.

**#5 — `local_schema_imports` allowlist (`#1360` refs / missing FIXMEs)** ⏳ not in this push
This one wasn't in the batch and touches our FIXME-ref convention (GH-ref vs `salesagent-xxxx`), so I'm leaving it out pending a decision rather than guessing. Will handle separately.

**#6 — `test_async_submitted_response` dropped `context`/`ext`** ✅ `c543dc43a`
Re-added `assert "context" in model_fields` / `assert "ext" in model_fields` beside the `task_id` assertions; confirmed both fields still exist on 5.7 `SyncCreativesSubmitted`.

### NITs

- **`creative_deadline` typed** ✅ `6c204623d` — declared `creative_deadline: datetime | None = None` beside `account`/`sandbox` on `CreateMediaBuySuccess`.
- **`Product` serializer warning** ✅ `b7604b079` — added a targeted `filterwarnings` (`Pydantic serializer warnings` / `UserWarning` / `pydantic.main`) after capturing the exact text, narrow enough that a real serializer regression still surfaces; `test_warning_filters.py` still passes.
- **`get_adcp_version()` deprecated** ✅ `99b5ac07c` — migrated the 3 test sites (plus a 4th reference in `a2a_server/README.md`) to `get_adcp_spec_version()`; same `3.1.0-beta.3` return.
- **`task_type` validation at the 3 admin webhook sites** ✅ `44a921ab7` — applied the shared helper at `creatives.py`, `operations.py` (approve + reject) for symmetry.

### Follow-up (separate PR)
Agreed on the cross-transport envelope-shape pass (MCP shipping `dry_run/context/ext: null` + wider advisory `errors[]`); leaving that out of this PR as you suggested.

### Verification
Full suite on the merged head, all green: unit 4953p/0f · integration 1995p/0f · e2e 90p/0f · admin 86p/0f · bdd 1522p/0f · ui 5p/0f; mypy/ruff/format clean; duplication baseline unchanged.

KonstantinMirin:
On **#5** (`local_schema_imports` allowlist refs) — I'd push back on the `salesagent-xxxx` part, and I've made the rest of the PR consistent with that position.

This is a public, multi-contributor repo. `salesagent-xxxx` are IDs from our local [beads](https://github.com/steveyegge/beads) tracker — they live outside this repository and don't resolve for a reviewer or any future contributor reading the code. GitHub issue/PR numbers are the durable, shared identifier everyone here can actually click through. So the convention we follow for **anything committed to the repo (code comments, commit messages)** is: reference GitHub issues/PRs, never beads IDs. The `#1360` refs you flagged are GitHub PR refs, so they're already correct by that rule, not a deviation.

Two concrete things on the back of this:

1. I filed #1442 for the `api_v1` bare-BaseModel migration (the follow-up from #4), and pointed those FIXME markers at `#1442` — `d3cf7f4c7`. So every new FIXME in this PR now uses a GitHub ref, none use beads IDs.
2. Your first sub-point on #5 is fair independent of the ref style: some `local_schema_imports` allowlist entries lack a `# FIXME` marker *at the source import site* (only the test carries them). Happy to add `# FIXME(#1360)` (or a fresh tracking issue) at each source location for traceability — just with the GitHub ref, not a beads id. Want me to fold that into this PR or take it as a follow-up?

(Heads up: our `CLAUDE.md` guard section still says `# FIXME(salesagent-xxxx)`, which is what cued the beads-id form in the first place — I'll get that corrected to the GitHub-ref convention so it stops happening.)

KonstantinMirin:
Folded #5 in rather than deferring:

- `838f15d70` — corrected the FIXME-ref convention in `CLAUDE.md` + the structural-guards / patterns-reference docs to `# FIXME(#<gh-issue>)` (that stale `salesagent-xxxx` wording is what cued the beads-id form in the first place).
- `d1db21bdd` — added `# FIXME(#1360)` at all 12 `local_schema_imports` source import sites (xandr, operations, schema_helpers, capabilities, creative_formats, products, dynamic_pricing, creative_agent_registry), so each allowlisted bypass is now traceable at the source, not just in the test. No behavioral change; guards + ruff/format green.

So the only open item from your review is the cross-transport envelope pass, which stays a separate PR as you flagged.

KonstantinMirin:
@ChrisHuie please re-check and let's merge if okay

ChrisHuie:
# Review — adcp 4.3.0 → 5.7.0 (head `d1db21b`)

The re-declared collapsed-envelope fields (`SyncCreativeResult`, `SyncCreativesResponse`, `SyncAccountsResponse`, `CreateMediaBuySuccess`, `SyncResponseAccount`) match the v3.1.0-beta.3 schemas; `Product` stays on the valid v1 `format_ids` arm; the `CreativeAction` enum→str normalization and the `products`/`signals` null-guards cover the consumed surface. No blockers. Findings below — F1–F3 are recommended fold-ins; the rest are tracked follow-ups and nits. Each item notes what it interacts with, what masks it today, and the forward risk.

Grounding: conformance items (F4–F6 and the field-drift matrix) are checked against the spec version this PR pins — `adcp==5.7.0` declares AdCP `3.1.0-beta.3` (`get_adcp_spec_version()`) — using the frozen schemas at `adcontextprotocol/adcp@v3.1.0-beta.3`, not `/schemas/latest` and not the SDK Python types (whose code-gen drops the required/optional information). F1 is assessed against live `/schemas/latest`, which is what that test fetches.

## SHOULD-FIX

### F1 — Schema-alignment test is blind to the two relocated request schemas
**Where:** `tests/unit/test_pydantic_schema_alignment.py:50-51` (map), `:84-90` (allowlist), `:108,118-119` (`load_json_schema`).
**Context:** This is the only test that diffs our Pydantic models against the live AdCP schema; `make quality` skips it (networked).
**Problem:** It requests `…/media-buy/sync-creatives-request.json` and `…/media-buy/list-creatives-request.json`; 3.1 relocated both to `…/creative/…`. The `media-buy/` paths now return **HTTP 404** (the `creative/` paths return 200).
**Interacts with:** the local cache, the CI cache state, and the allowlist. `load_json_schema` reads the cache first, then `pytest.skip()`s on HTTP error. On CI, `schemas/v1/` is gitignored → no cache → 404 → both models **skip**. Locally, the stale `schemas/v1/_…media-buy_*-creatives-request_json.json` copies (no `.meta` sidecar, unlike healthy siblings) are read instead → the test **passes against the pre-3.1 shape**. The `KNOWN_SCHEMA_LIBRARY_MISMATCHES` entries for those paths (`account_id`, `include_performance`, `include_sub_assets`) describe fields 3.1 no longer defines, so they are dead suppression keyed to the stale schema.
**What masks it:** the `Schema Contract` CI check is green while never validating these two request types, and local vs CI diverge silently because the cache is gitignored. So `SyncCreativesRequest`/`ListCreativesRequest` are never checked against the real 3.1 schema by any run.
**Forward risk:** a 404 is a silent skip, so the next time the spec relocates a schema this repeats with no signal. These lines predate this PR, but this is the spec-version bump, so it's the place to correct them.
**Fix:**
```diff
-    "/schemas/latest/media-buy/sync-creatives-request.json": SyncCreativesRequest,
-    "/schemas/latest/media-buy/list-creatives-request.json": ListCreativesRequest,
+    "/schemas/latest/creative/sync-creatives-request.json": SyncCreativesRequest,
+    "/schemas/latest/creative/list-creatives-request.json": ListCreativesRequest,
```
Re-derive the allowlist for both against the live `creative/` schemas (they appear to collapse to `set()`), drop the stale cache files, and fail-not-skip on a 404:
```diff
     except (httpx.HTTPError, json.JSONDecodeError) as e:
+        if isinstance(e, httpx.HTTPStatusError) and e.response.status_code == 404:
+            raise AssertionError(f"Schema ref no longer resolves (spec moved it?): {schema_ref}") from e
         pytest.skip(f"Could not download schema {schema_ref}: {e}")
```

### F2 — `FIXME(#1360)` references a closed, unrelated PR
**Where:** `tests/unit/test_architecture_local_schema_imports.py:49-66` + 12 source sites (`xandr.py`, `operations.py`, `schema_helpers.py`, `capabilities.py`, `creative_formats.py`, `products.py`, `dynamic_pricing_service.py`, `creative_agent_registry.py`).
**Context:** These markers give each allowlisted SDK-import bypass a clickable tracking ref.
**Problem:** #1360 is CLOSED and titled "Feature/e2e test" — unrelated to SDK-type-import tracking.
**Interacts with:** the related markers and issues. #1442 (used by the `no_bare_basemodel` markers) is the correct pattern but is scoped to the api_v1 `*Body` migration only; the SDK-import-bypass work tracks under #1388, or wants its own issue.
**What masks it:** no guard or hook enforces a FIXME-format/resolvability regex, so the wrong ref can't fail the build — it degrades silently.
**Forward risk:** a contributor following the marker lands on an unrelated closed PR, which is the opposite of the traceability the `salesagent-xxxx → #issue` convention change was meant to provide.
**Fix:** repoint the 12 source markers + the allowlist comment block:
```diff
-# FIXME(#1360): xandr adapter imports SDK types directly
+# FIXME(#1388): xandr adapter imports SDK types directly
```

### F3 — `filterwarnings` suppresses all Pydantic serializer warnings, not just `Product`
**Where:** `pytest.ini:46` — `ignore:Pydantic serializer warnings:UserWarning:pydantic\.main`.
**Context:** Added to silence a noisy `Product` RootModel-union serializer warning.
**Problem:** The `message` field is `re.match`-anchored, and every `PydanticSerializationUnexpectedValue` warning shares the prefix `"Pydantic serializer warnings:"` and originates in `pydantic.main`. So the filter matches them all, not just the `Product` arm.
**Interacts with:** every model that can emit a serializer warning (`Creative`, `MediaBuy`, …) across the whole suite, not the single `get_products` case it targets.
**What masks it:** these warnings are non-fatal (pytest's default action isn't `error`), so nothing flags the over-broad scope, and `test_warning_filters.py` (unchanged here) has no assertion covering this entry.
**Forward risk:** a real serializer regression in any model is silenced suite-wide going forward.
**Fix:** scope it to the test that exercises `GetProductsResponse`/`Product` serialization instead of the project-wide ini:
```diff
# pytest.ini
-    ignore:Pydantic serializer warnings:UserWarning:pydantic\.main
```
```python
# on the get_products serialization test that emits it:
@pytest.mark.filterwarnings("ignore:Pydantic serializer warnings:UserWarning:pydantic.main")
```

### F4 — `CreateMediaBuySuccess.valid_actions`/`.context` ride on inherited `extra='allow'`
**Where:** `src/core/schemas/_base.py:204-227`; emitted at `src/core/tools/media_buy_create.py:1524-1525`.
**Context:** 5.7 dropped 10 optional fields from the parent; three were re-declared locally (`account`, `sandbox`, `creative_deadline`).
**Problem:** `valid_actions` and `context` are not declared fields, yet production sets them.
**Interacts with:** the model's base class. Its MRO is `CreateMediaBuySuccess → CreateMediaBuyResponse1 → AdcpVersionEnvelope → AdCPBaseModel` — it extends the library base, not `SalesAgentBaseModel`, so it inherits `extra='allow'` even under `ENVIRONMENT=production`. CLAUDE.md Pattern #7's production `extra='ignore'` only applies to `SalesAgentBaseModel` subclasses, so it does not constrain this model.
**What masks it:** the fields serialize correctly in both dev and prod today (both `allow`), so no env or test catches the asymmetry, and the model-changeset notes have no row for `CreateMediaBuySuccess` removed fields.
**Forward risk:** the survival is incidental to the parent's `allow`, not a deliberate re-declaration. If a future SDK base flips to `forbid`/`ignore`, both fields vanish from the production wire silently. Both are valid 3.1 fields, so re-declaring them locally (matching how the other three were handled) makes the contract deliberate.
**Fix:**
```diff
 class CreateMediaBuySuccess(LibraryCreateMediaBuySuccess):
     account: ... | None = None
     sandbox: bool | None = None
     creative_deadline: datetime | None = None
+    valid_actions: list[MediaBuyValidAction] | None = None
+    context: ContextObject | None = None
```

### F5 — `SyncResponseAccount` declares spec-required fields as optional
**Where:** `src/core/schemas/account.py:122-125`; `tests/unit/test_sync_response_account_contract.py:69,91`.
**Context:** New local model replacing the removed SDK per-account type.
**Problem:** The 3.1 `sync-accounts-response` per-account item marks `brand`, `operator`, `action`, `status` required; the model declares all four `| None = None`, and the contract test codifies the laxer shape.
**Interacts with:** `_build_sync_result` (`accounts.py:310-341`), which takes all four as required args — so the required-ness currently lives in the helper signature rather than in the model the spec maps to.
**What masks it:** every production call site supplies all four, so the wire is correct today; no test exercises a construction path that omits them.
**Forward risk:** a future construction path that omits a spec-required field would pass model validation and emit a shape the spec rejects.
**Fix:**
```diff
-    brand: LibraryBrandReference | None = None
-    operator: str | None = None
-    action: str | None = None
-    status: str | None = None
+    brand: LibraryBrandReference
+    operator: str
+    action: str
+    status: str
```
(`billing` stays `str | None` — wire-equivalent to the billing-party enum's string value.) Update the two contract-test assertions to match.

## Follow-ups (track; not blocking)
- **`media_buy_status` never set; deprecated `status` emitted** — `media_buy_create.py:1524,2657,2808,3144,3644` set top-level `status`; no `media_buy_status=` site exists. 3.1 marks `status` deprecated (removed in 3.2) and its schema note warns of an MCP flat-serialization collision with the envelope `TaskStatus`. Conformant at 3.1.0-beta.3 (storyboards don't grade it); the forward risk is the 3.2 removal plus the flat-serialize collision, so populating `media_buy_status` now de-risks that bump.
- **Cross-transport advisory `errors[]` shape** — MCP raw-model wrappers serialize advisory `errors[]` via `to_jsonable_python`, which bypasses the `None`-stripping serializer, so they ship `null` keys that A2A/REST drop via `model_dump(exclude_none=True)`. The mechanism is pre-existing, but 5.7's `Error` adds `issues`/`source`/`sdk_id`/`retry_after`, so the divergent null-key set widens with this bump (6 wrapper sites). Fits the deferred envelope-shape pass; the new fields are the reason to schedule it.
- **enum-normalization duplication** — `enum_value` is now used at `then_success.py:45`, but the same `x.value if isinstance(x, Enum) else str(x)` / `str(getattr(x, "value", x))` pattern is inlined ~13× across `tests/` (`then_payload.py` ×5, `uc006_sync_creatives.py` ×6, +2) and 2× in `src/` (`media_buy_delivery.py:885`, `products.py:721`). The DRY/duplication guard can't see these (each copy is sub-threshold), so they persist silently; `uc006_sync_creatives.py` is a natural fold-in since this PR already touches it.
- **New 3.1 fields not yet handled** (all optional, so not a conformance gap today): `list_creatives_request.{include_purged,include_webhook_activity,webhook_activity_limit}`, `sync_accounts_request.Accounts.{notification_configs,preferred_reporting_protocol}`, `MediaBuyFeatures.committed_metrics_supported`.

## Confirmed correct
- The re-declared and deleted fields all match the v3.1.0-beta.3 schemas; the `GetMediaBuyDeliveryRequest` field deletions (`time_granularity`, `include_window_breakdown`, `include_package_daily_breakdown`) are provided by the SDK 5.7 parent.
- Enum comparisons are safe across the consumed surface (the `adcp-client-python#913` plain-`Enum` hazard only bites when a bare string is compared against an enum member): `adcp_a2a_server.py:938`, `creatives.py:155`, `dashboard_service.py:91` compare against `.value` with string operands (ORM columns / wire JSON); `media_buy_delivery.py:343,766` and `media_buy_list.py:486` are ORM-string or member-vs-member; the client `TaskResult.status` is a `StrEnum`. `SyncCreativeResult.action` is the only `Enum|str` field SDK-wide and it's normalized.
- Auth-error `recovery="terminal"` is unchanged on the wire (sourced from the `_default_recovery` ClassVar, not the SDK table); the rewritten docstrings match 5.7's `STANDARD_ERROR_CODES`.
- `task_type` wire/label split: `metadata["task_type"]` keeps the original label; the validated copy goes only to the webhook payload. All 5 `create_mcp_webhook_payload` call sites pass `task_type` correctly.
- Security: no auth, tenant-isolation, SSRF, or webhook-signature change; `tldextract` is a transitive dep of adcp 5.7 with no salesagent call sites and no runtime public-suffix-list fetch (bundled snapshot, refresh disabled).

## Field-drift matrix (4.3 → 5.7)

| Model · field | 4.3 | 5.7 | spec 3.1 | salesagent | verdict |
|---|---|---|---|---|---|
| `SyncResponseAccount` · brand, operator, action, status | (SDK class) | class removed | per-item **required** | new local model, all 4 optional | F5 |
| `CreateMediaBuySuccess` · valid_actions, context | present | removed | optional | emitted via parent `extra='allow'`, not declared | F4 |
| `CreateMediaBuySuccess` · status / media_buy_status | status | +media_buy_status | both optional; `status` deprecated | emits `status` only | F6 |
| `SyncCreativeResult` · assigned_to, platform_id, status, action | present | removed | optional | re-declared | handled |
| `SyncCreativesResponse` · dry_run, context, ext | present | removed | optional | re-declared | handled |
| `GetMediaBuyDeliveryRequest` · time_granularity, include_window_breakdown | local | SDK-provided | optional | deletion correct | handled |
| `GetProductsResponse.products` / `GetSignalsResponse.signals` | list | `… \| None` | — | guarded / re-declared required | handled |
| `Brand` · trademarks | 27 fields | +1 (28) | optional | unused (name/tagline only) | safe |

## Nits
- No guard enforces `enum_value` usage or null-guarding, so a future `creative.status == CreativeStatus.approved` (a plain-`Enum` field) would silently be `False`. The PR adds the helper but nothing forces its adoption at new call sites.
- `no_bare_basemodel` / `local_schema_imports` matchers miss aliased imports (`BaseModel as BM`, `Product as Prod`) — low likelihood given the `Library*` convention, but an unmodeled bypass form.
- `_normalize_action_to_str` annotated `-> str` but returns `str | None` (`creative.py:376`); the field type backstops it, so no runtime escape.
- Two `enum_value` imports route through `src.core.helpers` (which pulls the adapter graph) rather than the zero-dep `src.core.enum_helpers` (`adapters/base.py:88`, `targeting_capabilities.py:380`) — that's what forces the lazy in-body import.
- Stale class docstring `test_creative_agent_fallback.py:318-322` describes the pre-5.7 drop-with-warning behavior while the tests now assert passthrough.
- `then_user_assets_priority_over_generated` (`uc006:5389`): the leak-check secondary assertion is structurally always-true (the production storage branch can't produce a leaked role), so the url-containment check is the only discriminator; pointing the generative mock at a distinct role/url would make the priority check independently meaningful.
- Orphaned `ctx["expected_legacy_fields"]` write (`uc_get_products_inventory.py:135`) — dead after the consumer step was deleted.
- `CreativeStatusEnum` (local) duplicates SDK `CreativeStatus` (values align today; SDK adds `archived`) — a future value rename would desync the two.

KonstantinMirin:
Round-2 review folded in — pushed `d1db21b..d88e5726`. Point by point:

**SHOULD-FIX**
- **F1** `937643521` — the schema-alignment test now reads schemas vendored from the frozen 3.1 reference `adcontextprotocol/adcp@04f59d2d5`, not live `/schemas/latest`. A missing/relocated schema now fails instead of silently skipping, so `sync/list-creatives` (relocated to `creative/`) are actually validated again.
- **F2** `8f8f421d2` — repointed the `local_schema_imports` FIXME refs `#1360` → `#1388`.
- **F3** `beb658e50` — scoped the Pydantic serializer `filterwarnings` from suite-wide to the get_products serialization tests (confirmed sole emitter), so a real serializer regression elsewhere still surfaces.
- **F4** `3d2060ef8` — declared `valid_actions` + `context` on `CreateMediaBuySuccess`, grounded in the pinned `create-media-buy-response` success variant.
- **F5** `d88e57263` — `SyncResponseAccount` now enforces the spec-required `brand`/`operator`/`action`/`status` (pinned `sync-accounts-response`); verified runtime-safe via the sync_accounts integration + uc011 BDD suites.
- **nits** `db0034756` — `_normalize_action_to_str` return type, stale fallback-test docstring, orphaned ctx write, and `enum_value` import hoisted to the zero-dep module.

F4/F5 schema-conformance now lives in `test_pydantic_schema_alignment.py::TestResponseModelAlignment` (reusing the pinned loader), not hand-rolled in the contract tests.

**Follow-ups (non-blocking), filed:**
- #1447 — the plain-`Enum`→`StrEnum` workarounds (`CreativeStatusEnum`, enum-normalization dup, `enum_value` guard). Root cause is `adcontextprotocol/adcp-client-python#913` (fixed, unreleased) / #946 (5.7.1 deferred).
- #1448 — production handling for the new optional 3.1 fields.
- #1449 — the cross-transport advisory `errors[]` envelope pass (the separate PR flagged in review).

Intentionally not tracked: `media_buy_status` (conformant at 3.1.0-beta.3; only a 3.2 concern) and the uc006 INV-6 leak-assertion nit (low value).

ChrisHuie:
Re-checked the R2 fixes at d88e572 — F1–F5 and the nits are all in, and the full suite is green on the PR head (0 failed across unit/integration/e2e/admin/bdd/ui). A deeper pass on the account-sync area turned up one buyer-reachable bug plus two spec/consistency items. F5 is correct — `SyncResponseAccount.brand` should stay required; nothing below asks to revert it.

**1. Brandless account entry crashes sync_accounts (buyer-reachable)**
`SyncAccountsRequest.accounts` is `list[Accounts | Accounts3]`, and the `Accounts3` arm makes `brand` optional — so `{"account": {...}, "operator": "..."}` validates with `brand=None`. `_extract_natural_key` (accounts.py:424-425) then does `brand.domain` with no guard → unhandled `AttributeError` → 500 (it's not a `ValueError`, so it never normalizes to a clean 400). It's the first call per entry (accounts.py:482), and there's no test for a brandless entry.

```python
def _extract_natural_key(entry):
    brand = entry.brand
    if brand is None:
        raise AdCPValidationError(
            "Account sync requires brand.domain; account-reference entries (no brand) are not supported."
        )
    brand_domain = brand.domain
    ...
```
(Worth a BR-UC-011 / BR-RULE-055..062 check on whether the account-reference variant should be rejected vs implemented.)

**2. SyncCreativesResponse.creatives is optional but 3.1 requires it**
creative.py:480 declares `creatives: list[SyncCreativeResult] = []`. The pinned `sync-creatives-response.json` (04f59d2d5) success variant is `required: ["creatives"]` — the same invariant F5 enforced on `SyncAccountsResponse.accounts`. Not a live wire bug (both callers pass `creatives`), but the model currently permits an invalid shape.

```diff
- creatives: list[SyncCreativeResult] = []  # type: ignore[assignment]
+ creatives: list[SyncCreativeResult]
```
(Grep for any no-arg `SyncCreativesResponse(` relying on the default first; a `RESPONSE_ALIGNMENTS` entry would enforce it executably like F4/F5.)

**3. Auth docstring overstates spec grounding**
exceptions.py:402 says the default `error_code` is `AUTH_TOKEN_INVALID` "per the AdCP 3.1 spec," but that code isn't in the 3.1 error-code enum (64 codes; only `AUTH_REQUIRED`) — the next sentence already notes it's project-specific. Wording-only; the wire is unchanged and pre-existing.

```diff
- Default error_code is AUTH_TOKEN_INVALID per the AdCP 3.1 spec; ...
+ Default error_code is AUTH_TOKEN_INVALID. This code is project-specific: it is in
+ neither the AdCP 3.1 error-code enum nor adcp 5.7 STANDARD_ERROR_CODES (both define
+ AUTH_REQUIRED); it reaches the wire only via the _SPEC_CODES allowlist.
```

**4. SyncResponseAccount.brand required vs nullable column (defensive note — no change needed)**
F5 is correct, keep it. Noting only that the close path (accounts.py:644) passes `db_acct.brand` (column is `nullable=True`) into the now-required field. Not reachable today: the only writer of `Account.principal_id` (which the close path filters on) is sync-create, which always sets a non-null brand. If you want belt-and-suspenders, skip/log a null-brand row in the close branch — not by relaxing the model.

**Not this PR (#1447 scope):** the inline enum→str normalization is in ~5 src sites, not the 2 the issue lists. Adding `creative_formats.py:137`, `creatives/listing.py:440`, `creative.py:215` to #1447 would make that follow-up complete (creative_formats.py:137 is a clean `enum_value(...)` swap; listing.py:440 uses `else f` not `str(f)`, so check the element type first).

One process item: the branch is behind main and currently CONFLICTING (`.duplication-baseline` + a one-line import in `schema_helpers.py`), so the `pull_request` CI is being skipped — worth a merge/rebase + suite re-run on the merged state before merging, especially since #1439 reworks account-reference coercion in this same area.

Verified at d88e572: full suite green (0 failed / 6 suites); items 1–4 confirmed by constructing the models locally and against the pinned 04f59d2d5 schemas. Did not test the merged-with-main state.

KonstantinMirin:
@ChrisHuie — thanks for the deep R3 pass on the account-sync area. All four findings plus the process item are folded in, grounded against the pinned 3.1 reference (`adcontextprotocol/adcp@04f59d2d5`), and the suite is green on the merged-with-main state. Point by point:

### 1. Brandless account entry crashes `sync_accounts` ✅ `901bd400`
Confirmed and fixed. `_extract_natural_key` now None-guards `brand` and raises `AdCPValidationError` → wire `VALIDATION_ERROR` / 400 / `correctable`, instead of the `AttributeError`→500 you traced. Grounding: pinned `account/sync-accounts-request.json` makes each entry `required: ["brand","operator","billing"]` with no account-reference arm — the optional-brand `Accounts3` union member is an SDK artifact generated 9 days *after* the pin, so per the spec we reject. Coverage: a wire-envelope regression on A2A/REST **and** a hand-authored companion `BR-UC-011-account-validation.feature` (all 4 transports). On MCP the FastMCP TypeAdapter rejects it at the boundary before `_impl` — the scenario asserts each transport's actual envelope rather than faking uniformity.

### 2. `SyncCreativesResponse.creatives` required ✅ `dee5b778`
Fixed — dropped the `= []` default so the field is required, matching the pinned `creative/sync-creatives-response.json` success variant (`required: ["creatives"]`). Added a companion `BR-UC-006-creatives-invariants.feature` for the all-failed boundary (every item `action="failed"`, still the success variant carrying `creatives[]`, no operation-level `errors`) — the analog UC-011 already had and UC-006 lacked.

### 3. Auth docstring overstates spec grounding ✅ `179fdf84`
Corrected. Two small adjustments to your suggested wording, both verified: there's no `_SPEC_CODES` allowlist — `AUTH_TOKEN_INVALID` reaches the wire by *passthrough* (it's deliberately absent from `ERROR_CODE_MAPPING`); and the async webhook path *does* enforce `STANDARD_ERROR_CODES`, downgrading non-standard codes to `SERVICE_UNAVAILABLE`. The docstring now says exactly that. (Confirmed against the pinned enum: 64 codes, only `AUTH_REQUIRED`.)

### 4. `brand` required vs nullable column ✅ kept as-is
Agreed — F5 stays; `SyncResponseAccount.brand` remains required. The close-path concern isn't reachable today (the only writer of `Account.principal_id` is sync-create, which always sets a non-null brand). Tracked the optional belt-and-suspenders null-skip as a low-priority follow-up rather than relaxing the model.

### 5. enum→str normalization (your "~5 sites, not 2" note) ✅ `d82ddef10` `0b3ede7e0`
Rather than just widen the follow-up scope, I treated this as the DRY defect it is and routed the hand-rolled sites through the existing `enum_value()` helper (byte-identical per site; `listing.py`'s `else f` clears because `FieldModel` is an `Enum` and the consumer takes `list[str]`), plus a structural guard so a re-introduced hand-rolled `x.value if isinstance(...)` fails the build. The scan also surfaced a **latent bug**: `media_buy_create.py` derived `delivery_type` via `str(plain_enum)` → `"DeliveryType.guaranteed"`, which failed the `== "guaranteed"` check and silently gave guaranteed products the *non-guaranteed* (PRICE_PRIORITY) GAM config instead of STANDARD. Fixed with a real `_create_media_buy_impl` regression test. The upstream SDK `Enum`→`StrEnum` work stays as its own follow-up (gated on adcp-client-python#913/#946).

### Systemic backstop for the F4/F5/#2 class — `41f77e80` `2f2619594`
F4/F5 and your #2 were all the same shape: a response model silently permitting an under-specified shape because no one hand-added the check. So `RESPONSE_ALIGNMENTS` is now **machine-generated** from the pinned `required` arrays over every implemented library-grounded response model, with a self-enumerating coverage gate (a new unregistered grounded model fails the build). That pass surfaced and fixed two more real F5-class bugs — `GetProductsResponse.products` and `ListAccountsResponse.accounts` were optional/defaulted but are `required` in the pinned spec.

### Process: behind-main + CONFLICTING ✅ `34e8fab5`
Merged `main` in (including #1439's account-ref coercion consolidation) and resolved the `.duplication-baseline` + `schema_helpers.py` conflicts. A 3-way review of every co-modified file caught a silent auto-merge corruption git didn't flag: #1372's DB-host `localhost`→`127.0.0.1` fix had been reverted by a later `DATABASE_URL` export, so the merged `test-stack.sh` connected to the wrong endpoint — restored. The branch is now `MERGEABLE` and the full `pull_request` CI runs on the merged state (it was being skipped before). Also bumped aiohttp/cryptography/python-multipart/starlette for fresh security advisories that were blocking the required Security Audit check (those versions are on main too).

Green across unit/integration/bdd/e2e/admin. Ready for another look.

ChrisHuie (approved):

PR #1390, opened 2026-06-05
refactor: relocate pre-commit hooks to appropriate layers (PR 4 of #1234)

mkostromin-sigma:
Closes #1234.

## Summary

Part of **#1234 — PR 4 of 6**. Implements the layered pre-commit model (D27/D31/D44): fast local commits, heavier enforcement in pre-push + CI.

**Before → after:** 37 commit-stage hooks → **12 commit + 11 pre-push** (zero headroom under D27 ceiling).

### Hook relocation (`.pre-commit-config.yaml`)
- **Commit stage (12):** formatters + hygiene only — `ruff format`, `ruff check`, trailing whitespace, EOF, YAML/JSON, large files, merge conflicts, AST, debug statements, `repo-invariants`, `gitleaks`, `detect-secrets`.
- **Pre-push stage (11):** medium-cost checks — `adcp-contract-tests`, `mcp-contract-validation`, `mcp-tool-schema-alignment`, route conflicts, link validation, tenant-auth guard, duplication, migration heads, parameter alignment, type-ignore ratchet, duplication baseline.
- **Deleted from pre-commit (enforcement preserved elsewhere):** 5 grep hooks → AST guards; 4 checks → CI-only via `make quality-ci`; 6 redundant/no-op hooks removed.
- **Directives (mandatory per D31+D44):** `minimum_pre_commit_version: 3.2.0`, `default_install_hook_types: [pre-commit, pre-push]`.

### No enforcement gap (`Makefile` / `make quality-ci`)
Extended **before** deleting hooks so CI Quality Gate never loses coverage:
- `check-gam-auth-support`, `check-response-attribute-access`, `check-roundtrip-tests` → `make quality-ci`
- `check-code-duplication` remains in Quality Gate path

### AST structural guards (5 new `arch_guard` tests)
| Deleted/moved hook | Replacement guard |
|---|---|
| `no-tenant-config` | `test_architecture_no_tenant_config.py` |
| `enforce-jsontype` | `test_architecture_jsontype_columns.py` |
| `check-rootmodel-access` | `test_architecture_no_defensive_rootmodel.py` |
| `enforce-sqlalchemy-2-0` (partial) | existing `test_architecture_query_type_safety.py` |
| `check-import-usage` | `test_architecture_import_usage.py` |

Also adds `test_architecture_pre_commit_hook_count.py` (D27 ceiling 10–12) and `.pre-commit-coverage-map.yml` (hook → enforcement traceability).

### Shared test infrastructure
[...]

PR #1400, opened 2026-06-09
chore: consolidate version anchors and harden Dockerfile (PR 5 of #1234)

mkostromin-sigma:
Closes #1234.

## Summary

Part of **#1234 — PR 5 of 6**. Single-source version anchors across all repo surfaces + container hardening in Dockerfile (D34). Intentionally **does not** bump black/ruff `target-version` (deferred per ADR-008 / D28).

### Version anchor consolidation
| Surface | Before | After |
|---|---|---|
| Python runtime | scattered `3.12` strings | `.python-version` + `python-version-file` in `_setup-env` |
| Postgres CI services | `postgres:15` | `postgres:17-alpine` (5 jobs in `ci.yml`) |
| uv tool | mixed pins | `UV_VERSION=0.11.15` in Dockerfile + CI `env` |
| Drift surfaces | `creative-agent-stack.sh`, `architecture.md`, `LINK_VALIDATION_CI.md` | aligned to canonical tags |

### Dockerfile hardening
- `ARG PYTHON_VERSION` + `ARG PYTHON_BASE_DIGEST` with **`@sha256:`** digest pin on `python:3.12-slim`
- `COPY --from=uv` named build stage (`ghcr.io/astral-sh/uv:${UV_VERSION}`) — fixes Quickstart `COPY --from=ghcr.io/...:${UV_VERSION}` failure
- `ARG SOURCE_DATE_EPOCH` stub for PR6 reproducible builds
- **`USER app:app`** non-root runtime

### Structural guards (3 new `arch_guard` tests)
- `test_architecture_uv_version_anchor.py` — Dockerfile `UV_VERSION` ↔ `_setup-env` ↔ CI `env.UV_VERSION`
- `test_architecture_postgres_image_anchor.py` — every postgres ref uses canonical `17-alpine`
- `test_architecture_dockerfile_digest_pinned.py` — base image digest + non-root `USER` present

### ADR-008
- `docs/decisions/adr-008-target-version-bump.md` — documents deferral of black/ruff `py311→py312` to post-#1234 follow-up
- `docs/decisions/README.md` — index entry

> **Canonical BDD sharding (PR #1379):** 2 parallel shards, greedy scenario-count assignment, **20** rendered `CI / …` check names. Older PR comments mentioning 3/4 shards or 21/22 checks are superseded.
## Stack order (#1234)

Merge in order: **#1370 → #1372 → #1379 → #1424 (PR3-3) → #1390 (PR4) → this PR → #1401 (PR6)**.

**Review focus** — commits unique to PR5:

```bash
git log ci/pr4-hook-relocation..ci/pr5-version-consolidation
```

→ `79c44271` (anchors + Dockerfile) · `b587f052` (digest from `docker pull`) · `75b64a4c` (named `uv` COPY stage)

```bash
git diff ci/pr4-hook-relocation..ci/pr5-version-consolidation --stat
```

→ 11 files, +246/−34

## Regression guards (intentionally unchanged)

- **PR2–PR4:** pre-commit layering, `uv.lock` single-source, 20 frozen CI checks, 2 greedy BDD shards
- **PR5 does not touch:** harden-runner, cosign/Trivy, release signing, hook count

## Expected impact

- Eliminates silent drift between local Docker, CI Postgres services, and compose helpers
- Production image runs as non-root with digest-pinned base (supply-chain baseline for PR6 signing/Trivy)
- Guards fail CI immediately on future version anchor regressions

## Test plan

- [x] CI green on this branch
- [x] `uv run pytest tests/unit/test_architecture_uv_version_anchor.py tests/unit/test_architecture_postgres_image_anchor.py tests/unit/test_architecture_dockerfile_digest_pinned.py -v`
- [x] `uv run pytest tests/unit/ -m arch_guard -x`
- [x] Quickstart job passes (validates Dockerfile `COPY --from=uv` + digest pin)

Closes the version-consolidation slice of #1234 (PR 5 of 6).


mkostromin-sigma:
@ChrisHuie — this PR is ready for your review.

**Context:** Part of the #1234 CI/pre-commit rollout (stacked chain). CI is green on this branch; review comments from earlier PRs in the stack have been addressed.

**Suggested review focus:**
- Confirm the PR-specific diff scope matches the description (use the `git log` range in the PR body)
- Regression guards from earlier stack PRs are preserved (called out in the PR body)
- Structural `arch_guard` tests cover any enforcement moved out of pre-commit

Happy to walk through the stack merge order or split anything if useful. Thanks!

mkostromin-sigma:
@ChrisHuie — **PR5 review / merge note (#1400)** *(supersedes prior review ping)*

CI is green. Part of #1234 — merge **after #1390**.

### Your action for this PR
- Review + approve + merge when the stack through PR4 is on `main`.
- **Review focus:** PR5-only diff (`git log ci/pr4-hook-relocation..ci/pr5-version-consolidation`) — version anchors, Dockerfile digest + non-root `USER`, 3 structural guards.

### After **you** merge this PR
| Who | Task |
|-----|------|
| **You** | **Nothing required** — no branch-protection or org-settings change |
| **Author (me)** | Rebase #1401 (PR6) onto `main` |

### Still pending later (Chris — after **#1401** merges)
See #1401 comment for the full post-merge admin checklist (dependency graph, `Security / Dependency Review`, secret scanning, harden-runner block follow-up, OpenSSF badge).

Thanks!

KonstantinMirin:
Nice work — the artifact itself is clean: non-root `USER app:app`, digest-pinned python base, every `postgres:` at `17-alpine`, and `UV_VERSION=0.11.15` consistent across all three guarded surfaces. The consolidation goal is met for today's values, and the ADR-008 `py311→py312` deferral is internally consistent (no half-bump).

My findings are almost all about one thing: the new guards enforce a narrower invariant than their docstrings promise, so future drift can slip past them green. One real functional regression aside, nothing here is a live defect — it's about the protection the guards actually buy.

### 1. One real regression: dangling `inputs.python-version` in the cache key
`.github/actions/_setup-env/action.yml:32` still interpolates `${{ inputs.python-version }}`, but this PR deletes that input. It now resolves to empty, so the key becomes `…-uv-v4-py-…` — the Python-version dimension is gone, and a future `.python-version` bump would reuse a stale cache. Suggest keying off `hashFiles('.python-version')` or the resolved setup-python output.

### 2. Pattern: guards are narrower than their docstrings
All three new guards advertise a broad property but check a narrow syntactic slice — the unmatched surfaces are silently exempt. Illustrations (not an exhaustive list — the fix is to widen the checks, not patch these sites):
- **Postgres guard** (`_architecture_helpers.py:184`) scans only `Dockerfile*`/compose/workflows/`conftest`. It doesn't see the `.sh`/`.md` files this PR itself re-pinned — `scripts/creative-agent-stack.sh:59`, `.github/LINK_VALIDATION_CI.md:23`, `docs/development/architecture.md:327` (also `.claude/skills/agent-db/agent-db.sh:24`). A future `postgres:16` in any of those passes CI.
- **Digest guard** (`test_architecture_dockerfile_digest_pinned.py:30`) says "every external FROM must include `@sha256:`", but only `FROM python:` can ever produce a violation. The tag-only `FROM ghcr.io/astral-sh/uv:${UV_VERSION}` (Dockerfile:10) is exempt, as would be any future `FROM node:…`.
- **Non-root guard** (same file:45) blocks `USER root`/`USER 0` but not `USER root:root` / `USER 0:0`. Current `USER app:app` is fine; the guard just wouldn't catch a root-with-group regression.

The most robust close would be a single whole-tree deny-list ("no `postgres:<N>` literal outside the canonical anchor") rather than per-glob allow-lists, plus flagging *any* external `FROM` without a digest.

### 3. Smaller items
- **Dead helpers imply coverage that isn't there:** `iter_python_version_anchors` and `assert_anchor_consistency` (`_architecture_helpers.py:133,269`) aren't called by any test — and the former's docstring claims `test_architecture_uv_version_anchor.py` uses it, but that test hand-rolls a narrower check instead. Wiring the guard to call the helpers would delete ~70 lines and widen coverage to `target-version`/`tox.ini`.
- **`security.yml:22,44`** runs `setup-uv` with no `version:`, floating uv instead of pinning to `0.11.15` — same drift class this PR targets, in an unguarded surface.
- **Postgres guard** hard-codes `CANONICAL_TAG = "17-alpine"` (one more copy of the value); asserting `len({tags}) == 1` would be drift-proof and literal-free, matching the uv guard's approach.

mkostromin-sigma:
**Stack sync:** rebased onto PR4 tip `ce20bdbd` (includes PR3-3 `#1424` + full PR4 stack). PR5-only commits unchanged (3 commits). CI re-running.

mkostromin-sigma:
**Rebased** → tip `bb4baed9` (3 PR5-only commits on full stack). Mergeable; CI running.

mkostromin-sigma:
@KonstantinMirin — PR5 review items addressed in `37dd37d0`:

1. **Cache key** — fixed upstream in PR3-3 (`hashFiles('.python-version')`)
2. **Guard widening**:
   - postgres: whole-tree scan (scripts/docs included); single-tag uniqueness
   - Dockerfile: all external `FROM` require digest; `USER root:root`/`0:0` blocked
   - uv image digest-pinned via `UV_IMAGE_DIGEST`
3. **Helpers wired**: `test_python_version_anchors_consistent` uses `iter_python_version_anchors` (ADR-008 py311 target-version excluded)
4. **security.yml**: `UV_VERSION` env + pinned `setup-uv` version on both jobs

Rebased on PR4 tip. CI re-running.

mkostromin-sigma:
@KonstantinMirin — **rebased + review items addressed**, tip `2a1fc24f`, CI green (Summary ✅). Merge after #1390.

### Rebase
Cherry-picked PR5-only commits (6) onto rebased PR4 tip. Clean apply, no manual conflict resolution.

### Your review items (2026-06-11) — addressed

| Item | Commit / location |
|------|-------------------|
| Dangling `inputs.python-version` in cache key | Fixed upstream in PR3-3: `hashFiles('.python-version')` in `_setup-env/action.yml` |
| Postgres guard whole-tree scan (scripts/docs included) | `7f243e06` |
| Digest guard — all external `FROM` require `@sha256:` (not just `python:`) | `7f243e06` |
| Non-root guard — `USER root:root` / `USER 0:0` forms | `7f243e06` |
| `iter_python_version_anchors` / `assert_anchor_consistency` wired into guards | `3956b6cb` |
| `security.yml` — `setup-uv` pinned to `UV_VERSION=0.11.15` | in PR5/PR6 stack |
| Postgres guard — single-tag uniqueness (`len({tags}) == 1`) instead of hard-coded canonical | `7f243e06` |

Ready for re-review / merge after #1390 lands.

KonstantinMirin:
## Review — PR5 version-anchor consolidation

First, credit where it's due: the round-1 items got real work. The cache key no longer references the deleted `python-version` input; the digest guard now flags *every* external `FROM` (not just `FROM python:`) and ships a `_KNOWN_BAD_FROM` self-test; the `USER` guard handles `root:root`/`0:0`; `security.yml` pins uv; and the postgres guard dropped its hard-coded `CANONICAL_TAG` for a drift-proof `len(by_tag) == 1`. The digest and non-root guards are now genuinely sound.

The consolidation *values* are also correct — on tracked files, postgres is `17-alpine` everywhere, uv is `0.11.15`, Python is single-sourced via `.python-version`. That part of the goal is met.

I'm holding the PR on two **root-level** issues, not line nits. Both are the same patterns from round 1 — they didn't get fixed, they moved to new sites. I'd rather we close the *pattern* once than do a round 3.

### Root cause A — "single-source" is implemented as "copy + guard," which isn't single-source

The goal is one version, no drift. But for uv and postgres the approach is: copy the literal into N files, then add a test asserting they're equal. A value that lives as N literals kept-equal-by-a-test hasn't been single-sourced — we've added a tripwire on top of the duplication, and now the **tripwire's correctness** is the new single point of failure. That's exactly where every defect in this PR lives.

You already did it right once in this same PR: Python via `.python-version` + `python-version-file`. CI *reads the file* — there's no second copy, so no guard is even needed. That's the target shape.

What I'm asking for:
- Apply the `.python-version` shape to uv where feasible: one canonical `UV_VERSION`, **referenced**, not copied into `Dockerfile:8`, `_setup-env/action.yml`, `ci.yml:20`, `security.yml:11`.
- Where a literal genuinely must be duplicated, the guard has to pin the *whole* dependency, not just the number. Right now `astral-sh/setup-uv` is pinned to **v4.2.0** in `_setup-env/action.yml:17` and **v5.4.0** in `security.yml:25,49`. The uv binary is unified; the installer action drifts — in the guard's blind spot, because the guard only checks the version literal. That's the same drift class the PR exists to kill.

### Root cause B — a guard must enforce exactly what its docstring claims, and must be proven to fail

Every defect in both rounds is one shape: **the guard checks something narrower, wider, or different than it advertises.** Round 1 it was "scan-set too narrow" (missed `.sh`/`.md`). The fix widened the postgres scan to the whole tree — overshooting the other way — and the same gap reopened elsewhere. Two concrete problems right now:

1. **The postgres guard is non-hermetic and fails on this branch.** `iter_postgres_image_refs` (`tests/unit/_architecture_helpers.py:234-256`) does `repo.rglob("*")` and ignores `.gitignore`, so it scans gitignored `.claude/code-review/**` notes that quote `postgres:15/16/17.9` as prose and fails with `assert 6 == 1`. It's green on a clean CI checkout and red locally — its verdict is decoupled from committed drift.
   ```
   uv run pytest tests/unit/test_architecture_postgres_image_anchor.py::test_postgres_image_unified -v
   ```
2. **The python-version guard has zero coverage of the Dockerfile and CI** — the surfaces this PR changed. Its only Dockerfile pattern is `FROM python:<literal>` (`_architecture_helpers.py:152`), but the PR templated the Dockerfile to `FROM python:${PYTHON_VERSION}` + `ARG PYTHON_VERSION=3.12`. `iter_python_version_anchors(repo_root())` yields **nothing** from the Dockerfile or any workflow. Bumping the Dockerfile python out of sync with `.python-version` passes the guard green. The one property the PR most needs to guard isn't guarded.

Two rules close this entire class — and you already applied both to the digest/USER guards, which is why those are the sound ones:

- **Scan the committed surface, never the working tree.** Drive every guard's file enumeration from `git ls-files`, not `Path.rglob`. This single change removes the non-hermetic failure and every future variant of it.
- **Every guard ships a mutation self-test + a non-vacuity assertion** — a known-bad input it must reject, and "I found ≥1 anchor on every surface I claim to cover." The digest guard has `_KNOWN_BAD_FROM`; the anchor guards need the equivalent (e.g. an `ARG PYTHON_VERSION=` pattern, and an assertion the Dockerfile contributed an anchor).

### One DRY consequence to fold in while you're here

The shared helper `assert_anchor_consistency` was adopted in the uv guard only; the postgres (`test_architecture_postgres_image_anchor.py:18-27`) and python (`test_architecture_uv_version_anchor.py:55-69`) siblings still hand-roll the same equal-all-values logic. Route all three through the helper — otherwise the next change to it silently misses two of three guards. Also: `iter_compose_files` (`_architecture_helpers.py:140`) now has zero callers — delete it.

---

To be explicit about what "done" looks like, so this is one more round and not two:
1. uv referenced from one source (or guard widened to pin the installer action too);
2. all guard scans driven by `git ls-files`;
3. python-version guard actually covers the templated Dockerfile + CI, with a non-vacuity assertion;
4. all three anchor guards call `assert_anchor_consistency`; dead `iter_compose_files` removed;
5. `uv run pytest tests/unit/ -m arch_guard` green on a clean checkout *and* with local untracked files present.

The underlying ask: fix the two patterns once and grep for every other instance yourself, rather than patching the cited lines and waiting for the next site to surface.

mkostromin-sigma:
## Ready for re-review — round 2 feedback addressed

@KonstantinMirin — thanks for the detailed review. CI is green on `88a3882a`. This round closes both root-cause patterns you called out.

### Root cause A — single-source (uv + installer pin)

- **`.uv-version`** is now the canonical uv version (same shape as `.python-version`).
- **`.github/actions/_install-uv`** reads `.uv-version` at runtime — no copied literals in `ci.yml`, `security.yml`, or `_setup-env`.
- **`astral-sh/setup-uv`** is pinned once (`v5.4.0` SHA) in `_install-uv`; workflows call `./.github/actions/_install-uv`, not the action directly.
- **`Dockerfile`** keeps `ARG UV_VERSION=…` as the one unavoidable duplicate for BuildKit `FROM`; guard keeps it in sync with `.uv-version`.

### Root cause B — guards enforce what they claim

- **Postgres + python anchor scans** use `git ls-files` (not `Path.rglob`) — hermetic with local untracked files.
- **Python guard** covers templated Dockerfile via `ARG PYTHON_VERSION=…` (not `FROM python:${PYTHON_VERSION}` alone).
- **Non-vacuity** assertions on all anchor surfaces (including Dockerfile contributing an anchor).
- **Mutation self-tests** added (same class as `_KNOWN_BAD_FROM` on digest guard):
  - postgres tag drift
  - uv / python version drift (`.uv-version` / `.python-version` vs Dockerfile ARG)
  - templated `FROM` without `ARG PYTHON_VERSION`
  - setup-uv referenced outside `_install-uv`

### DRY

- All three anchor guards route through **`assert_anchor_consistency`**.
- Dead **`iter_compose_files`** removed.

Please take another look when you have a moment — happy to adjust if anything still misses the intent.

KonstantinMirin (changes requested):
## Review: version anchor consolidation

The consolidation itself is good. There's no live drift: postgres is `17-alpine` everywhere, Python is `3.12`, uv is `0.11.15`, and the suite is green. If this PR were only the anchor cleanup, it'd be a straightforward approve.

The problem is the part that's the whole point of the PR — the three guards that are supposed to *keep* it from drifting. Two of them pass when I feed them real drift. So I'm going to say the same three things I always end up saying, because this PR is a clean instance of all three.

### 1. The guard doesn't test what it says it tests

**Postgres guard is blind to drift within a file.** `assert_anchor_consistency` (`tests/unit/_architecture_helpers.py:499`) uses `re.search`, which compares only the *first* postgres tag in each file. `ci.yml` has four postgres services. Mutate the second one and the guard passes:

```bash
sed -i '' '260s/postgres:17-alpine/postgres:15-alpine/' .github/workflows/ci.yml
uv run pytest tests/unit/test_architecture_postgres_image_anchor.py::test_postgres_image_unified -q   # PASSES
```

Mutate the *first* ref and it fails. So detection depends on line order, not on whether the repo is consistent. Three of the four CI services are unguarded.

**Python guard scans none of the files it claims to.** `_python_anchor_candidate` (`tests/unit/_architecture_helpers.py:247-255`) tests an absolute path with `.startswith(".github/workflows/")`, which is always false. Run it:

```bash
uv run python -c "from pathlib import Path; import tests.unit._architecture_helpers as h; r=Path('.').resolve(); print(sorted({str(p.relative_to(r)) for p,_ in h.iter_python_version_anchors(r)}))"
# -> ['.python-version', 'Dockerfile', 'mypy.ini', 'pyproject.toml']  — zero workflows, zero actions
```

The docstring says it scans workflows and composite actions. It scans neither — which is exactly the surface this PR reorganized. The "no hardcoded UV_VERSION" guards have the same gap: they iterate `.github/workflows/` only, not the new `.github/actions/` layer you just created.

The fix for both is the same: the per-match iterators already exist (`iter_postgres_image_refs`, `iter_python_version_anchors` use `finditer`). Drive the consistency check off every match instead of re-reading the file and taking the first one, and add a non-vacuity assertion so an empty scan can never read as "consistent." Use repo-relative paths in the candidate check.

(Same class, lower priority: the `requires-python` exemption at `tests/unit/test_architecture_uv_version_anchor.py:166` keys on the value `"3.11"`, so downgrading `>=3.12` to `>=3.11` passes. Exempt the `target-version` key, not the string.)

### 2. The duplication is what made the guards wrong

The uv/python extraction regex is restated in four places: the live guard, the `_KNOWN_BAD_*` fixtures, and the sibling `test_architecture_helpers_contract.py`. That's not a style nit — it's the cause of #1. The mutation self-tests validate a *copy* of the detector, so they stay green while the real guard is blind. A self-test that exercises a duplicate of the thing under test isn't testing the thing.

You already did this correctly for postgres (`_PG_TAG_PATTERN = _PG_IMAGE_LITERAL.pattern`). Do the same for uv and python: one named constant in the helper, and have the self-tests import the real detector.

### 3. So, the order of operations

This is the same note as the last few PRs, so let me make it concrete as a sequence rather than a list of sites:

1. **De-duplicate first.** One source for each extraction regex and each canonical value, in `_architecture_helpers.py`. Everything else references it.
2. **Then make the test test the real thing.** Self-tests import the real detector; consistency checks run over every match; add the non-vacuity assertions.
3. **Then prove it.** For each guard, mutate the actual target file (not an in-memory snippet), confirm the guard fails, revert. If a guard can't be made to fail by breaking its invariant, it isn't a guard.

If you do step 1 properly, steps 2 and 3 mostly fall out — the duplication is what's hiding the blind spots.

Smaller stuff that's fine to fold in: digest pins (`UV_IMAGE_DIGEST`, `PYTHON_BASE_DIGEST`) aren't guarded at all; `_setup-env` provisions Python twice (once directly, once via `_install-uv`'s `setup-uv`); and ADR-008 lists three surfaces that don't exist (`tox.ini` python, Dockerfile postgres, `[tool.black].target-version`).

mkostromin-sigma:
@KonstantinMirin — thanks for the detailed review. CI is green on the latest commit (`1a08edb9`). Addressed all three recurring points:

### 1. Guards now test what they claim

**Postgres / `assert_anchor_consistency`:** switched from `re.search` (first match only) to `finditer` per file. Intra-file drift is covered by `test_postgres_anchor_detector_catches_intra_file_drift` and `test_assert_anchor_consistency_flags_intra_file_drift`.

**Mutation proof (reverted after):**
```bash
sed -i '' '260s/postgres:17-alpine/postgres:15-alpine/' .github/workflows/ci.yml
uv run pytest tests/unit/test_architecture_postgres_image_anchor.py::test_postgres_image_unified -q  # FAILS
# ci.yml shows ['15-alpine', '17-alpine']
```

**Python scan surface:** `_python_anchor_candidate` now uses repo-relative paths (`path.relative_to(repo)`), not absolute `path.as_posix()`. Added `test_python_anchor_scan_includes_github_yaml_surfaces` to assert workflows **and** actions are in the scan surface.

**UV / python-version hardcode guards:** `iter_hardcoded_uv_version_env` and new `iter_hardcoded_python_version_yaml` enumerate via `git ls-files` and cover `.github/workflows/` **and** `.github/actions/`.

**Non-vacuity:** empty scans cannot pass as consistent (`assert matches`, `assert refs`, etc.).

**ADR-008 exemption:** keyed on `anchor_kind == "target-version"`, not the string `"3.11"`.

### 2. De-duplication — single source for regex

`uv_version_pattern_map()` and `python_version_pattern_map()` in `_architecture_helpers.py` (same pattern as existing `_PG_TAG_PATTERN`). Self-tests and `test_architecture_helpers_contract.py` import the real pattern maps — no duplicated regex in fixtures.

### 3. Order of operations followed

1. De-duplicated regex/canonical values in helpers
2. Self-tests exercise the real detector; consistency checks run over every match
3. Mutation proof on real target files (ci.yml postgres tag, Dockerfile `PYTHON_VERSION`) — guards fail, changes reverted

### Smaller items folded in

- **Digest pins:** `assert_dockerfile_digest_args_present()` guards `UV_IMAGE_DIGEST` and `PYTHON_BASE_DIGEST` ARGs (+ mutation self-test)
- **`_setup-env`:** removed duplicate `setup-python` step — Python is provisioned once via `_install-uv` / `setup-uv` + `python-version-file`
- **`iter_compose_files`:** removed (superseded by `git ls-files`)

**Not in this commit (doc-only, happy to fold if you want):** ADR-008 still mentions `tox.ini` python and Dockerfile postgres surfaces that don't exist in the repo; `[tool.black].target-version` should read `[tool.ruff].target-version`.

Ready for re-review when you have a moment.

mkostromin-sigma:
Follow-up: ADR-008 surface list corrected in latest commit (tox.ini python, Dockerfile postgres, and `[tool.black]` references removed).

KonstantinMirin (changes requested):
**Status: changes requested.** The critical work from the last round is genuinely done — I want to be clear about that before the rest, because it was good work and I verified it properly. The remaining items are a smaller set, but two of them are lines I hold firmly: new work doesn't introduce duplication, and a fix doesn't introduce new gaps. Both happened in this commit, so I'm asking for one more pass.

### First — the round-1 critical is fixed, and I confirmed it the hard way

I didn't take the commit message on trust. I mutated the real files and watched the live guards fail:

- Every one of the four `ci.yml` postgres services now triggers on drift (not just the first) — `assert_anchor_consistency` moving to `finditer` did the job.
- The Python scan now actually covers workflows **and** composite actions — injecting a hardcoded `python-version` into `_setup-env/action.yml` fails the guard, as it should.
- `requires-python`, `mypy.ini`, and `.python-version` downgrades are all caught now, and the legitimate ruff `target-version = "py311"` exemption still passes.
- Removing either digest ARG from the Dockerfile fails the new guard.

That's the substance, and it holds. Thank you for the thorough self-tests.

### Two things I can't wave through

These are the two standing rules, not new opinions.

**1. The duplication fix relocated the duplication instead of removing it.**

The change asked of last round was to stop re-implementing the workflow-scan loop. This commit answered it by adding `iter_hardcoded_python_version_yaml` — which is the same body as `iter_hardcoded_uv_version_env`, differing only by the regex and one skip line:

```
# _architecture_helpers.py — these two are structurally identical
iter_hardcoded_uv_version_env(repo)        # scan -> _github_yaml_candidate -> read -> per-line regex
iter_hardcoded_python_version_yaml(repo)   # same, with one extra "python-version-file" skip
```

And `_pg_image()` was copied into `test_architecture_helpers_contract.py` — the same commit removed cross-file regex duplication in one place and re-introduced it in another. The duplicate-block count came out flat.

In this codebase DRY is a correctness rule, not a style preference, and it applies most strictly to *new* code. A duplicated scan loop means the next person who fixes one copy misses the other. The fix here is to extract one parametrized helper and call it twice:

```python
def iter_hardcoded_yaml_anchor(repo, regex, *, skip_substr=None): ...
```

— and move `_pg_image` into the shared helper so both tests import it. Carryover from last round in the same spirit: the `ARG PYTHON_VERSION` regex is still declared twice (`_architecture_helpers.py:223` and `:246`); please collapse to one.

**2. The exemption fix over-corrected and opened a new blind spot.**

Re-keying the ADR-008 carve-out from the value `"3.11"` to `anchor_kind == "target-version"` correctly fixed the `requires-python` problem I raised — good. But it now exempts `target-version` from *every* value check, and ADR-008 says that value must stay exactly `py311`. So:

```bash
# target-version py311 -> py310 now PASSES the guard; it should not
```

A fix shouldn't quietly widen a hole while closing another. Keep the runtime-consistency exemption keyed on the anchor, but add a positive assertion that `target-version == "3.11"`, so a downgrade or a stray bump is still caught.

### Lower priority (fold in while you're here)

- ADR-008 lists three surfaces that don't exist: `tox.ini` as a Python surface (no python pin there), the `Dockerfile` for postgres (no postgres ref), and `[tool.black].target-version` (there's no `[tool.black]` section — only ruff). The doc should match what the code anchors.
- `_PG_IMAGE_LITERAL` is unanchored — it matches `postgres:NN` inside markdown prose. Green today only because every ref is `17-alpine`.
- `iter_git_tracked_files` still calls `git ls-files` with no error handling, and more guards now route through it.

### What I'm asking for

One more pass that drives the duplicate count *down*, not sideways, and re-tightens the target-version check. Once the scan loop is a single helper, `_pg_image` is shared, and `target-version` is pinned again, I'll re-run the mutation checks and approve. The guard work underneath this is solid — it deserves a finish that holds the same bar.

mkostromin-sigma:
@KonstantinMirin — ready for re-review. CI is green on `ba5e106f`.

This addresses your [review #4519255005](https://github.com/prebid/salesagent/pull/1400#pullrequestreview-4519255005) items:

**Required (round 3)**
- **DRY scan loop:** `iter_hardcoded_yaml_anchor()` is the single implementation; `iter_hardcoded_uv_version_env` / `iter_hardcoded_python_version_yaml` are thin wrappers.
- **Shared postgres helper:** `postgres_image_ref()` lives in `_architecture_helpers.py`; contract + postgres guard tests import it (no local `_pg_image` copies).
- **Single `ARG PYTHON_VERSION` regex:** `_DOCKERFILE_PYTHON_VERSION_PATTERN` feeds both `_PY_VERSION_PATTERNS` and `_DOCKERFILE_PYTHON_VERSION_RE`.
- **ADR-008 target-version re-pinned:** runtime anchors still compare against `.python-version` (`3.12`), while `assert_adr008_target_version_pinned()` positively asserts deferred ruff `target-version` stays `py311` (`3.11`); downgrade to `py310` fails.

**Lower priority (folded in)**
- **ADR-008 surface list:** corrected in `97997942` (no tox.ini python, no Dockerfile postgres, ruff-only target-version).
- **`_PG_IMAGE_LITERAL` anchored:** context-prefixed patterns (`FROM` / `image:` / `PG_IMAGE=` / quotes / backticks / docker-run line); prose-only refs ignored (`test_postgres_scan_ignores_unanchored_prose`).
- **`iter_git_tracked_files`:** `git ls-files` errors wrapped in `RuntimeError`.
- **DRY python scan:** `iter_python_version_anchors` delegates to `extract_python_version_anchors`.

**Mutation proofs (local, reverted after)**
- `ci.yml` second postgres service `17-alpine` → `15-alpine` → guard **FAIL**
- `pyproject.toml` `py311` → `py310` → guard **FAIL**
- unanchored prose `postgres:15` in a sentence → guard **ignores**

Please take another look when you have a moment.

KonstantinMirin (approved):
**Approving.** This is solid work, and I want to acknowledge how it got here: across three rounds you didn't just patch the named sites — you removed the duplication at the root (one `iter_hardcoded_yaml_anchor`, a shared `postgres_image_ref`, a single `_DOCKERFILE_PYTHON_VERSION_PATTERN`), re-pinned `target-version` with a real guard instead of a broadened exemption, and anchored the postgres detector so it stops matching prose. I verified the important ones by mutation rather than by reading — every live guard now fails when I break its invariant (`target-version py311→py310` and `→py313` both caught, all four `ci.yml` postgres services, the python downgrades, the digest ARG removal). The duplicate-block count went *down*, not sideways. Thank you for taking the DRY and test-integrity notes seriously.

There are four small things left. **None of them block this PR** — it's correct and mergeable as-is. But I'd rather they get fixed before merge or filed as tracked issues now, so they don't quietly slip. My order of preference:

**1. The digest guard doesn't actually guard the digest (recommend fixing).**
`test_dockerfile_base_image_digest_pinned` only checks that a digest ARG *exists* and is 64 hex characters. It never checks the digest corresponds to the version it's pinning. So if someone bumps `PYTHON_VERSION` or `UV_VERSION` and forgets the digest, the guard stays green while the image builds on the old digest — which is exactly the failure the digest pin exists to prevent. A guard that reads as protection but checks only shape is worse than no guard, because it buys false confidence. Two honest options: (a) rename the test/helper to say what it really does (presence + shape only), or (b) make it real — assert the digest matches the resolved version (even a stored expected-digest map keyed by version would do). I'd take (b) if it's cheap, (a) if it isn't, but I don't want it sitting there implying more than it does.

**2. ADR-008 names a file that doesn't exist (recommend fixing — it's one line).**
The Postgres surface list says `compose.yaml`; this repo uses `docker-compose*.yml`. The whole point of this PR is that the documented surfaces match reality, so the ADR should name the actual files. Trivial, but it's the same class of drift we just spent three rounds closing.

**3. `_install-uv` silently provisions Python (fine as a follow-up).**
The composite action named "Install uv" also sets up Python via `python-version-file`. It works and is correctly single-sourced — the issue is only that the name and `description:` hide that responsibility, so the next person reading the workflow won't expect Python to come from there. Either rename it (`_setup-python-and-uv`) or call it out in the action's `description:`.

**4. `anchor_kind` label delimiters are mixed (optional/cosmetic).**
The label set mixes hyphen (`requires-python`, `target-version`) and underscore (`python_version`). It's defensible since each label mirrors its config key, but a one-line comment saying that would save the next reader a double-take.

Could you either fold #1 and #2 into this PR before merge, or open tracking issues for whichever you defer? I just want them recorded somewhere durable rather than living only in this thread. Items #3 and #4 are genuinely fine as backlog. Nice work — happy to see it land.

mkostromin-sigma:
## Follow-up tracker — Part B helper trim blocked on this PR

[#1459](https://github.com/prebid/salesagent/pull/1459) Part B (trim unused `_architecture_helpers` exports) was **deferred** until **this PR merges** — see [ChrisHuie review on #1459](https://github.com/prebid/salesagent/pull/1459#issuecomment-4745123707) and #1455 sequencing.

### Helpers this PR still needs (do not delete in follow-up)
| Helper | Consumer |
|--------|----------|
| `iter_python_version_anchors` | `tests/unit/test_architecture_uv_version_anchor.py` |
| `iter_postgres_image_refs` | `tests/unit/test_architecture_postgres_image_anchor.py` |
| `iter_workflow_files` | `tests/unit/test_architecture_uv_version_anchor.py` |
| `iter_compose_files` | absent on this branch; grep after merge before deleting anything |

### After **this PR merges** → open small follow-up PR
```bash
git fetch upstream
git checkout -b chore/trim-dead-architecture-helpers upstream/main

# For each candidate: confirm zero consumers outside _architecture_helpers.py
git grep -n "<helper_name>" -- tests/ ':!tests/unit/_architecture_helpers.py'

# Remove only confirmed-dead helpers; update module docstring
make quality

gh pr create --repo prebid/salesagent \
  --title "chore: trim dead _architecture_helpers after #1400 (#1455 Part B)" \
  --body "Closes remaining #1455 Part B. Consumers verified post-#1400."
```

**Constraints:**
- Merge #1400 first (this PR)
- Then #1459 can merge independently (Part A only; helpers already restored)
- Part B follow-up is a separate PR on fresh `main` — not stacked on #1459
- If grep finds no dead helpers post-merge, close follow-up with a note instead of deleting

/cc @ChrisHuie for visibility on deferred Part B plan.

mkostromin-sigma:
Folded review follow-up **#2**: ADR-008 postgres surface list no longer references `compose.yaml` (repo uses `docker-compose*.yml` only). Items #1, #3, #4 deferred to post-merge follow-up.

PR #1413, opened 2026-06-10
fix: Wrong number of arguments in a class instantiation

patmmccann:
To fix this without changing functionality, update the test to instantiate `DynamicPricingService` with the required constructor argument using a lightweight mock dependency. This keeps the test focused on `_calculate_product_pricing` behavior while satisfying the class contract and preventing `TypeError`.

In `tests/unit/test_format_id_spec_compliance.py`, edit the `test_pricing_service_receives_format_id_objects` function around line 192:
- Create a `MagicMock()` (or similar) as the required init dependency.
- Pass it to `DynamicPricingService(...)` instead of calling with no arguments.

No new imports are needed because `MagicMock` is already imported on line 10.

### DIFF
--- a/tests/unit/test_format_id_spec_compliance.py
+++ b/tests/unit/test_format_id_spec_compliance.py
@@ -189,7 +189,8 @@ def test_pricing_service_receives_format_id_objects():
     """
     from src.services.dynamic_pricing_service import DynamicPricingService
 
-    service = DynamicPricingService()
+    pricing_dependency = MagicMock()
+    service = DynamicPricingService(pricing_dependency)
 
     mock_product = MagicMock()
     mock_product.format_ids = [


ChrisHuie:
The constructor fix alone doesn't restore this test. It's xfail(strict=True) for the triple-isinstance behavior, but after this change it still fails before reaching that code: _calculate_product_pricing() requires tenant_id, country_code, min_exposures, and cutoff_date (src/services/dynamic_pricing_service.py:81-88), so the xfail trips on a TypeError rather than the documented bug.

Completing the signature fix would actually fail the suite: with a mock session the query path returns no metrics, _default_pricing() is returned, and the assertion ("base_rate" in result or result is not None) passes regardless of the isinstance behavior — so the test XPASSes under strict=True. As written, the assertion cannot detect the bug it documents.

Suggested: fix both signatures (the constructor parameter is db_session, so pricing_dependency is a misleading name), drop the strict xfail, and assert concrete behavior; track the triple-isinstance cleanup separately.


**BLOCKER**
— silences a strict-xfail; completing the fix fails the suite. test_format_id_spec_compliance.py:178
Ran it: still raises TypeError: missing 4 required positional arguments (dynamic_pricing_service.py:81), so the strict=True xfail trips on signature rot, not the documented bug — and the Codex bot's P2 saying so is unaddressed. Ran the complete fix: returns a default dict, the assertion "base_rate" in result or result is not None passes → XPASS → suite fails. Fix: rewrite with a real assertion and drop the xfail, or delete the sentinel.


**SHOULD-FIX**
— two sibling xfails are also rotten (ran --runxfail): test_format_resolver_finds_format_without_agent_url patches a dead target (format_resolver.py:43 is a function-local import); test_product_format_ids_are_always_format_id_objects dies on missing required Product fields before its assertion.


NIT
— pricing_dependency misnames a db_session; the "max 10 mocks/file" hook CLAUDE.md cites doesn't exist.

PR #1416, opened 2026-06-10
fix: Mismatch in multiple assignment

patmmccann:
The fix is to unpack all values returned by `_render_feature` (3 values) and ignore only the ones not needed in `verify_features`.  
Best minimal fix: update line 1600 in `scripts/compile_bdd.py` from two-target unpacking to three-target unpacking, keeping `expected` and discarding the extra return values with throwaway variables. This avoids changing functionality and only resolves the arity mismatch.

Concretely in `verify_features(...)`, replace:

- `expected, _new = _render_feature(...)`

with:

- `expected, _new, _unused = _render_feature(...)`

No imports, helper methods, or dependency changes are required.

### DIFF
--- a/scripts/compile_bdd.py
+++ b/scripts/compile_bdd.py
@@ -1597,7 +1597,7 @@ def verify_features(adcp_req_path: Path) -> bool:
         text = source_path.read_text()
         feature = parse_feature_file(text)
         uc_key = _extract_uc_key(source_path.name)
-        expected, _new = _render_feature(feature, traceability, uc_key, source_path.name, commit_sha)
+        expected, _new, _unused = _render_feature(feature, traceability, uc_key, source_path.name, commit_sha)
         actual = output_path.read_text()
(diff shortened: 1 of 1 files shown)


ChrisHuie:
The unpack looks correct — _render_feature returns a 3-tuple and this was the last 2-target unpack (the other call sites unpack 3/4/6 correctly), so it raises ValueError on the first verified file.

Two related items in the same file: 
(1) compile_feature returns 4 values but is annotated -> tuple[str, str, list[dict]]; scripts/ isn't mypy-scanned so it's latent — should be tuple[str, str, list[dict], set[str]]. 
(2) verify_features re-renders with the wholesale renderer, but most compiled files are --merge-generated (which preserves LEGACY scenarios the wholesale renderer drops), so --verify will report false STALE against them; worth fixing before wiring --verify into CI. The "Re-run: --all" hint at the bottom has the same mismatch.

Naming nit: _unused -> _scenario_ids to match the other call site.

Issue #1419, opened 2026-06-11
ci: run the BDD e2e_rest 5th transport in CI as an additive step in #1372's e2e job


KonstantinMirin:
## Summary

CI never runs the BDD **`e2e_rest`** transport (the real-HTTP 5th transport). The local in-network runner is currently the only thing that exercises it. This issue tracks closing that gap **as a small additive step into the active #1234 CI pipeline**, not by replacing it.

## Why CI never runs e2e_rest today

In both the current `test.yml` and @mkostromin-sigma's authoritative `ci.yml` (#1372), the `bdd-tests` job runs `pytest tests/bdd/` with **no `BDD_E2E_ENABLED` and no live stack → 4 transports only** (`impl`/`a2a`/`mcp`/`rest`). The 5th transport requires a running server + `BDD_E2E_ENABLED=true`.

## Local vs CI — not the same problem

The **in-network runner** (PR #1420) exists to solve **host-port collisions on a shared dev machine** — it reaches Postgres/the server by service name with zero published ports so many stacks run concurrently. A CI runner is an ephemeral, isolated VM with no collision pressure, so CI does **not** need the in-network *orchestration*. Local and CI are different orchestration layers:

- **Local:** in-network runner replaces `run_all_tests` (GHA `services:`/composite actions can't run locally).
- **CI:** #1372's GHA-native authoritative pipeline stays the source of truth.

**Parity lives in the shared layer, not the orchestration.** Both drive the *same* `tox` envs + `conftest` gating + harness/dispatcher + ledger. The orchestration differs; the test definitions are identical. That is the anti-drift contract.

## Relationship to the active #1234 CI overhaul (@mkostromin-sigma)

Not stranded, not a competing pipeline — active work:

- **#1370** (PR2, merged) — uv.lock single source / test isolation
- **#1372** (PR3, open) — authoritative `ci.yml` (replaces `test.yml`), composite actions (`_postgres`/`_pytest`/`_setup-env`), **frozen required-checks guard** (`test_architecture_required_ci_checks_frozen.py`)
- **#1379** (PR3-2, open) — parallel BDD CI shards
- **#1390 / #1400 / #1401** (PR4–6, open)

#1372 already stands up the full `docker-compose.e2e.yml` stack in its `e2e-tests` job. That is where the 5th transport belongs.

## Plan — Option B: one additive in-network step inside #1372's `e2e-tests` job

Add a single step to the existing `e2e-tests` job (which already has the stack up):

[...]

Issue #1418, opened 2026-06-11
test(bdd): retire the e2e_rest xfail ledger via transport-aware harness setup


KonstantinMirin:
## Summary

The BDD suite's 5th transport (`e2e_rest` — real HTTP through nginx to a separate live server) carries a **312-entry xfail ledger** (`tests/bdd/e2e_rest_known_failures.txt`). These scenarios pass on all four in-process transports (`impl`/`a2a`/`mcp`/`rest`) and fail **only** over `e2e_rest`, because their setup uses in-process mock injection a separate server process cannot see.

This is a leaky abstraction: transport awareness has bled into the scenario/step-def layer. The fix is transport-aware harness setup, with one real source of truth per concept.

## Why it matters

`e2e_rest` is the only transport that can't cheat — it must go through the server's real surfaces. The ledger is **tracked debt, not a permanent fixture**; it should shrink to ~0 as the harness gains transport-aware setup.

## Core invariant

The Gherkin scenario and the step definition stay transport-agnostic. The harness env realizes setup **intent** through the single real source-of-truth surface per concept. No in-process mock is the source of truth for any e2e-capable scenario.

## Source of truth per concept

| Concept | Real surface | e2e realization |
|---------|-------------|-----------------|
| creative formats | live creative agent | capture into a persisted checked-in fixture (refresh only when formats change); `ADCP_TESTING` path + harness read the same fixture |
| products / properties / principals / tenant billing | server DB | seed rows into the server DB |
| adapter delivery numbers | Mock adapter reading `DeliverySimulationConfig` from server DB | write the sim-config row; recover the stranded mechanism first |

## Ledger breakdown (312)

- **Formats** — 141 (uc005/uc006/get_products)
- **Adapter delivery** — 127 (uc004)
- **Account / billing** — 44 (uc011)

## Design + plan

Full design: `docs/test-redesign/e2e-rest-ledger-retirement.md` (added in the in-network-runner PR).

Tracked locally in beads under epic `salesagent-x0nl` with children for: DeliverySimulationConfig recovery, formats-cache capture, transport-aware env dispatch, and per-scenario triage. Triage must confirm tractability **empirically via a harness run** (uc011 is partly DB-backed already, yet all 44 remain — they touch something else in-process).

## Notes

- The ledger grows when `main` adds/renames scenarios (e.g. the recent merge added 19) and shrinks as the harness mechanisms land. This is expected; re-run triage after every `main` merge.
[...]

Issue #1270, opened 2026-05-04
refactor: add adapter_type discriminator to all reconciled *ProductConfig schemas


jrosendahl-opt:
Sub-issue of #1239 (Phase 2 of #996). Depends on #1267, #1268, #1269 being complete.

## Scope

Add `adapter_type: Literal[<adapter_type>]` discriminator field to each reconciled schema:
- `MockProductConfig` → `adapter_type: Literal["mock"]`
- `BroadstreetProductConfig` → `adapter_type: Literal["broadstreet"]`
- `GAMProductConfig` → `adapter_type: Literal["google_ad_manager"]`

This enables a Stage-3 discriminated union (`Annotated[Union[...], Field(discriminator="adapter_type")]`) to be a one-line addition in Phase 3 of #996.

## Acceptance criteria

- [ ] All three reconciled schemas have `adapter_type: Literal[...]` field
- [ ] Cross-adapter rejection test: `MockProductConfig.model_validate(broadstreet_config_dict)` raises `ValidationError`
- [ ] `make quality` clean

## Dependencies
- #1267 (Mock), #1268 (Broadstreet), #1269 (GAM) must be merged first

## Related
- Parent: #1239
- Phase 1 PR: #1007

PR #1420, opened 2026-06-11
ci: in-network Docker test runner + recover e2e_rest BDD 5th transport

KonstantinMirin:
## What

A fully **in-network** test runner: all six suites (unit, integration, bdd, admin, e2e, ui) run inside the Docker Compose network with **zero published host ports**, reaching Postgres/the server by service name (`postgres:5432`, `proxy:8000`). This removes the host-port collision surface that makes the host runner flaky on a crowded Docker host, and lets many stacks run concurrently. Proven equivalent to the host `run_all_tests` in both test counts and results.

It also **recovers and wires the `e2e_rest` BDD transport** (the real-HTTP 5th transport): scenarios dispatch over httpx → nginx → live server, with factory data written to the server's DB.

## Key changes

- `run_all_tests.sh` — the in-network runner (build creative image → bring up stack → `docker compose run --use-aliases tests tox -e <suites>` → extract JSON reports). Default = all 6 suites, serial.
- `run_all_tests_host.sh` — the previous host runner, preserved.
- `docker-compose.e2e.yml` — zero published host ports; in-network `tests` + pinned `creative-agent` services (runner profile). `docker-compose.e2e.ports.yml` — ports overlay for host paths only.
- `Dockerfile.test` — test image (bookworm, `uv sync --group dev --extra ui-tests`, prebaked Playwright chromium).
- `tox.ini` — env-controlled BDD worker count (`BDD_XDIST_N`: host `auto`, in-network `0`); per-suite DB split for e2e/ui.
- `tests/harness/dispatchers.py`, `tests/harness/_base.py`, `tests/bdd/conftest.py` — `RestE2EDispatcher`, env `e2e_config`/server-DB binding, `e2e_stack` detection, per-scenario reset-to-baseline.
- `src/app.py` — agent-card scheme honors `X-Forwarded-Proto` (nginx).

## Test integrity: e2e_rest fixture must error, not skip

`tests/bdd/conftest.py`: when `BDD_E2E_ENABLED=true` but the live stack is unreachable, the `ctx` fixture now **raises (ERROR)** instead of `pytest.skip()`. `e2e_rest` is only parametrized when the flag is set, so a missing stack means an explicitly-requested transport could not run — a skip there is a false green. Verified: with no stack, the affected e2e_rest scenarios report errors / 0 skips.

## Known limitation — the e2e_rest xfail ledger (tracked)

`e2e_rest` carries a 312-entry `xfail(strict=False)` ledger (`tests/bdd/e2e_rest_known_failures.txt`): scenarios that pass on all four in-process transports but fail only over real HTTP, because their setup uses in-process mock injection a separate server can't see. This is a leaky abstraction; the fix is transport-aware harness setup (one real source of truth per concept). Design: `docs/test-redesign/e2e-rest-ledger-retirement.md`. Tracked in #1418.

## Relationship to the #1234 CI overhaul (#1372) — local runner vs CI pipeline

This PR is a **local** test runner; it does not touch CI. It is complementary to @mkostromin-sigma's active #1234 CI series, **not** a competing pipeline.

- **This PR (local):** the in-network runner replaces `run_all_tests` for local/dev use. Its reason to exist is **host-port collisions on a shared dev machine** — it reaches Postgres/the server by service name with zero published ports, so many stacks run concurrently. GitHub Actions `services:`/composite actions cannot run locally, so #1372's machinery does not cover this use case.
- **#1372 (CI):** the authoritative GHA-native `ci.yml` (composite actions, `services:` Postgres, sharding, frozen required-checks guard) stays the source of truth for CI. A CI runner is ephemeral and isolated, so it has no port-collision pressure and does not need the in-network orchestration.

**Parity is preserved at the shared layer:** both orchestrations drive the same `tox` envs + `conftest` gating + harness/dispatcher + ledger. Orchestration differs by environment; test definitions are identical.

**One real gap remains — CI never runs the `e2e_rest` 5th transport** (the `bdd-tests` job in both `test.yml` and #1372's `ci.yml` runs 4 transports, no `BDD_E2E_ENABLED`). The agreed fix is **additive into #1372**: a single step in #1372's existing `e2e-tests` job that calls `docker compose -f docker-compose.e2e.yml run --use-aliases tests tox -e bdd` — reusing this PR's in-network plumbing (DNS, `--use-aliases` webhook callback, `-n0`) and leaving #1372's frozen check names untouched. Tracked in #1419; to be coordinated with @mkostromin-sigma. This migration is intentionally out of scope here.

## Duplication baseline note (honest, tracked)

This PR bumps the `tests` duplication baseline 98 → 99. It is **not** new duplication: #1420's files appear in **0** R0801 groups. pylint R0801 counts emergent clusters, so adding test files reshuffles clustering and recounts a pre-existing, byte-identical e2e webhook duplicate pair as its own group (`origin/main`=98 and this tree=99 in the same env). Root cause is pre-existing e2e-test debt (missing shared e2e fixtures) surfaced by #1372's count-based gate — tracked and to be fixed in #1423. The bump is documented inline in `.duplication-baseline`.

## Verification

Full in-network 6-suite run (post-merge with `origin/main`), all green:

| suite | passed | failed | xfailed |
|-------|-------:|-------:|--------:|
| unit | 4875 | 0 | 26 |
| integration | 1995 | 0 | 40 |
| bdd | 1604 | 0 | 9622 |
| admin | 86 | 0 | 3 |
| e2e | 90 | 0 | 2 |
| ui | 5 | 0 | 0 |

Equivalent to the host `run_all_tests` on the five non-bdd suites; bdd is the host's 4-transport result plus the additive `e2e_rest` 5th transport.

Follow-up work is tracked in #1418 (xfail-ledger retirement) and #1419 (CI in-network migration). Neither is closed by this PR.


ChrisHuie:
# Review — #1420 (verified at head `c3d8e60c`)

The mechanism holds up, and it's a real improvement: this recovers a genuine **5th BDD transport** (`e2e_rest`) that was dormant on `main` — `RestE2EDispatcher` (`tests/harness/dispatchers.py:188`) sends real HTTP through nginx → `UnifiedAuthMiddleware` → `resolve_identity()` → DB token lookup → `_impl`, reusing the in-process REST contract (`build_rest_body`/`parse_rest_response`), so it exercises the full stack rather than echoing a mock. The shared-DB isolation (per-scenario `TRUNCATE … RESTART IDENTITY CASCADE`, `conftest.py:2530`) and the hard-error-not-skip `ctx` fixture (`conftest.py:2560`, *"Refusing to skip — a skipped e2e test is a false green"*) are both the right calls. Suites run in-network by service name, sidestepping the host-port race. CI is green on this head and the in-network runner's verdict logic (`exit $RC` from `tox -e "$SUITES"`, no `coverage` env in the list) does **not** mask a failing suite.

Nothing below is a hard merge-blocker. Items 1–2 are worth folding in before merge; 3–4 are cheap correctness/coverage gaps; 5 is a latent false-green trap; the rest are nits, and the last section is the verification trail.

## 1. The blanket `strict=False` flip neutralizes the `#1270` validation-gap tripwires

**Where:** `tests/bdd/conftest.py:2387-2392`, inside the e2e_rest block.

**Context:** an `xfail(strict=True)` *fails the suite* if the test unexpectedly passes (an "xpass") — that's the signal to remove a stale xfail once production catches up. `strict=False` records the xpass and moves on. The collection hook is one `for item in items:` loop spanning `:378→:2392`, so per-transport markers added mid-loop are in `item.own_markers` by the time this block runs.

**Problem:** the second sub-block strips **every** xfail marker off an `[e2e_rest]` item and re-adds a single `strict=False` one, keeping only `xfails[0].kwargs.get("reason")`:

```python
xfails = [m for m in item.own_markers if m.name == "xfail"]
if xfails:
    item.own_markers = [m for m in item.own_markers if m.name != "xfail"]
    item.add_marker(pytest.mark.xfail(reason=xfails[0].kwargs.get("reason", "e2e_rest xfail"), strict=False))
```

This silently down-grades the e2e_rest `strict=True` markers authored a few hundred lines up — including the two documented production gaps:

- `conftest.py:1475-1481` — `FIXME(#1270): e2e_rest: Docker does not validate sampling_method — invalid value succeeds` (`strict=True`)
- `conftest.py:1502-1508` — `FIXME(#1270): … does not validate date range — invalid cases succeed` (`strict=True`)

These aren't environment-flake xfails; they assert *"production is wrong here, tell me when it's fixed."* Because `e2e_rest` only goes live in **this** PR (`BDD_E2E_ENABLED` was never `true` on `main`), the flip doesn't regress a working tripwire — but it disarms it at the very moment it would start working. When `#1270` is fixed, the scenario will xpass, `strict=False` swallows it, and the `FIXME` lingers forever. For items that are *both* in the ledger and carry a `:1475`/`:1502` strict marker, `xfails[0]` is the strict one, so the surviving reason is the validation-gap text and the ledger's "mock-incompatible" reason is dropped — two contradictory characterizations collapsed into one.

**Fix:** scope the non-strict policy to the ledger; leave authored `strict=True` markers alone so they can graduate.

```diff
-            xfails = [m for m in item.own_markers if m.name == "xfail"]
-            if xfails:
-                item.own_markers = [m for m in item.own_markers if m.name != "xfail"]
-                item.add_marker(
-                    pytest.mark.xfail(reason=xfails[0].kwargs.get("reason", "e2e_rest xfail"), strict=False)
-                )
+            # Only ledger-driven xfails are non-strict (live-stack xpass must not
+            # fail CI). A strict=True marker authored for a specific transport gap
+            # is a graduation tripwire — leave it strict so a fixed gap surfaces.
+            has_strict = any(m.name == "xfail" and m.kwargs.get("strict") for m in item.own_markers)
+            if nodeid in _E2E_REST_KNOWN_FAILURES and not has_strict:
+                pass  # ledger marker (added above) is already strict=False
```

(If the live stack genuinely produces *non-deterministic* xpasses on strict markers, that's worth saying explicitly per-marker rather than blanket-flipping — otherwise the flip is indistinguishable from hiding a real graduation.)

## 2. `./run_all_tests.sh ci` is broken; the old contract moved without updating callers

**Where:** `run_all_tests.sh:49,99` vs `run_all_tests_host.sh:19` and ~15 callers.

**Problem:** the new in-network runner reads `$1` as a tox suite-list (`SUITES="${1:-…}"` → `dc run … tox -e "$SUITES"`). On `main`, `run_all_tests.sh:19` was `MODE=${1:-ci}`. The old script was copied verbatim to `run_all_tests_host.sh` (it still self-identifies as `Usage: ./run_all_tests.sh ci …` in its header), but nothing else was repointed, so the documented gate now fails:

```
$ ./run_all_tests.sh ci
# → tox -e ci → "provided environments not found in configuration file: ci"
```

Callers still on the old contract: `Makefile:20,40` (`quality-full`, `test-full`), `CLAUDE.md:417,428,521,522,540`, `.claude/rules/patterns/testing-patterns.md`, `docs/development/{GETTING_STARTED,README}.md`, `scripts/README.md`, and — most visibly to a user — the pre-commit hook error message (`.pre-commit-hooks/check_migration_completeness.py:108`, *"Only ./run_all_tests.sh ci catches migration bugs"*) and the DB test-skip messages (`tests/conftest_db.py:69,72`, *"Run: ./run_all_tests.sh ci"*). All now point at a command that errors.

(Two minor follow-ons: the new `run_all_tests.sh` header docstring says `Usage: scripts/test-in-network.sh` — a path that doesn't exist; and the in-network runner drops the `scripts/security-audit.sh` step the host runner still runs, so the canonical local gate no longer runs uv-secure.)

**Fix:** either repoint the Makefile targets + docs at `run_all_tests_host.sh`, or have `run_all_tests.sh` recognize the `ci`/`quick` MODE tokens and delegate (cleanest, since `ci` ≈ "all suites" already). Note targeted runs (`run_all_tests.sh ci tests/path -k name`, used in several docs) need a home either way.

## 3. The `get_protocol` X-Forwarded-Proto change ships without a test, under a `ci:` title

**Where:** `src/app.py:297-310`.

**Problem:** this is the only production behavior change in the PR — the A2A agent-card scheme now derives from `X-Forwarded-Proto` instead of the localhost heuristic. A single-line revert turns no test red: nothing in `tests/` references `X-Forwarded-Proto`/`forwarded_proto`, and the one test importing `_create_dynamic_agent_card` (`tests/e2e/test_a2a_regression_prevention.py`) builds its own expected URL and never sets the header. All four branches (header present / comma-chain first-hop / invalid value → heuristic / absent → heuristic) are uncovered. `get_header_case_insensitive` returns `str | None`, so this is a pure unit test — no DB, no Docker. Separately, a production scheme fix under a `ci:` prefix won't surface in the release-please changelog's Bug Fixes; consider `fix:` (the security surface itself is bounded — see nit below).

**Fix:** add a parametrized unit test over the four branches.

## 4. `.duplication-baseline` 98→99: the justifying comment is factually wrong

**Where:** `.duplication-baseline:2`; `tests/e2e/test_a2a_webhook_payload_types.py:201`, `tests/e2e/test_delivery_webhooks_e2e.py:85`.

**Context:** pylint R0801 flags duplicate code blocks ≥ `min-similarity-lines`; the baseline is the allowed count.

**Problem:** the inline comment says the changed files *"appear in ZERO R0801 groups."* Running R0801 on both trees (base `68fd9066` = 98, head = 99) shows the new group **is** those two files — the PR added an identical `webhook_host = os.getenv("ADCP_WEBHOOK_HOST", "localhost")` / `webhook_url = f"http://{webhook_host}:{port}/webhook"` block to *both* fixtures, which crossed the similarity floor. The comment's *conclusion* (pre-existing fixture debt, tracked in #1423 — which is open) is correct; its stated fact isn't. Since the PR is editing both copies anyway, extracting a shared webhook-capture fixture would drop the count instead of bumping it.

**Fix:** correct the comment (the files *do* appear in R0801 groups; the duplication is pre-existing and the +1 is the threshold-crossing recount), or fold in the fixture extraction.

## 5. The `BDD_E2E_ENABLED ⟹ -n0` invariant is unguarded — a silent 5th-transport drop

**Where:** `tox.ini:117-130` (`-n {env:BDD_XDIST_N:auto}`) + `docker-compose.e2e.yml:139` (`BDD_XDIST_N: "0"`).

**Problem:** by the env's own comment, under xdist `BDD_E2E_ENABLED` doesn't reach the worker's `pytest_generate_tests`, so `e2e_rest` is silently dropped from collection; the PR correctly forces `-n0` in-network. But nothing *enforces* the coupling. If anyone runs the in-network bdd suite with `BDD_E2E_ENABLED=true` and a parallel `BDD_XDIST_N`, all 293 e2e_rest scenarios silently vanish and the suite goes green having never run the transport. Crucially, the `ctx` hard-error (your own anti-false-green guard) can't catch this — collection never happens, so the fixture never executes. This is the same false-green class the `ctx` guard exists to prevent, one level up. (I did not verify the *mechanism* — env vars normally inherit to xdist workers, so the real cause may be a controller/worker collection mismatch under `--dist loadfile` — but the mitigation and the missing guard are independent of the cause.)

**Fix:** assert the invariant in `pytest_configure` — if `BDD_E2E_ENABLED == "true"` and the session is distributed, raise (turn the silent drop into a hard error).

## Nits

- **X-Forwarded-Proto trust** (`src/app.py:304`): blast radius is only the *advertised* agent-card `url` scheme — no redirect/cookie-`Secure`/OAuth/SSRF sink consumes it (traced). nginx always sets the header (`config/nginx/nginx-multi-tenant.conf:121…`) and `Fly-Forwarded-Proto` wins on the Fly target (map at `:27-42`), so spoofing is bounded to a scheme downgrade on a non-Fly proxy that doesn't overwrite the header. The comment's claim *"the admin app already trusts X-Forwarded-Proto"* checks out (`src/admin/app.py:187`, `ProxyFix(x_proto=1)`). Optional: prefer `Fly-Forwarded-Proto` in-app, or note the precedence dependency.
- **Duplicate dep** (`pyproject.toml:100` vs `:108`): `pytest-xdist` is now declared twice in `dev` (`>=3.5.0` and `>=3.0.0`); collapse to one.
- **Stale runner headers**: `run_all_tests.sh:35` says `Usage: scripts/test-in-network.sh`; `run_all_tests_host.sh:7` still says `./run_all_tests.sh`.
- **Doc count drift**: `docs/test-redesign/e2e-rest-ledger-retirement.md:4,8` say "293", but `:89` and the live ledger say "312".
- **Shell `set` flags** differ across the four runners (`-uo` / `-eo` / `-eo` / `-euo`).
- **`TRUNCATE` re-seed** (`_reset_e2e_db`): `TenantFactory` auto-creates only USD; worth confirming no running e2e_rest scenario depends on a startup-seeded row the harness doesn't recreate (EUR/GBP `CurrencyLimit`, the bootstrap `ci-test-principal` token).

## Ledger framing (informational)

The 312-entry ledger (`tests/bdd/e2e_rest_known_failures.txt`) is an honest, tracked retirement (beads `salesagent-x0nl` + a per-mechanism breakdown), not a skip-to-green — and it nets *more* coverage than `main`. The one structural gap: ~20 `[…-invalid]` entries are *validation-rejection* cases whose Then-step asserts an error is returned and never touches a mock, so they're not "mock-incompatible" — they fail because the server accepts invalid input over REST (the documented `#1270` gap; `RestE2EDispatcher` sends raw JSON and the server validates). Worth splitting those out of the "mock-incompatible" bucket so a future server-side validation fix isn't permanently masked (it'd xpass under `strict=False`). Related: `dispatchers.py:240-254` wraps a non-JSON 5xx as a bare `AdCPError`, so a server *crash* satisfies a loose "expect an error" assertion — keep that in mind when triaging the ledger empirically. Unlike the duplication baseline and the guard allowlists, the ledger has no shrink-ratchet or stale-entry test; a fitness function (every nodeid resolves to a collected item; count is monotonic-non-increasing) would give it the same governance.

## Verification performed

At head `c3d8e60c` (clean tree): read the full collection hook, both runners, the e2e_rest dispatcher, `tox.ini`/`docker-compose.e2e.yml` env wiring, the nginx map chain, and the `get_protocol` flow. Confirmed `tox config -e ci` errors; `git show origin/main:run_all_tests.sh` = `MODE=${1:-ci}`; `BDD_E2E_ENABLED` set `true` only in `docker-compose.e2e.yml:113`, absent from `origin/main` except the `tox.ini:30` `pass_env` line (→ `e2e_rest` dormant on `main`). pylint R0801 on both trees → 98/99, new group = the two webhook fixtures. **Not run:** any Docker suite (CI green on this head taken as given); a live e2e_rest run to triage the ~20 `[…-invalid]` entries individually (their *test-side* mock-independence is read from the step defs; the *server-side* accept-invalid behavior is from the author's own `#1270` strict markers); the xdist `BDD_E2E_ENABLED` propagation mechanism; Fly's edge `Fly-Forwarded-Proto` behavior.

KonstantinMirin:
**Re: review — all items addressed at `d21a6c14b`** (pushed on top of the `c3d8e60c` you reviewed; CI green at that head). Thanks for the thorough pass — every finding turned out actionable, and three of them flushed out more than the note itself.

**1. `strict=False` flip neutralizing the #1270 tripwires** — Fixed in `cb8aa2400`. Confirmed your read: it's one loop, the tripwire markers (`:1475`/`:1502`) are in `own_markers` by the time the e2e_rest block runs. Rather than the "leave untouched if ledger-and-not-strict" form, I kept the single-marker collapse but made its strictness the OR of the sources: `strict = next((m for m in xfails if m.kwargs.get("strict")), None)` → `strict=strict is not None`, and the surviving `reason` is the strict marker's when both exist — so the ledger+tripwire collision now resolves to the *actionable* (strict, validation-gap) characterization instead of the mock-incompatible one. Regression test drives the real `pytest_collection_modifyitems` via a faithful stub item (`tests/unit/test_bdd_e2e_rest_xfail_policy.py`): asserts the sampling/date-range tripwires stay strict and ledger entries stay non-strict.

**2. `./run_all_tests.sh ci` broken** — Fixed in `5f101e758`. `run_all_tests.sh` now recognizes the MODE tokens: bare/`ci` → all six suites in-network; `quick` and targeted `ci <path> [args]` **delegate to `run_all_tests_host.sh`** (the host runner already implements no-Docker quick + the marker-filtered targeted run, so no reimplementation). A `RUN_ALL_TESTS_RESOLVE_ONLY` seam lets a unit test assert the contract without a stack (`tests/unit/test_run_all_tests_contract.py`, 5 cases). The two follow-ons are also fixed: the stale `scripts/test-in-network.sh` header + host-runner self-id (`d21a6c14b`), and the dropped `security-audit.sh` step is restored (`8c8a52183`).

**3. `get_protocol` X-Forwarded-Proto untested, under `ci:`** — Fixed in `c895b4974`, committed as `fix:` so it lands in the changelog (the original production change is already published upstream of this branch under `ci:`, so I couldn't retitle it — this commit carries the categorization + the coverage). Parametrized unit test over all four branches + the localhost heuristic leg; host fixed to a non-localhost value so a `http` outcome proves the header won (a revert to the heuristic reds it).

**4. `.duplication-baseline` comment factually wrong** — Fixed in `10ebf05f3` by folding in the extraction you suggested: `tests/e2e/_webhook_capture.py::run_webhook_capture_server`, with all three e2e receiver fixtures routed through it (the reference test pins `host="127.0.0.1"`). That removed your group **and** a latent reference-fixture group (`tests` 99 → **97**), and the false comment is deleted rather than corrected.

**5. `BDD_E2E_ENABLED ⟹ -n0` unguarded** — Fixed in `bf583bf58`. `pytest_configure` now raises `pytest.UsageError` when `BDD_E2E_ENABLED == "true"` and the session is distributed (`config.option.numprocesses` non-zero), pointing at `BDD_XDIST_N=0`. Cause-independent, as you noted. 7-case test.

**Nits:**
- X-Forwarded-Proto trust: agreed it's bounded to a scheme-downgrade on a non-Fly proxy with no redirect/cookie/SSRF sink; left as-is, noting the `Fly-Forwarded-Proto` precedence as possible future hardening.
- Duplicate `pytest-xdist` dep → collapsed to `>=3.5.0` (`d8791fa06`).
- Stale runner headers + doc count drift (293 vs 312) → `d21a6c14b` (live count reconciled to 308 — see below).
- Shell `set` flags → standardized to `set -euo pipefail` across all four runners, with `-e` guards on `run_all_tests.sh`'s RC capture (`0e8a3dc49`).
- `TRUNCATE` re-seed → `a1808152f`. Investigated: no running e2e_rest scenario depends on the wiped rows — the harness self-seeds identity via `PrincipalFactory` (not `ci-test-token`), the EUR scenarios are all xfail/unwired as e2e_rest today, GBP is unused. Filed the EUR-seeding requirement for when those wire up (`salesagent-jdy1`) and pinned the USD-only baseline with a guard test.

**Ledger framing:**
- Fitness function → `cc0ce08e7`: count is monotonic-non-increasing (ceiling) + every nodeid resolves to a collected item. The resolve check immediately caught **4 stale entries** from parametrize renames (312 → 308).
- The ~20 `[…-invalid]` validation-rejection cases mis-bucketed as "mock-incompatible": agreed — that reclassification (and the `dispatchers.py:240-254` crash-as-error concern) is server-side `#1270` territory and lives on the stacked harness PR #1430 (`salesagent-itn6`), where the ledger reaches 47 and the buckets get split with gate evidence. Not changed on this branch.

Two honest deltas from your suggestions: on **#1** I used the OR-strictness collapse rather than the exact diff (same intent — strict tripwires graduate — but it avoids pytest's multi-marker ambiguity); on **#3** the `ci:`→`fix:` is a new commit, not a retitle.

ChrisHuie:
# Re-review — #1420 (verified at head `d21a6c14b`)

Every item from the prior review is addressed at `d21a6c14b`, verified against the `c3d8e60c..d21a6c14b` diff and by driving the real production code, not by reading the commit messages. The two fixes that matter most are pinned by tests that exercise the real symbol — `test_run_all_tests_contract.py` subprocesses the actual `run_all_tests.sh` through the `RUN_ALL_TESTS_RESOLVE_ONLY` seam, and `test_bdd_e2e_rest_xfail_policy.py` imports and calls the real `pytest_collection_modifyitems` — and each reddens when the production line under test is mutated. Item #4 was over-delivered: rather than fixing the false comment, the webhook bootstrap was extracted to `tests/e2e/_webhook_capture.py` (3 callers), eliminating the duplicate pair, so the `tests` duplication baseline is a net **decrease 98→97** (fresh `pylint` R0801: `src=36`/`tests=97`) and the comment is gone.

No blockers. CI is green on this head; `make quality` passes locally (4797 passed). The findings below live in the fix commits and the original e2e_rest plumbing — two are worth folding in, the rest are nits.

## Prior-review items

| # | Item | Status |
|---|------|--------|
| 1 | `strict=False` neutralized the #1270 tripwires | **Fixed** `cb8aa2400`. OR-strictness collapse (`conftest.py:2408-2418`), scoped to `if is_e2e_rest`; explicit `strict=True` preserved (`xfail_strict` unset, so nothing relies on the ini default). The 4 both-set overlaps (`date_range…-invalid`, `principal_ownership…-invalid`) fail on a **server-enforcement** gap, not mock-incompatibility — so each still xpasses when the gap closes and the tripwire fires. |
| 2 | `./run_all_tests.sh ci` broken | **Fixed** `5f101e758`. `case` restores the MODE tokens (`ci`→all-six in-network; `quick`/`ci <path>`→`exec` host runner); the ~15 downstream callers work again transitively; security-audit step restored (`8c8a52183`). |
| 3 | X-Forwarded-Proto untested, under `ci:` | **Fixed** `c895b4974`. Separate `fix:` commit (changelog) + 4-branch + heuristic unit test. Production logic byte-identical to the reviewed head; the derived scheme has a single sink (the agent-card `url`), and a spoof on a non-overwriting proxy is bounded to a scheme downgrade in the spoofer's own card. |
| 4 | `.duplication-baseline` comment false | **Over-delivered** `10ebf05f3`. Duplication eliminated (not re-allowlisted); net **98→97**; false comment removed. |
| 5 | `BDD_E2E_ENABLED ⟹ -n0` unguarded | **Fixed** `bf583bf58`. `pytest_configure` raises `UsageError` (`conftest.py:101-116`), robust across `None`/`0`/`"0"`/`int`/`"auto"`. |
| nits | xdist dep · TRUNCATE re-seed · ledger fitness · shell flags | **Fixed**: dup `pytest-xdist` collapsed (`d8791fa06`); USD-only baseline pinned (`a1808152f`); ledger fitness added (`cc0ce08e7`), which caught 4 stale entries (312→308); all runner-script headers standardized to `set -euo pipefail`. Two sub-nits remain partial — below. |

## How the e2e_rest error path relates to in-process REST

`RestE2EDispatcher` reuses the in-process REST *contract* (`build_rest_body`/`parse_rest_response`/`parse_rest_error`) but re-implements the *error* path, and the two diverge on a ≥400 response:

| ≥400 response | `RestDispatcher` (in-process, `:154-162`) | `RestE2EDispatcher` (`:240-254`) |
|---|---|---|
| JSON error body | `parse_rest_error(...)` → coded error; **sets `wire_error_envelope=body`** | `parse_rest_error(...)` → coded error; **`wire_error_envelope` unset** |
| Non-JSON body (5xx crash / nginx HTML) | falls to outer `except` → bare error | crash-wrap → bare `AdCPError`, **no `error_code`** |

Finding A and the `wire_error_envelope` nit are the two cells where e2e_rest diverges.

## Should-fix

### A. e2e_rest reports a server crash as a correct rejection
**Where:** `tests/harness/dispatchers.py:243-253`; loose assertion at `tests/bdd/steps/domain/uc004_delivery.py:2606-2614`.
**Problem:** the crash-wrap is deliberate ("wrap as AdCPError so Then steps detect the error type and xfail spec-production gaps") — but it builds an `AdCPError` with no `error_code`, and the `"invalid"` Then-step asserts only `isinstance(error, (AdCPError, ValidationError))` (`:2613`) with no mock and no code check. So a genuine 5xx crash (or nginx HTML error) satisfies "the scenario correctly rejected invalid input." A well-formed 4xx routes through `parse_rest_error` and yields a coded error, and the narrower `error "<CODE>"` step (`:2617-2625`) checks `error.error_code` (`:2625`), so only the loose `"invalid"` assertions on a non-JSON response are fooled.
**Interacts with:** the graduation path in item #1 — these same `-invalid` scenarios are ledgered today because the server *accepts* the input (200, no error). When #1270/ownership land and they graduate off the ledger, a later 5xx regression on that input would false-pass the loose step (and could trip a strict tripwire as a false xpass).
**Fix:** require a non-empty `error_code` in the `"invalid"` step (or return a distinguishable error type the step rejects). The author scoped the broader reclassification to the stacked harness PR; if the hardening goes there too, add an in-tree marker at `dispatchers.py:243` so the deferral is discoverable — right now nothing in-tree records it.

### B. `e2e_host()` is introduced but bypassed at 7 sibling sites
**Where:** helper at `tests/e2e/conftest.py:25` (used 6× in-module); inline copies at `test_a2a_adcp_compliance.py:262`, `test_a2a_endpoints_working.py:26`, `test_a2a_regression_prevention.py:26`, `test_admin_bdd_e2e.py:75`, `test_landing_pages.py:34`, `tests/harness/admin_accounts.py:149`, `tests/ui/conftest.py:24`.
**Problem:** this PR adds `e2e_host()` (`os.getenv("ADCP_TEST_HOST", "localhost")`) and adopts it within `conftest.py`, but the same expression stays inlined at 7 sibling sites the PR also touched (2 via `os.environ.get`, 4 embedded in f-string URLs). A consolidation helper adopted in only its own module is the partial-extraction smell — the 8th copy of a concept the same PR just named.
**Fix:** import `e2e_host` at the 7 sites — `from tests.e2e.conftest import find_free_port` already establishes that conftest helpers are importable here — or move the host/URL helpers to the importable `tests/e2e/utils.py` (already in this PR).

## Nits

**Dead-ish / unwired**
- `McpE2EDispatcher`/`A2AE2EDispatcher` (`dispatchers.py:264,273`, registered `:297-298`) only `raise NotImplementedError`, and `E2E_MCP`/`E2E_A2A` are never parametrized (`conftest.py:2518` appends only `E2E_REST`). Not dead code in a vacuum — the enum members pre-exist on `main` while `main`'s `DISPATCHERS` had no entry for them (a latent `KeyError`), so this PR *completes* the dict with honest placeholders. To make them load-bearing now, add a test asserting `set(DISPATCHERS) == set(Transport)`; otherwise they can defer to the PR that wires those transports.

**Comment / doc accuracy**
- `conftest.py:2388-2393` — the comment lead reads "make EVERY e2e_rest xfail non-strict … this also clears the strict-xfail scenarios that now pass," contradicting `:2402-2407` and the code that now preserves `strict=True`. Related: the surviving tripwire `reason` is an earlier date-range C4 reason rather than the #1270 one the comment names, because the `T-UC-004-boundary-date-range` strict rows (`conftest.py:1170-1182`) match on `any(s in nodeid …)` (`:1047,1060`) and `rest-start_date equals end_date` is a substring of the `e2e_rest-…` nodeid — `strict=True` still holds, only the reason text differs.
- The by-mechanism breakdown still totals **312** (`e2e_rest_known_failures.txt:13`; `docs/test-redesign/e2e-rest-ledger-retirement.md:89`) while the live ledger and the fitness ceiling are **308** — the 4 pruned entries weren't subtracted from the buckets.
- `scripts/test-in-network.sh` is named as the launcher in 5 comments but doesn't exist (`docker-compose.e2e.yml:94,188`, `docker-compose.e2e.ports.yml:4`, `Dockerfile.test:11`, `scripts/test-stack.sh:83`); the in-network orchestration is inline in `run_all_tests.sh`. The header reference was fixed; these five weren't.
- New test docstrings and the contributor-facing retirement doc reference local task ids (9 `salesagent-…`) that don't resolve for outside readers; the public `#1423` is already used in the PR body.

**Consolidation / DRY**
- `WebhookReceiver.do_POST` (`test_adcp_reference_implementation.py:39`) and `DeliveryWebhookReceiver.do_POST` (`test_delivery_webhooks_e2e.py:39`) are near-byte-identical (differ only by one local var name and a docstring word). The extraction folded the server bootstrap but left these two receiver bodies duplicated; a shared default capture handler would fold them (the a2a handler genuinely differs).
- The six-suite list is materialized in 4 places with no guard tying them: `run_all_tests.sh:61`, `run_all_tests_host.sh:48`, `tox.ini:6`, `tests/unit/test_run_all_tests_contract.py:21` (two use a different element order). A 7th suite added to `tox.ini` would silently not run in-network. Derive `ALL_SUITES` from `tox -l`, or assert the script list equals tox's `env_list`.

**Test hygiene**
- `test_bdd_e2e_enabled_xdist_guard.py:28,41` parametrizes `numprocesses` as the strings `"auto"`/`"logical"`, which xdist resolves to an `int` before `pytest_configure` runs — the real shapes are `int ≥ 0` and `None`. The guard is correctly pinned for the real failure mode (the resolved int is covered and reddens under mutation); the string cases are surplus.
- `test_e2e_tenant_currency_baseline.py` asserts on `TenantFactory`/`CurrencyLimitFactory` attributes, not on `_reset_e2e_db` output — a defensible CI-visible proxy and a real tripwire, but the docstring frames it as guarding the reset path directly.
- `_webhook_capture.py:40,44` binds the receiver to `0.0.0.0`; the reference fixture previously bound `127.0.0.1` (its callback URL host is still pinned `127.0.0.1`). Harmless for a loopback-reached receiver and it unifies the three fixtures, but the loopback→all-interfaces relaxation is worth a note.
- `RestE2EDispatcher` never sets `result.wire_error_envelope` (`dispatchers.py:254`), unlike in-process REST (`:161`). Latent — no Then-step reads it today — but the project's mandated error-verification authority, `assert_envelope_shape(result.wire_error_envelope, …)`, is therefore unavailable on the new transport; populate it from `response.json()` in the JSON-error branch.

## Ledger classification (informational)
At least 4 ledger entries labeled "mock-incompatible" (`date_range…-invalid`, `principal_ownership…-invalid`) actually fail because the real server doesn't enforce validation/ownership (the #1270 gap), confirming the prior review's critique. The mislabel doesn't mask graduation — the strict tripwires are the real enforcement — but a future reader sees only "mock-incompatible." A one-line FIXME on the ledger block pointing at the reclassification follow-up would make the deferral discoverable.

## Verification performed
At head `d21a6c14b` (clean tree, asserted before and after): checked all 5 prior items against the `c3d8e60c..d21a6c14b` diff; read both REST dispatchers side by side and confirmed the ≥400 divergence; drove the real `pytest_collection_modifyitems`/`pytest_configure` and the `run_all_tests.sh` seam; confirmed `xfail_strict` unset and the collapse scoped to `is_e2e_rest`; read the loose vs coded `"invalid"` Then-steps (`uc004_delivery.py:2606-2625`) and the `s in nodeid` fragment match (`conftest.py:1047`); read the ledger overlaps and confirmed they are server-enforcement scenarios; confirmed the `E2E_*` enum members pre-date this PR and `main`'s `DISPATCHERS` lacked them; confirmed conftest helpers are importable (`find_free_port` on `main`); ran the 6 new unit files (23 passed) and `make quality` (4797 passed, 1 pre-existing networked skip); ran `pylint` R0801 → `src=36`/`tests=97` (matches the committed baseline; `main`'s tests base was 98); enumerated the 7 `ADCP_TEST_HOST` inline sites and the 6 `e2e_host()` callers; confirmed `run_all_tests_host.sh` keeps `nounset` on throughout (the `set +eo` toggles flip only `e`/`pipefail`); CI green on this exact head (29 checks). **Not run:** the in-network Docker stack — the live e2e_rest pass/xfail/xpass is unverified here (CI does not run `e2e_rest`; the new unit tests are the only CI-level proof, and they are faithful). The crash-as-error and missing-`wire_error_envelope` findings are traced from code, not observed on a live 5xx.

KonstantinMirin:
**Re: re-review — both should-fix + all nits addressed** (head `eee998a6`, rebased onto current main).

## Should-fix
- **A — crash-as-rejection:** the loose `"invalid"` Then-step (`uc004_delivery.py`) now requires a client-side rejection — a specific `error_code` (not `INTERNAL_ERROR`) and a non-5xx status — so a non-JSON 5xx wrapped as a bare `AdCPError` no longer passes as "correctly rejected." Faithful unit test (`tests/unit/test_e2e_rest_dispatcher.py`) drives the real `_assert_partition_or_boundary` with crash / 4xx-coded / pydantic errors; uc004 `-invalid` is green across all 4 in-process transports (**48 passed / 154 xfailed / 0 failed**). `faeaa0b0`.
- **B — e2e_host:** imported at all 7 sibling sites (the 2 `os.environ.get` + 4 f-string URL sites + `admin_accounts.py`); `ADCP_TEST_HOST` now appears only in the `e2e_host()` helper + docstrings. `14e458a5`.

## Nits
- **wire_error_envelope:** `RestE2EDispatcher` now sets it on the structured-JSON error branch (parity with in-process `RestDispatcher`), so `assert_envelope_shape()` works on e2e_rest. `faeaa0b0`.
- **DISPATCHERS completeness:** `test_dispatchers_cover_every_transport` asserts `set(DISPATCHERS) == set(Transport)`. `faeaa0b0`.
- **Ledger buckets:** reconciled to the live **308** (formats 140 / adapter-delivery 124 / account-billing 44) in both the `.txt` header and the retirement doc. `03b17fce`.
- **test-in-network.sh:** the 5 stale refs repointed to `run_all_tests.sh`. `185a47c4`.
- **Webhook DRY:** extracted a shared `WebhookCaptureHandler` into `_webhook_capture.py`; both receivers subclass it (a2a handler left as-is). `8ba09835`.
- **Six-suite list:** `test_six_suite_list_is_single_sourced` asserts the three script lists equal `tox.ini` `env_list` (as sets). `644f7255`.
- **Stale xfail-policy comment** (`conftest.py:2388`) corrected to match the strict-preserving code. `eaea65c6`.
- **Local ids:** the retirement doc + a test docstring now anchor on the public **#1423** (epic) / **#1430** (harness); internal sub-task ids roll up to #1423. (Mandated `FIXME(salesagent-*)` guard-allowlist comments and pre-existing `salesagent-18h.*` refs are out of scope.) `d5eaf207`.
- **Test hygiene:** xdist guard params reduced to the real `int` shapes `[1,4]`; deliberate `0.0.0.0` bind documented; currency-baseline docstring reframed as the CI-visible proxy it is. `8ba09835`.

Also **merged current `origin/main`** (the only conflict was `.duplication-baseline` → `{src:36, tests:97, scripts:0}`: main's new `scripts` key + this branch's `tests:97`), so the PR is mergeable again. `make quality` green locally (4805 passed).

ChrisHuie:
# Re-review — #1420 (verified at head `eee998a69`)

Round-2 items all land, verified by driving the real collection/dispatcher code at this
head (not commit messages): A's crash discrimination, B's `e2e_host` adoption, the
`wire_error_envelope` parity, the DISPATCHERS/six-suite guards, and the webhook extraction
are all present. No blockers.

One item worth folding in (one line), then nits.

## Worth folding in
`_UC004_GENUINE_XFAIL_ROWS` matches with bare `any(s in nodeid for s in substrings)`
(conftest.py:1415) and its substrings drop the bracket the other transport guards use
(`is_rest = "[rest-" in nodeid`, :400). So `"rest-interval=0 (below minimum)"` matches
`…[e2e_rest-interval=0 (below minimum)-…]`. Driving the real collection at this head,
6 ledger entries (attribution interval=0 / unit=weeks / model=last_click; reporting-dims
geo / limit=0 / limit-negative) come out `strict=True` with an "a2a/mcp/rest … (impl passes)"
reason — contradicting the block's own comment at :2407 ("Ledger-only items … stay
non-strict"), and the dedicated attribution/reporting-dims blocks deliberately exclude
e2e_rest. The date-range / sampling / ownership tripwires also surface a leaked in-process
reason instead of their own. Guarding the loop fixes both (re-collected to confirm: the 6
flip to `strict=False`, the tripwires keep `strict=True` and regain their own reasons):

```python
for tag, substrings, reason in _UC004_GENUINE_XFAIL_ROWS:
    if is_e2e_rest:
        break  # C4 rows describe the in-process a2a/mcp/rest boundary; e2e_rest is
               # governed by the ledger + the dedicated FIXME blocks below
    if tag in marker_names and (not substrings or any(s in nodeid for s in substrings)):
        item.add_marker(pytest.mark.xfail(reason=reason, strict=True))
        break
```

Latent today (CI doesn't run e2e_rest live; these currently xfail because the server
doesn't validate), so no current failure — it bites at the #1270 graduation.

## Nits
- The `status < 500` clause in the uc004 "invalid" step (uc004_delivery.py:2623) has no
  test: the dispatcher test only drives a 500/INTERNAL_ERROR crash, which both clauses
  reject, so dropping `status < 500` alone stays green. A structured 503 (the only shape
  that clause uniquely rejects) is untested. Add a 503 case if you want it pinned.
- `assert_envelope_shape` is named in the dispatcher comment / test docstring as "works on
  this transport," but nothing calls it on an e2e_rest envelope and the test's flat fake
  body wouldn't satisfy it. Soften the wording or add a real two-layer-envelope assertion.
- Server-DB-URL is resolved 3–4× with no shared helper; the post-reset diagnostic
  (conftest.py:409) uses only `ADCP_TEST_DB_HOST`/`PORT`, not `E2E_DATABASE_URL`, so
  in-network it targets the wrong host (print-only, so logs not outcomes). A sibling
  `e2e_db_url()` would mirror `e2e_host()`.
- The ~18 `-invalid]` ledger entries are server-validation gaps, not "mock-incompatible";
  a one-line breadcrumb on the ledger block pointing at the reclassification follow-up
  would make the deferral discoverable.

## Verification performed
Head `eee998a69`, isolated worktree: drove the real `pytest_collection_modifyitems`
(`BDD_E2E_ENABLED=true`, `--collect-only`, `-n0`) and read the final markers; applied the
loop guard and re-collected to confirm the fix; mutation-tested the uc004 status clause;
ran the dispatcher / agent-card / xfail-policy unit files (14 passed). Not run: the
in-network Docker stack, so the live xfail/xpass of the 6 items is inferred from the open
validation gap.

KonstantinMirin:
Pushed `8a9fa3c6`. All five items addressed — and verifying in-network turned up the real reason the e2e_rest delivery transport was dormant, fixed here too.

## Review items
- **1 — marker leak:** Guarded the `_UC004_GENUINE_XFAIL_ROWS` loop with `if not is_e2e_rest` so a bare `rest-…` row can't tail-match an `[e2e_rest-…]` nodeid. The old regression test only checked the strict bool, so it never caught the leak — strengthened it to drive the real collection hook over the six substring-collision ledger rows and assert both `strict=False` and the ledger reason (mutation-verified: it fails without the guard).
- **status<500:** Removed that clause. That check was the actual bug — it read a *reconstructed* exception, and reconstruction drops the HTTP status, so a genuine 400 `INVALID_REQUEST` looked like a 500 crash. The "invalid" Then-step now asserts on the real two-layer **wire envelope** captured per-transport (`ctx["wire_error_envelope"]`): a rejection is a non-`INTERNAL_ERROR`, non-transient, non-auth code with `recovery` correctable/terminal. Dispatcher unit test rewritten to that discriminator.
- **assert_envelope_shape:** the wire envelope is now genuinely consumed by the Then-step; comment corrected.
- **e2e_db_url:** added the shared helper; routed `live_server` and the post-reset diagnostic through it (it was reading `ADCP_TEST_DB_HOST/PORT` and ignoring `E2E_DATABASE_URL` in-network).
- **-invalid] ledger:** breadcrumbed (they're server-validation gaps, not mock-incompatible).

## What in-network verification surfaced (also fixed here)
- The e2e_rest delivery scenarios were all **401-ing** — a subdomain DRY bug: the ORM `TenantFactory.subdomain` kept underscores (`pub-test_tenant`) while the identity builder hyphenated (`pub-test-tenant`), and e2e_rest authenticates by sending the subdomain as `x-adcp-tenant`. A single `subdomain_for()` fixes it.
- With auth working, the `-invalid]` cases exposed a genuine gap: the REST boundary returned FastAPI's raw 422 instead of the AdCP envelope. Added a `RequestValidationError` → `INVALID_REQUEST` handler (per `error-code.json`, schema-constraint failures map to `INVALID_REQUEST`).
- BR-RULE-092 **INV-5** (campaign unit ⇒ interval 1) had no validator — added one in the delivery `_impl`; graduated the now-closed C10/C4 campaign markers.

## Verification
Unit green (4935 passed, 0 failed); full in-process bdd green across all four transports (1522 passed, 0 failed). The subdomain fix unblocks ~150 more e2e_rest scenarios that now pass — those need per-scenario inspection (scenario → business rule → use case → siblings → production) before their ledger entries are retired, so that's tracked as a follow-up rather than a bulk ledger edit here.

Note: the full CI workflow on this fork PR is waiting on a maintainer to approve the run.

ChrisHuie:
# Re-review — #1420 (verified at head `8a9fa3c62`)

Verified by running the real code at this head, not by reading commit messages. All five prior-review items are addressed; the two new production changes (`RequestValidationError → INVALID_REQUEST`, BR-RULE-092 INV-5) are spec-grounded and work. No blockers in the diff. The one thing gating this PR is the merge conflict, not a finding below.

**Merge state:** the PR is `CONFLICTING`/`DIRTY` against `main` (`.duplication-baseline`, `pyproject.toml`, `run_all_tests.sh`, `tests/bdd/steps/generic/_dispatch.py`). GitHub is therefore skipping every `pull_request` workflow — only the three `pull_request_target` checks ran, so there is no Test Suite / Quality Gate / Unit / Integration / BDD / E2E verdict on this head. Rebase/merge `main` to clear it before CI can validate.

## Prior items — addressed (verified)
- **Marker leak:** `if not is_e2e_rest` guard landed; driving the real collection hook, the six substring-collision ledger rows now resolve `strict=False` with the ledger reason. The strengthened regression test is mutation-verified (re-enabling the leak reds it across all six).
- **status<500 / wire discriminator:** the reconstruction-based check is gone; the "invalid" Then-step now reads the real two-layer `wire_error_envelope`. `e2e_db_url()` and `subdomain_for()` consolidations are complete; ledger breadcrumb added.
- **INV-5 + REST handler:** `AdCPInvalidRequestError` → `INVALID_REQUEST` / HTTP 400 / `recovery=correctable` (confirmed). Calling the validator directly rejects `unit=campaign, interval=2`; the boundary scenario passes on all four transports (REST returns 400 with the INVALID_REQUEST envelope). The Duration "interval must be 1 when unit is campaign" constraint is in the 3.1.0-beta.3 schema verbatim, and `INVALID_REQUEST` is the spec code for schema-constraint violations.

## Should-fix

### 1. `_assert_wire_rejection` accepts a server crash as a field rejection
**Where:** `tests/bdd/steps/domain/uc004_delivery.py` (`_assert_wire_rejection`, the `code not in _NON_REJECTION and recovery in ("correctable","terminal")` predicate).
**Problem:** a base `AdCPError` and `AdCPConfigurationError` both serialize to the wire as `{code: SERVICE_UNAVAILABLE, recovery: terminal}` (`INTERNAL_ERROR`/`CONFIGURATION_ERROR` are remapped to `SERVICE_UNAVAILABLE` by `ERROR_CODE_MAPPING`, and the base `_default_recovery` is `terminal`). `SERVICE_UNAVAILABLE` is not in `_NON_REJECTION`, and `terminal` is accepted — so driving the real assertion, a `{SERVICE_UNAVAILABLE, terminal}` envelope **passes** as "the field was correctly rejected." The docstring's claim that "a server crash … can never satisfy this" is not upheld. `{SERVICE_UNAVAILABLE, transient}` and `{INTERNAL_ERROR, terminal}` are correctly rejected; only the terminal-server-fault shape slips through. Latent today (the invalid-field scenarios emit `INVALID_REQUEST`, not crashes), but it removes the oracle the helper exists to provide.
**Fix:** treat a server-fault wire code as a non-rejection regardless of recovery, and add a case that pins it:
```python
_NON_REJECTION = {"INTERNAL_ERROR", "SERVICE_UNAVAILABLE", "AUTH_REQUIRED", "AUTH_TOKEN_INVALID"}
```
(or key the discriminator on HTTP status ≥ 500 rather than the code/recovery pair). Add a unit case feeding `{SERVICE_UNAVAILABLE, terminal}` and asserting it is rejected, so the gap stays closed.

### 2. Schema-failure error code differs across transports
**Where:** `src/app.py` `request_validation_error_handler` vs the MCP/A2A schema-rejection paths.
**Context:** the new handler maps a request-body `RequestValidationError` to `INVALID_REQUEST`. MCP and A2A reach a pydantic `ValidationError` (a `ValueError`) → `normalize_to_adcp_error` → `VALIDATION_ERROR` for the same malformed body.
**Problem:** a buyer gets `INVALID_REQUEST` on REST and `VALIDATION_ERROR` on MCP/A2A for the identical input. The spec documents both codes for schema-layer failures and the conformance storyboards don't grade which is canonical, so neither is non-conformant — but the three transports should agree. The handler docstring also states it brings REST into line with MCP/A2A, which is inaccurate on the code.
**Fix:** pick one code for request-body schema rejection and apply it on all three transports (the BDD feature pins `INVALID_REQUEST`; if that's the target, the MCP `except ValidationError` and the A2A `ValueError → VALIDATION_ERROR` mapping need to change too). This is a typed-family sweep across the tool wrappers, not just this tool — confirm against the BDD scenarios that pin `VALIDATION_ERROR` before changing.

## Nits
- **Stale campaign markers + suggestion mismatch.** INV-5 works (boundary passes), but `T-UC-004-attr-campaign-invalid` (strict=True) and the partition `campaign_interval_not_one` row stay xfailed with now-false reasons ("validation not implemented in _impl" / "production accepts invalid configs"). The dedicated scenario fails *only* on the suggestion text: the scenario asserts the suggestion contains `"interval must be 1"`, but the validator emits `"set interval to 1 when unit is 'campaign'"`. Align them and graduate both markers (like the boundary row):
  ```python
  suggestion="interval must be 1 when unit is 'campaign'",
  ```
- **Validation before auth.** `_validate_attribution_window(req.attribution_window)` runs before `require_identity`/`require_principal_id` in `_get_media_buy_delivery_impl`, so an unauthenticated caller with a bad campaign window gets `INVALID_REQUEST` instead of the auth error. Not a disclosure issue (the validator reads only `req`, and the constraint is public), and not spec-mandated — but `accounts.py` documents the opposite ordering ("require principal first so the canonical auth message surfaces"). Move the validate call after the auth gate for consistency.
- `getattr(window.unit, "value", window.unit)` reimplements the existing `resolve_enum_value()` helper (already imported-from in this module).
- `_validate_attribution_window(attribution_window)` has no parameter type annotation, so mypy can't check the attribute accesses.
- The `request_validation_error_handler` `loc` filter drops `body`/`query`/`path` at any position; only the leading segment is the FastAPI location prefix. No current trigger (no top-level request field is named that), but a body field so named would lose its `field` hint.

## Verification performed
At head `8a9fa3c62`, isolated worktree, running the real code (not commit messages):
- Drove the real BDD collection (`BDD_E2E_ENABLED=true`, `--collect-only`, `-n0`) → the six leaked ledger rows are now `strict=False`; mutation-tested the guard (re-enabling the leak reds the strengthened test).
- Called `_validate_attribution_window` directly → rejects `interval=2/campaign` with `INVALID_REQUEST`.
- Ran the campaign scenarios against a real Postgres: the **boundary** scenario passes on impl/a2a/mcp/rest (REST → 400 `INVALID_REQUEST`); the dedicated `attr-campaign-invalid` and partition `campaign_interval_not_one` **xfail** (strict satisfied — the dedicated fails on the suggestion-text mismatch above), confirmed on two separate databases.
- Drove the real `_assert_wire_rejection`: `{SERVICE_UNAVAILABLE, terminal}` accepted (the should-fix above), `{SERVICE_UNAVAILABLE, transient}` and `{INTERNAL_ERROR, terminal}` rejected.
- Confirmed `AdCPInvalidRequestError` → 400 / correctable; ran the new/changed unit files (pass).

**Not run:** the full `tox -e bdd` / `run_all_tests.sh ci` suite — the PR is DIRTY so CI is not running, and I ran targeted scenarios rather than the whole suite. The `SERVICE_UNAVAILABLE/terminal` hole's reachability in a live delivery path is traced, not observed. The `INVALID_REQUEST`-vs-`VALIDATION_ERROR` canonical-code question is ungraded by the conformance storyboards.

KonstantinMirin:
## Re-review response — should-fixes + nits in; the error-code sweep is scoped out (with reasons)

@ChrisHuie — thanks for the thorough passes. Short version: both should-fixes and all nits are addressed, the full in-network suite is green, and should-fix #2 turned out to be the visible edge of a repo-wide error-code question that we **deliberately do not try to close in this PR**. Laying out done vs deferred so you can review the bounded part and treat the rest as direction.

### What this PR actually is
The one real change is the **in-network Docker test setup** — running the suites *inside* the compose network so every dependency resolves by service name. Everything else was meant to be small fixes. Those fixes uncovered that **`e2e_rest` (the 5th, real-HTTP BDD transport) was never actually running** on main. Making it run was the right move — and it's what surfaced everything below. Good outcome, but the tail of it (error-code canonicalization) is genuinely unbounded, so we drew a line.

### Should-fixes + nits — done, self-contained
- **#1 — `_assert_wire_rejection` accepted a server fault** (`0d141c44a`): added `SERVICE_UNAVAILABLE` to the non-rejection set (it's what `INTERNAL_ERROR`/`CONFIGURATION_ERROR` remap to, with `terminal` recovery). Mutation-verified — `{SERVICE_UNAVAILABLE, terminal}` now raises; real `{*, correctable}` rejections still pass.
- **Nit — campaign suggestion + dedicated scenario** (`7a95372ee`): suggestion aligned to `interval must be 1…`; corrected the now-false marker reasons — the real cause is #1462 (the in-process path drops `attribution_window.post_click`, so INV-5 can't fire there; it *is* implemented).
- **Nit — DRY + annotation + loc** (`0a9f6f521`): `enum_value()` for the unit extraction; `_validate_attribution_window` annotated; `loc` strips only the leading `body/query/path` prefix.

### Should-fix #2 — what we found, and where we stopped
You said: pick one code, confirm against the BDD first, it's a typed-family sweep not just this tool. We did the confirm and traced it to the pinned AdCP schema:
- The graded `error-compliance.yaml` splits the codes by **failure semantics**: `schema_invalid` (missing/malformed/wrong-type) → `INVALID_REQUEST` (mandated for missing `idempotency_key`); a value/business-rule violation on a **well-formed** payload → `VALIDATION_ERROR` (canonical). **Grading accepts both** for the value cases — so it's consistency/canonicality, not non-conformance, exactly as you called it.
- The AdCP JSON schema defines the relevant enums identically (plain enums, no per-field code), so a differing pin is a derivation choice. **Our BDD is internally inconsistent** — the same "invalid enum value" class is `INVALID_REQUEST` in attribution_window but `VALIDATION_ERROR` in UC-006/UC-018.
- **This PR introduced one of those mis-derivations** — the new INV-5 + `RequestValidationError` handler pinned attribution_window as `INVALID_REQUEST`. So we corrected our own regression: attribution_window value/enum/range/INV-5 now emit canonical `VALIDATION_ERROR` across all transports — feature + `_validate_attribution_window` + REST handler (`b35fc230e`, `7a95372ee`), made **upstream in adcp-req** (branch `fix/attribution-window-error-code-validation`) and mirrored locally so it's regen-reproducible.

**Deferred — deliberately out of scope:** the rest is a repo-wide canonicalization — ~34 value/range/duplicate cases across UC-002/003/004/017/023/026 pinned `INVALID_REQUEST` (canonical `VALIDATION_ERROR`), one structural UC-018 case the other way, plus the general REST/A2A error-type classifier that makes all transports agree. It's **pre-existing** (those scenarios are unchanged from before this PR and pass), it's an all-at-once coordinated sweep with its own completeness/verification needs, and it isn't what this PR is about. Not denying it needs doing — it's tracked with the full work-list. It just shouldn't ride in the PR that introduces the in-network setup.

So for review: the items above are bounded and self-contained. #2's general form is **direction** — happy to align on the target code/policy, but the sweep itself is a separate change.

### Verification
Full in-network run at `0a9f6f521`, all 6 suites green, **0 failures / 0 errors**:
`unit 5115 · integration 2065 · bdd 1227 (7534 xfailed, 160 xpassed) · admin 86 · e2e 90 · ui 5`. `e2e_rest` now runs end-to-end; the attribution change is verified over the real server, and the ui inventory-smoke fix lands.

_(Correcting an earlier reply: the unit tests it referenced — `test_e2e_rest_dispatcher.py` / `test_bdd_e2e_rest_xfail_policy.py` — were removed; they faked transport behavior. Error-path assertions now live in the BDD harness/oracles.)_

ChrisHuie (approved):

PR #1424, opened 2026-06-11
fix(ci): operational helper layering and guard completeness (PR 3-3 of #1234)

mkostromin-sigma:
Closes #1234.

## Summary

Follow-up to merged #1372 (authoritative CI pipeline). Closes remaining architecture gaps from the post-merge review thread on #1372:

- Operational scripts import from `scripts/ci/`, not `tests/`
- Summary guard iterates **all** `summary.needs` jobs + pins `REQUIRED_SUMMARY_GATES` floor (PR #1299 class)
- Type Check runs `make typecheck` (CI/local parity)
- Smoke job no longer duplicates skip-decorator grep (`TestNoSkippedTests` is sole authority)
- Migration parse failures surface instead of vanishing from the graph
- `port_scan_start` requires explicit `int` pid; `_GETPID` bound at import for harness safety

**Head:** `55a119dc` · **Rebased on `main` after #1379 merge** · **Diff vs `main`:** 3 commits, ~16 files

## Changes

| Area | Change |
|------|--------|
| Script layering | `scripts/ci/migration_helpers.py`, `scripts/ci/workflow_helpers.py` |
| Pre-commit hook | imports shared migration helpers (no hand-synced duplicate) |
| Summary guard | `test_summary_gates_every_required_job` + `REQUIRED_SUMMARY_GATES` |
| CI/local parity | Type Check → `make typecheck`; drop unused `_setup-env` input |
| E2E fixture | `TypeError` on non-int pid; import-time `_GETPID` bind |

## Stack position

Merge **after ~~#1379~~ (merged)**. #1390 (PR4) rebased onto this branch.

Part of: #1234

## Review status

- **ChrisHuie** — no blockers; SHOULD-FIX 1 (summary floor) addressed in `55a119dc`
- **KonstantinMirin** — PAT-01…04 closure posted on merged #1372

## Test plan

- [x] Arch guards pass locally
- [x] CI green on rebased push (`55a119dc`)

### DIFF
--- a/.duplication-baseline
+++ b/.duplication-baseline
@@ -1,4 +1,5 @@
 {
   "src": 36,
-  "tests": 98
+  "tests": 98,
+  "scripts": 0
 }
--- a/.github/actions/_setup-env/action.yml
+++ b/.github/actions/_setup-env/action.yml
@@ -9,10 +9,6 @@ inputs:
     description: uv version to install.
     required: false
     default: "0.11.15"
-  dependency-group:
-    description: Dependency group passed to uv sync.
-    required: false
-    default: "dev"
 runs:
   using: composite
   steps:
@@ -36,4 +32,4 @@ runs:
 
     - name: Install dependencies
       shell: bash
-      run: uv sync --group "${{ inputs.dependency-group }}"
+      run: uv sync --group dev
--- a/.github/actions/_setup-env/action.yml
+++ b/.github/actions/_setup-env/action.yml
@@ -28,7 +28,7 @@ runs:
         path: |
           ~/.cache/uv
           .venv
-        key: ${{ runner.os }}-uv-v4-py${{ inputs.python-version }}-${{ hashFiles('uv.lock') }}
+        key: ${{ runner.os }}-uv-v4-py${{ hashFiles('.python-version') }}-${{ hashFiles('uv.lock') }}
 
     - name: Install dependencies
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
           uv-version: ${{ env.UV_VERSION }}
       - name: Run mypy
-        run: uv run python -m mypy src --config-file=mypy.ini
+        run: make typecheck
 
   schema-contract:
     name: Schema Contract
@@ -138,13 +138,6 @@ jobs:
           paths: tests/smoke/
           tb: short
           extra_args: -v
-      - name: Check for skipped tests
-        run: |
-          if grep -rE '@pytest\.mark\.skip(\(|[[:space:]]|$)' tests/ --include="test_*.py" | grep -v "skip_ci"; then
-            echo "Found skip decorators in tests!"
-            exit 1
-          fi
-          echo "No skip decorators found"
 
   unit-tests:
     name: Unit Tests
--- a/.pre-commit-hooks/check_code_duplication.py
+++ b/.pre-commit-hooks/check_code_duplication.py
@@ -76,7 +76,8 @@ def main() -> int:
     print("Scanning for code duplication (pylint R0801)...")
     src_count = count_duplications("src/")
     tests_count = count_duplications("tests/")
-    current = {"src": src_count, "tests": tests_count}
+    scripts_count = count_duplications("scripts/")
+    current = {"src": src_count, "tests": tests_count, "scripts": scripts_count}
 
     # Read baseline
     baseline = read_baseline(baseline_file)
@@ -85,21 +86,21 @@ def main() -> int:
     if baseline is None:
         print(f"No baseline found. Creating {BASELINE_FILE}:")
         print(f"  src/   = {src_count} duplicate blocks")
-        print(f"  tests/ = {tests_count} duplicate blocks")
+        print(f"  scripts/ = {scripts_count} duplicate blocks")
         write_baseline(baseline_file, current)
         return 0
 
     # Handle --update-baseline flag
     if args.update_baseline:
         print(
-            f"Updating baseline: src/ {baseline.get('src', '?')} -> {src_count}, tests/ {baseline.get('tests', '?')} -> {tests_count}"
+            f"Updating baseline: src/ {baseline.get('src', '?')} -> {src_count}, tests/ {baseline.get('tests', '?')} -> {tests_count}, scripts/ {baseline.get('scripts', '?')} -> {scripts_count}"
         )
         write_baseline(baseline_file, current)
         return 0
 
     # Compare counts
     failed = False
-    for scope in ("src", "tests"):
+    for scope in ("src", "tests", "scripts"):
         baseline_count = baseline.get(scope, 0)
         current_count = current[scope]
 
--- a/.pre-commit-hooks/check_migration_completeness.py
+++ b/.pre-commit-hooks/check_migration_completeness.py
@@ -8,34 +8,14 @@ runs against a real PostgreSQL database.
 Only `./run_all_tests.sh ci` catches migration bugs.
 """
 
-import ast
 import sys
 from pathlib import Path
 
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
 
-def _is_merge_migration(functions: dict[str, ast.FunctionDef]) -> bool:
-    """Check if this is a merge migration (both upgrade and downgrade are empty).
-
-    Merge migrations reconcile multiple alembic branch heads. They have no
-    schema changes — both upgrade() and downgrade() are intentionally empty.
-
-    NOTE: This duplicates logic in tests/unit/_migration_helpers.py:is_merge_migration().
-    Pre-commit hooks run via ``python script.py`` where sys.path[0] is the
-    script directory, not the project root — so ``from tests.unit...`` is not
-    importable.  Keep both implementations in sync when changing empty-body
-    detection logic.
-    """
-    if "upgrade" not in functions or "downgrade" not in functions:
-        return False
-
-    for name in ("upgrade", "downgrade"):
-        for stmt in functions[name].body:
-            if isinstance(stmt, ast.Pass):
-                continue
-            if isinstance(stmt, ast.Expr) and isinstance(stmt.value, ast.Constant):
-                continue
-            return False
-    return True
+from scripts.ci.migration_helpers import is_empty_body, is_merge_migration, parse_function, parse_migration_tree
 
 
 def check_migration_file(path: Path) -> list[str]:
@@ -44,41 +24,28 @@ def check_migration_file(path: Path) -> list[str]:
     Returns list of error messages (empty = OK).
     """
     errors = []
-    source = path.read_text()
-
     try:
-        tree = ast.parse(source, filename=str(path))
-    except SyntaxError as e:
-        errors.append(f"{path}: SyntaxError: {e}")
+        tree = parse_migration_tree(path)
+    except Exception as exc:
+        errors.append(f"{path}: {exc}")
         return errors
 
-    # Find upgrade() and downgrade() functions
-    functions = {}
-    for node in ast.walk(tree):
-        if isinstance(node, ast.FunctionDef) and node.name in ("upgrade", "downgrade"):
-            functions[node.name] = node
-
-    # Merge migrations have intentionally empty upgrade() + downgrade() — skip them
-    if _is_merge_migration(functions):
+    if is_merge_migration(tree):
         return errors
 
-    if "upgrade" not in functions:
+    upgrade = parse_function(tree, "upgrade")
+    downgrade = parse_function(tree, "downgrade")
+
+    if upgrade is None:
         errors.append(f"{path}: missing upgrade() function")
 
-    if "downgrade" not in functions:
+    if downgrade is None:
         errors.append(f"{path}: missing downgrade() function")
 
-    # Check for non-empty bodies (not just `pass` or docstring-only)
-    for name, node in functions.items():
-        body = node.body
-        # Filter out docstrings and pass statements
-        meaningful = [
-            stmt
-            for stmt in body
-            if not (isinstance(stmt, ast.Pass))
-            and not (isinstance(stmt, ast.Expr) and isinstance(stmt.value, ast.Constant))
-        ]
-        if not meaningful:
+    for name, node in (("upgrade", upgrade), ("downgrade", downgrade)):
+        if node is None:
+            continue
+        if is_empty_body(node):
             errors.append(f"{path}: {name}() is empty (only pass/docstring) — must contain migration logic")
 
     return errors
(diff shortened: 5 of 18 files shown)


ChrisHuie:
## Review — operational helper layering + guard completeness (verified at head `4ae7ce433`)

Reviewed against **`2ff10d1c..4ae7ce433`** — i.e. this PR's *net delta on top of #1379*, not the combined GitHub "Files changed" view. Because #1424 stacks on #1379 (`main 68fd9066 → …#1379… → 2ff10d1c → …#1424… → 4ae7ce433`), GitHub's diff folds in all of #1379; once #1379 merges, #1424's diff collapses to exactly the 16 files reviewed here.

**No blockers.** This is a clean, well-tested PR that does two things well: (1) moves the migration/workflow helpers out of `tests/` into the operational `scripts/ci/` layer so the pre-commit hook can finally *import* `is_merge_migration` instead of carrying its own copy; (2) replaces three silent-failure paths with loud ones — including the e2e port allocator's opaque `id(pid)` fallback, now an explicit type contract. The move is atomic (zero stale references; every consumer resolves), and the de-duplication is genuine (old files deleted, the hook's duplicate deleted, all "keep in sync" notes gone).

All three `SHOULD-FIX` items are guard-strength / DRY-completeness refinements with no active production impact today — and two of them connect directly to the two `SHOULD-FIX` items from the #1379 review.

<details>
<summary><b>Verified sound (with evidence)</b></summary>

| Area | Verification |
|---|---|
| Atomic move | Zero surviving references to `tests.unit._migration_helpers` / `tests.unit.workflow_helpers` across `*.py/*.sh/*.yml/*.md/*.toml/*.cfg/*.ini`; all 5 operational modules import; `parents[2]` still resolves to repo root after the move |
| Hook de-dup | `python3 .pre-commit-hooks/check_migration_completeness.py <file>` resolves `from scripts.ci.migration_helpers import …` under bare interpreter (exit 0 on valid, exit 1 + clear message on a deliberately broken migration); duplicate `_is_merge_migration` deleted |
| No-Quiet-Failures ×3 | `parse_migration_tree` raises `MigrationParseError` instead of swallowing `SyntaxError`; `load_ci_workflow` validates file/dict/`jobs`; `port_scan_start` raises `TypeError` on non-int pid (paired test flipped, mutation-confirmed) |
| CI/local parity | `make typecheck` == `uv run mypy src/ --config-file=mypy.ini` — same mypy, config, target as the replaced inline CI line; `dependency-group` input removal has **zero** callers in `.github/` |
| Skip consolidation | Removing the CI grep step is **not** a coverage loss: `TestNoSkippedTests.test_no_skip_decorators` covers the same `@pytest.mark.skip` surface (in fact broader), runs in the same `smoke-tests` job, and there are zero real decorator skips in the tree |
| Test suites @ head | `test_architecture_*` family + boundary guards 277 passed/0 failed; touched guards+smoke 44/0; consumer/guard unit tests for the moved helpers all green; full CI green at `4ae7ce433` |

</details>

---

### SHOULD-FIX 1 — the generalized summary gate dropped the #1299 regression floor

**Where:** `tests/unit/test_architecture_ci_suite_coverage.py` — `test_summary_gates_bdd_and_e2e` → `test_summary_gates_every_required_job`.

**Problem:** The rewrite is a real improvement — it now checks that *every* job listed in `summary.needs` is real and has its `needs.<job>.result` inspected. But the old test also hard-asserted that `bdd-tests` **and** `e2e-tests` are *members of* `summary.needs`, and the new one doesn't. That membership floor was the regression coverage for PR #1299 (a suite that ran but wasn't aggregated, so CI stayed green while main went red).

No active hole today — `summary.needs` currently includes both and both `.result`s are checked (verified). And a *fully deleted* job would still be caught by `test_required_check_names_are_frozen` (its rendered name vanishes). But the specific #1299 vector — a suite that **keeps running yet is removed from `summary.needs`** — now passes every guard (mutation-confirmed: drop `e2e-tests` from `needs` + its result line → new guard PASSES, old guard FAILED). Branch protection isn't enforced by any in-repo guard, so this is the only floor for that vector, and it's gone.

**Fix (keep the generalization, restore the floor):** add a module-level constant — mirroring `REQUIRED_RENDERED_CHECKS` in the frozen-checks guard — and assert it's a subset of `needs`:

```diff
+# A suite that runs but is absent from summary.needs leaves the gate soft (the
+# silent-breakage class this module exists for). The loop below only validates
+# what's already listed, so pin the required gate set explicitly.
+REQUIRED_SUMMARY_GATES = {"unit-tests", "integration-tests", "e2e-tests", "bdd-tests", "admin-ui-tests"}
```
```diff
     @pytest.mark.arch_guard
     def test_summary_gates_every_required_job(self):
-        """summary must depend on AND fail for every job in summary.needs."""
+        """summary must include every suite gate AND fail for every job in summary.needs."""
         workflow = load_ci_workflow()
         summary = workflow["jobs"]["summary"]
         needs = summary["needs"]
         check_text = " ".join(str(step.get("run", "")) for step in summary.get("steps", []))

         assert isinstance(needs, list) and needs, "summary.needs must list every upstream gate job."
+        missing = REQUIRED_SUMMARY_GATES - set(needs)
+        assert not missing, (
+            f"summary.needs dropped required suite gate(s) {sorted(missing)}. A suite that runs but is "
+            "absent from summary.needs leaves CI green on its failure."
+        )
         for required in needs:  # unchanged below: every listed job must be real + result-checked
             assert required in workflow["jobs"], (
                 f"summary.needs lists unknown job '{required}'. Update summary.needs or add the missing job definition."
             )
             token = f"needs.{required}.result"
             assert token in check_text, (
                 f"summary's result-check step does not inspect '{token}'. "
                 f"A failing {required} job would not fail CI even though it is listed in needs[]."
             )
```

---

### SHOULD-FIX 2 — make the skip-enforcement "single source of truth" actually enforceable

**Where:** `tests/unit/test_architecture_ci_suite_coverage.py` — `test_smoke_tests_do_not_duplicate_skip_guard` (+ the two skips in `tests/unit/test_architecture_single_migration_head.py`).

**Problem:** The PR removes the workflow skip-grep and declares `TestNoSkippedTests` the single source of truth — sound, but the new guard has two soft spots:

1. **Matches on the step *name*** (`"skipped tests" not in step.name`). The deleted step was identifiable by its `run:` content (the `@pytest.mark.skip` grep), not its name — so the exact step can be re-added under any other label and the guard passes (mutation-confirmed: re-added as *"Verify no test exclusions"* → PASS).
2. **Nothing pins that the SSoT exists.** `TestNoSkippedTests.test_no_skip_decorators` is defined only in `tests/smoke/test_smoke_basic.py` (a file this PR doesn't touch). If it's later renamed/deleted, skip enforcement silently evaporates while this guard stays green.

**Fix — match on content, and pin the survivor:**

```diff
     @pytest.mark.arch_guard
     def test_smoke_tests_do_not_duplicate_skip_guard(self):
         """Skip-decorator enforcement belongs in the smoke suite, not a workflow grep step."""
         smoke_job = load_ci_workflow()["jobs"]["smoke-tests"]
-        step_names = " ".join(str(step.get("name", "")) for step in smoke_job.get("steps", []))
-        assert "skipped tests" not in step_names.lower(), (
-            "smoke-tests must not grep for skip decorators; TestNoSkippedTests is the single source of truth."
-        )
+        step_text = " ".join(
+            f"{step.get('name', '')} {step.get('run', '')}" for step in smoke_job.get("steps", [])
+        )
+        assert "@pytest.mark.skip" not in step_text, (
+            "smoke-tests must not grep for skip decorators in a workflow step; "
+            "TestNoSkippedTests is the single source of truth."
+        )
+
+    def test_skip_guard_single_source_of_truth_exists(self):
+        """The SSoT the workflow delegates to must exist, or enforcement is unguarded."""
+        from tests.smoke.test_smoke_basic import TestNoSkippedTests
+
+        assert hasattr(TestNoSkippedTests, "test_no_skip_decorators"), (
+            "TestNoSkippedTests.test_no_skip_decorators is the declared single source of truth for skip enforcement."
+        )
```

**Optional, in-theme (behavior-neutral today):** `single_migration_head.py:55,73` end with `pytest.skip("No merge migration in graph.")` where their single-parent sibling (`:64`) uses `pytest.fail`. The branch is unreachable today (merge migrations exist in the graph, so the tests pass rather than skip), but the imperative-skip form is invisible to *every* current guard. Converting just these two to `pytest.fail("…merge-head roundtrip logic is unexercised.")` removes the silent-pass path. Note: **don't** add a blanket `pytest.skip(` guard — 50 of the 81 imperative skips in the suite are legitimate env-gated skips (server/creds unavailable), and would be false positives.

---

### SHOULD-FIX 3 — DRY: the two downgrade-target resolvers are still byte-identical *(carried from #1379)*

**Where:** `scripts/ci/migration_helpers.py` — `resolve_roundtrip_downgrade_target` vs `expected_heads_after_roundtrip_downgrade`.

This is the same item as #1379 `SHOULD-FIX 1` — it's pre-existing (byte-identical at the base) and **not new in #1424**. Flagging it here only because this PR relocates and edits this exact file (it trimmed both functions' docstrings/comments), which makes it the natural home for the dedup if you'd rather land it here than on #1379. Pure scope call.

**Problem:** Identical `head_revision is None` preamble + identical file-scan loop, differing only in return shape. And `resolve_…` has a dead branch — `if len(down_revisions) == 1: return down_revisions[0]` immediately followed by an identical `return down_revisions[0]`. (Not caught by `check_code_duplication.py` — pylint R0801 scores the file 10.00/10; the two functions diverge before its similarity threshold, so this is a human-judgment DRY call per the repo invariant.)

**Fix (gold standard — semantics verified arm-by-arm):** one private resolver; both public functions become thin shapers; the dead branch collapses; keep the merge-head rationale inline (it's the *only* reason the two differ).

```diff
+def _downgrade_parents_for_head(head_revision: str | None = None) -> list[str]:
+    """Resolve the head revision and return its down_revisions (parent list)."""
+    if head_revision is None:
+        heads = get_migration_heads()
+        if len(heads) != 1:
+            msg = f"Expected exactly 1 migration head, found {sorted(heads)}"
+            raise ValueError(msg)
+        head_revision = next(iter(heads))
+    for path in get_migration_files():
+        revision, down_revisions = extract_revision_info(path)
+        if revision != head_revision:
+            continue
+        if not down_revisions:
+            msg = f"Cannot downgrade from base revision {head_revision}"
+            raise ValueError(msg)
+        return list(down_revisions)
+    msg = f"Migration file not found for head revision {head_revision}"
+    raise ValueError(msg)
+
+
 def resolve_roundtrip_downgrade_target(head_revision: str | None = None) -> str:
     """Return an explicit Alembic downgrade target one step back from head."""
-    if head_revision is None:
-        ...                         # preamble + scan loop (duplicated) + dead branch
-    raise ValueError(f"Migration file not found for head revision {head_revision}")
+    # At a merge head, downgrading to any one parent undoes the merge and restores
+    # all branch tips; the first parent is a valid single target for `alembic downgrade`.
+    return _downgrade_parents_for_head(head_revision)[0]


 def expected_heads_after_roundtrip_downgrade(head_revision: str | None = None) -> set[str]:
     """Return alembic_version heads expected after CI roundtrip downgrade from head."""
-    if head_revision is None:
-        ...                         # preamble + scan loop (duplicated)
-    raise ValueError(f"Migration file not found for head revision {head_revision}")
+    # Merge heads restore every branch tip; single-parent heads land on their one parent.
+    return set(_downgrade_parents_for_head(head_revision))
```

---

### NITs

- **`tests/e2e/conftest.py:19` — the `_GETPID` comment overstates its trigger.** "unit suite autouse mocks leak through" — no autouse fixture anywhere patches `os.getpid` (all autouse fixtures swept; the only `os.getpid` patches are `with`-scoped harness self-tests that reverse on exit). The bind-at-import is genuinely useful defensive code (proven to survive a `patch("os.getpid")` in-process), but the stated mechanism doesn't exist. Suggest softening to the real rationale, e.g. *"Bind the real os.getpid at import so a later patch("os.getpid") in the same process can't leak a non-int into the pid-is-None default path."*
- **`check_code_duplication.py` no longer covers these helpers.** Its scan-set is `src/` + `tests/` only, so moving operational logic into `scripts/ci/` removes it from R0801 coverage. Moot for the SHOULD-FIX 3 dup (R0801 wouldn't catch it anyway), but future duplication under `scripts/ci/` is now unguarded — worth adding `scripts/` to the scan-set in a follow-up now that real logic lives there.
- **`.pre-commit-hooks/check_migration_completeness.py:29` — broad `except Exception`.** It's report-and-*fail* (loud), so not a quiet-failure — but narrowing to `except MigrationParseError` (the only documented raise) would let a genuinely unexpected error crash with a traceback instead of being downgraded to a per-file lint line.
- **Minor:** `workflow_helpers.load_ci_workflow` and friends use bare `dict` where `dict[str, Any]` is knowable (`from __future__ import annotations` is already present, so it's free); and the refactored hook still has no test exercising its own glue (`check_migration_file`/`main`) — both pre-existing, optional.

---

### Verification performed
- **Ran:** the touched-guard + smoke suites (44 passed) and the full `test_architecture_*` family + boundary guards (277 passed) at `4ae7ce433`; the moved-helper consumer unit tests (all green); import-resolution sweep across all 5 operational modules + a repo-wide stale-reference grep (zero); `pylint R0801` on `migration_helpers.py` (10.00/10); the pre-commit hook on valid + deliberately-broken migrations; `summary.needs` introspection (bdd+e2e gated today); mutation tests for the summary gate (both directions), the skip-dup guard (name-evasion), and `test_non_int_pid_raises` (revert→fail); decorator-skip census (zero) and imperative-skip census (81 total / 50 legit env-gated). All CI checks green for this head.
- **Not run:** `./run_all_tests.sh ci` end-to-end locally (relied on the green CI run at this head); a live `alembic upgrade/downgrade` roundtrip against Postgres (no new migration files in this delta; the `Migration Roundtrip` CI job is green); the full e2e suite for the harness↔`port_scan_start` autouse ordering interaction (the import-time `_GETPID` bind makes the default-arg path order-independent, which is the only behavior that matters here).

mkostromin-sigma:
@ChrisHuie — **SHOULD-FIX 1 addressed** in `4c4b2913`: restored `REQUIRED_SUMMARY_GATES` floor alongside the generalized `summary.needs` loop (PR #1299 regression class). SHOULD-FIX 2–3 noted for follow-up (skip SSoT existence test; migration helper DRY — can land here or on #1379 per your scope call).

CI re-running on latest push.

mkostromin-sigma:
**Rebased on `main` after #1379 merge** (`19d3efd5`). PR3-3-only delta: 3 commits → tip `55a119dc`. Conflicts resolved; mergeable again. CI running.

mkostromin-sigma:
@ChrisHuie — All review items from your PR3-3 review are addressed in `9f2f9300`:

- SHOULD-FIX 2: skip guard matches step content + SSoT existence test
- SHOULD-FIX 3: shared `_downgrade_parents_for_head` in migration_helpers
- Cache key: `hashFiles('.python-version')` in `_setup-env`
- NITs: `_GETPID` comment, `scripts/` in duplication scan, `MigrationParseError`-only hook, merge-head tests use `pytest.fail`

CI re-running on this tip.

ChrisHuie (approved):

PR #1417, opened 2026-06-10
feat: UC-002/003 validation wiring + canonical error-code reconciliation on adcp 5.7

KonstantinMirin:
## Summary

This PR began as "wire UC-002/UC-003 budget/currency/spend validation BDD scenarios" and grew into the full **adcp 5.7 upgrade + the validation/error-coherence/transport-dispatch reconciliation it forced**. The inflection point was the **adcp SDK 4.3 → 5.7 bump**: once the SDK's error vocabulary, request/response contracts, and validation surface shifted, "wire budget validation" stopped being isolatable and became a whole-surface reconciliation. The budget work (the original seed) is the earliest part of the branch and is intact.

It is large but **internally coherent and green**, and it's being landed as one unit so the branch stops diverging from other in-flight work. Follow-on cleanups are scoped as separate small PRs (see Follow-ups).

## Scale & how to review

~195 commits, ~756 files (+76k/−26k). Most of the file count is `tests/fixtures/` (pinned adcp schemas) and `tests/unit`. Suggested review path — by theme, in this order:

1. **adcp 4.3 → 5.7 upgrade** — the root change everything else follows from.
2. **Error-code reconciliation** — `tests/bdd/features/BR-UC-*.feature` + the matching production raise-sites; all expected codes now map to the pinned AdCP `error-code.json` enum.
3. **Production behavior** — `src/core/tools/media_buy_create.py`, `media_buy_update.py`, `exceptions.py`, `auth*.py`.
4. **BDD harness / wire dispatch** — `tests/bdd/steps/`, `tests/harness/`.
5. **Guards** — `tests/unit/test_architecture_*.py`.

## What's included (by theme)

- **adcp SDK 4.3.0 → 5.7.0** upgrade and the contract/type fallout.
- **Budget / currency / spend validation** (the original scope): wired BR-RULE-008 budget-positive, minimum-budget, currency scenarios for UC-002/003; extracted `validate_budget_positive()` to `financial_validation.py`.
[...]

Issue #1129, opened 2026-03-16
security: SSRF via outbound agent URL handling


sigma-antonnash:
### Severity
High

### Status
Confirmed

### Description
Server-side signals-agent integrations accept tenant-admin supplied `agent_url` values and use them for outbound network calls without SSRF-safe target restrictions. In the unified deployment, the relevant external admin routes are mounted under `/admin`, so the live paths are `/admin/tenant/<tenant_id>/signals-agents/add` and `/admin/tenant/<tenant_id>/signals-agents/<id>/test`. Testing on the current local deployment confirmed that loopback and link-local metadata URLs are accepted as provided and later requested by the server during the test flow. 

Evidence:
request:
1) Authenticate and create a signals agent with `agent_url=http://127.0.0.1:8000/health` via `POST /admin/tenant/default/signals-agents/add`
2) Create a second signals agent with `agent_url=http://169.254.169.254/latest/meta-data/`
3) Trigger `POST /admin/tenant/default/signals-agents/<id>/test` for both stored agents
4) Inspect `signals_agents` rows and application logs
response/output:
- both add requests returned `HTTP/1.1 302 Found`
- loopback URL was stored verbatim and `POST /admin/tenant/default/signals-agents/<id>/test` returned `HTTP/1.1 200 OK`
- application logs showed a server-side `GET /health` during that test
- metadata URL was stored verbatim and `POST /admin/tenant/default/signals-agents/<id>/test` returned `HTTP/1.1 504 Gateway Time-out`
- application logs showed attempted outbound requests to `http://169.254.169.254/latest/meta-data` and `http://169.254.169.254/latest/meta-data/mcp`

Code:
- `salesagent-main-latest/src/app.py:271-275`
- `salesagent-main-latest/src/admin/app.py:397`
- `salesagent-main-latest/src/admin/blueprints/signals_agents.py:95-123`
- `salesagent-main-latest/src/admin/blueprints/signals_agents.py:184-205`
- `salesagent-main-latest/src/admin/blueprints/signals_agents.py:238-267`
- `salesagent-main-latest/src/core/signals_agent_registry.py:343-363`

```
agent_url = request.form.get("agent_url", "").strip()
...
agent = SignalsAgent(
    tenant_id=tenant_id,
    agent_url=agent_url,
    name=name,
    ...
)
...
result = asyncio.run(registry.test_connection(agent.agent_url, auth=auth, auth_header=agent.auth_header))
```

### Impact
[...]

PR #1430, opened 2026-06-12
test: transport-aware BDD harness — retire e2e_rest ledger 312→7 (#1418)

KonstantinMirin:
> Built on `ci/in-network-test-runner` (#1420, merged 2026-06-18) — the diff now contains only the harness-wiring work. Tracking issue: #1418.

## What

Makes the BDD harness **transport-aware** so the `e2e_rest` transport (real HTTP to a separate live server) can run scenarios that previously only worked in-process. Retires the bulk of the `e2e_rest` xfail ledger by fixing the real cause rather than parking failures.

**Core invariant:** the Gherkin scenario and step definitions stay transport-agnostic; the harness `env` realizes setup *intent* through the one real source-of-truth surface per concept. No in-process mock is the source of truth for any e2e-capable scenario.

| Concept | Real surface (what e2e now writes to) |
|---|---|
| creative formats | reference fixture captured from the pinned creative agent (the server reads the same fixture in test mode; the in-process harness resolver falls back to the same catalog) |
| products / principals / tenant / billing | the server's Postgres (seeded rows) |
| production DB reads inside e2e scenarios | `_db_scope_for` points production's engine at the server DB for the scenario duration (`_production_db_pointed_at`) — every e2e-capable `_harness_env` branch, structurally |
| adapter delivery numbers | a DB row the mock adapter reads (`DeliverySimulationConfig`) |
| adapter fault injection | DB-backed `test_behavior` flags (`fail_on_create` / `fail_on_update` / `fail_on_upload`) via one shared `_raise_injected_failure` |
| genuinely unrealizable setups | declared `E2EUnsupportedSetup` at the env (reasoned xfail), not a bare nodeid |

## Result

**`e2e_rest` ledger: 312 → 11.** Final in-network gate (all six suites): **green, 0 failed** (`innet_060726_1150`; two e2e tests hit a transient 429 from the public creative agent and pass on re-run).

- Wave 3 (#1418): 312 → 47 (163 graduated, 98 env-owned `E2EUnsupportedSetup`, 4 stale).
- `salesagent-jdy1` — **fully closed**: M1 REST-422 wire shape, M3 idempotent tenant seeding, M4 webhook-observability tag declarations pre-merge; M2 resolved by the #1417-merge graduations plus the wrong-DB fix below (47 → 21 together with the main/#1417 merges).
- This branch absorbed **two semantic merges**: `c81760acd` (origin/main) and `7d789cddd` (PR #1417 media-buy-validation-refactor), each reconciled diff-by-diff.
- Items 1–4 of the ledger plan: the **wrong-DB class closed structurally** (`_db_scope_for` across all 14 e2e-capable branches), uc011 agent identities authenticated (+ a structural guard banning swallowed Given dispatch errors), uc002 ext-o/ext-p auto-approval seeding, uc004 roas/cpa Then steps wired (entry converted to a tag-declared gap, then implemented — see below). 21 → 12, all graduations confirmed in-network (`innet_050726_2030`).
- ext-q upload mock-fidelity closed (fail_on_upload + catalog format + the async format-resolution bug below): 12 → **11**.

**Remaining 11, all ticketed:** 9 `salesagent-itn6` real spec-production validation gaps (live server accepts input the spec rejects) + 2 inline-xfail production gaps (seller attribution default, push_notification_config acknowledgement). Side ledgers: 13 uc004 webhook tag-family e2e xpasses unlocked by the DB repoint await per-scenario retirement (`salesagent-e2aw`); an obsolete buyer_refs boundary xpass awaits retirement (`salesagent-8efn`); `salesagent-93ou` (`update_media_buy` FK twin) remains open.

## Production bugs the work flushed out (fixed here, repro-first)

1. **`sync_creatives` FK-violation 500** — lenient mode inserted assignment rows for creatives that *failed* validation. Now existence-gated; per-creative failure reported. (`src/core/tools/creatives/_assignments.py`)
2. **`TenantFactory` subdomain split-brain** — DB row wrote `pub-test_tenant`, identity dict produced `pub-test-tenant`; one bug = 142 failures. Unified via `tenant_subdomain()`.
3. **`ui` suite-order lottery** — the `ui` suite relied on a boot-seeded tenant the bdd suite truncates. The fixture now owns its precondition.
4. **Pending-approval path skipped creative validation** — missing `creative_ids` returned a pending *success* (silently dropped assignments) and format mismatch emitted `VALIDATION_ERROR` while the auto path emits `CREATIVE_REJECTED` for the same input. Both paths now share `_pre_validate_package_creatives`, rejecting identically **before anything persists**; pinned by REST wire-envelope regression tests + a structural guard.
5. **`_get_format_spec_sync` never worked inside the async server** — bare `asyncio.run()` raises in a running event loop, so every server-side format lookup failed and valid catalog formats were rejected as unknown. Now uses `run_async_in_sync_context`.
6. **Delivery `aggregated_totals` missing spec scalars** — `roas`, `cost_per_acquisition`, `conversion_value`, `conversions` (BR-RULE-220 INV-2 / `core/delivery-metrics.json`) are now computed with omit-on-zero semantics; the graded scenario passes on all transports.

## Key mechanisms added

- `tests/harness/_realize.py` — `realize_e2e` dispatch + `E2EUnsupportedSetup` declaration.
- `_production_db_pointed_at` + `_db_scope_for` — production's cached engine follows the scenario's real DB on every e2e-capable branch.
- `DeliverySimulationConfig` ORM + repository + migration; mock adapter reads the row first.
- 54-format reference fixture + `scripts/refresh-reference-formats.py`; in-process and e2e format resolution agree by construction.
- DB-backed adapter fault injection incl. `fail_on_upload`; structural guards: swallowed-Given-error ban, shared-creative-validation call-site pin.

## Verification

Final gate `innet_060726_1150` @ `6b5c0575a`: unit / integration / bdd / admin / ui green; bdd **1802 passed, 0 failed**, 11 ledger xfails, 16 xpasses (all pre-tracked retirement items); e2e 89 passed + 2 transient external-429 failures that pass on re-run. Every ledger removal is backed by in-network gate evidence; no speculative removals.


ChrisHuie:
**#1430 — verified at `c214bfa45`.** The transport-aware harness holds: the production-DB repoint disposes + restores under a hard serial guard (`tests/bdd/conftest.py:126`, `:2989`), `realize_e2e` re-raises real errors rather than masking them (`tests/harness/_realize.py:82`), and the migration is single-head/reversible. Two things gate merge, one is a real bug; the rest is test-infra and nits.

**Merge blocker — stale #1417 embed.** This branch merged #1417 at `cecc90e12`, but #1417 has advanced 16 commits to `cd3391f1` (empty-update→INVALID_REQUEST, suggestion-parity one-shared-boundary, dup step-def removal). Most of this diff is #1417's, and diffing against current #1417 makes its newer fixes read as deletions here — e.g. `_build_get_media_buy_delivery_request` exists only in `cd3391f1`, and `performance.py` has no commit unique to this branch. Rebase onto current #1417 (or land #1417 first), then review; the error/suggestion/`media_buy_update` surface belongs to #1417's review, not this one.

**High — `sync_creatives` still 500s + leaks across principals.** `creatives` has a composite PK `(creative_id, tenant_id, principal_id)` (`models.py:666`), but the existence gate `get_creative_by_id` (`creative.py:450`) filters tenant-only ("no principal filter"), while the INSERT uses the requester's `principal_id` (`_assignments.py:~250`). A principal referencing a `creative_id` owned only by another principal passes the gate on the other's row → INSERT of `(C, tenant, requester)` → raw FK `IntegrityError`/500; and `_assignments.py:161-162` reads the other principal's `agent_url`/`format`, surfacing it in the requester's `assignment_errors` — a cross-principal read on a P0 boundary. The recent fix closed the absent-creative case, not this sibling. `_sync.py:215` already principal-scopes the parallel lookup under a `# SECURITY` comment.

```diff
-    def get_creative_by_id(self, creative_id: str) -> Creative | None:
+    def get_creative_by_id(self, creative_id: str, principal_id: str) -> Creative | None:
```
```diff
-                creative_row = assignment_repo.get_creative_by_id(creative_id)
+                creative_row = assignment_repo.get_creative_by_id(creative_id, principal_id)
```

Cross-principal references then resolve to `None` → the existing lenient-skip / strict-raise path reports them instead of 500-ing. No multi-principal test exists — add one asserting a clean `SyncCreativesResponse` when a principal references another's `creative_id`.

**Medium — transient format-agent errors become terminal `CREATIVE_REJECTED`.** The async fix (`_get_format_spec_sync`, `media_buy_create.py:271`) is correct, but its `except Exception: return None` now swallows the typed transient errors `get_format` raises (`creative_agent_registry.py:504/509/527/531`) → `:362` treats `None` as unknown → terminal rejection. A rate-limited or timed-out creative agent tells the buyer their creative is permanently rejected instead of `SERVICE_UNAVAILABLE`.

```diff
     try:
         return run_async_in_sync_context(registry.get_format(agent_url, format_id))
+    except AdCPError:
+        raise
     except Exception as e:
         logger.warning(f"Could not fetch format {format_id} from {agent_url}: {e}")
         return None
```

**Medium — the 312→11 ledger isn't graded in CI.** `BDD_E2E_ENABLED` is set only in `docker-compose.e2e.yml` (manual `run_all_tests.sh`); no workflow runs the `e2e_rest` transport, and it is a hard error under the xdist the CI bdd job uses (`conftest.py:126`). Removals are backed by the `innet_*` runs, not CI — a still-failing entry can be dropped (edit the ledger + `EXPECTED_LEDGER` together) and pass all of CI. Either add a required in-network job or state in the ledger header that graduation is manual-gate-verified (and drop "self-enforcing"). Relatedly the fitness ceiling is inert: `_LEDGER_CEILING = 309` vs an actual `11` (raised from 308) — set it to `len(EXPECTED_LEDGER)`.

**Medium — per-scenario e2e engine never disposed.** `tests/harness/_base.py:1071` creates `self._e2e_engine`; `__exit__` closes the session but never `dispose()`s it. Pre-existing, but graduating 312→11 makes ~300 more scenarios build e2e envs per run — accumulation toward `max_connections`.

```diff
+        if self._e2e_engine is not None:
+            self._e2e_engine.dispose()
+            self._e2e_engine = None
```

**Medium — "agree by construction" leaks on the default format.** The default `fmt="display_300x250"` (`given_media_buy.py:1996`) is not in `reference_formats.json` (which has `display_300x250_image`); it resolves in-process via the seeded `_format_specs` but the server rejects it over e2e — the ext-q scenario had to hand-override to the real id. Reseed the harness/product defaults to a real catalog id so a new scenario can't pass in-process and mask over e2e.

**Delivery `aggregated_totals` — spec-checked, contract-safe.** The new scalars are all real `AggregatedTotals` fields in adcp 5.7.0 / spec 3.1.0-beta.3 (`extra="allow"`, omit-on-absent — a measured `0` is still emitted), so the additions don't break existing shapes; `roas = Σconversion_value/Σspend`, `cost_per_acquisition = Σspend/Σconversions`, both denominator-guarded (`media_buy_delivery.py:563`). One nit: `BR-RULE-220 INV-2` is cited but appears only in-branch, not in the spec or the UC-004 obligations — cite the authoritative rule or reframe the comment as a platform choice.

**Nits.** The new `test_architecture_*` guards cite internal ids (`salesagent-kk15`, `0wby`) instead of `# FIXME(#<gh-issue>)` — those don't resolve for outside contributors. `_log_safe` strips only CR/LF (adequate for the CodeQL alert). `delivery_simulation_configs` is read un-gated on every mock delivery poll (`mock_ad_server.py:1193`) with no production writer — gate the read behind `ADCP_TESTING`. The share-validation guard counts calls module-wide, not per-branch, so both calls landing in one branch would still pass.

`--- Verified at c214bfa45:`
- **Ran:** `alembic heads` (single head `64f0fff7d954`); adcp SDK field check against 3.1.0-beta.3 (`AggregatedTotals` defines roas/cost_per_acquisition/conversion_value/conversions/clicks); `git grep`/`git log -S` attribution (`_build_get_media_buy_delivery_request` only in `cd3391f1`; `performance.py` unchanged on this branch; the FK gate + composite PK); `git grep BDD_E2E_ENABLED -- .github/` (none); ledger count 11 vs ceiling 309; fixture parse (`display_300x250_image` present, `display_300x250` absent); traced `get_creative_by_id`→INSERT→FK, `get_format` transient raises, and the `xfail` report hook.
- **Not run:** the in-network BDD gate — the 312→11 pass/fail rests on the `innet_*` runs cited in the commits, not observed here; a live multi-principal repro for the FK-500 (traced end-to-end, not observed).

KonstantinMirin:
@ChrisHuie thanks for the thorough review — everything verified and addressed. Finding → commit:

**Merge blocker — stale #1417 embed:** merged the current #1417 head (`385388920`, which already contains main's #1506/#1543) in `60ff7fb58`. Done as a semantic per-file merge of the 23-commit delta rather than a textual one; the three real conflicts were in exactly the files this branch also touched (`mock_ad_server`, `given_media_buy`, the type-ignore baseline). Notable reconciliation: both branches had independently extracted the injected-failure raise into a helper — kept this branch's flag-based API (the new `fail_on_upload` site needs it) with #1417's suggestion-promotion body. The diff should now read clean against #1417.

**High — `sync_creatives` cross-principal FK-500 + leak:** confirmed live exactly as you traced (gate passes on the other principal's row, INSERT dies on `ForeignKeyViolation`). Fixed in `555069ffe` with your sketch — `get_creative_by_id` now takes the full composite key — plus a repro-first multi-principal integration test (REST wire, observed the 500 before the fix) and a structural guard requiring `principal_id` on buyer-path `select(Creative)` lookups (`admin_*` exempt by design).

Follow-through on why this survived the suite: the BR-RULE-034 scenarios existed but were dormant (never wired into the harness), and the storyboard has no cross-principal *assignment* case at all. `e76061fad` wires the BR-RULE-034 family (which surfaced and fixed two latent step bugs the dormancy was hiding), and `4fbbe8b58` adds a hand-authored local scenario covering the assignment-reference surface until the upstream storyboard grows the equivalent. Both pass on all transports including `e2e_rest` in-network.

**Medium — transient agent errors → terminal `CREATIVE_REJECTED`:** `eb5bba06e`, your diff plus tests pinning both directions (typed transient error propagates; genuinely-unknown format still returns None).

**Medium — ledger not graded in CI / inert ceiling:** `a39ce36a4`. Ceiling is now `len(EXPECTED_LEDGER)` (binds at 11), and the ledger header states explicitly that graduation is manual-gate-verified with `innet_*` evidence — "self-enforcing" phrasing dropped. A required in-network CI job is the remaining option; leaving that as a separate infra decision.

**Medium — e2e engine never disposed:** `8f9118a93`.

**Medium — default format "agree by construction" leak:** agreed, and deliberately **deferred** — flipping the harness/product default off the synthetic `display_300x250` touches a seam referenced across ~84 files and needs its own pass with a full-BDD gate rather than riding this batch. Tracked with a full implementation plan.

**Nits:** `dd5f09079` — internal tracker ids in comments replaced with PR references, the `delivery_simulation_configs` read is now gated behind `ADCP_TESTING`, the share-validation guard counts per-branch (one call inside the manual-approval branch, one outside), and the roas/cpa comment now cites the response schema definitions directly instead of the in-branch rule id.

Verification: every fix landed repro-first where a repro was expressible, and the full in-network gate on the final head (`innet_080726_2131`) is green across all six suites — bdd 1819 passed / 0 failed (the newly wired scenarios pass against the live server), e2e 91/0, unit 5391/0, integration 2149/0. The 16 remaining bdd xpasses are the known retirement work-list (webhook tag-family + the obsolete buyer_refs boundary), tracked separately.

ChrisHuie:
**#1430 re-review — verified at `a9f11e98`.** All eight items from the prior round are addressed in code: the sync-path cross-principal fix, the transient-error propagation, the disposed e2e engine, the bound ledger ceiling + exact-set lock, and the nits all landed; the aggregated_totals scalars are spec-correct (formulas match `get-media-buy-delivery-response.json` in 3.1.0-beta.3, omit-when-uncomputable). Two things gate merge — one is a real cross-principal defect that the sync-path fix didn't reach; the rest is CI governance, test strength, and prose.

**Merge blocker — the cross-principal creative fix stops at `sync_creatives`; the `create_media_buy` path is unfixed and the new guard can't see it.** `creatives` has composite PK `(creative_id, tenant_id, principal_id)`, but three lookups on the buyer create path filter tenant-only:

- `media_buy_create.py:384` (`_validate_creatives_before_adapter_call`, both create paths + the approve path)
- `media_buy_create.py:2992` (auto-approval assignment)
- `media_buy_create.py:3747` (manual-approval branch)

All three are `select(DBCreative).where(DBCreative.tenant_id == …, DBCreative.creative_id.in_(…))`, fed by buyer-supplied `package.creative_ids`, with no principal-ownership check anywhere earlier on the path. So principal P1 referencing P2's `creative_id` in the same tenant matches P2's row, and:

- **Leak** — an invalid foreign creative surfaces P2's `status`/`format`/`agent_url` in the buyer-facing `CREATIVE_REJECTED` text (`:414`, `:426`).
- **FK-500 + DB-text disclosure** — a valid foreign creative passes validation, then `DBAssignment(principal_id=P1, creative_id=C2)` (`:3026`) violates `fk_creative_assignments_creative_composite` (the assignment FK needs the full `(creative_id, tenant_id, principal_id)` triple). There's no `IntegrityError` handler around the insert, so it hits the outer `except Exception` → `raise AdCPAdapterError(f"Failed to create media buy: {str(e)}")` (`:4269`) → the buyer gets a `SERVICE_UNAVAILABLE`/transient 502 carrying the raw psycopg2 constraint text, and a buyer-input error is misclassified as a transient server fault.

This is the same class fixed for `sync_creatives`, and `17db1a026` actually extended this tenant-only validation onto the pending path while the sibling was being fixed principal-scoped. The reason it passes CI: `test_architecture_creative_lookup_principal_scoped` scans only `creative.py` (and its matcher models `Creative.<attr> ==`/`filter_by`, not the `DBCreative` alias or `.in_()`), and `test_architecture_no_raw_select` doesn't see `select(DBCreative)` because it resolves the aliased name against `ORM_MODEL_NAMES`. The guard reads "class closed" while three sites violate it.

Fix — thread `principal_id` into the validation and scope all three selects (foreign refs then resolve to "not found" and take the existing `VALIDATION_ERROR`/correctable path, matching `_assignments.py:101`); `identity.principal_id` is in scope on the buyer paths and `media_buy.principal_id` on the approve path:

```diff
-def _validate_creatives_before_adapter_call(packages, tenant_id, session=None):
+def _validate_creatives_before_adapter_call(packages, tenant_id, principal_id, session=None):
     ...
-    stmt = select(DBCreative).where(
-        DBCreative.tenant_id == tenant_id, DBCreative.creative_id.in_(list(all_creative_ids))
-    )
+    stmt = select(DBCreative).where(
+        DBCreative.tenant_id == tenant_id,
+        DBCreative.principal_id == principal_id,
+        DBCreative.creative_id.in_(list(all_creative_ids)),
+    )
```

Then widen the guard (scan the buyer-path modules, resolve import aliases, match `.in_()`) so it reddens on these sites until fixed, and add a create-path scenario that references another principal's `creative_id`. This should block merge: it's a reachable cross-principal isolation break on the primary buyer path, the fix is small, and the PR's own guard currently reports the class closed. Exploitability is low — it requires knowing another principal's `creative_id` within the tenant, and the leak is creative metadata rather than asset content — but that bounds how easily it's triggered, not whether it should ship.

**Medium — the `e2e_rest` ledger is graded but not merge-blocking.** The `bdd-in-network` job runs, but it isn't a required check and isn't in the required `Summary` job's `needs:`/validation (which lists `bdd-tests` but not `bdd-in-network`). So a de-ledgered-but-still-failing scenario reddens only a non-blocking check while `Summary` stays green and the PR can merge. Add `bdd-in-network` to `Summary.needs` + its result check (or to the branch ruleset's required checks) to give the anti-cheat teeth. Related: the exact-set lock + ratchet cover only the 11 nodeid entries — the scenarios moved to `E2EUnsupportedSetup` env declarations and the conftest `is_e2e_rest` tag-xfails have no lock or monotonic count, so a still-failing scenario relocated there leaves the tracked count silently. Consider giving that set the same ratchet.

**Medium — the transient-error fix isn't verified on the wire.** `except AdCPError: raise` (`media_buy_create.py:332`) is correct, but the only test asserts `pytest.raises(AdCPServiceUnavailableError)` on `_get_format_spec_sync` directly, while the docstring claims the buyer sees `SERVICE_UNAVAILABLE` on the wire. A future intermediate handler swallowing it would leave the test green. Add one create-flow REST test asserting `assert_envelope_shape(result.wire_error_envelope, "SERVICE_UNAVAILABLE", recovery="transient")`.

**Medium — the roas/cpa Then steps recompute the expected value from production's own output.** `uc004_delivery.py:1352,1379` derive `expected = total_conversion_value / total_spend` by summing production's per-delivery output, so a same-source extraction bug passes. The Given already fixes the inputs (`conversions=10`, `conversion_value=500` → `roas=2.0`, `cpa=25.0`); assert against those literals (`pytest.approx(2.0)` / `(25.0)`) instead.

**Medium — the cross-principal repro and the local isolation scenario under-assert.** `test_creative_sync_behavioral.py:688` reddens on a fix revert via the FK-500 (a valid pin), but despite `_or_leak` in its name it never asserts the owner's `status`/`format`/`agent_url` is absent from the response; and the local uc006 scenario asserts only `len(rows) == 0` with no positive-control assignment in the same request (every assignment-creating uc006 scenario is XFAIL), so a zero-rows result from an unrelated downstream gate would also pass. Add a same-principal assignment that must create a row, and assert the foreign creative's fields don't appear in `assignment_errors`. (`test_creative_sync_behavioral.py:734` also uses `get_db_session()` in the test body.)

**Medium — `_log_safe` wasn't swept to the create path.** The log-injection fix is applied in `_assignments.py` but the buyer `creative_id` is logged unsanitized at `media_buy_create.py:3011,3015,3021,3036` — same CWE-117 class. Hoisting `_log_safe` to a shared util and applying it there closes the siblings.

**Medium — the roas/cpa spec-grounding comment overclaims conformance coverage.** `media_buy_delivery.py:655-656` grounds the value in a "uc004 aggregated-roas-and-cpa storyboard scenario," but that's a local BDD scenario — the value is conformance-ungraded on the general delivery path (the delivery storyboard grades schema/type only; value-grading lives in the capability-gated `performance_buy_flow[_roas]` scenarios). Reword to the accurate grounding so the comment doesn't read as spec-graded coverage it doesn't have.

**Nits.**
- `principal_id: str | None` on `get_creative_by_id`/`_process_assignments` is always non-None in production — type it `str` so a future caller that drops it trips mypy rather than silently matching `IS NULL`.
- "terminal CREATIVE_REJECTED" in `media_buy_create.py:337` (and `test_create_media_buy_behavioral.py:810`) — `CREATIVE_REJECTED` is `correctable`, not terminal; the code and assertions are right, only the prose is off.
- `BR-RULE-220 INV-2` citation still in `test_delivery_schema_contracts.py:78` (the production comment was corrected; the test wasn't); the `core/delivery-metrics.json` reference alongside it is the real grounding.
- Stale `_get_mock_formats()` references in `uc005_format_id_roundtrip.py:18,20` (renamed to `_get_reference_formats`). Migration `64f0fff7d954`: `idx_delivery_sim_tenant` on `[tenant_id]` duplicates the composite-PK leading column, and `media_buy_id` carries no FK (orphan rows if a buy is deleted — fine for a test-seeding table, worth a deliberate note).

**Confirmed sound.** The sync-path fix (`get_creative_by_id(creative_id, principal_id)`) closes both the leak and the FK-500 on that path; the BR-RULE-034 isolation steps genuinely grade isolation (the read-back uses a principal-blind query then asserts `principal_id`). The realize dispatch catches only `E2EUnsupportedSetup` and lets real errors propagate; the e2e path writes and reads the server DB. The `DeliverySimulationConfig` seam is production-safe (`ADCP_TESTING` gate is the first statement in the read; no `src/` writer for the fault-injection flags; the repo is tenant-scoped). Migration is reversible and single-head with the model matching exactly. The duplication baseline decreased (src 36→35, tests 89→87).

**Mergeable state.** Currently DIRTY — the conflict in `tests/bdd/test_uc018_list_creatives.py` is from #1551 wiring the UC-018 cross-principal scenario on main; it's complementary to this PR's UC-006 one (UC-018 uses real `wire_response` id-set matching and is the stronger of the two), so the resolution should preserve both surfaces. The green CI predates #1551, so it needs to re-run after the merge is resolved.

`--- Verified at a9f11e98:`
- **Ran:** traced the three create-path selects (`:384/:2992/:3747`) → assignment insert (`:3026`) → composite FK → the `except Exception` re-raise (`:4269`), and confirmed no principal-ownership gate exists earlier on the path; confirmed the two guards are blind to these sites by inspecting their scan set + matchers; the ruleset's required checks (`E2E Tests`, `Summary`, `Unit Tests`) and the `Summary` job's `needs:` (no `bdd-in-network`); the aggregated_totals formulas against 3.1.0-beta.3; the ledger exact-set lock (11 entries) and the disposed engine; `alembic heads` (single) and the migration model match; the duplication baseline via the R0801 checker; the `#1551` merge conflict via `merge-tree`.
- **Not run:** a live two-principal repro of the create-path leak/FK-500 (traced end-to-end from the schema + code, not observed — the FK violation is a deterministic consequence of the composite PK + the tenant-only insert); the in-network `bdd-in-network` stack (the 312→11 pass over real HTTP rests on the cited in-network runs, not observed here); `alembic downgrade` and the full suite.

KonstantinMirin:
**Re-review response — all items landed at `5a19bcb02`** (17 commits since `a9f11e98`). Every item from both rounds is addressed in code; one deliberate wire-code change you haven't seen is flagged at the bottom. Full suite on the CI box at this head: unit 5461, integration 2170, bdd in-process 1613, bdd in-network 356, e2e 91, admin 86, ui 5 — all green, xpass delta vs the prior gate none/none. Verification command log: `.claude/notes/verification-commands-20260710-execute-batch.md`.

**Merge blocker — create-path cross-principal scoping** → `8a2ffeaeb`. All three selects you cited (`:384`, manual- and auto-assignment blocks) are principal-scoped via a new `CreativeRepository.get_by_ids(creative_ids, principal_id)`; `update_media_buy` no longer calls `admin_get_by_ids` on a buyer path; `_fetch_creative_approvals` (list path) is scoped too. Foreign refs resolve to the uniform not-found and take the correctable path, matching `_assignments.py`. The guard is rewritten as you specified: scans `repositories/creative.py` + all of `src/core/tools/**`, resolves import aliases (`DBCreative`/`CreativeModel`/`DBAssignment`), per-QUERY granularity, `.in_()` matching, and a second detector bans `admin_*` repo calls from buyer tool code (explicit shrink-only allowlist replaces the name exemption). Widening the guard surfaced three more sites beyond your list — `CreativeAssignmentRepository.get_existing`, the update find-or-create lookup, and the approval-exec creative reload — all fixed in the same commit. Repro-first multi-principal wire tests cover auto/manual create (FK path + both leak variants), update, and assert the not-found envelope with owner-field-absence markers; the live run before the fix reproduced the FK-500-as-transient-502 you traced.

**bdd-in-network not merge-blocking + unlocked xfail hatches** → `70e0de9d8`. The job is in the required `Summary`'s `needs:` + result validation, and a new exact-set guard (`test_architecture_e2e_rest_escape_hatches.py`) locks both relocation routes you called out — the 20 conftest `is_e2e_rest` routes and the 3 `E2EUnsupportedSetup` declarations — with meta-tests on the live detectors. This push is the first run with the check required.

**Transient-error fix not verified on the wire** → `e4da87dfe`. Went past the single test you asked for: the fetch is now ONE helper (`format_resolver.fetch_format_spec`; `_get_format_spec_sync` delegates), so the `_validation.py` bare-except rewrap you'd have hit next is gone; the `run_async_in_sync_context` bridge no longer re-runs a consumed coroutine on coroutine-raised `RuntimeError`; and the sync per-item handler re-raises transient-recovery AdCPErrors so a 429 is `RATE_LIMITED` on the wire on both tools (correctable/terminal stay per-item — unknown-format semantics unchanged). Wire-envelope tests assert `RATE_LIMITED` and `SERVICE_UNAVAILABLE` (`recovery=transient`) on create_media_buy REST and sync_creatives, plus the transport matrix. A new guard bans sync `<x>.get_format` call sites outside `format_resolver`/the registry so the copies can't re-diverge.

**roas/cpa steps recompute from production output** → `2d90dd6bc`. Then steps assert the Given-derived literals (`pytest.approx(2.0)` / `(25.0)`). The `BR-RULE-220 INV-2` citations are gone from the step, `test_delivery_schema_contracts.py`, and the production comment, re-grounded in the pinned response schema; the comment no longer claims storyboard grading.

**Under-asserting cross-principal tests** → `2d525e1ce` (+ `9de142fbd`). The sync test now proves the `_or_leak` claim (owner `status`/`format`/`agent_url` absence asserted) and carries a same-request positive control that must create a row; the DB-less transient tests moved to unit behind one shared registry seam; `get_db_session()` read-backs replaced with env read-backs. Relatedly, `9de142fbd` fixes the false in-code claim behind that test's shape: `creatives=[] + assignments` no longer returns bare success — the merge synthesizes `action='failed'` (errors + `assignment_errors`, status omitted) / `'unchanged'` (+`assigned_to`) entries per the pinned response schema, whose success branch forbids a response-level errors array.

**`_log_safe` not swept** → `8a9e9b5e9`. Hoisted to `logging_config.py`, applied at all 20 unsanitized `creative_id` sites in `media_buy_create.py`.

**Nits** — all landed: `principal_id` strict `str` on `get_creative_by_id`/`_process_assignments` (`d910b3e3a`, which also delegates 4 duplicate cross-repo lookups to their owning repositories); "terminal CREATIVE_REJECTED" prose corrected to correctable and stale `_get_mock_formats` refs fixed (`312f3b229`); the PK-redundant `idx_delivery_sim_tenant` is dropped by new migration `823974a5553e` (committed migrations stay immutable; roundtrip-verified, single head).

**Mergeable state** → `bff2fdbe4` merges current main; the #1551 UC-018 conflict is resolved preserving BOTH cross-principal surfaces (UC-018's `wire_response` id-set matching and UC-006's), with `6b1334ac7` isolating the UC-018 BR-RULE-034 scenarios in a fresh per-scenario tenant.

**One wire-code change you haven't reviewed** — `287c93099`: the strict-mode assignment gate now emits `CREATIVE_NOT_FOUND` (was `VALIDATION_ERROR`) for unowned/unknown creative_ids. Spec-checked against the pinned enum, which defines `CREATIVE_NOT_FOUND` and mandates uniform treatment for unowned ids; only the SDK's helper table lacks the code, so the wire-safe set is now SDK baseline + pinned-spec supplement. Red-then-green wire repro included. The create/update surfaces deliberately still emit `CREATIVE_REJECTED` — those cells are storyboard-graded, so the reconciliation goes upstream first (tracked) rather than churning graded behavior here.

Not run by me: your live two-principal create-path repro concern is now covered by observed runs — the FK-500 and both leaks were reproduced live before the fix and are pinned by the wire tests. Outstanding known flake: the e2e ref-impl test syncs against the live public creative agent and 429s under repeated suite load (passed this run; tracked with a repoint-to-pinned-agent fix direction).

ChrisHuie:
**#1430 re-review — verified at `0085f3c04`.** Every item from the prior round (F1–F9) is addressed in code, and I re-verified the load-bearing ones empirically rather than by inspection — the cross-principal fix, the transient-error wire behavior, and the ledger lock all redden a real test when the fix is reverted. The new `CREATIVE_NOT_FOUND` wire-code change is spec-grounded and correct. No blocker in this round. What remains is a cluster of small in-PR consistency items and a set of genuinely pre-existing issues that should be filed separately rather than charged against this PR.

## Prior round — closed, and mutation-verified

- **Create-path cross-principal (the prior merge-blocker) — closed.** All buyer-tool creative lookups are principal-scoped through `CreativeRepository.get_by_ids(creative_ids, principal_id)` (create pre-validate `media_buy_create.py:364`, execute-approved `:1114`, auto/manual assign `:3009`/`:3772`, update gate `media_buy_update.py:206`, sync `_assignments.py:97`, list `media_buy_list.py:563`). Reverting the `principal_id` predicate reddens all five `test_cross_principal_creative_scoping.py` cases (auto-approve, manual-approval, status-leak, format-leak, update) and the sync repro (`test_creative_sync_behavioral.py:793`) fails with the exact original `ForeignKeyViolation` 500. The rewritten guard resolves the `DBCreative`/`CreativeModel`/`DBAssignment` aliases, matches `.in_()`, is per-query, and its two detectors both redden on mutation (drop `principal_id` → query detector RED; inject `admin_get_by_ids` into a buyer tool → admin-call detector RED). `push_creative_to_existing_buy`'s allowlisted `admin_get_by_id` is legitimate — it gates on `status in {approved, active}` and re-derives the principal from `creative.principal_id`.
- **`bdd-in-network` merge-blocking — closed.** It's now in the required `Summary` job's `needs:` (`ci.yml:575`) and its result is validated (`:597`, `exit 1` on non-success). A de-ledgered-but-still-failing scenario now reddens `Summary`.
- **Transient error on the wire — verified, not just asserted.** `test_format_fetch_transient_errors.py` asserts `assert_envelope_shape(result.wire_error_envelope, RATE_LIMITED/SERVICE_UNAVAILABLE, recovery="transient")` on real REST wire for both `create_media_buy` and `sync_creatives`; three independent mutations (swallow in `fetch_format_spec`, disable the sync re-raise, swallow on the create path) each redden the right tests. The single-fetch-path consolidation and the `run_async_in_sync_context` loop-narrowing are both pinned.
- **roas/cpa (F5)** now assert the Given-derived literals `pytest.approx(2.0)`/`(25.0)` and the scenario is graduated out of xfail (passes on mcp/rest/a2a). **Cross-principal under-assertion (F6)** is fixed: the `_or_leak` test now asserts the owner's `format`/`status` are absent from the whole wire response and carries a same-request positive control that must create a row. **`_log_safe` (F7)** is hoisted to `logging_config.py` and the CodeQL-495 site is wrapped. **Nits (F9):** strict `principal_id: str`, corrected `CREATIVE_REJECTED` prose, and the redundant `idx_delivery_sim_tenant` dropped via a new reversible migration (single head).
- Both cross-principal isolation surfaces survive the #1551 merge and are live (non-vacuous, non-xfail-masked): UC-018 exact-set `returned == owned` with a non-empty anti-vacuous anchor, and UC-006's DB read-back. The e2e_rest escape-hatch lock (20 xfail routes + 3 unsupported-setup declarations) is an exact set with live-detector meta-tests; the 11-entry ledger is a genuine product/spec-gap set. The e2e creative-agent pinning is a fidelity improvement — the live public host is blackholed to `127.0.0.1` (fail-loud) and the stack raises if the pinned image build fails (not a skip).

## `CREATIVE_NOT_FOUND` (new since the last round) — spec-sound, one gap to finish

The change is correct and well-grounded: `CREATIVE_NOT_FOUND` is a real correctable wire code in the pinned 3.1.0-beta.3 enum; the SDK's `ErrorCode` enum contains it and only the `STANDARD_ERROR_CODES` helper table lags, so the `WIRE_STANDARD_CODES = STANDARD_ERROR_CODES ∪ _SPEC_SUPPLEMENT_CODES` supplement is the right seam, and it reaches the wire untranslated with a real REST wire test. The enum's anti-enumeration mandate is satisfied everywhere — no path distinguishes "owned by another principal" from "does not exist."

The gap: the enum mandates that code **uniformly for any `creative_id` not owned by the calling account, across all operations** (the description names no specific tool). Today only the strict sync-assignment path emits it. Create/update still emit `CREATIVE_REJECTED` (`media_buy_create.py:374`, `media_buy_update.py:210`) and the lenient sync-synthesis emits `VALIDATION_ERROR` (`_assignments.py:334`) for the same not-found condition — and `CREATIVE_REJECTED`'s own enum description is "creative failed content policy review," which is semantically wrong for a lookup miss. This is a net improvement over the prior state (everything was `CREATIVE_REJECTED`/`VALIDATION_ERROR`; the sync path is now correct), and the create/update deferral is documented — so it's a follow-up, not a blocker. But two comments now falsely claim the paths already agree (`_assignments.py:332` and `_processing.py:45` say `VALIDATION_ERROR` "matches the strict-mode raise," which `287c93099` falsified), and the lenient-synthesis path isn't mentioned in the deferral at all.

## In-PR items (small, recommend folding into this PR)

1. **Fix the two stale "matches the strict-mode raise" comments** (`_assignments.py:332`, `_processing.py:45`) — they now contradict the code. And either align the lenient-synthesis path to `CREATIVE_NOT_FOUND` or fold it into the same tracked deferral as create/update (right now it's silently on `VALIDATION_ERROR`).
2. **Wire-code prose drift** — `to_wire_error_code`'s docstring (`exceptions.py:157`) and `build_two_layer_error_envelope`'s (`:960-961`) still say the result "lands in `STANDARD_ERROR_CODES`," and a `# ... wire → INVALID_REQUEST` comment in `test_error_format_consistency.py:755` describes pre-`287c93099` behavior. The runtime is correct (`WIRE_STANDARD_CODES`); the prose will mislead the next maintainer into "fixing" the passthrough back into a regression.
3. **Canonical helper bypassed inline** — `8a2ffeaeb` introduced `CreativeRepository.get_by_ids(creative_ids, principal_id)` and then hand-inlined the same principal-scoped `select(Creative).where(tenant, principal, creative_id.in_(...))` at `media_buy_list.py:563` and `media_buy_create.py:1114`. Both are correctly scoped (no security issue), but they duplicate the canonical query the PR just centralized — route them through `get_by_ids`. (R0801 can't see these; the duplication baseline dropped while they were added.)
4. **`preview_creative` builds a raw identity dict** (`creative_agent_registry.py:923`: `{"agent_url": agent_url, "id": format_id}`) though `FormatId` is imported in the same file. The object shape is correct, but the comment claims it "keeps the canonical `agent_url`" while the raw dict passes it un-normalized — `FormatId(agent_url=..., id=...).model_dump(mode="json")` is what would actually canonicalize it. Worth confirming the pinned reference agent tolerates the trailing-slash form before switching. The same raw shape recurs at 4 other sites (a shared serializer is a separate follow-up).
5. **`CREATIVE_NOT_FOUND` has a REST-only wire test** (`test_creative_sync_behavioral.py:327`); its sibling `PACKAGE_NOT_FOUND` is parametrized across all transports and the transient path got all four. Since it's the one non-SDK-standard passthrough code, an MCP/A2A wire assertion is worth adding (parity holds by construction today, but nothing reddens if a boundary re-adds a `STANDARD_ERROR_CODES` gate).
6. **Two new integration tests read back via `get_db_session()`** (`test_creative_sync_behavioral.py:1027`, `:1067`) instead of the `env.query()` seam the sibling new tests use — the repository-pattern guard only enforces `session.add`, so this passes but is the pattern the F6 round otherwise removed.
7. **The new principal-scope guard's query matcher misses accumulator-style queries** — `stmt = select(Creative).where(tenant); stmt = stmt.where(creative_id)` isn't flagged, only the single chain is. That style is already used in the scanned file (`creative.py:121`), so a future edit could reintroduce an unscoped lookup the guard reports as caught. Worth a matcher fix + a positive meta-test.
8. **Ledger comment contradicts F5** (`e2e_rest_known_failures.txt:95-99`): the roas/cpa retirement note still says "production computes neither roas nor cpa nor conversion_value," which this PR falsified. Reword to reflect graduation.
9. **Deferral trackers reference internal ids** (`exceptions.py:42`, the new guard file, a `media_buy_create.py` FIXME) that don't resolve for outside contributors — the earlier round's guards used GitHub/PR refs. Swap the new ones to a GH issue or an inline explanation.

## Pre-existing — please file separately (not this PR's scope, but worth tracking)

- **SSRF on the creative-agent connection path** — a buyer-supplied `agent_url` (from `format_id` in `sync_creatives`) reaches `create_mcp_client`/`registry.get_format` with no `check_url_ssrf`, though the webhook, property-list, and signals-agent paths all validate. This is present at the merge-base (`2b487cc47`, i.e. main) — not introduced here — but it's the highest-impact item surfaced: an authenticated advertiser can point it at internal/metadata addresses. Worth a dedicated issue + fix threading `check_url_ssrf` into that seam.
- **Principal-scope disease outside the guard's scan set** — tenant-only `creative_id` lookups at `gam/managers/orders.py:649` and `services/media_buy_status_scheduler.py:182` (and a dead `queries.py:168` helper) match the F1 shape but sit outside the guard's `creative.py` + `src/core/tools/` scan roots. Pre-existing (introduced 2026-03-14); the ids there are server-derived, so the direct exploit doesn't apply (residual is UUID-collision). Widening the guard's scan roots would prevent regressions.
- **`_log_safe` sweep is partial** — `_processing.py`, `_sync.py`, `media_buy_update.py`, `listing.py` still log buyer `creative_id`/`agent_url` unwrapped. Low impact: the production `JSONFormatter` `json.dumps` escapes newlines (no log-forging in prod), so this is a dev-format / static-CodeQL residual — either finish the sweep or close the alerts with that justification.
- **Recovery-oracle grades the pre-mapping code, not the wire code** (`test_architecture_error_recovery_enum_conformance.py`) — so remapped classes (`AdCPConfigurationError` → SERVICE_UNAVAILABLE) can carry a recovery that diverges from the wire code's spec recovery without the oracle noticing. Pre-existing; this PR changed no recovery class. The pinned schema fixture is also 16 codes behind the SDK pin (64 vs 80) with zero current impact.

`--- Verified at 0085f3c04:`
- **Ran:** the full guard suite (485 passed); mutation-probes reverting the `get_by_ids`/`get_creative_by_id` principal scoping (5+1 tests redden), the transient re-raises (3 mutations redden), the CREATIVE_NOT_FOUND class (envelope test reddens), the ledger exact-set (reddens on a bogus nodeid); the principal-scope guard's two detectors on mutation; F5/F6 and both isolation scenarios run fresh on mcp/rest/a2a; `ci.yml` Summary `needs:`/validate wiring; the `CREATIVE_NOT_FOUND` enum description and recovery against the pinned 3.1.0-beta.3 schema; the SSRF and out-of-scan-scope provenance via `git show` at the merge-base + blame.
- **Not run:** the in-network (e2e_rest) stack and the full Docker suite (CI is green on this head, including BDD In-Network); a live two-principal SSRF proof (traced statically); MCP/A2A wire for `CREATIVE_NOT_FOUND` (parity holds by construction, no test exists to run); fresh-DB provenance for the new integration tests (ran against a persistent DB).

KonstantinMirin:
**Re-review response — all items landed at `0dfd90fc3`** (10 commits since `0085f3c04`). Every in-PR item is addressed in code, the recovery-oracle item is taken in-PR rather than filed separately (rationale below), and the three genuinely pre-existing items are filed/annotated on their canonical issues after a dedup pass over the tracker. Three wire-behavior changes you haven't reviewed are flagged inline. Full suite on the CI box at `183b8419f` (the one commit after it is comment-only): unit 5482, integration 2184, bdd in-process 1613, bdd in-network 356, e2e 91, admin 86, ui 5 — zero failures, xpass counts unchanged vs the prior gate.

**Item 1 — stale comments + the lenient path** → `925d4de6d`. Went with alignment rather than deferral: the lenient sync-synthesis now emits `CREATIVE_NOT_FOUND` for creative-not-found entries (**wire change** — parity with the strict raise; other per-item failure kinds keep `VALIDATION_ERROR`), and both stale "matches the strict-mode raise" comments are corrected. The create/update deferral is now visible in code — `FIXME(#1598)` at all three `CREATIVE_REJECTED` sites — with the upstream reconciliation filed as #1598.

**Item 2 — wire-code prose drift** → `587f41791`. Reworded to `WIRE_STANDARD_CODES` at your four sites plus two more the sweep found (`exceptions.py:623` and `a2a_server/adcp_a2a_server.py:204`).

**Item 3 — inline selects** → `1f85dd630`. Both sites routed through `CreativeRepository.get_by_ids`; the two AST guards were rewritten to pin the stronger repository-routing shape rather than just the principal predicate.

**Item 4 — preview_creative identity** → `183b8419f`. Probed the pinned agent empirically for trailing-slash tolerance first, then switched to `FormatId(...).model_dump(mode="json")` (**wire change** — the AnyUrl-normalized form). Regression tests pin both wire forms. A shared serializer for the remaining hand-rolled `{"agent_url", "id"}` sites is tracked as a follow-up.

**Item 5 — REST-only wire grading** → `c25590a07`. One correction to the premise: the transient-error tests were *also* REST-only (their parametrization was over error kinds, not transports). Both surfaces now run the full `[REST, MCP, A2A]` wire matrix — the strict `CREATIVE_NOT_FOUND` envelope and the transient `RATE_LIMITED`/`SERVICE_UNAVAILABLE` envelopes — so a boundary re-adding a `STANDARD_ERROR_CODES` gate reddens the matrix, not just REST.

**Item 6 — test read-backs** → `939559b16`. Both moved inside the env block onto the `env.query()` seam.

**Item 7 — guard accumulator blind spot** → `0f8f6ba75`. The matcher now merges accumulator-style chains (`stmt = select(...); stmt = stmt.where(...)`) into one query before grading, with two red-first meta-tests.

**Item 8 — ledger note** → `5ff6d8a47`. Reflects the roas/cpa graduation.

**Item 9 — internal tracker ids** → `0dfd90fc3`. One correction: your `exceptions.py:42` cite already carried a GH ref — but the same file, the new guard, and `media_buy_create.py` did have other new-in-branch internal ids; all now reference GitHub (#1602 for the recovery-audit deferral, #1119 for the repository-pattern FIXME, #1430 for review-round provenance). Pre-existing ids elsewhere are untouched debt.

**Recovery oracle — taken in-PR, not filed separately** → `8053d56aa`. You filed it under pre-existing; the envelope mechanic is indeed pre-existing, but the oracle itself is this branch's artifact, so the blind spot was ours to fix here. The oracle now grades the recovery of the **emitted wire code** (covering remapped classes and the base `AdCPError`), and closing it surfaced three reclassifications — **wire changes you haven't reviewed**, each spec-cited against the pinned 3.1.0-beta.3 enumMetadata:
- `CONFIGURATION_ERROR` promoted to a `_SPEC_SUPPLEMENT_CODES` passthrough, recovery terminal ("the buyer cannot resolve a seller-side deployment misconfiguration and MUST NOT auto-retry") — previously demoted to `SERVICE_UNAVAILABLE`/transient;
- `AdCPNotFoundError` recovery corrected to correctable (its wire code is `INVALID_REQUEST`, pinned correctable);
- the base crash-wrap now carries transient (wire `SERVICE_UNAVAILABLE`) — your exact `SERVICE_UNAVAILABLE`+`terminal` divergence example, closed.

The per-raise `recovery=` override audit (~23 sites) and the remaining demoted code (`BILLING_NOT_SUPPORTED`) are #1602.

**Pre-existing items — dedup pass done, filed on their canonical homes:**
- **SSRF**: the creative-agent connection path is now explicitly named on **#1589** (the secure-fetch umbrella) with the full buyer-input → `registry.get_format` → raw-httpx trace; #1129 turned out to cover only the signals-agent admin flows, so no new issue — but the comment flags that this path is advertiser-reachable, unlike the admin-facing cases the issue body cites.
- **Out-of-scan principal-scope sites**: filed as **#1601** (gam orders, status scheduler, and the dead `queries.py` helper marked for deletion; sibling isolation issue #1318 referenced).
- **Log-injection residual**: annotated on **#1450**, with two corrections from re-verification — the sanitizer that issue references (`log_safety.loggable`) doesn't exist in the tree (the real seam is `log_safe` @ `logging_config.py:231`), and `listing.py` is clean while `_sync.py:109` logs buyer-controlled `webhook_url` unsanitized.

Verification: every fix in the batch landed with its own red-first evidence where expressible (the multi-transport matrix, the accumulator meta-tests, the wire-form regression tests); the full seven-suite gate above ran on the CI box at `183b8419f`, and the trailing comment-only commit re-ran ruff plus the touched guard suite green. CI on this head is green including the required `bdd-in-network` job; the PR is conflict-free against main.

ChrisHuie (approved):

PR #1431, opened 2026-06-12
fix(test): read step files as UTF-8 in BDD guard (Windows crash)

pmezzich:
## Summary
The `test_architecture_bdd_no_direct_call_impl` guard reads step files with the platform default encoding. On Windows that's cp1252, and the UTF-8 box-drawing section headers in the step files crash the guard with `UnicodeDecodeError` before it checks anything — the unit suite can't run on a Windows checkout. Pin `encoding='utf-8'` on both `read_text()` calls.

CI (Linux, UTF-8 default) was unaffected — this only fixes local dev on Windows.

## Test plan
- [x] `pytest tests/unit/test_architecture_bdd_no_direct_call_impl.py` passes on Windows after the fix (crashed before)

### DIFF
--- a/tests/unit/test_architecture_bdd_no_direct_call_impl.py
+++ b/tests/unit/test_architecture_bdd_no_direct_call_impl.py
@@ -74,7 +74,7 @@ def _scan_bdd_steps() -> list[str]:
     for py_file in sorted(_BDD_STEPS_DIR.rglob("*.py")):
         if py_file.name.startswith("__"):
             continue
-        source = py_file.read_text()
+        source = py_file.read_text(encoding="utf-8")
         source_lines = source.splitlines()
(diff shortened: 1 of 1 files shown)


pmezzich:
I have read the IPR Policy

ChrisHuie:
looks good. FYI

The same latent bug exists broadly: 64 files under tests/unit/ use bare read_text(), several of which also rglob + parse source files (e.g. test_architecture_bdd_step_module_reachability.py, _bdd_guard_helpers.py). This PR correctly fixes only the one that crashes this guard. The complete systemic fix would be a lint/pre-commit rule banning bare read_text() in tests/ — appropriate as a separate follow-up, not a reason to hold this one.

ChrisHuie (approved):

PR #1437, opened 2026-06-12
fix(ci): cache creative agent build with buildx GHA cache (#1189)

mkostromin-sigma:
## Summary
- Add buildx + GHA cache (`scope=adcp-creative-agent-${ADCP_PIN}`) to the creative integration shard so cold CI runs reuse Docker layers instead of rebuilding from scratch every time.
- Retry tarball fetch and docker build (3 attempts, exponential backoff) to absorb transient network flakes.
- Document CI cache behavior and local fallback in `docs/development/ci-pipeline.md`.

Closes #1189.

## Test plan
- [ ] CI creative integration shard completes on cold run (buildx cache upload)
- [ ] Second CI run on same pin shows faster creative agent startup (cache hit)
- [ ] Local `./scripts/creative-agent-stack.sh up` still uses plain `docker build` (no GHA env required)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      actions: write # GHA cache for buildx (creative shard only)
+      actions: write # GHA cache upload for buildx (used by creative shard step only)
     timeout-minutes: 15
     strategy:
       fail-fast: false
@@ -219,6 +219,9 @@ jobs:
       - name: Set up Docker Buildx
         if: matrix.group == 'creative'
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+      - name: Expose GHA cache runtime to buildx
+        if: matrix.group == 'creative'
+        uses: crazy-max/ghaction-github-runtime@04d248b84655b509d8c44dc1d6f990c879747487 # v4.0.0
       - name: Start pinned creative agent
         if: matrix.group == 'creative'
         run: ./scripts/creative-agent-stack.sh up
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,6 +175,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      actions: write # GHA cache for buildx (creative shard only)
     timeout-minutes: 15
     strategy:
       fail-fast: false
@@ -215,6 +216,9 @@ jobs:
       - uses: ./.github/actions/_postgres
         with:
           database-url: ${{ env.DATABASE_URL }}
+      - name: Set up Docker Buildx
+        if: matrix.group == 'creative'
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
       - name: Start pinned creative agent
         if: matrix.group == 'creative'
         run: ./scripts/creative-agent-stack.sh up
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      actions: write # GHA cache upload for buildx (used by creative shard step only)
+      packages: write # push adcp-creative-agent to ghcr.io (creative shard step only)
     timeout-minutes: 15
     strategy:
       fail-fast: false
@@ -216,14 +216,17 @@ jobs:
       - uses: ./.github/actions/_postgres
         with:
           database-url: ${{ env.DATABASE_URL }}
-      - name: Set up Docker Buildx
+      - name: Log in to GitHub Container Registry
         if: matrix.group == 'creative'
-        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
-      - name: Expose GHA cache runtime to buildx
-        if: matrix.group == 'creative'
-        uses: crazy-max/ghaction-github-runtime@04d248b84655b509d8c44dc1d6f990c879747487 # v4.0.0
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Start pinned creative agent
         if: matrix.group == 'creative'
+        env:
+          CREATIVE_AGENT_GHCR_IMAGE: ghcr.io/${{ github.repository }}/adcp-creative-agent
         run: ./scripts/creative-agent-stack.sh up
       - uses: ./.github/actions/_pytest
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      packages: write # push adcp-creative-agent to ghcr.io (creative shard step only)
+      packages: read # pull adcp-creative-agent from ghcr.io (publish runs on main via publish-creative-agent.yml)
     timeout-minutes: 15
     strategy:
       fail-fast: false
--- a/.github/workflows/publish-creative-agent.yml
+++ b/.github/workflows/publish-creative-agent.yml
@@ -1,0 +1,27 @@
+name: Publish creative agent image
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      - uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push pinned creative agent
+        env:
+          CREATIVE_AGENT_GHCR_IMAGE: ghcr.io/${{ github.repository }}/adcp-creative-agent
+        run: ./scripts/creative-agent-stack.sh publish
--- a/docs/development/ci-pipeline.md
+++ b/docs/development/ci-pipeline.md
@@ -128,6 +128,28 @@ commit without authentication.
 cleanly. Check the `community_points` / `users` table FK ordering — this was the
 failure that prompted pinning (April 2026).
 
+The pin lives only in [`scripts/creative-agent-stack.sh`](../../scripts/creative-agent-stack.sh)
+— both CI and `run_all_tests.sh` call that script so local and CI cannot diverge.
+
+### CI build cache and retries
+
+On ephemeral GitHub Actions runners, local `/tmp` tarball reuse and `docker image
+inspect` guards do not survive between runs. The creative shard therefore:
+
+1. Runs `docker/setup-buildx-action` (creative matrix leg only) before
+   `creative-agent-stack.sh up`.
+2. Uses `docker buildx build --cache-from/to type=gha` inside the script when
+   `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` are present (CI). Cache scope
+   is `adcp-creative-agent-${ADCP_PIN}` so a pin bump invalidates stale layers.
+3. Falls back to plain `docker build` locally — no CI-only env required for
+   `run_all_tests.sh`.
+
+Tarball fetch (`curl -f`) and docker build each retry up to 3 times with
+exponential backoff to absorb transient `ECONNRESET` flakes.
+
+The `integration-tests` job grants `actions: write` so buildx can upload GHA
+cache entries (same pattern as [`release-please.yml`](../../.github/workflows/release-please.yml)).
+
 ### Creative agent infrastructure
 
 The agent runs in its own Docker network (`creative-net`) with a separate
--- a/docs/development/ci-pipeline.md
+++ b/docs/development/ci-pipeline.md
@@ -136,11 +136,14 @@ The pin lives only in [`scripts/creative-agent-stack.sh`](../../scripts/creative
 On ephemeral GitHub Actions runners, local `/tmp` tarball reuse and `docker image
 inspect` guards do not survive between runs. The creative shard therefore:
 
-1. Runs `docker/setup-buildx-action` (creative matrix leg only) before
-   `creative-agent-stack.sh up`.
+1. Runs `docker/setup-buildx-action` and `ghaction-github-runtime` (creative
+   matrix leg only) before `creative-agent-stack.sh up`. The runtime action
+   exposes `ACTIONS_RUNTIME_TOKEN` to shell steps — without it, `type=gha` cache
+   backends cannot authenticate (unlike `docker/build-push-action`, which injects
+   the token internally).
 2. Uses `docker buildx build --cache-from/to type=gha` inside the script when
-   `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` are present (CI). Cache scope
-   is `adcp-creative-agent-${ADCP_PIN}` so a pin bump invalidates stale layers.
+   `ACTIONS_RUNTIME_TOKEN` is present. Cache scope is
+   `adcp-creative-agent-${ADCP_PIN}` so a pin bump invalidates stale layers.
 3. Falls back to plain `docker build` locally — no CI-only env required for
    `run_all_tests.sh`.
 
--- a/docs/development/ci-pipeline.md
+++ b/docs/development/ci-pipeline.md
@@ -131,27 +131,29 @@ failure that prompted pinning (April 2026).
 The pin lives only in [`scripts/creative-agent-stack.sh`](../../scripts/creative-agent-stack.sh)
 — both CI and `run_all_tests.sh` call that script so local and CI cannot diverge.
 
-### CI build cache and retries
+### CI image cache (ghcr.io) and retries
 
 On ephemeral GitHub Actions runners, local `/tmp` tarball reuse and `docker image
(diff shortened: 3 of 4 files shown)


mkostromin-sigma:
@ChrisHuie — ready for review. Tip `087455a7`, CI green (Summary ✅).

**Scope (3 files, based on `main`):**
- `scripts/creative-agent-stack.sh` — buildx GHA cache (`scope=adcp-creative-agent-${ADCP_PIN}`) + retry on tarball fetch and docker build
- `.github/workflows/ci.yml` — `setup-buildx-action` on creative integration shard only; `actions: write` for cache upload
- `docs/development/ci-pipeline.md` — cache/retry documentation

Local `./scripts/creative-agent-stack.sh up` still uses plain `docker build` (no GHA env required).

Closes #1189. Independent of the #1234 stack — can merge anytime.

ChrisHuie:
Reviewed at head `087455a7`. The retry/hardening half of this is solid, but the headline — cross-run buildx layer caching — does not actually engage on CI in its current form. Details below, with the evidence so it's reproducible.

## What works

- `_retry` correctly wraps both the tarball fetch and the build (3 attempts, exponential backoff). With `set -euo pipefail` already at the top of the script, a mid-stream `curl` failure in `curl -sLf … | tar xz` propagates through the pipe, so the retry fires on exactly the `ECONNRESET` flakes #1189 calls out. A persistent failure still fails the job (bare call under `set -e`). This meets the retry acceptance criterion.
- `curl -sL` → `curl -sLf` is a real improvement: an HTTP 4xx/5xx from the archive host now fails fast and retries instead of piping an error body into `tar`.

## Blocking: the GHA cache path never runs on CI

On the creative integration job for this head (CI run 27431888867), the build step logged:

```
[creative-agent] fetching adcp@ca70dd1e2a6c
[creative-agent] docker build adcp-creative-agent (adcp@ca70dd1e2a6c)
```

That second line is the `else` branch of `_build_image` (plain `docker build`). The buildx branch logs `buildx build … gha cache scope=…`, which never appears. There are zero `importing cache` / `exporting cache` / `CACHED` lines in the whole job log, and the build ran ~2m20s (17:30:42 → 17:33:04 to health) — i.e. the same full, uncached rebuild #1189 is trying to eliminate. `setup-buildx-action` does run (a `docker-container` builder is created and torn down), so we pay the buildx setup cost for no benefit.

Root cause — this gate is always false here:

```sh
if [ -n "${ACTIONS_CACHE_URL:-}" ] && [ -n "${ACTIONS_RUNTIME_TOKEN:-}" ]; then
```

`ACTIONS_RUNTIME_TOKEN` / `ACTIONS_CACHE_URL` are GitHub-internal runtime values that are **not** exposed to plain `run:` steps — only JavaScript actions see them, via the Actions toolkit. That is precisely why `release-please.yml` works: it uses `docker/build-push-action`, which injects the runtime token internally. A raw `docker buildx build` in a shell script does not get it, so "same pattern as release-please.yml" (PR body + docs) does not transfer. To use `type=gha` from an inline step you have to expose the runtime first, canonically `crazy-max/ghaction-github-runtime`. Refs: https://docs.docker.com/build/cache/backends/gha/ and https://github.com/orgs/community/discussions/42856.

A second, independent problem with the same gate: GitHub's Actions cache moved to service **v2** (legacy v1 API shut down 2025-04-15; buildx ≥ 0.21 required). Under v2 the relevant variable is `ACTIONS_RESULTS_URL`; `ACTIONS_CACHE_URL` is the v1 name. So even after the runtime is exposed, keying the gate on `ACTIONS_CACHE_URL` may stay false. Ref: https://github.com/docker/buildx/pull/3001.

Net against #1189's acceptance criteria: the retry criterion is met, but "warm-cache build step < 30s, verified by running CI twice on an unchanged pin" is not — the cache is never populated or read.

## Suggested direction

1. Expose the runtime to the creative leg before `creative-agent-stack.sh up`, SHA-pinned for pinact:
   ```yaml
   - name: Expose GHA cache runtime to buildx
     if: matrix.group == 'creative'
     uses: crazy-max/ghaction-github-runtime@<sha>   # pin a tagged release
   ```
2. Gate on `ACTIONS_RUNTIME_TOKEN` (present under both v1 and v2) rather than the v1-only `ACTIONS_CACHE_URL`.
3. Verify empirically — the unchecked test-plan box ("second CI run on same pin shows a cache hit") is the one that matters. Two consecutive runs on an unchanged pin should show the `buildx build … gha cache scope=…` line, `importing cache` on run 2, and a build step well under 30s.

Alternatively, `docker/build-push-action` (as in `release-please.yml`) sidesteps the runtime-token plumbing entirely — but it moves build logic out of the shared single-source script that #1189 explicitly wants to keep as the one source of the pin, so the in-script approach plus a runtime-export step is probably the better fit here.

## Minor

- `ci.yml`: `actions: write  # … (creative shard only)` — job-level permissions can't be matrix-scoped, so this is granted to every integration leg, not just creative. Functionally harmless, but the comment overstates the scoping.
- The new docs section states the cache runs "when `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` are present (CI)". They are not present in a `run:` step, so that paragraph documents intended-but-not-actual behavior until the above is addressed.

mkostromin-sigma:
@ChrisHuie — blocking review addressed at `b93186c6`, CI green (Summary ✅, run [27571023376](https://github.com/prebid/salesagent/actions/runs/27571023376)).

### What changed

1. **`ghaction-github-runtime@v4.0.0`** (SHA-pinned) added in `ci.yml` on the creative matrix leg, between `setup-buildx-action` and `creative-agent-stack.sh up`. Exposes `ACTIONS_RUNTIME_TOKEN` to shell `run:` steps.
2. **`creative-agent-stack.sh`** — buildx GHA cache gate now keys on `ACTIONS_RUNTIME_TOKEN` only (dropped the v1-only `ACTIONS_CACHE_URL` check). Comment documents why.
3. **`ci-pipeline.md`** — updated to describe the runtime-export requirement and corrected the env-var gate.
4. **Minor** — `actions: write` comment clarified as job-level (not matrix-scoped).

### Verification

Creative integration shard passed on this head. The build step should now log `buildx build … gha cache scope=…` instead of the plain `docker build` else-branch.

**Warm-cache check:** this is the first CI run after the fix, so it populates cache. A second consecutive run on the same pin should show `importing cache` and a build step well under 30s — happy to trigger a re-run if you want that confirmed before merge.

Ready for re-review.

ChrisHuie:
Re-reviewed at head `b93186c6`; re-ran the creative integration job on the unchanged pin (run 27571023376, attempt 2). The cache now engages and hits, but the measured warm time does not meet #1189's performance criteria.

**Cache engagement (was the prior blocker).** `ghaction-github-runtime@v4.0.0` exposes `ACTIONS_RUNTIME_TOKEN`, `ACTIONS_CACHE_URL`, and `ACTIONS_RESULTS_URL` (the v2-service var) — gate passes, buildx authenticates to the v2 backend. Attempt 2 is a confirmed hit: stages `#12`–`#30` all `CACHED`, compile skipped.

**Measured impact:**

| Scenario | `Start pinned creative agent` step | Job total |
|---|--:|--:|
| No cache (per #1189) | ~2 min | ~6.5 min |
| Attempt 1 — cold/populate | 9m45s | 14m40s |
| Attempt 2 — warm/hit | 1m57s | 6m54s |

Warm breakdown: fetch ~3s · build (all `CACHED`) ~3s · `#31 exporting to docker image format` (`--load`) ~1m32s · `#33` cache re-export ~13s · health ~6s.

**Root cause:** warm ≈ no-cache (~2 min). `type=gha` requires the buildx `docker-container` driver, so the image is materialized into the daemon via `--load` before the agent runs; that `--load` (~1m32s) ≈ the compile the cache removes. Net wash warm; cold ~5× slower (the `mode=max` export). The `release-please.yml` pattern is build-only (no `--load`) — it doesn't transfer to a build-then-run job.

**#1189 acceptance criteria:**
- warm step < 30s → not met (1m57s; `--load` is the floor)
- cold ≤ ~2m30s → not met (9m45s)
- retry ≥3 / `curl -f` → met; fixes the ECONNRESET flake
- CI↔local parity, SHA-pin, frozen-checks → met

**Cache budget:** `mode=max` makes the creative cache ~2.6 GB; the repo is already ~10.1 GB against the 10 GB limit. This entry risks LRU eviction between runs (an evicted "warm" run falls back to 9m45s) and evicts other caches.

**Path forward:** the retry/`curl -f` change stands on its own. For the caching goal, take #1189's stated escalation — pre-build the pinned image, push to `ghcr.io` keyed on `ADCP_PIN`, `docker pull` in CI; drops npm and the `--load` tax. Or split the retry/hardening into its own PR to merge now and rework the cache separately.

*Verification* — Read: diff at `b93186c6`, cold + warm creative logs, `actions/caches` + `cache/usage` API. Ran: re-triggered the creative job on the unchanged pin (attempt 2); step timings from the jobs API. Not measured: `main`-baseline timing (~2 min is #1189's figure, corroborated by the ~1m52s compile in attempt 1); the ghcr.io path.

mkostromin-sigma:
@ChrisHuie — re-review at `c512a3e`, CI green (Summary ✅, run [27636999555](https://github.com/prebid/salesagent/actions/runs/27636999555)).

**Review feedback applied (head `c512a3e`):**
- Replaced buildx GHA layer cache with pin-keyed ghcr.io images (`CREATIVE_AGENT_GHCR_IMAGE` + `docker manifest inspect` → pull on warm, build+push on cold)
- Retries preserved (`curl -f`, exponential backoff on tarball fetch / pull / build / push)
- Local fallback unchanged (plain `docker build` when `CREATIVE_AGENT_GHCR_IMAGE` unset)
- `packages: write` + `docker/login-action` on creative shard only

**CI verification (3 consecutive creative runs on unchanged pin `ca70dd1e2a6c`):**

| Run | Start pinned creative agent step | Path |
|-----|----------------------------------|------|
| Attempt 1 (cold) | ~2m44s (17:50:42 → 17:53:26) | fetch + build + push (3× fail → WARN) |
| Re-run 1 | ~2m43s (18:42:34 → 18:45:17) | fetch + build + push (3× fail → WARN) |
| Re-run 2 | ~2m43s (18:52:31 → 18:55:14) | fetch + build + push (3× fail → WARN) |

All three runs logged `docker build …` — no `pulling … from ghcr.io`. Root cause: `docker push ghcr.io/prebid/salesagent/adcp-creative-agent:ca70dd1e2a6c` fails all 3 retries on every run; package does not exist in GHCR yet (404 on package API). Fork PR `GITHUB_TOKEN` appears unable to publish the org package — script correctly falls back to local image via WARN.

**#1189 warm criterion (< 30s):** not verifiable on this fork PR until the image is published (likely first `main` CI run after merge, then second run should show `docker pull` + step well under 30s). Cold path matches baseline (~2m43s compile, no buildx `--load` tax).

Ready for re-review. Happy to re-run warm verification post-merge if helpful.

ChrisHuie:
Re-reviewed at `c512a3e`, and re-ran the creative shard on the unchanged pin (run 27636999555). The ghcr.io switch removed the buildx `--load` tax and the GHA cache-budget problem — good. But the cache still can't populate or be verified here, for a structural reason that needs a different publish model rather than a token tweak.

## Root cause (verified)

The creative *test* shard builds **and** pushes on every run. Two structural facts defeat that:
1. Fork PRs get `packages: write` downgraded to read-only, so the push fails: `denied: installation not allowed to Create organization package` (run 27636999555). The package is never created (`404`), so every run is a cold build.
2. Even after a publish, a *private* GHCR package can't be read by a fork PR's token.

So the cache is invisible to exactly the PRs under test, and all three runs measured were cold (~2m43s).

## Fix: publish from `main`, pull everywhere, public image

**1. New workflow `.github/workflows/publish-creative-agent.yml`** — publishes from trusted main context, self-skips when the tag exists:

```yaml
name: Publish creative agent image
on:
  push:
    branches: [main]
  workflow_dispatch:
permissions:
  contents: read
  packages: write
jobs:
  publish:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
      - uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
      - name: Build and push pinned creative agent
        env:
          CREATIVE_AGENT_GHCR_IMAGE: ghcr.io/${{ github.repository }}/adcp-creative-agent
        run: ./scripts/creative-agent-stack.sh publish
```

**2. `publish` subcommand in `creative-agent-stack.sh`** — dispatch is a `case`, so update both the branch and the usage string:

```sh
cmd_publish() {
    [ -n "${CREATIVE_AGENT_GHCR_IMAGE:-}" ] || { echo "CREATIVE_AGENT_GHCR_IMAGE required" >&2; return 1; }
    if _ghcr_image_exists; then
        echo "[creative-agent] $(_ghcr_ref) already published; nothing to do"
        return 0
    fi
    _ensure_tarball
    _retry 3 _build_for_ghcr
    _retry 3 _push_to_ghcr      # bare: a publish job should fail loudly if it cannot publish
}

case "${1:-}" in
    up) cmd_up ;;
    down) cmd_down ;;
    url) cmd_url ;;
    publish) cmd_publish ;;
    *) echo "usage: $0 {up|down|url|publish}" >&2; exit 2 ;;
esac
```

**3. Make the creative test shard pull-only** — keep the login (authenticated pulls dodge anonymous rate limits and work for public images from forks), drop write:

```yaml
    permissions:
      contents: read
      packages: read            # was: packages: write
...
      - name: Log in to GitHub Container Registry
        if: matrix.group == 'creative'
        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
      - name: Start pinned creative agent
        if: matrix.group == 'creative'
        env:
          CREATIVE_AGENT_GHCR_IMAGE: ghcr.io/${{ github.repository }}/adcp-creative-agent
        run: ./scripts/creative-agent-stack.sh up
```

**4. Rewrite `_ensure_image` as pull-only with a local-build fallback** — the current code hard-fails under `set -euo pipefail` if `manifest inspect` passes but the pull then dies (a GHCR read blip would break CI):

```sh
_ensure_image() {
    if docker image inspect "$IMAGE" >/dev/null 2>&1; then
        return 0
    fi
    if [ -n "${CREATIVE_AGENT_GHCR_IMAGE:-}" ] && _ghcr_image_exists; then
        if _retry 3 _pull_from_ghcr; then
            return 0
        fi
        echo "[creative-agent] WARN: ghcr pull failed; building locally" >&2
    fi
    _ensure_tarball
    _retry 3 _build_image_local
}
```

The test path now never pushes. **Graceful degradation:** in every package state (absent / private / public) the shard either pulls or builds locally — no `denied`, no failures — so this restructure is safe to merge before the maintainer steps below; it just stays at today's cold behavior until the image is live and public. The *publish workflow* is the one that will fail loudly on `main` if package creation is blocked — that is the intended signal for step 1 below, and it does not affect PR checks.

## Maintainer prerequisites (org/repo side — not contributor code)

1. **First publish + creation check.** `GITHUB_TOKEN` → GHCR publishing is already in use here — `ghcr.io/prebid/salesagent` is public with release tags pushed via release-please's `GITHUB_TOKEN` login — so a sibling image uses the same mechanism. The one first-time unknown is auto-creation of the *new* package name: trigger the publish workflow once (`workflow_dispatch` on a main-repo branch) and confirm it creates `adcp-creative-agent` rather than returning `denied … Create organization package`. If denied, enable Actions package creation in Org → Packages (or have an admin pre-create + link the empty package once), then re-run.
2. **Set the package Public** (one time) so fork PRs can pull. Until then the shard just builds locally everywhere (green, no speedup).

## Cache key — already correct

Tagging on `${ADCP_PIN}` is right: the image is built from the pinned source tarball, so a pin bump produces a new tag and a natural miss.

## Verification + the `<30s` caveat

Not verifiable on a fork PR. After the package is public, trigger any PR's creative shard and confirm the log shows `pulling … from ghcr.io` and measure the step. The pull skips the ~2 min compile, but a non-trivial image pull may still land above 30s — measure the real number before claiming #1189's `<30s`; if it is large, slimming the image (multi-stage / smaller base) is a sensible follow-up.

## Lower-effort path to land value now

The `_retry` + `curl -sLf` hardening fixes the ECONNRESET flake independent of any caching. Splitting it into its own PR (script retry only, no `ci.yml` registry changes) would merge cleanly now while the caching is reworked.

(New-workflow housekeeping: SHA-pin all `uses:` for pinact, keep permissions minimal for zizmor; it runs only on `push: main`, so it adds no PR-required check and does not touch the frozen-required-checks list.)

mkostromin-sigma:
@ChrisHuie — re-review requested at `038b4383`, CI green (Summary ✅, run [27705951614](https://github.com/prebid/salesagent/actions/runs/27705951614)).

### Review feedback applied

Addressing [#4730960553](https://github.com/prebid/salesagent/pull/1437#issuecomment-4730960553) (publish-from-main model):

1. **New `.github/workflows/publish-creative-agent.yml`** — publishes from trusted `main` / `workflow_dispatch` via `creative-agent-stack.sh publish`; self-skips when tag `${ADCP_PIN}` already exists. SHA-pinned `uses:`; `persist-credentials: false` on checkout (zizmor clean).
2. **`publish` subcommand** in `creative-agent-stack.sh` — build+push with fail-loud retries; only path that pushes to GHCR.
3. **CI creative shard pull-only** — `integration-tests` now `packages: read`; login + `CREATIVE_AGENT_GHCR_IMAGE` + `up` unchanged. Test path never pushes.
4. **`_ensure_image` rewritten** — pull when public tag exists; WARN + local `docker build` on pull blip or absent/private package (graceful degradation under `set -euo pipefail`).
5. **`ci-pipeline.md`** — publish/pull model, maintainer prerequisites, fork-PR caveat, and <30s measurement caveat documented.

Retry + `curl -sLf` hardening from earlier rounds preserved.

### CI verification (this head)

| Check | Result |
|-------|--------|
| Summary | ✅ |
| Integration (creative) | ✅ (~8m job; fork PR expected cold path — local `docker build`, no push) |
| pinact / zizmor | ✅ |

Warm-cache pull (`pulling … from ghcr.io`, step <30s) remains **post-merge**: requires `workflow_dispatch` publish on `main`, public `adcp-creative-agent` package, then a second CI run on unchanged pin. Happy to measure and report that after merge if helpful.

Ready for re-review.

ChrisHuie (approved):

PR #1439, opened 2026-06-12
refactor: consolidate account reference coercion

### LINKED ISSUE #1435: refactor: extract to_account_reference() helper; consolidate dict→AccountReference coercion
## Problem

Dict→`AccountReference` coercion is open-coded in two different idioms, and `schema_helpers.py` is missing the `to_account_reference` peer to its existing `to_*_reference` family:

- `src/a2a_server/adcp_a2a_server.py:114` — a local `_coerce_account_reference()` (dict → `AccountReference.model_validate`), called once at `:1634` (`_handle_sync_creatives_skill`)
- `src/routes/api_v1.py:260` — `LibraryAccountReference(**body.account)` (the `**unpack` form)

`src/core/schema_helpers.py` already exposes the convention:
- `to_context_object` (`:21`), `to_reporting_webhook` (`:39`), `to_brand_reference` (`:57`), `to_property_list_reference` (`:77`)

`to_account_reference` is the missing sibling. (Both idioms above work today — `AccountReference(**dict)` and `AccountReference.model_validate(dict)` are equivalent for this RootModel — so this is a consolidation for DRY/consistency, not a bug fix.)

## Fix

1. Add `to_account_reference()` to `src/core/schema_helpers.py`. Match the signature shape of the `to_*_reference` family — accept `dict | AccountReference | None`, return `AccountReference | None`, with `isinstance` guards.

   **One wrinkle vs the existing helpers:** `AccountReference` is a `RootModel` (a single `root` field wrapping a `oneOf`), unlike `BrandReference`/`PropertyListReference`, which take field kwargs and are constructed `Model(**dict)`. For the dict branch, use the canonical RootModel entry `AccountReference.model_validate(value)` — the same call the existing `_coerce_account_reference` already uses (`adcp_a2a_server.py:126`) — rather than field-unpacking. So model the body on `_coerce_account_reference`, not on `to_brand_reference`.

2. Replace the two sites:
   - `adcp_a2a_server.py` — have `_coerce_account_reference` delegate to the new helper (or replace its single call at `:1634` directly), then delete `_coerce_account_reference`.
   - `api_v1.py:260` — swap `LibraryAccountReference(**body.account)` for `to_account_reference(body.account)`.

## Acceptance criteria

- [ ] `to_account_reference()` exists in `schema_helpers.py` and is the single place that does dict→`AccountReference` coercion.
- [ ] `grep -rn "AccountReference.model_validate\|AccountReference(\*\*" src/` returns only the helper definition.
- [ ] `make quality` passes.

## Context

Split from #1259. The original issue catalogued several open-coded coercion sites; an account-resolution refactor since then collapsed most of them into typed request parameters, leaving the two surviving idioms above. Small and mechanical — templated by the four existing `to_*_reference` helpers, with the one RootModel wrinkle noted above.

Jeshua09090:
## Summary
- add the missing shared `to_account_reference()` schema helper
- replace the A2A and REST dict-to-`AccountReference` coercion sites
- extend regression coverage for dict, typed, and `None` inputs

## Testing
- `uv run pytest tests/unit/test_sync_creatives_a2a_account.py tests/unit/test_rest_api_endpoints.py -q` (17 passed)
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run mypy src/ --config-file=mypy.ini`
- `uv run python .pre-commit-hooks/check_code_duplication.py`

`make quality` is not directly available in this Windows environment because `make` is not installed. Its unit-test step reaches pre-existing Windows-only path/encoding guard failures; CI Linux is the authoritative full run.

Closes #1435

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -66,6 +66,7 @@ from src.core.exceptions import (
     normalize_to_adcp_error,
 )
 from src.core.resolved_identity import ResolvedIdentity
+from src.core.schema_helpers import to_account_reference
 from src.core.schemas import CreativeStatusEnum
 from src.core.tool_context import ToolContext
 from src.core.tool_error_logging import record_boundary_error
@@ -109,21 +110,6 @@ from src.core.version import get_version
 from src.services.protocol_webhook_service import get_protocol_webhook_service
 
 logger = logging.getLogger(__name__)
-
-
-def _coerce_account_reference(account: Any) -> Any:
-    """Coerce a raw dict account param to AccountReference at the A2A boundary.
-
-    A2A clients may send account as a plain dict. The MCP wrapper's TypeAdapter
-    handles this automatically, but the A2A raw path passes dicts through. This
-    wraps dicts in the SDK AccountReference Pydantic model so downstream code
-    gets a typed object.
-    """
-    if account is None or not isinstance(account, dict):
-        return account
-    from adcp.types import AccountReference as LibraryAccountReference
-
-    return LibraryAccountReference.model_validate(account)
 
 
 def _dict_to_value(d: dict) -> struct_pb2.Value:
@@ -1631,7 +1617,7 @@ class AdCPRequestHandler(RequestHandler):
             validation_mode=parameters.get("validation_mode", "strict"),
             push_notification_config=parameters.get("push_notification_config"),
             context=context,
-            account=_coerce_account_reference(parameters.get("account")),
+            account=to_account_reference(parameters.get("account")),
             identity=identity,
         )
 
--- a/src/core/schema_helpers.py
+++ b/src/core/schema_helpers.py
@@ -95,6 +95,7 @@ def to_account_reference(account: dict[str, Any] | AccountReference | None) -> A
     if isinstance(account, AccountReference):
         return account
     if isinstance(account, dict):
+        # AccountReference is a RootModel, so validate the whole value instead of field-unpacking.
         return AccountReference.model_validate(account)
     return None  # Fallback for unexpected types
 
--- a/src/core/schema_helpers.py
+++ b/src/core/schema_helpers.py
@@ -13,7 +13,14 @@ Philosophy:
 from typing import Any
 
 from adcp import GetProductsResponse, Product
-from adcp.types import BrandReference, ContextObject, ProductFilters, PropertyListReference, ReportingWebhook
+from adcp.types import (
+    AccountReference,
+    BrandReference,
+    ContextObject,
+    ProductFilters,
+    PropertyListReference,
+    ReportingWebhook,
+)
 
 from src.core.schemas.product import GetProductsRequest
 
@@ -71,6 +78,24 @@ def to_brand_reference(brand: dict[str, Any] | BrandReference | str | None) -> B
         return BrandReference(domain=brand)
     if isinstance(brand, dict):
         return BrandReference(**brand)
+    return None  # Fallback for unexpected types
+
+
+def to_account_reference(account: dict[str, Any] | AccountReference | None) -> AccountReference | None:
+    """Convert dict to AccountReference for adcp compatibility.
+
+    Args:
+        account: Account reference as dict, AccountReference, or None
+
+    Returns:
+        AccountReference or None
+    """
+    if account is None:
+        return None
+    if isinstance(account, AccountReference):
+        return account
+    if isinstance(account, dict):
+        return AccountReference.model_validate(account)
     return None  # Fallback for unexpected types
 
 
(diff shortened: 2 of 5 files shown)


Jeshua09090:
I have read the IPR Policy

ChrisHuie:
Reviewed by multiple agents and ran the offline gate locally on 794ca5a (HEAD asserted,
clean tree), not from CI's checks:
- ruff (format + check) and mypy clean on the changed files
- full unit suite: 4785 passed, 0 failed
- duplication guard (pylint R0801, ratcheting) passes — the 5th to_*_reference helper
  doesn't trip it
- deleted _coerce_account_reference leaves no orphaned references

The helper is correct: model_validate is the right entry for a RootModel, both call
sites are replaced, __all__ updated. No regression detectable by the offline gate.

One thing to address before merge — error-path coverage:
- The only tests are happy-path. Neither the helper test nor any call-site test exercises
  an invalid account (a oneOf violation — e.g. both account_id and brand, or neither).
- That matters most at the changed REST site: src/routes/api_v1.py swapped
  LibraryAccountReference(**body.account) → to_account_reference(body.account) on
  POST /media-buys/delivery. Happy-path is equivalent, but **-unpack and model_validate
  can differ on malformed input, and that path has no test. Please add one (valid +
  malformed) through the route, and an invalid-input case on the helper itself.

Nit:
- One-line comment on to_account_reference noting it uses model_validate (not Model(**dict)
  like the sibling to_*_reference helpers) because AccountReference is a RootModel — so a
  later "consistency" edit doesn't regress it.


(Integration/e2e/admin not run locally — low value for a pure coercion change; noted for
completeness.)

Jeshua09090:
Addressed in 6673ebca:
- added invalid-input coverage for `to_account_reference({})`
- added REST `/api/v1/media-buys/delivery` coverage for valid account coercion before identity enrichment
- added REST malformed-account coverage to assert the validation envelope and that core delivery is not called
- documented why `to_account_reference` uses `AccountReference.model_validate(...)` for the RootModel case

Verified locally:
- `uv run pytest tests/unit/test_sync_creatives_a2a_account.py tests/unit/test_rest_api_endpoints.py -q`
- `uv run ruff format --check src/core/schema_helpers.py tests/unit/test_sync_creatives_a2a_account.py tests/unit/test_rest_api_endpoints.py`
- `uv run ruff check src/core/schema_helpers.py tests/unit/test_sync_creatives_a2a_account.py tests/unit/test_rest_api_endpoints.py`
- `uv run mypy src/core/schema_helpers.py src/routes/api_v1.py src/a2a_server/adcp_a2a_server.py --config-file=mypy.ini`

ChrisHuie:
Thanks for the PR. Updating #1434 as follow-up since it references the old symbol

To get CI green
------------------------
tests/unit/test_rest_api_endpoints.py:149 trips the "no split mock assertions" guard
(test_architecture_weak_mock_assertions.py): a bare assert_called_once() plus a .call_args
read. Rewrite to the atomic form — it also strengthens the test by pinning the whole
coerced AccountReference instead of just brand.domain:
        assert response.status_code == 200
        expected_account = LibraryAccountReference.model_validate(
            {"brand": {"domain": "example.com"}, "operator": "op-1", "sandbox": False}
        )
        mock_enrich.assert_called_once_with(_MOCK_IDENTITY, expected_account)
        assert mock_impl.call_args.kwargs["identity"] is enriched_identity
Verify: uv run pytest tests/unit/test_architecture_weak_mock_assertions.py tests/unit/test_rest_api_endpoints.py -q


Optional polish (not blocking)
------------------------------
test_malformed_account_returns_validation_error: per tests/CLAUDE.md, new error tests should
pin recovery via the canonical helper (production emits "correctable" here):
        from tests.helpers import assert_envelope_shape
        assert response.status_code == 400
        assert_envelope_shape(response.json(), "VALIDATION_ERROR", recovery="correctable")
        mock_enrich.assert_not_called()
        mock_impl.assert_not_called()

Jeshua09090:
Fixed in ed2f45ca:
- replaced the split mock inspection with an atomic assert_called_once_with() that pins the full coerced AccountReference
- updated the malformed-account test to use the canonical envelope helper and pin recovery=correctable

Local verification:
- uv run pytest tests/unit/test_sync_creatives_a2a_account.py tests/unit/test_rest_api_endpoints.py -q (20 passed)
- focused weak-mock scan for tests/unit/test_rest_api_endpoints.py returns no violations
- uv run ruff format --check tests/unit/test_rest_api_endpoints.py
- uv run ruff check tests/unit/test_rest_api_endpoints.py

The full architecture guard is not reliable locally on Windows because its allowlist paths are POSIX-style; the new Linux CI run will be the authoritative full check.

ChrisHuie (approved):

PR #1443, opened 2026-06-12
fix(agent-db): resolve PROJECT_DIR to the worktree root, not the skill dir

ChrisHuie:
## Problem

`agent-db.sh` (the per-worktree throwaway Postgres used by integration tests) resolves `PROJECT_DIR` only one level up from the script directory:

```bash
SCRIPT_DIR="$( dirname "${BASH_SOURCE[0]}" )"   # .claude/skills/agent-db
PROJECT_DIR="$( cd "$SCRIPT_DIR/.." && pwd )"    # → .claude/skills   ❌
```

The script sits three levels below the worktree root (`agent-db/` → `skills/` → `.claude/` → root), so `basename "$PROJECT_DIR"` is the literal string `"skills"` for **every** worktree. That breaks what the tool promises ("each agent gets its own container on a unique port"):

- **No per-worktree isolation.** `CONTAINER_NAME` collapses to a single shared `agent-pg-skills` instead of `agent-pg-<worktree>`, so concurrent worktrees contend on one container.
- **Stale-port desync.** The per-worktree state file lands under `.claude/skills/`, and `up` re-exports the port saved in that file without checking the container's *live* published port. A leftover state file (e.g. port `50000`) while the container is actually on another port (e.g. `58239`) makes integration runs fail with connection-refused.
- The generated `.agent-db.env` was not gitignored, and a copy had been accidentally committed.

## Fix

- Resolve `PROJECT_DIR` three levels up to the actual worktree root, so `WORKTREE_ID` is the real worktree name, `CONTAINER_NAME` is genuinely per-worktree, and the state file lands at the worktree root.
- Gitignore `.agent-db.env`.
- Untrack the accidentally-committed `.claude/skills/.agent-db.env`.

## Verification

```
$ basename "$(cd .claude/skills/agent-db/../../.. && pwd)"
salesagent          # was: skills
```

The container name now follows the worktree (`agent-pg-<worktree>`), so worktrees no longer share a container and the exported `DATABASE_URL` matches the running instance.


KonstantinMirin (approved):
Verified the fix:

- `PROJECT_DIR` now resolves three levels up to the worktree root, so `WORKTREE_ID`/`CONTAINER_NAME` (`agent-pg-<worktree>`)/`STATE_FILE` are genuinely per-worktree — confirmed `basename` yields the worktree name, not the literal `skills`.
- Used consistently across all derived values in the script.
- `.agent-db.env` is gitignored and the accidentally-committed copy is untracked.

Root-cause fix, minimal and clean. LGTM.

PR #1457, opened 2026-06-17
refactor(test): adopt iter_call_expressions in architecture guards (Pattern 1b, #1234)

### LINKED ISSUE #1453: refactor(test): adopt iter_call_expressions in architecture guards (Pattern 1b, #1234)

---

### Problem

Same duplication class as Pattern 1a, different helper. `iter_call_expressions` in `tests/unit/_architecture_helpers.py` shipped in #1390 but has **zero external adopters**; **~33** `test_architecture_*.py` files still inline `ast.walk` + `isinstance(node, ast.Call)`. Same duplication pattern deferred to this ticket — **highest priority** per [#1390 review follow-up](https://github.com/prebid/salesagent/pull/1390#pullrequestreview-4510040035).

Merged baseline: `upstream/main` @ #1390.

### Class-level done check (reviewer runnable)

All three must pass on PR tip:

```bash
# 1. No hand-rolled Call walk in arch guards (only helper + regression guard exempts)
rg -n 'isinstance\([^)]*ast\.Call\)' tests/unit/test_architecture_*.py \
  | rg -v 'no_handrolled_call_walk|_architecture_helpers|helpers_contract|known-bad|probe'
# → expect: empty (or only non-detection uses documented in PR)

# 2. Regression guard green + catches probe
uv run pytest tests/unit/test_architecture_no_handrolled_call_walk.py -v

# 3. Helper contract tests green
uv run pytest tests/unit/test_architecture_helpers_contract.py -k iter_call -v
```

### Required deliverables

| # | Deliverable | Model |
|---|-------------|-------|
| 1 | Migrate every **Call-detection loop** in `test_architecture_*.py` to `iter_call_expressions(tree)` or `iter_call_expressions(subtree, name=...)` | Mechanical, like allowlist migration |
| 2 | **Self-closing regression guard** `test_architecture_no_handrolled_call_walk.py` | Mirror `test_architecture_no_handrolled_allowlist_diff.py`: AST-scan, exempt list, probe self-test |
| 3 | Contract tests for unfiltered + name-filtered `iter_call_expressions` | Extend `test_architecture_helpers_contract.py` |
| 4 | `make quality` green | Standard |

### Migration rules (prevents partial fixes)

- **Migrate:** inner loops whose purpose is finding `ast.Call` nodes
- **Keep:** outer `ast.walk` that filters `FunctionDef` / `Import` / `Assign` first — only replace the **inner** Call loop with `iter_call_expressions(func_node)`
- **Do not migrate:** `test_transport_agnostic_impl.py`, `test_no_toolerror_in_impl.py`, other non-`test_architecture_*` unit tests; `parse_module` / `repo_root()` adoption (separate concern, not Pattern 1b)

### Out of scope (will not re-litigate)

- Pattern 1a / allowlist helper (closed)
- Moving pre-push hooks to CI (#1454)
- Coverage-map parser (#1455)
- skipif / hook-count (#1456)
- Changing `iter_call_expressions` type hint (`ast.Module` → `ast.AST`) unless needed for mypy

### Suggested PR

- Branch: `refactor/guard-iter-call-expressions`
- Title: `refactor(test): adopt iter_call_expressions in architecture guards (Pattern 1b, #1234)`
- Closes #1453

### When to implement

| Field | Value |
|-------|-------|
| **Stack position** | After **PR5 #1400 merge** recommended (same `test_architecture_*.py` files PR5 expanded) |
| **Blockers** | None hard — independent of #1454, #1455, #1456 |
| **Blocks** | Nothing |
| **Branch from** | `upstream/main` **after PR5 merge** to avoid rebase conflicts with PR5 review-fix (`88a3882a`: `_architecture_helpers.py`, uv/postgres anchor guards) |
| **Priority** | Highest per #1390 review — but **schedule after PR5** for cleaner diff, not because of functional dependency |

**Suggested timing:** Can start in parallel with #1456 / #1454 once PR5 is on main. Do **not** bundle into PR6 rebase (#1401).

mkostromin-sigma:
## Summary
- Migrate architecture guards from hand-rolled `ast.walk` + `isinstance(..., ast.Call)` loops to the shared `iter_call_expressions()` helper.
- Add self-closing regression guard `test_architecture_no_handrolled_call_walk.py`.
- Extend helper contract tests for filtered/unfiltered/subtree `iter_call_expressions`.
- Extract `iter_architecture_guard_trees()` to share guard file scanning with the allowlist-diff guard (DRY).

Closes #1453

## Documented non-detection `isinstance(ast.Call)` uses (intentionally kept)
- Decorator checks: `isinstance(dec, ast.Call)` in BDD guards
- Raise.exc checks: `no_value_error_in_impl`, `adapter_raise_site_coverage`, `no_error_flattening`, `no_base_notfound_raise`
- Type checks on assert/compare nodes: `bdd_assertion_strength`, `bdd_no_request_in_then`, `bdd_no_trivial_assertions`
- Multi-type scanners: `sdk_idempotency_seam`, `auth_helper_context_wiring`, `no_raw_thread_registry`
- Allowlist guard helper: `no_handrolled_allowlist_diff` (`sorted(...)` detection)

## Test plan
- [x] `uv run pytest tests/unit/test_architecture_no_handrolled_call_walk.py -v`
- [x] `uv run pytest tests/unit/test_architecture_helpers_contract.py -k iter_call -v`
- [x] grep done-check (documented exceptions only)
- [x] `make quality`

### DIFF
--- a/tests/unit/_architecture_helpers.py
+++ b/tests/unit/_architecture_helpers.py
@@ -161,7 +161,7 @@ def find_plain_json_column_violations(tree: ast.Module) -> list[int]:
     return lines
 
 
-def iter_call_expressions(tree: ast.Module, name: str | None = None) -> Iterator[ast.Call]:
+def iter_call_expressions(tree: ast.AST, name: str | None = None) -> Iterator[ast.Call]:
     """Yield Call nodes, optionally filtered by callable name."""
     for node in ast.walk(tree):
         if not isinstance(node, ast.Call):
@@ -179,16 +179,18 @@ def iter_call_expressions(tree: ast.Module, name: str | None = None) -> Iterator
 def iter_architecture_guard_trees(
     *,
     exempt: Iterable[Path] | None = None,
-) -> Iterator[tuple[Path, ast.Module]]:
-    """Yield ``(repo_relative_path, parsed_tree)`` for each ``test_architecture_*.py`` module."""
+) -> Iterator[tuple[ast.Module, Path]]:
+    """Yield ``(parsed_tree, repo_relative_path)`` for each ``test_architecture_*.py`` module."""
     repo = repo_root()
     skip = set(exempt or ())
     for path in sorted((repo / "tests" / "unit").glob("test_architecture_*.py")):
         rel = path.relative_to(repo)
         if rel in skip:
             continue
-        tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
-        yield rel, tree
+        tree = safe_parse(path)
+        if tree is None:
+            continue
+        yield tree, rel
 
 
 # ---------------------------------------------------------------------------
--- a/tests/unit/_architecture_helpers.py
+++ b/tests/unit/_architecture_helpers.py
@@ -176,6 +176,21 @@ def iter_call_expressions(tree: ast.Module, name: str | None = None) -> Iterator
             yield node
 
 
+def iter_architecture_guard_trees(
+    *,
+    exempt: Iterable[Path] | None = None,
+) -> Iterator[tuple[Path, ast.Module]]:
+    """Yield ``(repo_relative_path, parsed_tree)`` for each ``test_architecture_*.py`` module."""
+    repo = repo_root()
+    skip = set(exempt or ())
+    for path in sorted((repo / "tests" / "unit").glob("test_architecture_*.py")):
+        rel = path.relative_to(repo)
+        if rel in skip:
+            continue
+        tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
+        yield rel, tree
+
+
 # ---------------------------------------------------------------------------
 # File iteration
 # ---------------------------------------------------------------------------
--- a/tests/unit/test_architecture_a2a_test_uses_factory.py
+++ b/tests/unit/test_architecture_a2a_test_uses_factory.py
@@ -19,6 +19,8 @@ from pathlib import Path
 
 import pytest
 
+from tests.unit._architecture_helpers import iter_call_expressions
+
 
 def _a2a_test_files() -> list[Path]:
     repo_root = Path(__file__).resolve().parents[2]
@@ -31,13 +33,12 @@ def _a2a_test_files() -> list[Path]:
 def _find_resolved_identity_calls(path: Path) -> list[int]:
     tree = ast.parse(path.read_text(), filename=str(path))
     lines: list[int] = []
-    for node in ast.walk(tree):
-        if isinstance(node, ast.Call):
-            func = node.func
-            if isinstance(func, ast.Name) and func.id == "ResolvedIdentity":
-                lines.append(node.lineno)
-            elif isinstance(func, ast.Attribute) and func.attr == "ResolvedIdentity":
-                lines.append(node.lineno)
+    for node in iter_call_expressions(tree):
+        func = node.func
+        if isinstance(func, ast.Name) and func.id == "ResolvedIdentity":
+            lines.append(node.lineno)
+        elif isinstance(func, ast.Attribute) and func.attr == "ResolvedIdentity":
+            lines.append(node.lineno)
     return lines
 
 
--- a/tests/unit/test_architecture_adapter_raise_site_coverage.py
+++ b/tests/unit/test_architecture_adapter_raise_site_coverage.py
@@ -92,9 +92,7 @@ def collect_pytest_raises_classes() -> set[str]:
         if tree is None:
             continue
         for node in iter_call_expressions(tree, name="raises"):
-            func = node.func
-            is_raises = isinstance(func, ast.Attribute) or isinstance(func, ast.Name)
-            if not is_raises or not node.args:
+            if not node.args:
                 continue
             first = node.args[0]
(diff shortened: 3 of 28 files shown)


ChrisHuie:
Reviewed against `e1a3aab9` in an isolated worktree. The migration is complete and faithful — verified empirically, not just by inspection: 294/294 architecture guards green, the highest-churn migration (`bdd_no_dict_registry`) is result-equivalent to a clean rewrite across 10 adversarial inputs, and the new regression guard's oracle is real (reverting a migration reddens it). Nice work — two items worth addressing before merge, neither breaks correctness today.

## SHOULD-FIX 1 — regression guard matcher (`test_architecture_no_handrolled_call_walk.py`)

**Crash:** `test.operand.args[1].value.id` (line 66) and `test.args[1].value.id` (line 90) read `.value.id` after checking only that `args[1]` is an `ast.Attribute`. A dotted base like `isinstance(node, mod.ast.Call)` makes `.value` an `Attribute` with no `.id` → `AttributeError` that aborts the whole scan (reproduced). Guard it:

```diff
- and test.operand.args[1].value.id == "ast"
+ and isinstance(test.operand.args[1].value, ast.Name)
+ and test.operand.args[1].value.id == "ast"
```

(mirror the same at line 90.) The sibling `no_handrolled_allowlist_diff` guard already does the `isinstance`-first check in `_is_set_subtraction` — worth matching that pattern here.

**Completeness:** the matcher models only the two canonical shapes. These reintroduction forms pass silently (each verified against the real matcher): pre-collected `nodes = list(ast.walk(t))`, a statement before the `if` guard, `from ast import walk`, a compound `if isinstance(n, ast.Call) and ...:`, and `type(n) is not ast.Call`. Either broaden the structural match (anchor on any `ast.For` over `ast.walk(...)` whose body contains an `isinstance(<target>, ast.Call)` test — not just `body[0]`, allow `BoolOp` operands), or add a docstring line listing the deliberately out-of-scope shapes so an empty result isn't read as total coverage.

## SHOULD-FIX 2 — `test_architecture_bdd_no_dict_registry.py:50-59`

The `for node in iter_call_expressions(func)` at line 58 is nested inside `for node in ast.walk(func)` at line 50 and rebinds `node` (O(N²) re-walk + shadowed iterator; correct only via the early `return`). It's result-equivalent today, but split into two sibling loops with a distinct variable — the way `weak_mock_assertions.py` / `bdd_no_trivial_assertions.py` handle the same combined Assign+Call walk.

## NITs

- **`test_architecture_query_type_safety.py:88`** — `value = node.func.value` can `AttributeError` on a bare `in_()` (the `name=` filter matches both `in_()` and `x.in_()`); re-add `if not isinstance(node.func, ast.Attribute): continue` before line 88. (`in_` path only — `filter_by` uses `node.keywords`, so it's safe.)
- **`test_architecture_no_handrolled_call_walk.py` self-tests (lines 130 / 150 / 170)** write a probe whose filename matches the scan glob, and `iter_architecture_guard_trees` parses with no `try/except` — latent flake under `-n auto`, and a `SIGKILL`'d run leaves a probe that fails the guard on the next run. Prefer `tmp_path` with a name that doesn't match `test_architecture_*`.
- **`iter_architecture_guard_trees` (`_architecture_helpers.py:179`)** yields `(rel, tree)` — reversed vs `iter_module_trees`'s `(tree, rel)` — and uses raising `ast.parse` where the existing `safe_parse` would also remove the probe flake above.
- **`iter_call_expressions` (`_architecture_helpers.py:164`)** is typed `ast.Module`, but 4 new callers pass subtree (`FunctionDef`) nodes → widen to `ast.AST`.
- **`test_architecture_adapter_raise_site_coverage.py:96`** — `is_raises` is a tautology after the `name="raises"` filter (the helper only yields `Name`/`Attribute` funcs); dead clause.
- **`test_architecture_helpers_contract.py`** — the `name=` test never pins the receiver-agnostic match (`h()` and `x.h()` coexisting under `name="h"`).

All line numbers are at `e1a3aab9`.

mkostromin-sigma:
Addressed all items from https://github.com/prebid/salesagent/pull/1457#issuecomment-4743825440:

**SHOULD-FIX**
- [x] **1a** — `isinstance(..., ast.Name)` guard before `.value.id` in both matcher branches (mirrors `no_handrolled_allowlist_diff` pattern)
- [x] **1b** — Added module docstring "Deliberately out of scope" listing the five documented shapes + `test_call_walk_guard_allows_out_of_scope_shapes` self-test
- [x] **2** — Split nested loops in `bdd_no_dict_registry` into sibling `walk_node` / `call_node` loops

**NITs**
- [x] **3** — Re-added `isinstance(node.func, ast.Attribute)` guard before `node.func.value` in `query_type_safety`
- [x] **4** — Self-tests now exercise matcher functions on synthetic AST (no `test_architecture_probe_*` files)
- [x] **5** — `iter_architecture_guard_trees` yields `(tree, rel)` + uses `safe_parse`
- [x] **6** — `iter_call_expressions(tree: ast.AST)`
- [x] **7** — Removed tautologicaluse
- [x] **8** — Contract test pins receiver-agnostic `name="h"` match (`h()` + `x.h()`)

CI green on latest push. Ready for re-review.

ChrisHuie (approved):

PR #1458, opened 2026-06-17
ci: move pre-push-only checks to make quality-ci (Pattern 2, #1234)

### LINKED ISSUE #1454: ci: move pre-push-only checks to make quality-ci (Pattern 2, #1234)

---

### Problem

Four checks are **pre-push only** — bypassable via `--no-verify` or missing hook install. **None run in `make quality-ci`** (CI Quality Gate). #1390's layered model doc says CI is authoritative; these cheap checks must live there too.

**Not a #1390 regression:** CI never ran pre-commit before #1390 either. This issue closes a **documented enforcement gap**.

### Checks (4 + coverage-map gap)

| Hook ID | Script | Notes |
|---------|--------|-------|
| `check-route-conflicts` | `.pre-commit-hooks/check_route_conflicts.py` | No args; loads Flask app |
| `type-ignore-no-regression` | `.pre-commit-hooks/check_type_ignore_count.py` | Scans `src/` vs baseline |
| `check-docs-links` | `.pre-commit-hooks/check_docs_links.py` | Auto-discovers markdown |
| `no-hardcoded-urls` | `.pre-commit-hooks/check_hardcoded_urls.py` | **Must pass filenames** — empty argv is a no-op |

**Also fix:** `no-hardcoded-urls` is missing from `.pre-commit-coverage-map.yml` today — add entry when wiring CI.

### Class-level done check

```bash
# All four scripts referenced in quality-ci
rg 'check_route_conflicts|check_type_ignore_count|check_docs_links|check_hardcoded_urls' Makefile

# Coverage map shows ci-step for all four
rg 'pre-push \+ ci-step' .pre-commit-coverage-map.yml | wc -l   # ≥4 relevant entries
rg 'no-hardcoded-urls' .pre-commit-coverage-map.yml             # must exist

# CI path unchanged except quality-ci content
make quality-ci   # exits 0 locally
```

Update `docs/development/ci-pipeline.md` (~lines 193–194) to list all four alongside existing `check-gam-auth-support` entries.

### Makefile addition (exact intent)

```makefile
uv run python .pre-commit-hooks/check_route_conflicts.py
uv run python .pre-commit-hooks/check_type_ignore_count.py
uv run python .pre-commit-hooks/check_docs_links.py
uv run python .pre-commit-hooks/check_hardcoded_urls.py $$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null)
```

No change to `.github/workflows/ci.yml` job graph — Quality Gate already runs `make quality-ci`.

### Out of scope

- Moving hooks back to commit stage (D27 ceiling 12)
- Adding hooks to `pre-commit run` in CI (project rule: `make quality-ci` is authoritative, not `pre-commit run --all-files`)
- Pattern 1b / coverage-map parser / skipif fixes

### Merge order

Land **before or with** #1455 — parser test will assert `enforced_by: pre-push + ci-step` for these four.

### Suggested PR

- Branch: `ci/quality-ci-authoritative-checks`
- Title: `ci: move pre-push-only checks to make quality-ci (Pattern 2, #1234)`
- Closes #1454

### When to implement

| Field | Value |
|-------|-------|
| **Stack position** | After #1234 stack (PR5 #1400 + PR6 #1401) merge **or in parallel with #1456** |
| **Blockers** | None from other follow-ups |
| **Blocks** | **#1455** — coverage-map parser expects `pre-push + ci-step` + `Makefile::quality-ci` for all four hooks |
| **Branch from** | `upstream/main` after PR6 merge (or rebase onto main if started earlier) |
| **Why before #1455** | #1455 Part A parser asserts referential integrity against `quality-ci` entries you add here |

**Suggested timing:** Land before #1455. Can parallel #1456 and #1453.

mkostromin-sigma:
## Summary
- Add four pre-push-only checks to `make quality-ci`: route conflicts, type-ignore ratchet, docs links, hardcoded URLs
- Update `.pre-commit-coverage-map.yml` (`pre-push + ci-step` for all four; add missing `no-hardcoded-urls`)
- Document new checks in `docs/development/ci-pipeline.md`

Closes #1454

## Test plan
- [x] `make quality-ci` passes locally
- [x] CI Quality Gate green on PR

### DIFF
--- a/.pre-commit-coverage-map.yml
+++ b/.pre-commit-coverage-map.yml
@@ -39,14 +39,17 @@ mcp-contract-validation:
   enforced_by: ci
   location: .github/workflows/ci.yml::schema-contract
 check-docs-links:
-  enforced_by: pre-push
-  location: stages_prepush
+  enforced_by: pre-push + ci-step
+  location: stages_prepush + Makefile::quality-ci
 check-route-conflicts:
-  enforced_by: pre-push
-  location: stages_prepush
+  enforced_by: pre-push + ci-step
+  location: stages_prepush + Makefile::quality-ci
 type-ignore-no-regression:
-  enforced_by: pre-push
-  location: stages_prepush
+  enforced_by: pre-push + ci-step
+  location: stages_prepush + Makefile::quality-ci
+no-hardcoded-urls:
+  enforced_by: pre-push + ci-step
+  location: stages_prepush + Makefile::quality-ci
 no-skip-integration-v2:
   enforced_by: deleted
   location: dead-code (tests/integration_v2/ does not exist)
--- a/.pre-commit-hooks/check_hardcoded_urls.py
+++ b/.pre-commit-hooks/check_hardcoded_urls.py
@@ -68,8 +68,12 @@ def main(filenames: list[str]) -> int:
     Check all provided files for hardcoded URLs.
 
     Returns:
-        0 if no violations, 1 if violations found
+        0 if no violations, 1 if violations found or no files provided
     """
+    if not filenames:
+        print("check_hardcoded_urls: no files provided", file=sys.stderr)
+        return 1
+
     all_violations = []
 
     for filename in filenames:
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ quality-ci:
 	uv run python .pre-commit-hooks/check-gam-auth-support.py
 	uv run python scripts/hooks/check_response_attribute_access.py $$(find src -name '*.py')
 	uv run python .pre-commit-hooks/check_roundtrip_tests.py
+	uv run python .pre-commit-hooks/check_route_conflicts.py
+	uv run python .pre-commit-hooks/check_type_ignore_count.py
+	uv run python .pre-commit-hooks/check_docs_links.py
+	uv run python .pre-commit-hooks/check_hardcoded_urls.py $$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null)
 
 quality:
 	$(MAKE) quality-ci
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,9 @@ quality-ci:
 	uv run python .pre-commit-hooks/check_route_conflicts.py
 	uv run python .pre-commit-hooks/check_type_ignore_count.py
 	uv run python .pre-commit-hooks/check_docs_links.py
-	uv run python .pre-commit-hooks/check_hardcoded_urls.py $$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null)
+	@files=$$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null); \
+	test -n "$$files" || { echo "check_hardcoded_urls: no files matched find glob"; exit 1; }; \
+	uv run python .pre-commit-hooks/check_hardcoded_urls.py $$files
 
 quality:
 	$(MAKE) quality-ci
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,7 @@ quality-ci:
 	uv run python .pre-commit-hooks/check_route_conflicts.py
 	uv run python .pre-commit-hooks/check_type_ignore_count.py
 	uv run python .pre-commit-hooks/check_docs_links.py
-	@files=$$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null); \
-	test -n "$$files" || { echo "check_hardcoded_urls: no files matched find glob"; exit 1; }; \
-	uv run python .pre-commit-hooks/check_hardcoded_urls.py $$files
+	uv run python .pre-commit-hooks/check_hardcoded_urls.py $$(find templates static -type f \( -name '*.html' -o -name '*.js' \) 2>/dev/null)
 
 quality:
 	$(MAKE) quality-ci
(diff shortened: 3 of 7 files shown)


ChrisHuie (commented):
Reviewed against the #1454 intent and verified the wiring on the PR head. The four checks are correctly added to `make quality-ci`, and I confirmed each is load-bearing (exits non-zero on a real violation). The `pass_filenames` handling is right — the three `pass_filenames: false` hooks are invoked with no args, and `no-hardcoded-urls` is fed filenames, so the empty-argv no-op called out in #1454 is avoided. The Quality Gate job runs `make quality-ci`, so the checks engage in CI; coverage-map entries are consistent and the previously-missing `no-hardcoded-urls` entry is added.

One in-scope fix, plus a couple of notes:

**1 · Doc label is inaccurate for the four added checks** — `docs/development/ci-pipeline.md`
They're added under the heading "CI-only checks absorbed into `make quality-ci`", but the coverage-map this PR adds marks all four `pre-push + ci-step`, and the layer table just above lists docs/routes as pre-push examples. The four pre-existing entries under that heading are genuinely CI-only; the four added are dual-stage. Suggested rewrite:

    Checks invoked directly by `make quality-ci` (not via `pre-commit run --all-files` in the Quality Gate):
    - CI-only (no commit/pre-push stage): check_code_duplication, check-gam-auth-support,
      check_response_attribute_access, check_roundtrip_tests.
    - Also enforced at pre-push (dual-stage, `pre-push + ci-step` in the coverage map):
      check_route_conflicts, check_type_ignore_count, check_docs_links, check_hardcoded_urls.

**2 · `make quality` can now leave an unstaged edit to `.type-ignore-baseline`** (heads-up, not a blocker)
`check_type_ignore_count.py` auto-rewrites the tracked baseline when the `src/` type-ignore count decreases (no flag required). Since `make quality` runs `quality-ci`, a local `make quality` can now dirty the tree — the same behavior `check_code_duplication` already has with `.duplication-baseline` in `quality-ci`. Harmless in CI (the write is ephemeral). Might be worth a one-line note in CONTRIBUTING that `make quality` may update these baseline files.

**3 · Latent no-op (minor)** — `check_hardcoded_urls.py` is fed `$(find templates static … 2>/dev/null)`; the script returns 0 on empty argv, so if those directories ever move the suppressed `find` error would make the step pass vacuously. Matches the command #1454 prescribed, so optional hardening.

**Pre-existing (not introduced here) — filed as #1466:** the `no-hardcoded-urls` matcher only flags `fetch('/api/...')`, so a hardcoded `fetch('/admin/...')` in `templates/inventory_profiles_list.html` passes the check, and `src/landing/templates/` sits outside both the glob and the hook's `files:` regex. Promoting the check to a CI-authoritative gate makes widening the matcher more valuable.

Verification: ran each check's failing oracle (broken internal doc link; `type: ignore` count increase; injected hardcoded `fetch`) and each exits non-zero; the `find` glob resolves to 77 files; coverage-map YAML is valid with no existing parser consuming it; `.github/workflows/ci.yml` is untouched; CI is green on the head commit.

mkostromin-sigma:
Addressed review https://github.com/prebid/salesagent/pull/1458#pullrequestreview-4526824830 (@ChrisHuie):

- **Doc:** split CI-only vs dual-stage checks in `ci-pipeline.md`
- **Docs:** noted baseline auto-rewrite (`.duplication-baseline`, `.type-ignore-baseline`) in `contributing.md`
- **Hardening:** `make quality-ci` fails if the hardcoded-urls find glob is empty (Makefile guard before invoking the hook)

CI green on `f0081ca`. Ready for approve.

ChrisHuie:
Thanks for the quick turnaround. The rebase dropped the unrelated `.beads/.migration-hint-ts` change and the branch is now current with `main`. I re-verified the three earlier points still hold: the docs CI-only / dual-stage split matches the coverage map, the baseline note is in `contributing.md`, and the `no-hardcoded-urls` guard fails loud on an empty glob while still catching a real violation. CI is green.

Two optional polish items — neither blocking, fine to fold in now or defer to a follow-up:

1. **`Makefile`** — the new empty-glob guard is on the `check_hardcoded_urls` line only. The sibling `check_response_attribute_access.py $(find src -name '*.py')` one line above has the same silent-pass-on-empty-argv behavior and stays unguarded (lower risk, since `src/` is never empty). If you want the "find-fed argv must not silently no-op" invariant applied consistently, either guard that line the same way, or have both scripts exit non-zero on empty argv (covers every caller in one place).

2. **`docs/development/ci-pipeline.md`** — the Dual-stage list names hooks by script (`check_route_conflicts`, `check_hardcoded_urls`) while the coverage map it links to uses hook ids (`check-route-conflicts`, `no-hardcoded-urls`). Matching the id form makes the cross-reference exact. Cosmetic.

ChrisHuie (approved):

mkostromin-sigma:
Thanks again for the review and approve, @ChrisHuie — both optional polish items from [your follow-up comment](https://github.com/prebid/salesagent/pull/1458#issuecomment-4752434291) are addressed on the latest head:

1. **Empty-argv / silent no-op** — `check_hardcoded_urls.py` and `check_response_attribute_access.py` now exit non-zero when given no files, so the invariant holds for every caller (including the Makefile `find`-fed lines). The Makefile-only shell guard for hardcoded URLs was removed in favor of the script-level check.

2. **Dual-stage hook ids in docs** — `docs/development/ci-pipeline.md` now uses the coverage-map hook ids (`check-route-conflicts`, `type-ignore-no-regression`, `check-docs-links`, `no-hardcoded-urls`) instead of script filenames.

CI is green on `67d413d1` (including Security Audit / pip-audit after a transitive bump of `pydantic-settings` and `zeep` in `uv.lock`).

Since the PR is approved, planning to merge shortly unless you spot anything else.

PR #1460, opened 2026-06-17
fix(ci): tighten check_no_skip_tests and hook-count guard (Pattern 4, #1234)

### LINKED ISSUE #1456: fix(ci): tighten check_no_skip_tests and hook-count guard (Pattern 4, #1234)

---

### Problem

Two **latent defects** in checks ported/consolidated by #1390 — not enforcement gaps, but wrong-by-construction logic with missing self-tests.

### Problem 1 — `check_no_skip_tests` regex

Current in `.pre-commit-hooks/check_repo_invariants.py`:

```python
pattern = re.compile(r"@pytest\.mark\.skip(?!_ci)")
```

`@pytest.mark.skipif` **matches** (`skipif` starts with `skip`). Repo already uses `skipif` in integration tests (`tests/integration/test_mcp_client_util.py`).

**Fix:** `@pytest\.mark\.skip(?!if|_ci)` or equivalent that matches only bare `skip`.

**Allowed markers after fix:** `@pytest.mark.skipif`, `@pytest.mark.skip_ci`  
**Forbidden:** `@pytest.mark.skip` (bare)

### Problem 2 — hook-count guard self-test gaps

`test_architecture_pre_commit_hook_count.py`: `_count_commit_stage_hooks` logic is **correct** for D31 (hooks without `stages` inherit `default_stages`). **Do not change counting semantics** — only add missing self-tests:

- Default-stages fallback (hook with no `stages` key → counted as commit-stage)
- Under-ceiling detection (counter returns `< COMMIT_STAGE_MIN`; wire into assertion or document that main test covers live config)

`test_architecture_repo_invariants.py` covers `check_no_fn_calls` only — add parallel tests for `check_no_skip_tests`.

### Class-level done check

```bash
uv run pytest tests/unit/test_architecture_repo_invariants.py \
               tests/unit/test_architecture_pre_commit_hook_count.py -v
make quality

# Manual sanity: skipif not flagged
python -c "
import re; p=re.compile(r'@pytest\.mark\.skip(?!if|_ci)')
assert not p.search('@pytest.mark.skipif(True, reason=\"x\")')
assert p.search('@pytest.mark.skip')
"
```

### Out of scope

- Changing repo-invariants hook stage (stays commit-stage)
- Changing D27 MIN/MAX constants
- Pattern 1b (#1453) / CI gap (#1454) / coverage-map parser (#1455)

### Suggested PR

- Branch: `fix/ci-skipif-hook-count-guards`
- Title: `fix(ci): tighten check_no_skip_tests and hook-count guard (Pattern 4, #1234)`
- Closes #1456
- **Independent** — good first PR (~3 files, <60 lines)

### When to implement

| Field | Value |
|-------|-------|
| **Stack position** | First follow-up after #1234 code PRs — **or in parallel** with finishing PR5/PR6 |
| **Blockers** | None — fully **independent** of #1453, #1454, #1455 |
| **Blocks** | Nothing |
| **Branch from** | `upstream/main` (current main is fine; no need to wait for PR5/PR6 merge) |
| **Why first** | ~3 files, <60 lines; fixes latent skipif false-positive — good quick win while stack merges |

**Suggested timing:** Start anytime after #1390 is on main. Safe to land before, during, or after PR5 (#1400) / PR6 (#1401) merge.

mkostromin-sigma:
## Summary
- Fix `check_no_skip_tests` regex so `@pytest.mark.skipif` is not falsely flagged (`skip(?!if|_ci)`).
- Add self-tests for skip detection (bare skip vs skipif) in `test_architecture_repo_invariants.py`.
- Add hook-count self-tests for default-stages fallback and under-ceiling detection.

Closes #1456

## Test plan
- [x] `uv run pytest tests/unit/test_architecture_repo_invariants.py tests/unit/test_architecture_pre_commit_hook_count.py -v`
- [x] `make quality`

### DIFF
--- a/.pre-commit-hooks/check_repo_invariants.py
+++ b/.pre-commit-hooks/check_repo_invariants.py
@@ -10,6 +10,17 @@ import re
 import sys
 from pathlib import Path
 
+_SKIP_RE = re.compile(r"@pytest\.mark\.skip(?!if|_ci)")
+
+
+def is_forbidden_skip_line(line: str) -> bool:
+    """True when *line* contains a bare ``@pytest.mark.skip`` decorator.
+
+    ``@pytest.mark.skipif`` and ``@pytest.mark.skip_ci`` are allowed.
+    Shared by the pre-commit hook and the Smoke Tests gate.
+    """
+    return bool(_SKIP_RE.search(line))
+
 
 def check_no_skip_tests(files: list[Path]) -> list[str]:
     """Forbid bare @pytest.mark.skip in test files.
@@ -17,13 +28,12 @@ def check_no_skip_tests(files: list[Path]) -> list[str]:
     ``@pytest.mark.skipif`` and ``@pytest.mark.skip_ci`` are allowed (conditional /
     CI-specific skips). Bare ``skip`` without justification is forbidden.
     """
-    pattern = re.compile(r"@pytest\.mark\.skip(?!if|_ci)")
     out: list[str] = []
     for filepath in files:
         if "tests/" not in str(filepath) or not filepath.name.startswith("test_"):
             continue
         for lineno, line in enumerate(filepath.read_text().splitlines(), 1):
-            if pattern.search(line):
+            if is_forbidden_skip_line(line):
                 out.append(f"{filepath}:{lineno}: @pytest.mark.skip forbidden (use skip_ci with justification)")
     return out
 
--- a/.pre-commit-hooks/check_repo_invariants.py
+++ b/.pre-commit-hooks/check_repo_invariants.py
@@ -13,7 +13,7 @@ from pathlib import Path
 
 def check_no_skip_tests(files: list[Path]) -> list[str]:
     """Forbid @pytest.mark.skip without skip_ci justification."""
-    pattern = re.compile(r"@pytest\.mark\.skip(?!_ci)")
+    pattern = re.compile(r"@pytest\.mark\.skip(?!if|_ci)")
     out: list[str] = []
     for filepath in files:
         if "tests/" not in str(filepath) or not filepath.name.startswith("test_"):
--- a/.pre-commit-hooks/check_repo_invariants.py
+++ b/.pre-commit-hooks/check_repo_invariants.py
@@ -12,7 +12,11 @@ from pathlib import Path
 
 
 def check_no_skip_tests(files: list[Path]) -> list[str]:
-    """Forbid @pytest.mark.skip without skip_ci justification."""
+    """Forbid bare @pytest.mark.skip in test files.
+
+    ``@pytest.mark.skipif`` and ``@pytest.mark.skip_ci`` are allowed (conditional /
+    CI-specific skips). Bare ``skip`` without justification is forbidden.
+    """
     pattern = re.compile(r"@pytest\.mark\.skip(?!if|_ci)")
     out: list[str] = []
     for filepath in files:
--- a/tests/smoke/test_smoke_basic.py
+++ b/tests/smoke/test_smoke_basic.py
@@ -181,6 +181,7 @@ class TestNoSkippedTests:
                 if line
                 and "pytest.skip(" not in line  # Allow runtime skips
                 and "skip_ci" not in line  # Allow CI-specific skips
+                and "skipif" not in line  # @pytest.mark.skipif is a conditional skip — allowed
                 and "skip_pattern" not in line  # Exclude this test file
                 and "test_no_skip" not in line  # Exclude this test
                 and "test_no_hardcoded_credentials" not in line  # Exclude disabled test
--- a/tests/smoke/test_smoke_basic.py
+++ b/tests/smoke/test_smoke_basic.py
@@ -1,8 +1,16 @@
 """Basic smoke tests that don't require running servers."""
 
+import importlib.util
 from pathlib import Path
 
 import pytest
+
+_REPO_INVARIANTS_HOOK = Path(__file__).resolve().parents[2] / ".pre-commit-hooks" / "check_repo_invariants.py"
+_spec = importlib.util.spec_from_file_location("check_repo_invariants", _REPO_INVARIANTS_HOOK)
+assert _spec and _spec.loader
+_check_repo_invariants = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_check_repo_invariants)
+_is_forbidden_skip_line = _check_repo_invariants.is_forbidden_skip_line
 
 
 class TestCriticalImports:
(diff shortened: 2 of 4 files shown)


ChrisHuie (commented):
**Reviewed at `0f81e76de`.** The regex fix is correct — verified against every skip form (bare `skip`, `skip(...)`, `skipif`, `skip_ci`) and by running the hook over the full tree (no false positives today). Four items below; **#1 is the main one to address before merge**, the rest are smaller.

### ✅ Correct
- `@pytest\.mark\.skip(?!if|_ci)` flags bare `@pytest.mark.skip` / `@pytest.mark.skip(...)` and correctly allows `@pytest.mark.skipif` and `@pytest.mark.skip_ci`.
- Lint / format / mypy clean; the live hook run is green on the current tree.

---

### 🔴 1 · `allows_skipif` doesn't actually exercise the fix
`tests/unit/test_architecture_repo_invariants.py` → `test_repo_invariants_skip_detector_allows_skipif`

The fixture uses the **assignment form**, which contains no `@pytest.mark.skip…` literal — so neither the old nor the new regex ever matches it, and the test passes identically against the *buggy* regex:

```python
'skip_no_agent = pytest.mark.skipif(True, reason="no agent")\n@skip_no_agent\n...'
```

Quick proof: revert only the regex (`check_repo_invariants.py` → `r"@pytest\.mark\.skip(?!_ci)"`) and run this test — it still **passes**, so it isn't guarding the change.

The `(?!if|_ci)` lookahead only affects the **decorator** form `@pytest.mark.skipif(...)`. Assert on that instead — built via concatenation so the literal doesn't trip the smoke-test grep on this file (the `catches_bare_skip` test already uses this trick):

```python
@pytest.mark.arch_guard
def test_repo_invariants_skip_detector_allows_skipif(tmp_path) -> None:
    ok_file = tmp_path / "tests" / "integration" / "test_probe.py"
    ok_file.parent.mkdir(parents=True)
    skipif_dec = "@pytest.mark." + "skipif"  # keep the literal out of this file's source
    ok_file.write_text(
        f'{skipif_dec}(True, reason="conditional")\ndef test_probe():\n    pass\n',
        encoding="utf-8",
    )
    hits = check_no_skip_tests([ok_file])
    assert not hits, "check_no_skip_tests must not flag @pytest.mark.skipif"
```

With the regex reverted, this version **fails**; with the fix it passes — i.e. it now actually locks the behavior.

---

### 🟠 2 · The same `skipif` false-positive lives in the Smoke Tests gate
`tests/smoke/test_smoke_basic.py` → `TestNoSkippedTests.test_no_skip_decorators`

This is a second enforcer of the no-skip rule, and it runs as a required `Smoke Tests` CI check. It greps `@pytest.mark.skip` and filters out `skip_ci` / `pytest.skip(` — but **not `skipif`**:

```text
# plant tests/_probe.py containing:  @pytest.mark.skipif(True, reason="x")
# the gate fails:
AssertionError: Found skip decorators:
  tests/_probe.py:@pytest.mark.skipif(True, reason="x")
```

So after this PR the hook *allows* decorator-form `skipif` while this gate still *rejects* it — the two disagree on the exact form the issue says to allow. It's dormant only because the two existing `skipif` sites use the assignment form (no leading `@`); the first `@pytest.mark.skipif(...)` **decorator** anyone adds turns CI red. (This is what commit 2 worked around by avoiding the literal — the root cause is here.)

One line in the filter chain fixes it:

```python
and "skipif" not in line  # @pytest.mark.skipif is a conditional skip — allowed
```

Verify: plant the temp file above; `test_no_skip_decorators` fails before, passes after.

Scope: strictly this is beyond what #1456 specs (it only covers the hook), so a follow-up is legitimate if you'd rather keep this PR minimal — but it's one line, in a gate you already touched, and it's the difference between "the hook allows `skipif`" and "`skipif` is actually allowed everywhere." Your call.

---

### 🟡 3 · Hook-count self-tests (minor)
`tests/unit/test_architecture_pre_commit_hook_count.py`

- `test_commit_hook_counter_counts_hooks_without_explicit_stages` sets `default_stages` explicitly, but the real `.pre-commit-config.yaml` omits that key and relies on the counter's hardcoded fallback (already covered by `test_pre_commit_hook_count_within_ceiling`). Drop the key so the test exercises the path the live config actually uses:
  ```python
  cfg = {"repos": [{"hooks": [{"id": "ruff"}, {"id": "mypy"}]}]}
  assert _count_commit_stage_hooks(cfg) == 2
  ```
- `test_commit_hook_counter_detects_under_ceiling`: the `< COMMIT_STAGE_MIN` assertion is implied by the `== COMMIT_STAGE_MIN - 1` assertion above it. Drop it, or make it assert the guard actually fires:
  ```python
  with pytest.raises(AssertionError):
      assert count >= COMMIT_STAGE_MIN
  ```

---

### 🟡 4 · Docstring nit
`check_repo_invariants.py` — `check_no_skip_tests`'s docstring could note that `skipif` / `skip_ci` are allowed, so the intent reads clearly.

---

Net: the fix is correct. Please tighten **#1** so the self-test locks the behavior, and decide fold-in vs. follow-up on **#2**. #3/#4 are polish. Thanks for the clean, focused PR!

mkostromin-sigma:
Addressed all review items in `4b99f20f`:

### 🔴 1 · `allows_skipif` self-test locks the regex fix
`test_repo_invariants_skip_detector_allows_skipif` now asserts on **decorator-form** `@pytest.mark.skipif(...)` (built via concatenation). Reverting the regex to `(?!_ci)` fails this test; the fix passes.

### 🟠 2 · Smoke Tests gate aligned
`TestNoSkippedTests.test_no_skip_decorators` now excludes lines containing `skipif`, matching the hook allowance for conditional skips.

### 🟡 3 · Hook-count self-tests
- `test_commit_hook_counter_counts_hooks_without_explicit_stages` — dropped explicit `default_stages` so the test exercises the live-config fallback path.
- `test_commit_hook_counter_detects_under_ceiling` — replaced redundant `< COMMIT_STAGE_MIN` with `pytest.raises(AssertionError)` on the ceiling guard.

### 🟡 4 · Docstring
`check_no_skip_tests` docstring documents that `skipif` / `skip_ci` are allowed.

**Verified locally:** 8 affected tests pass; `make quality` green.

@ChrisHuie ready for re-review when CI completes.

ChrisHuie (commented):
Thanks for the quick turnaround — re-verified the updates at `4b99f20f7` by execution (not just inspection), and the review items are resolved:

- **`allows_skipif` now locks the fix** — reverting the regex to `(?!_ci)` makes the test **fail**; with `(?!if|_ci)` it passes. It now actually guards the change. ✅
- **Smoke gate aligned** — passes on a planted `@pytest.mark.skipif` decorator, still fails on a bare `@pytest.mark.skip`. ✅
- **`counts_hooks_without_explicit_stages`** now exercises the real fallback (a broken default reddens it); docstring documents the allowed forms. ✅
- 305 arch-guards green and the live hook is clean on the full tree — no regression.

Two **optional** polish items below — both non-blocking and fine as a follow-up (they're beyond what #1456 scoped).

### 1. Consolidate the two skip-matchers (recommended follow-up)
Skip policy is now enforced in two places: the hook regex and the smoke-test grep+filter. They agree today, but the smoke filter is a substring check over the **whole grep line**, so a bare `@pytest.mark.skip` whose file path or `reason=` text contains `skipif` would slip past it (the hook still catches it, so it's backstopped). Pulling both onto one matcher removes that edge, gives the smoke side a real regression oracle, and makes the `test_skip_guard_single_source_of_truth_exists` claim literally true:

```python
# .pre-commit-hooks/check_repo_invariants.py
_SKIP_RE = re.compile(r"@pytest\.mark\.skip(?!if|_ci)")

def is_forbidden_skip_line(line: str) -> bool:
    return bool(_SKIP_RE.search(line))
```
…then have `TestNoSkippedTests` call `is_forbidden_skip_line` per line instead of the bespoke grep + substring-exclusion list (most of those excludes fall away once the precise regex is the single matcher).

### 2. Drop the tautological assertion
In `test_commit_hook_counter_detects_under_ceiling`:
```python
with pytest.raises(AssertionError):
    assert count >= COMMIT_STAGE_MIN   # 9 >= 10 always raises → exercises Python's assert, not the guard
```
The `assert count == COMMIT_STAGE_MIN - 1` line above it is the real oracle; this block can be removed.

Core fix and self-tests are solid — the above is polish, not a blocker. Thanks again!

mkostromin-sigma:
Thanks for the re-review at `4b99f20f` — glad all four items are verified.

**Optional polish (#1 consolidate skip-matchers, #2 tautological assert):** deferring to a follow-up so this PR stays scoped to #1456 / Pattern 4. Will track under #1468 or a small chore PR post-merge.

CI is green on head; PR is mergeable. @ChrisHuie — ready for approval and merge when you have a moment. Closes #1456.

mkostromin-sigma:
Addressed the two optional polish items in `5518ea7f` (no deferral):

### 1. Consolidated skip-matchers
- Extracted `_SKIP_RE` + `is_forbidden_skip_line()` in `.pre-commit-hooks/check_repo_invariants.py`
- `check_no_skip_tests` and `TestNoSkippedTests.test_no_skip_decorators` both delegate to the same matcher (walks `tests/**/test_*.py`, no grep + substring exclusions)

### 2. Dropped tautological assert
- Removed `pytest.raises(AssertionError)` block from `test_commit_hook_counter_detects_under_ceiling`; `assert count == COMMIT_STAGE_MIN - 1` remains the oracle

Verified locally: 24 affected tests pass; `make quality` green.

@ChrisHuie ready for approval — all review items (required + optional) addressed.

ChrisHuie (approved):

PR #1461, opened 2026-06-17
fix: link ObjectWorkflowMapping on auto-approve path before push notification (#1378)

### LINKED ISSUE #1378: Bug: push notification webhook not delivered on auto-approve path — missing ObjectWorkflowMapping row
# Bug: Push notification webhook not delivered on auto-approve path — missing `ObjectWorkflowMapping` row

**Component:** `src/core/tools/media_buy_create.py` — `_create_media_buy_impl()` auto-approve path  
**Severity:** High — silent data loss; media buy returns `status: active` but the buyer's webhook is never called  

---

## Summary

When a media buy is auto-approved (no human review step), `_create_media_buy_impl` calls
`ctx_manager.update_workflow_step(step.step_id, status="completed")` to mark the step done.
This internally triggers `_send_push_notifications`, which queries `ObjectWorkflowMapping` to
find the webhook URL for the media buy. On the auto-approve path no `ObjectWorkflowMapping`
row is created before `update_workflow_step` is called, so the function returns early and the
webhook is never sent.

## Affected code

```python
# src/core/context_manager.py — _send_push_notifications
stmt = select(ObjectWorkflowMapping).filter_by(step_id=step.step_id)
mappings = session.scalars(stmt).all()

if not mappings:
    console.print(f"[yellow]No object mappings found for step {step.step_id}[/yellow]")
    return   # ← exits here silently on the auto-approve path
```

On the **manual-approval path** the mapping is created by the workflow machinery before the
step is completed, so it works correctly there. On the **auto-approve path** no code creates
the mapping before `update_workflow_step` is called.

## Root cause

`ContextManager` already exposes a [`link_workflow_to_object()`](src/core/context_manager.py)
method designed exactly for this purpose:

```python
def link_workflow_to_object(
    self,
    step_id: str,
    object_type: str,
    object_id: str,
    action: str | None = None,
    tenant_id: str | None = None,
) -> None:
    """Link a workflow step to an object after the step is created."""
    ...
```

This method is **never called anywhere in the codebase**. Instead, both the manual-approval
path and the auto-approve path contain inline `ObjectWorkflowMapping` inserts that duplicate
its logic — and the two copies have inconsistent error handling:

| Path | Location | Error handling |
|---|---|---|
| Manual-approval | `media_buy_create.py` ~line 2567 | Bare `with` block — exception propagates, failing the request |
| Auto-approve | `media_buy_create.py` ~line 3775 | `try/except` — exception is swallowed silently |

## Symptom

Media buy creation succeeds and returns HTTP 200 with `status: active`, but the buyer's push
notification webhook endpoint receives no POST request. No error is logged — the failure is
completely silent.

## Reproduction

1. Configure a principal with a push notification webhook URL.
2. Create a media buy for a product that auto-approves (no manual review workflow step).
3. Observe that the webhook endpoint receives no POST request despite the media buy
   returning `status: active`.

### Immediate (bug fix)

Insert an `ObjectWorkflowMapping` row on the auto-approve path **before** calling
`update_workflow_step`:

```python
# Link workflow step to media buy so _send_push_notifications can find the webhook URL
ctx_manager.link_workflow_to_object(
    step_id=step.step_id,
    object_type="media_buy",
    object_id=response.media_buy_id,
    action="create",
    tenant_id=tenant["tenant_id"],
)

# Mark workflow step as completed on success (triggers _send_push_notifications)
ctx_manager.update_workflow_step(step.step_id, status="completed")
```

### Follow-up (DRY / consistency)

Replace both inline `ObjectWorkflowMapping` inserts (manual-approval path and auto-approve
path) with calls to the existing `ctx_manager.link_workflow_to_object()` method. This:

- Eliminates the duplicated insert logic
- Gives both paths consistent error handling (the method already has its own `try/except`
  with rollback)
- Makes `link_workflow_to_object()` actually used, removing dead code risk

torbenbrodt:
## Summary

Fixes #1378 — push notification webhook silently dropped on auto-approve path.

## Root cause

On the auto-approve path, `_create_media_buy_impl` called `ctx_manager.update_workflow_step(step.step_id, status="completed")` without first creating an `ObjectWorkflowMapping` row. `update_workflow_step` internally triggers `_send_push_notifications`, which queries `ObjectWorkflowMapping` to find the webhook URL. With no row present it returned early silently — the webhook was never delivered.

### Bug fix (auto-approve path)
`src/core/tools/media_buy_create.py` — call `ctx_manager.link_workflow_to_object()` **before** `update_workflow_step(status="completed")` on the auto-approve path so `_send_push_notifications` can find the mapping row.

### DRY follow-up (manual-approval path)
Replace the inline `ObjectWorkflowMapping` insert on the manual-approval path (~line 2817) with a call to the same `ctx_manager.link_workflow_to_object()` method. This:
- Eliminates duplicated insert logic
- Gives both paths consistent error handling (the method has its own `try/except` with rollback)
- Removes the unused-dead-code risk from `link_workflow_to_object()` never being called

### Import cleanup
Remove `ObjectWorkflowMapping` from the import in `media_buy_create.py` — it is no longer referenced directly.

### Harness update
`tests/harness/media_buy_create.py` — delegate `link_workflow_to_object` to the real `ContextManager` implementation so integration tests actually persist the `ObjectWorkflowMapping` FK row (matching production behaviour).

### Tests
Two new integration tests in `tests/integration/test_create_media_buy_behavioral.py`:
- `test_auto_approve_calls_link_workflow_to_object` — asserts the method is called with correct args and **before** `update_workflow_step(status='completed')`
- `test_manual_approval_calls_link_workflow_to_object` — same assertion for the manual-approval path

## Testing
- `make quality` passes (unit tests green; mypy errors are pre-existing on `upstream/main`)
- New integration tests cover both paths

### DIFF
--- a/docs/test-obligations/UC-002-create-media-buy.md
+++ b/docs/test-obligations/UC-002-create-media-buy.md
@@ -297,6 +297,16 @@ Source: UC-002-main-mcp.md
 **Business Rule:** BR-RULE-018 INV-1
 **Priority:** P0
 
+#### Scenario: Auto-Approve -- ObjectWorkflowMapping Persisted Before Push Notification
+**Obligation ID** UC-002-MAIN-22
+**Layer** behavioral
+**Given** auto-approval path
+**When** the system completes the media buy and calls `update_workflow_step(status="completed")`
+**Then** an `ObjectWorkflowMapping` row linking the workflow step to the media buy is persisted in the database BEFORE `update_workflow_step` is called
+**And** `link_workflow_to_object` is invoked with `object_type="media_buy"` and the correct `object_id`
+**Business Rule:** BR-RULE-020 (regression: issue #1378 — silent webhook drop on auto-approve path)
+**Priority:** P0
+
 ---
 
 ### 3.6 Upgrade: Boundary Field Propagation (salesagent-7gnv)
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -107,7 +107,7 @@ from src.core.auth import (
     resolve_principal_or_raise,
 )
 from src.core.context_manager import get_context_manager
-from src.core.database.models import AdapterConfig, CurrencyLimit, MediaBuy, ObjectWorkflowMapping
+from src.core.database.models import AdapterConfig, CurrencyLimit, MediaBuy
 from src.core.database.models import Creative as DBCreative
 from src.core.database.models import CreativeAssignment as DBAssignment
 from src.core.database.models import MediaPackage as DBMediaPackage
@@ -2815,15 +2815,14 @@ async def _create_media_buy_impl(
                 logger.info(f"✅ Created {len(pending_packages)} MediaPackage records")
 
             # Link the workflow step to the media buy so the approval button shows in UI
-            with MediaBuyUoW(tenant["tenant_id"]) as wf_uow:
-                # FIXME(salesagent-9f2): workflow mapping should use a repository method
-                assert wf_uow.session is not None
-                mapping = ObjectWorkflowMapping(
-                    object_type="media_buy", object_id=media_buy_id, step_id=step.step_id, action="create"
-                )
-                wf_uow.session.add(mapping)
-                # UoW auto-commits on clean exit
-                logger.info(f"✅ Linked workflow step {step.step_id} to media buy")
+            ctx_manager.link_workflow_to_object(
+                step_id=step.step_id,
+                object_type="media_buy",
+                object_id=media_buy_id,
+                action="create",
+                tenant_id=tenant["tenant_id"],
+            )
+            logger.info(f"✅ Linked workflow step {step.step_id} to media buy")
 
             # Create creative assignments for manual approval flow
             # This must happen AFTER media packages are created so we have package_ids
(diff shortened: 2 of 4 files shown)


ChrisHuie:
Thanks for this — the fix is correct. Traced it: on `main` the only `ObjectWorkflowMapping(...)` on this path was the manual one, so the auto-approve path never created the row and `_send_push_notifications` early-returned. Adding the link before `update_workflow_step` is right, carries no duplicate-mapping risk, and folding the manual path onto the same method is a real improvement.

Three things:

**1. Blocking CI (one line).** Quality Gate fails only on `ruff format --check` for `tests/integration/test_create_media_buy_behavioral.py:744`; Summary is red only as a rollup. Run:

    uv run ruff format tests/integration/test_create_media_buy_behavioral.py

That formats the long comprehension and turns both green. (`make quality` includes the format check, so it catches this locally.)

**2. Test quality.** Both new tests assert that `link_workflow_to_object` was called (on the mock). But the harness no-ops `update_workflow_step`, the only caller of `_send_push_notifications` — so the delivery path this bug lived in never runs, and the tests would stay green if the webhook dropped again. Please add a production-state assertion that the row was actually persisted (the harness already runs the real `link_workflow_to_object`, so it's in the DB):

    from src.core.database.repositories.workflow import WorkflowRepository
    repo = WorkflowRepository(session, tenant_id=tenant.tenant_id)
    mapping = repo.get_latest_mapping_for_object("media_buy", result.response.media_buy_id)
    assert mapping is not None

That's a repository read (no `session.add`/`get_db_session` needed) and survives refactoring of how the mapping gets created. Fuller version — asserting the webhook is actually delivered — needs the harness to forward `push_notification_config` and run the real `update_workflow_step`, plus a registered webhook config and a patch on `get_protocol_webhook_service`.

**3. `Covers:` tags.** `UC-002-MAIN-01` is the generic happy-path obligation (already covered by `test_happy_path_auto_approved`) and `UC-002-CC-ADAPTER-ATOMICITY-03` is persistence-before-adapter — neither describes webhook delivery, and there's no obligation for auto-approve push notification. Either add one to `docs/test-obligations/UC-002-create-media-buy.md` and tag the new test with it, or drop the two tags.

torbenbrodt:
Thanks for the review @ChrisHuie , I have addressed the quality and test issues.

ChrisHuie (approved):

PR #1463, opened 2026-06-17
test: wire list_creative_formats format_id object-shape scenario on all transports

### LINKED ISSUE #1409: test: list_creative_formats returns format_id as {agent_url, id} objects
Part of #1247 (AdCP 3.1 compliance). **Good first issue** — a schema-shape assertion, likely green-on-wire under the 3.1 SDK.

## The behavior (v3.1 requirements Test `T-UC-005-storyboard-baseline-format-id-object-shape`)

```gherkin
Given the Buyer Agent calls list_creative_formats without filters
When the response returns a non-empty formats array
Then every entry's format_id should be an object carrying both agent_url and id
And no entry's format_id should be a bare string
```

Every `format_id` must be an object with both `agent_url` (the creative agent's URL) and `id` — bare-string format ids break the v3.1 federation contract.

## Links
- **Requirements (web):** https://main.d1tyvbf1at6iao.amplifyapp.com/#T-UC-005-storyboard-baseline-format-id-object-shape
- **Feature file (PR #1384):** https://github.com/prebid/salesagent/blob/c59ebcfeaddae3ffd45b27014f0d06e7111c9130/tests/bdd/features/BR-UC-005-discover-creative-formats.feature#L1093
- **Source obligation:** adcp creative/index.yaml · discover_formats

## Where to implement
- `list_creative_formats` impl / response under `src/core/tools/`.
- Wiring + asserting the `{agent_url, id}` object shape on every entry.

## Definition of done
- [ ] Every `list_creative_formats` entry exposes `format_id` as an object with both `agent_url` and `id`; none is a bare string.

## Notes
- The `{agent_url, id}` object shape is the 3.1 contract — cleanest **after PR #1399** (`adcp==5.7.0`) lands so the SDK types match.
- Natural intro to the format-id issues (`roundtrip-from-products`, `third-party-agent`).

mkomorski:
## What

Wires the AdCP v3.1 baseline scenario `T-UC-005-storyboard-baseline-format-id-object-shape` to real production across all four transports (IMPL/A2A/MCP/REST), flipping it from auto-xfail (missing step definitions) to passing. Every `list_creative_formats` entry's `format_id` is asserted to be an object carrying both `agent_url` and `id` — never a bare string (the v3.1 federation contract).

Closes #1409. Unblocked by #1399 (adcp 5.7.0 / spec 3.1.0-beta.3), now on `main`.

## Approach — federation-grade

Per the implementer trace on the issue, the contract is only *falsifiable* against serialized wire bytes — the typed payload is a required structured type and can never be a bare string by construction. Previously only REST surfaced the success-path wire; this PR extends the harness so every real-wire transport does:

`TransportResult.wire_response: dict | None` — the serialized success-path body:
- **REST** — HTTP JSON body
- **MCP** — `ToolResult.structured_content`
- **A2A** — full artifact DataPart (captured *before* the `message`/`success` strip)
- **IMPL** — `None` (no wire; the step serializes the typed payload via `model_dump(mode="json")`, exercising the production serializer)

The scenario asserts the `{agent_url, id}` object shape on the actual serialized bytes for REST/A2A/MCP and on the serializer output for IMPL — meaningful on every transport, tautological on none. The `wire_response` extension is reusable for the sibling format-id scenarios (`roundtrip-from-products`, `third-party-agent`).

## On falsifiability (open-question Q3)

`tests/unit/test_uc005_format_id_shape_falsifiable.py` proves the assertion *bites* (rejects a bare-string `format_id`) at the helper level. Q3 suggested an end-to-end registry mutation; that's brittle — production can't construct a bare-string `format_id` without bypassing the typed model. Instead, end-to-end falsifiability holds **by construction** on REST/A2A/MCP (real bytes + `type(fid) is dict`), and `tests/integration/test_harness_wire_response.py` pins that `wire_response` is real wire (REST `== raw_response.json()`), not a payload reconstruction — closing the only otherwise-unproven link.

## Verification

- Scenario green on all 4 transports (`impl`/`a2a`/`mcp`/`rest`) — xfail→pass
- Full UC-005 BDD: **613 passed / 225 xfailed / 0 failed** (no harness regressions)
- `wire_response` authenticity test: 3 passed
- `make quality`: **4977 passed, 0 failed** — no new allowlist entries
- Two structural guards (duplicate-step, count-only assertion) caught real issues during development; both fixed properly, not allowlisted

## Files

- `tests/harness/{transport,_base,dispatchers}.py` — capture success-path wire
- `tests/bdd/steps/generic/when_request.py` — surface `ctx["wire_response"]`; baseline `Given` reuses the canonical unfiltered-dispatch step
- `tests/bdd/steps/domain/uc005_format_id_shape.py` — new When/Then steps
- `tests/helpers/format_assertions.py` — shared wire assertion
- `tests/unit/test_uc005_format_id_shape_falsifiable.py` — falsifiability
- `tests/integration/test_harness_wire_response.py` — wire authenticity
- `tests/CLAUDE.md` — document the `wire_response` field per transport

### DIFF
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -401,6 +401,37 @@ MCP, or A2A — only those transports observe actual wire bytes.
 compatibility. Reconstruction is lossy — assert on `wire_error_envelope`
 (or `synthesized_error_envelope` for IMPL).
 
+### TransportResult.wire_response (success-path wire)
+
+`TransportResult` also exposes `wire_response: dict | None` — the **serialized
+success-path response body**, the success-path analogue of `wire_error_envelope`.
+Populated on success by the REST dispatcher (HTTP body) and by the A2A/MCP
+dispatchers **only when the env routes through `_run_a2a_handler` /
+`_run_mcp_client`** (which stash the wire). Legacy `_run_mcp_wrapper` and the
+direct `*_raw` wrappers do not stash, so `wire_response` is `None` there — as it
+is on error and on IMPL. Today only `CreativeFormatsEnv` reads it.
+Use it to assert the **actual serialized shape** a buyer receives (e.g. the v3.1
+`format_id` `{agent_url, id}` federation contract on `list_creative_formats`)
+rather than the typed `payload`, whose fields are already coerced to their
+declared types and so cannot catch a serialization regression.
+
+**Authenticity per transport:**
+
+| Transport | `wire_response` source | Notes |
+|-----------|------------------------|-------|
+| REST | HTTP JSON body (`response.json()`) | Real wire; equals `raw_response.json()`. |
+| MCP  | `ToolResult.structured_content` (real wire) | Stashed by `_run_mcp_client`. |
+| A2A  | Full artifact DataPart (real wire) | Stashed by `_run_a2a_handler` BEFORE the `message`/`success` strip, so top-level envelope fields are present. |
+| IMPL | `None` (no wire by definition) | Serialize the typed `payload` (`model_dump(mode="json")`) — exercises the production serializer, not transport framing. |
+
+As with `wire_error_envelope`, real wire-shape regressions are only observable on
+REST/MCP/A2A; IMPL's `model_dump` only exercises the serializer. See
+`tests/integration/test_harness_wire_response.py` (pins that the field is real
+wire, not a payload reconstruction) and
+`tests/bdd/steps/domain/uc005_format_id_shape.py` (uses it for the format_id
+federation contract; reusable by the `roundtrip-from-products` /
+`third-party-agent` siblings).
+
 ## Infrastructure
 
 | What you need | Command |
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -43,6 +43,7 @@ pytest_plugins = [
     "tests.bdd.steps.domain.uc004_delivery",
     "tests.bdd.steps.domain.uc002_create_media_buy",
     "tests.bdd.steps.domain.uc006_sync_creatives",
+    "tests.bdd.steps.domain.uc005_format_id_shape",
     "tests.bdd.steps.domain.uc011_accounts",
     "tests.bdd.steps.domain.admin_accounts",
     "tests.bdd.steps.domain.uc_get_products_inventory",
--- a/tests/bdd/features/BR-UC-005-discover-creative-formats.feature
+++ b/tests/bdd/features/BR-UC-005-discover-creative-formats.feature
@@ -1101,3 +1101,6 @@ Feature: BR-UC-005 Discover Creative Formats
     # unique identifier within that agent). Sellers returning bare-string format IDs
     # break the v3.1 federation contract.
     # discover_formats: format_id object shape is the federation contract
+    # The strict "every entry / never a bare string" form is mandated by the schema
+    # (core/format-id.json: required [agent_url, id]); the storyboard grades field_present on formats[0].
+    # @source repo=adcp ref=v3.1-04f59d2d5 commit=04f59d2d5 path=static/schemas/source/core/format-id.json
--- a/tests/bdd/steps/domain/uc005_format_id_shape.py
+++ b/tests/bdd/steps/domain/uc005_format_id_shape.py
@@ -1,0 +1,74 @@
+"""UC-005 baseline: list_creative_formats format_id object shape.
+
+Scenario ``T-UC-005-storyboard-baseline-format-id-object-shape`` (@baseline-conformance):
+every ``format_id`` returned by ``list_creative_formats`` must be an object carrying
+both ``agent_url`` and ``id`` — never a bare string. The "every entry / never a bare
+string" strictness is the **schema** contract (``core/format-id.json``: required
+[agent_url, id]); the storyboard's discover_formats step only grades ``field_present``
+on ``formats[0]``, so this scenario is intentionally stricter than the graded step.
+
+Wired to real production across all 4 transports (auto-parametrized; UC-005 →
+CreativeFormatsEnv). REST/A2A/MCP assert the actual serialized wire dict surfaced
+by the harness (``ctx["wire_response"]``); IMPL has no wire, so it asserts the
+production serializer output (``model_dump(mode="json")`` — exercises
+``NestedModelSerializerMixin``). The wire/serializer path is the only place the
+contract is falsifiable: the typed payload is a required structured type and can
+never be a bare string by construction.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from pytest_bdd import then, when
+
+from tests.bdd.steps._outcome_helpers import _require_response
+from tests.harness.transport import Transport
+from tests.helpers.format_assertions import assert_wire_format_id_is_object
+
+
+def _serialized_formats(ctx: dict) -> list[dict[str, Any]]:
+    """Return the formats array as the buyer sees it on the serialized wire.
+
+    REST/A2A/MCP expose the real success-path wire dict via ``ctx["wire_response"]``.
+    IMPL has no wire, so serialize the typed payload through the production
+    serializer — the same path that produces wire bytes for the other transports.
+    """
+    wire = ctx.get("wire_response")
+    transport = ctx.get("transport")
+    # Loud guard: a real-wire transport (REST/A2A/MCP) that didn't stash
+    # wire_response would otherwise fall through to the model_dump path and
+    # assert nothing on the wire — the silent tautology this scenario removes.
+    # A future sibling wired against a non-stashing env (e.g. media_buy_list,
+    # creative_sync, media_buy_update A2A/MCP) trips this instead of passing
+    # green. IMPL (and the non-parametrized None default) legitimately have no wire.
+    if wire is None and transport not in (None, Transport.IMPL):
+        raise AssertionError(f"{transport}: wire_response missing — env does not stash success-path wire")
+    if wire is not None:
+        return wire["formats"]
+    # IMPL has no wire — serialize the typed payload through the production
+    # serializer. _require_response preserves the diagnostic if a (reused) sibling
+    # scenario hit an error path, instead of a bare ctx["response"] KeyError.
+    return _require_response(ctx).model_dump(mode="json")["formats"]
+
+
+@when("the response returns a non-empty formats array")
+def when_response_returns_non_empty_formats(ctx: dict) -> None:
+    # Precondition guard (phrased as a When by the storyboard): the Given's
+    # dispatch must have returned a non-empty formats array before shape checks.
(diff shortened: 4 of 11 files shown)


mkomorski:
I have read the IPR Policy

ChrisHuie:
Reviewed across BDD, harness/test-integrity, error-wire, spec-conformance, code-patterns, and the structural guards, and ran it locally: the scenario passes on all four transports (`impl`/`a2a`/`mcp`/`rest` — 4 passed, not xfailed), the integration wire-authenticity test passes (3), `make quality` is 0-failed, and the `{agent_url, id}` object shape is confirmed against the 3.1.0-beta.3 schema (`core/format-id.json`: `required: [agent_url, id]`, "never a plain string"). Nicely-reasoned PR. No blockers — four refinements:

1. **`wire_response` is silently `None` on 5 non-stashing success paths** (`media_buy_list` A2A+MCP, `creative_sync` A2A, `media_buy_update` A2A+MCP) — `_run_mcp_wrapper` and the direct `*_raw` wrappers never set `_last_wire_response`, so `tests/CLAUDE.md`'s "Populated by REST/A2A/MCP dispatchers on success" is broader than the code. Latent today (only `CreativeFormatsEnv` reads it), but `_serialized_formats` silently falls back to the IMPL `model_dump` path when `wire_response` is `None`, so a future sibling scenario on one of those envs would lose real-wire authenticity with a green test. Suggest declaring `_last_wire_response` in `__init__` (it's set in `call_via` today, which forces the defensive `getattr` in the dispatchers — reading it directly makes a missed capture loud), and narrowing the doc to the envs that actually route through `_run_a2a_handler`/`_run_mcp_client`.

2. **`test_harness_wire_response`'s A2A/MCP cases assert only `isinstance(dict)` + `"formats" in`**, which a `payload.model_dump()` reconstruction also satisfies — yet the docstring says the test stops a refactor from "quietly substitut[ing] a reconstruction." Captured the real wire: A2A carries `success`/`message` and MCP carries `task_id`/`adcp_version` that a reconstruction would lack, so asserting those makes the oracle bite. (REST `wire_response == payload.model_dump()`, so the `== raw_response.json()` check there pins provenance, not reconstruction-difference — worth saying so in the docstring.)

3. **The step/module docstrings frame the conformance storyboard as grading this**, but the storyboard (`creative/index.yaml` `discover_formats`/`list_formats`) only grades `field_present` on `formats[0].format_id.{agent_url,id}`. The strict "every entry, never a bare string" form is mandated by the schema (`core/format-id.json`), not the storyboard — suggest citing the schema and noting the test is intentionally stricter than the graded step.

4. **`_serialized_formats`' IMPL fallback uses a bare `ctx["response"]` subscript**, which dodges the `no_response_subscript` guard and raises a bare `KeyError` if a reused sibling scenario hits an error path. `tests/bdd/steps/_outcome_helpers._require_response` is the existing accessor that preserves the diagnostic.

Minor: `type(fid) is dict` vs the `isinstance` convention elsewhere; the When-step-that-asserts reads like a Then; the `@T-UC-005-storyboard-…` tag isn't a real spec step id and this scenario lacks the `@source` line its siblings carry.

mkomorski:
Thanks for the thorough review @ChrisHuie — all four refinements + minors addressed in `e16db6c9`. Rundown:

**1. `wire_response` silently `None` on non-stashing success paths.** Declared `_last_wire_response` on `BaseTestEnv.__init__`, and the A2A/MCP dispatchers now read `env._last_wire_response` directly (dropped the defensive `getattr`). Narrowed the `tests/CLAUDE.md` note: populated on success by the REST dispatcher and by A2A/MCP **only when the env routes through `_run_a2a_handler` / `_run_mcp_client`** — `_run_mcp_wrapper` and the direct `*_raw` wrappers don't stash, so `wire_response` is `None` there (as on error and IMPL).

**2. Authenticity test didn't distinguish a reconstruction.** The A2A/MCP cases now assert envelope-only fields a `payload.model_dump()` would lack — A2A `success`/`message`, MCP `task_id`/`adcp_version` (confirmed absent from the payload-dump key set). REST keeps `== raw_response.json()`, and its docstring now states that REST `wire_response == payload.model_dump()`, so that check pins **provenance** (it's the real HTTP body), not a reconstruction-difference.

**3. Storyboard vs schema framing.** The step/module + helper docstrings now attribute the strict "every entry / never a bare string" form to the **schema** (`core/format-id.json`: `required: [agent_url, id]`), and note it's intentionally stricter than the storyboard's graded `field_present` on `formats[0]`. Added the missing `@source` line (pointing at the schema) to the scenario.

**4. `_serialized_formats`' bare `ctx["response"]`.** Now uses `_require_response(ctx)` in both the IMPL fallback and the When step, so a reused sibling scenario on an error path surfaces the diagnostic instead of a bare `KeyError`.

**Minors:** `type(fid) is dict` → `isinstance(fid, dict)` (matches the convention; still trivial-assert-guard safe); added a comment flagging the When step as a precondition guard phrased as a When. The IMPL-weakest point is left as the acknowledged in-code trade-off.

Re-verified: scenario green on all four transports, the wire-authenticity integration test passes (3, now with the envelope-field oracle), `make quality` 0-failed (4977 passed), and the BDD structural guards are green. Thanks again for the careful pass!

ChrisHuie:
# Re-review — #1463 (verified at head `e16db6c9`)

Re-checked the revision and ran it locally: the four points raised earlier on this PR are addressed and the mechanism holds. `_last_wire_response` is now declared on `BaseTestEnv.__init__` (`tests/harness/_base.py:347`) and read directly by the dispatchers (`tests/harness/dispatchers.py:134,205`); the `tests/CLAUDE.md` claim is narrowed to the stashing envs; the integration test pins the A2A (`success`/`message`) and MCP (`task_id`/`adcp_version`) envelope markers; `_require_response` replaces the bare `ctx["response"]` subscript and was extended to the When step; and the schema-vs-storyboard attribution + `@source` line are correct (path resolves at `04f59d2d5`). BDD `[impl][a2a][mcp][rest]` → 4 passed (not xfailed), integration → 3 passed, `make quality` → 0 failed, CI → 32/32.

In strong shape — nothing blocks. One optional forward-looking elevation, one minor touch; the rest is already right.

## 1. (Optional) The documented stashing contract is unenforced
- **Where:** `tests/CLAUDE.md` (the narrowed claim) ↔ `tests/bdd/steps/domain/uc005_format_id_shape.py:36-42` (`_serialized_formats`).
- **Context:** `wire_response` is populated only when an env routes through `_run_a2a_handler` / `_run_mcp_client`. The five non-stashing cells (`media_buy_list` A2A+MCP, `creative_sync` A2A, `media_buy_update` A2A+MCP) leave it `None` on success — now correctly documented.
- **Problem:** the contract is documented but has no failing oracle. A future sibling scenario wired against a non-stashing env would fall through the `model_dump` branch of `_serialized_formats` and assert nothing on the wire, with a green test — the tautology this work removes. Latent today (only `CreativeFormatsEnv` consumes `wire_response`).
- **Fix (propose, not applied):** a loud guard at the consumer so a missed capture can't pass silently (add `from tests.harness.transport import Transport`):
```diff
     wire = ctx.get("wire_response")
+    transport = ctx.get("transport")
+    if wire is None and transport not in (None, Transport.IMPL):
+        raise AssertionError(f"{transport}: wire_response missing — env does not stash success-path wire")
     if wire is not None:
         return wire["formats"]
     return _require_response(ctx).model_dump(mode="json")["formats"]
```
  `ctx["transport"]` is the `Transport` enum (`conftest.py:2492`); the impl-default path leaves it `None`, so both are excluded. Equivalently, a small test asserting `CreativeFormatsEnv` populates `wire_response` on A2A/MCP while a non-stashing env returns `None` turns the documented matrix into an oracle.

## Minor
- **Where:** `tests/integration/test_harness_wire_response.py:34-46`. The test asserts A2A/MCP carry their envelope markers; also asserting REST's `wire_response` does **not** carry them (it's the bare HTTP body) closes the per-transport shape contract symmetrically.

## Informational — already right
- The "reusable by `roundtrip-from-products` / `third-party-agent`" references are accurate: both exist as tagged scenarios (`BR-UC-005-discover-creative-formats.feature:1059,1077`), currently unwired — substantial cross-tool scenarios, reasonably their own follow-ups.
- Falsifiability holds end-to-end without an explicit mutation test: the REST/A2A/MCP legs serialize through the production serializer, so a `format_id`-flattening regression reddens the scenario directly.

## Verification performed
- **Ran (head `e16db6c9`, asserted, clean tree):** `make quality` → all checks passed, 0 failed; UC-005 scenario → `[impl][a2a][mcp][rest]` 4 passed, 0 xfailed; `test_harness_wire_response.py` → 3 passed (incl. the new envelope-marker assertions); `test_uc005_format_id_shape_falsifiable.py` → 4 passed.
- **Observed:** CI rollup 32/32 SUCCESS; `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (required-review gate, not code).
- **Verified prior fixes against the diff** `9892d53c..e16db6c9` item-by-item; confirmed the `@source` path exists at `adcontextprotocol/adcp@04f59d2d5`.
- **Not run:** full e2e/admin locally (relied on CI); the five non-stashing paths' live A2A/MCP dispatch (latent, no consumer today — reasoned from the return statements, not executed).

mkomorski:
Thanks for the careful re-review @ChrisHuie — both follow-ups applied in `7a1b9347`:

**1. Stashing contract now enforced (the optional elevation).** Added a loud guard at the consumer in `_serialized_formats`:
```python
wire = ctx.get("wire_response")
transport = ctx.get("transport")
if wire is None and transport not in (None, Transport.IMPL):
    raise AssertionError(f"{transport}: wire_response missing — env does not stash success-path wire")
```
A future sibling wired against a non-stashing env (`media_buy_list`, `creative_sync`, `media_buy_update` A2A/MCP) now trips loudly instead of falling through to the `model_dump` branch and asserting nothing on the wire — the documented matrix is an oracle now, not just prose. IMPL and the non-parametrized `None` default are excluded.

**2. REST symmetry (the minor).** `test_rest_wire_response_is_the_http_body` now also asserts the bare HTTP body does **not** carry the A2A/MCP envelope markers (`success`/`message`/`task_id`/`adcp_version`), closing the per-transport shape contract from both directions. The markers are DRY'd into a shared `ENVELOPE_MARKERS` class constant used by both tests.

Left as-is per your notes: IMPL-weakest (acknowledged trade-off), the sibling scenarios as their own follow-ups, and no explicit mutation test (falsifiability holds end-to-end through the production serializer).

Re-verified at `7a1b9347`: UC-005 scenario `[impl][a2a][mcp][rest]` → 4 passed (loud guard correctly silent for `CreativeFormatsEnv`); `test_harness_wire_response.py` → 3 passed (incl. the REST-absence assertion); `make quality` → 0 failed (4977 passed). Thanks again!

ChrisHuie (approved):

PR #1464, opened 2026-06-18
fix: serialize push_notification_config with mode='json' at MCP/A2A transport boundaries

### LINKED ISSUE #1377: Bug: create_media_buy crashes with StatementError when push_notification_config.url is a Pydantic AnyUrl object
# Bug: `create_media_buy` crashes with `StatementError` when `push_notification_config.url` is a Pydantic `AnyUrl` object

**Component:** `src/core/tools/media_buy_create.py` — MCP wrapper and `create_media_buy_raw` A2A wrapper
**Severity:** High — push notification config is never persisted; buyer's webhook is silently never called

---

## Summary

The MCP wrapper and the `create_media_buy_raw` A2A wrapper both serialise the
`PushNotificationConfig` Pydantic model to a plain `dict` using `.model_dump()` before
passing it to `_create_media_buy_impl`. Pydantic v2's default `model_dump()` (Python mode)
preserves `AnyUrl` fields as typed `AnyUrl` objects and enum fields as enum instances rather
than converting them to `str`. SQLAlchemy's `String` column type does not coerce these objects
and raises a `StatementError` at flush time.

Note: the protobuf-based A2A path (`adcp_a2a_server.py`) uses `json_format.MessageToDict()`
which always produces plain strings and is not affected.

## Root cause

Pydantic v2 has two serialisation modes:

```python
pnc = PushNotificationConfig(url='https://example.com/webhook')

pnc.model_dump()              # {'url': AnyUrl('https://example.com/webhook')}  ← AnyUrl object
pnc.model_dump(mode='json')   # {'url': 'https://example.com/webhook'}          ← plain str
```

The same applies to `authentication.schemes`: if the caller passes
`schemes=[AuthenticationScheme.Bearer]` (a Pydantic enum), `model_dump()` preserves the enum
instance while `model_dump(mode='json')` serialises it to its string value `"Bearer"`.

Both affected call sites use plain `.model_dump()`:

```python
# MCP wrapper (create_media_buy tool function)
pnc_dict = push_notification_config.model_dump() if push_notification_config else None

# A2A wrapper (create_media_buy_raw)
pnc_dict = (
    push_notification_config.model_dump()
    if isinstance(push_notification_config, PushNotificationConfig)
    else push_notification_config
)
```

The resulting `pnc_dict` contains `AnyUrl` / enum objects. When `_create_media_buy_impl`
later extracts `url = pnc_dict.get("url")` and passes it to
`DBPushNotificationConfig(url=url, ...)`, SQLAlchemy rejects the non-string value.

## Symptom

```
sqlalchemy.exc.StatementError: (builtins.TypeError)
  String column received <class 'pydantic.networks.AnyUrl'> value
```

The exception is raised inside the push notification config registration block. The media buy
itself may still be created, but the webhook URL is never stored, so no push notifications
are ever delivered.

## Reproduction

1. Register a principal with a push notification webhook URL.
2. Call `create_media_buy` via MCP or A2A with a `push_notification_config`.
3. Observe `StatementError` at the `session.flush()` call inside the push notification config
   registration block.

## Proposed fix

Use `mode='json'` at the transport boundary where the dict is created, so `AnyUrl` fields are
serialised to plain strings before they reach `_impl`:

```python
# MCP wrapper
pnc_dict = push_notification_config.model_dump(mode='json') if push_notification_config else None

# A2A wrapper
pnc_dict = (
    push_notification_config.model_dump(mode='json')
    if isinstance(push_notification_config, PushNotificationConfig)
    else push_notification_config
)
```

This is the correct fix at the transport boundary. It is consistent with the architecture
rule that `_impl` functions receive plain Python types, not transport-layer objects.

An alternative workaround — adding `str()` coercions inside `_create_media_buy_impl` — also
prevents the crash but treats the symptom inside business logic rather than fixing the
serialisation at the boundary where it belongs.

torbenbrodt:
## Summary

Fixes #1377

Pydantic v2's default `model_dump()` preserves `AnyUrl` fields as typed `AnyUrl` objects and enum fields as enum instances rather than converting them to `str`. SQLAlchemy's `String` column type does not coerce these objects and raises a `StatementError` at flush time when `push_notification_config.url` is persisted.

## Root Cause

```python
pnc = PushNotificationConfig(url='https://example.com/webhook')
pnc.model_dump()              # {'url': AnyUrl('https://example.com/webhook')}  ← AnyUrl object
pnc.model_dump(mode='json')   # {'url': 'https://example.com/webhook'}          ← plain str ✓
```

## Fix

Changed `.model_dump()` → `.model_dump(mode="json")` at both transport boundaries in `src/core/tools/media_buy_create.py`:

1. **MCP wrapper** (`create_media_buy` tool)
2. **A2A wrapper** (`create_media_buy_raw`)

The protobuf-based A2A path (`adcp_a2a_server.py`) uses `json_format.MessageToDict()` which always produces plain strings and is not affected.

## Verification

The regression tests were verified to:
- **FAIL** with the bug (plain `model_dump()`): `AssertionError: url must be a plain str after model_dump(mode='json'), got 'AnyUrl'`
- **PASS** with the fix (`model_dump(mode="json")`): all 5 tests green

## Spec Citation

This fix is at the transport boundary layer (Critical Pattern #5 in AGENTS.md). No AdCP protocol behavior is changed — this is a serialization correctness fix ensuring `_impl` receives plain Python types as the architecture requires.

## Test Obligations

New behavioral obligations added to `docs/test-obligations/UC-002-create-media-buy.md`:
- `UC-002-TRANSPORT-PNC-SERIALIZATION-01` — MCP wrapper serializes AnyUrl to plain str
- `UC-002-TRANSPORT-PNC-SERIALIZATION-02` — A2A wrapper serializes AnyUrl to plain str

New test file: `tests/unit/test_create_media_buy_behavioral.py` with 5 tests covering both obligations.

### DIFF
--- a/docs/test-obligations/UC-002-create-media-buy.md
+++ b/docs/test-obligations/UC-002-create-media-buy.md
@@ -1524,6 +1524,30 @@ Source: BR-RULE-026
 
 ---
 
+### Transport Boundary: push_notification_config Serialization
+
+#### Scenario: MCP Wrapper Serializes AnyUrl to Plain String
+**Obligation ID** UC-002-TRANSPORT-PNC-SERIALIZATION-01
+**Layer** behavioral
+**Given** a `create_media_buy` MCP request with `push_notification_config.url` set to a Pydantic `AnyUrl` value
+**When** the MCP wrapper serializes the `PushNotificationConfig` model to a dict before calling `_impl`
+**Then** the `url` field in the resulting dict is a plain `str`, not a Pydantic `AnyUrl` object
+**And** enum fields such as `authentication.schemes` are serialized to their string values
+**And** `_impl` receives a dict whose values are all plain Python types compatible with SQLAlchemy `String` columns
+**Priority:** P0 (regression: gh-#1377)
+
+#### Scenario: A2A Wrapper Serializes AnyUrl to Plain String
+**Obligation ID** UC-002-TRANSPORT-PNC-SERIALIZATION-02
+**Layer** behavioral
+**Given** a `create_media_buy` A2A request with `push_notification_config` as a `PushNotificationConfig` model instance
+**When** the A2A wrapper (`create_media_buy_raw`) serializes the model to a dict before calling `_impl`
+**Then** the `url` field in the resulting dict is a plain `str`, not a Pydantic `AnyUrl` object
+**And** enum fields such as `authentication.schemes` are serialized to their string values
+**And** `_impl` receives a dict whose values are all plain Python types compatible with SQLAlchemy `String` columns
+**Priority:** P0 (regression: gh-#1377)
+
+---
+
 ## Summary Statistics
 
 | Category | Count |
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -4293,7 +4293,7 @@ async def create_media_buy(
     # Serialize PushNotificationConfig model to dict for _impl (which accepts dict|None).
     # Use mode='json' so Pydantic v2 converts AnyUrl fields to plain str and enum fields
     # to their string values — plain model_dump() preserves typed objects that SQLAlchemy
-    # String columns cannot coerce, causing StatementError at flush time (gh-#1377).
+    # String columns cannot coerce, causing StatementError at flush time.
     pnc_dict = push_notification_config.model_dump(mode="json") if push_notification_config else None
     result = await _create_media_buy_impl(
         req=req,
@@ -4378,7 +4378,7 @@ async def create_media_buy_raw(
     # Serialize SDK model to dict for _impl (which uses dict-based config access).
     # Use mode='json' so Pydantic v2 converts AnyUrl fields to plain str and enum fields
     # to their string values — plain model_dump() preserves typed objects that SQLAlchemy
-    # String columns cannot coerce, causing StatementError at flush time (gh-#1377).
+    # String columns cannot coerce, causing StatementError at flush time.
     pnc_dict = (
         push_notification_config.model_dump(mode="json")
         if isinstance(push_notification_config, PushNotificationConfig)
--- a/src/core/tools/media_buy_create.py
+++ b/src/core/tools/media_buy_create.py
@@ -4290,8 +4290,11 @@ async def create_media_buy(
 
     identity = enrich_identity_with_account(identity, req.account)
 
-    # Serialize PushNotificationConfig model to dict for _impl (which accepts dict|None)
-    pnc_dict = push_notification_config.model_dump() if push_notification_config else None
+    # Serialize PushNotificationConfig model to dict for _impl (which accepts dict|None).
+    # Use mode='json' so Pydantic v2 converts AnyUrl fields to plain str and enum fields
+    # to their string values — plain model_dump() preserves typed objects that SQLAlchemy
+    # String columns cannot coerce, causing StatementError at flush time (gh-#1377).
+    pnc_dict = push_notification_config.model_dump(mode="json") if push_notification_config else None
     result = await _create_media_buy_impl(
         req=req,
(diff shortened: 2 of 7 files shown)


ChrisHuie:
Thanks for this, @torbenbrodt — the core fix is correct, so this is just about getting CI green.

**The fix is right.** `model_dump(mode="json")` is the correct call at the create wrappers and is spec-aligned (AdCP 3.1.0-beta.3: `url` is `string`/`format: uri`, `schemes` a string enum). Verified against `adcp==5.7.0` that plain `model_dump()` yields `AnyUrl`/enum while `mode="json"` yields plain `str`. It's genuinely create-only — `update_media_buy` carries the config inside its request model, already serialized by `context_manager.create_workflow_step`.

**Why CI is red — two failures, one cause.** The new `tests/unit/test_create_media_buy_behavioral.py` repeats the scaffolding of the existing `tests/unit/test_push_notification_forwarding.py` (same mock-context boilerplate, same two wrappers):
- Quality Gate → DRY duplication hook: `tests/ 90 → 93 (+3 blocks)`, all between the two files.
- Unit Tests → weak-mock-assertions guard: the 5 tests use `assert_called_once()` + `call_args` (allowlisted for the existing file, not a new one).

**The fix — share the scaffolding via a helper.** Per the repo's DRY rule, extract the shared setup into a helper both files use; that resolves the duplication at its source and lets the tests assert on a captured value (no `call_args` split):

1. Add `tests/helpers/create_media_buy_capture.py` with `capture_mcp_forwarded_pnc(pnc)` / `capture_a2a_forwarded_pnc(pnc)` (build the mock context, patch `_create_media_buy_impl` with a `side_effect` that records its kwargs, return the forwarded `push_notification_config`).
2. Point both `test_create_media_buy_behavioral.py` (all 5 tests + the `Covers:` tags stay) and `test_push_notification_forwarding.py` at the helper and assert on the returned dict, e.g. `assert isinstance(forwarded["url"], str)`.
3. Drop the two `test_push_notification_forwarding.py` entries from `WEAK_ASSERTION_ALLOWLIST` (they no longer use the split pattern) and the `test_create_media_buy_behavioral.py` entry from `BEHAVIORAL_MOCK_CONSTRUCTION_CAP` (the file no longer constructs mocks directly).

I ran this end-to-end: all 5 of your tests still pass, the forwarding tests pass, and the duplication / weak-mock / mock-cap / obligation-coverage guards are green (`tests/` back to baseline 90). Net it's ~155 fewer lines and retires two pre-existing allowlist entries. Happy to hand you the exact patch if useful.

**Suggestion (additive): a real-DB oracle.** The unit tests patch `_create_media_buy_impl`, so they prove the wrapper serializes correctly but never hit the `DBPushNotificationConfig` flush where the original crash fired. One integration test via `MediaBuyCreateEnv` (`tests/harness/media_buy_create.py`, real DB) driving `create_media_buy` with a real `PushNotificationConfig` and asserting the row persists with `url` as a string would be the test that fails if this regresses at the DB layer.

**Coordination.** The same `model_dump(mode="json")` change also lands in #1465, so worth coordinating which PR carries the hunk so the other rebases it out.

Minor: the `(gh-#1377)` suffix in the production-code comments can be dropped (git history covers traceability) — the explanatory text is good to keep.

---
Verified at `7ddf7bd1d` + the change above:
- Ran: the 7 wrapper tests (pass); the DRY duplication hook (`tests/` 90, baseline); the weak-mock, mock-cap, and obligation-coverage guards (pass); the `model_dump()` vs `mode="json"` type behavior against `adcp==5.7.0`.
- Not run: a live DB flush reproducing the original `StatementError`; the full suite.

torbenbrodt:
Thanks @ChrisHuie  - the CI tests are green now :)

ChrisHuie (approved):

PR #1465, opened 2026-06-18
test(e2e): repair cascading shape bugs in reference lifecycle (#1358)

### LINKED ISSUE #1358: test(e2e): fix cascading shape bugs in test_adcp_reference_implementation lifecycle test
`tests/e2e/test_adcp_reference_implementation.py::TestAdCPReferenceImplementation::test_complete_campaign_lifecycle_with_webhooks` is currently `@pytest.mark.xfail(strict=False)` (PR #1306, commit 
  `3f047b4bc`). The test contains pre-existing shape bugs in its request data that were masked when the broad boundary catch in `src/core/tools/media_buy_create.py` swallowed typed errors. With typed 
  `AdCPError` now propagating past `_impl` (PR #1306's architectural direction), the test progresses through real validation and hits successive bugs that need their own fixes.

  The test is named the "REFERENCE E2E TEST" and its docstring says "Use this as a template for all future E2E tests" — fixing it correctly restores its role as the canonical example.

  ## Known bugs (in order encountered)

  1. **Phase 1 (fixed in #1306)** — `pricing_option_id` was hardcoded to `"default"`; no production handler recognized that value. Fixed by extracting
  `product["pricing_options"][0]["pricing_option_id"]` from the `get_products` response.

  2. **Phase 3 (fixed in #1306)** — `format_id` was sent as a bare string (`"display_300x250"`); the schema requires `FormatReferenceStructuredObject` (`{agent_url, id}`). Fixed by extracting
  structured `format_ids` from the product response.

  3. **Phase 5 (open)** — `push_notification_config.authentication` is sent as `{"type": "none"}`. Schema requires `{schemes, credentials}` and explicitly forbids the `type` field. Source: helper
  builder default.

  4. **Phase 6+ (unverified)** — additional phases (reporting verification, webhook delivery) may surface more shape mismatches once Phase 5 is fixed. Cannot confirm until earlier phases pass.

  ## Contract

  A. Walk the test phase-by-phase, fix each shape bug at the call site by referencing the actual AdCP schema for that endpoint (not by guessing field names).

  B. Mirror the discovery pattern used in `tests/e2e/test_a2a_webhook_payload_types.py` for any inputs that depend on server state (pricing options, format ids, etc.).

  C. When the test passes end-to-end with real wire validation, remove the `@pytest.mark.xfail` marker on the test function.

  ## Acceptance criteria

  - [ ] `./run_all_tests.sh ci tests/e2e/test_adcp_reference_implementation.py -k test_complete_campaign_lifecycle_with_webhooks` passes without xfail
  - [ ] No test data hardcoded against assumed server state — all dynamic values discovered from prior phase responses
  - [ ] xfail marker removed from `test_complete_campaign_lifecycle_with_webhooks`

  ## References

  - Test file: `tests/e2e/test_adcp_reference_implementation.py:92` (function), `:91-107` (xfail marker)
  - Helper: `tests/e2e/adcp_request_builder.py` (build_creative, build_adcp_media_buy_request, etc.)
  - Discovery precedent: `tests/e2e/test_a2a_webhook_payload_types.py:29-52` (`_discover_product_and_pricing`)
  - Origin of cascade: PR #1306 (boundary catch revert in `src/core/tools/media_buy_create.py:1982`)

mkomorski:
## What & why

Fixes #1358. The reference E2E lifecycle test (`test_complete_campaign_lifecycle_with_webhooks`) was `xfail`'d because narrowing the error catch exposed a cascade of request-shape bugs the broad `except` had masked. Running the lifecycle honestly against the Docker stack surfaced and fixed each one. Scope is **test-only**: the `push_notification_config` URL serialization fix this originally carried landed independently in #1464, so it was dropped here on rebase.

## Changes (test-only)

- **Creative formats**: discover from `list_creative_formats` (what the creative agent validates against) instead of the product's bare `format_ids`. The product advertises `display_300x250` but the agent only knows `display_300x250_image`, so `sync_creatives` rejected both creatives (`action="failed"`, never persisted) and `list_creatives` came back empty.
- **Silent-pass hatches removed**: the Phase 2 early `return` on a missing `media_buy_id` and the Phase 6 "webhook optional" branch. Added an `action != "failed"` assertion so a sync rejection fails at sync, not as a downstream empty list.
- **Webhook receiver**: now uses the shared `tests/e2e/_webhook_capture.run_webhook_capture_server` helper; the dedicated delivery test uses the default callback host so the server can reach it.
- Removed the broad `xfail` marker on the lifecycle test.

## Webhook delivery (documented, not silently swallowed)

`update_media_buy`'s task-status webhook is never delivered: `_send_push_notifications` (`src/core/context_manager.py`) queries `object_workflow_mapping` in a separate session before the update's UnitOfWork commits that mapping → "No object mappings found" → never POSTs. The create/auto-approve path was fixed in #1461 (mapping linked before the notification fires); `update_media_buy` still needs the same treatment. Guarded by `test_update_media_buy_push_webhook_delivery` (`xfail(strict=True)`), which forces closure when the update path is fixed.

## Verification

- `test_complete_campaign_lifecycle_with_webhooks` → passes against the Docker stack
- `test_update_media_buy_push_webhook_delivery` → xfails (strict), documenting the delivery bug
- Full CI green on this head (unit, integration ×5, BDD, E2E, admin, CodeQL).

## Spec grounding

The `push_notification_config` `{url, authentication?}` shape and the `format_id` `{agent_url, id}` structured reference follow AdCP 3.1; no protocol-behavior contract changed.

### DIFF
--- a/tests/e2e/test_adcp_reference_implementation.py
+++ b/tests/e2e/test_adcp_reference_implementation.py
@@ -1,15 +1,7 @@
-"""
-AdCP V2.3 Reference E2E Test
+"""AdCP V2.3 reference E2E test — the template for new E2E tests.
 
-This is the REFERENCE implementation for all future E2E tests.
-It demonstrates:
-1. Proper use of NEW AdCP V2.3 format
-2. Full campaign lifecycle (discovery → creation → delivery → reporting)
-3. Mix of synchronous and asynchronous (webhook) responses
-4. Proper schema validation using helper utilities
-5. Creative workflow integration
-
-Use this as a template when adding new E2E tests.
+Full campaign lifecycle (discovery → create → creatives → delivery → update → verify) over
+sync and async (webhook) responses, using the request-builder helpers for schema correctness.
 """
 
 import json
@@ -31,59 +23,38 @@ from tests.e2e.adcp_request_builder import (
 
 
 class WebhookReceiver(WebhookCaptureHandler):
-    """Simple webhook receiver for testing async notifications."""
+    """Webhook receiver for async notification tests."""
 
     received_webhooks: list = []
 
 
 @pytest.fixture
 def webhook_server():
-    """Start a local webhook server for testing async notifications."""
-    # Loopback-only receiver: this reference test reaches the server over the
-    # host loopback, so pin the callback host rather than honoring ADCP_WEBHOOK_HOST.
+    # Loopback-pinned: this lifecycle reaches the server over the host loopback, so the
+    # callback host stays 127.0.0.1 rather than honoring ADCP_WEBHOOK_HOST.
     with run_webhook_capture_server(WebhookReceiver, WebhookReceiver.received_webhooks, host="127.0.0.1") as info:
         yield info
 
 
 class TestAdCPReferenceImplementation:
-    """Reference E2E test demonstrating full AdCP V2.3 workflow."""
+    """Reference E2E test demonstrating the full AdCP V2.3 workflow."""
 
     @pytest.mark.asyncio
     async def test_complete_campaign_lifecycle_with_webhooks(
         self, docker_services_e2e, live_server, test_auth_token, webhook_server
     ):
-        """
-        REFERENCE TEST: Complete campaign lifecycle with sync + async (webhook) responses.
-
-        This test demonstrates the CORRECT way to write E2E tests using AdCP V2.3 format.
-
-        Flow:
-        1. Discovery: Get products and formats
-        2. Create: Create media buy with webhook for async updates
-        3. Creatives: Sync creatives (sync response)
-        4. Delivery: Get delivery metrics (sync response)
-        5. Update: Update campaign budget (webhook notification)
-        6. Reporting: Verify webhook received update notification
-
-        Use this as a template for all future E2E tests!
-        """
+        """Discovery → create → creatives → delivery → update → verify, over sync + webhook responses."""
         print("\n" + "=" * 80)
         print("REFERENCE E2E TEST: Complete Campaign Lifecycle")
         print("=" * 80)
 
-        # Setup MCP client with both auth and tenant detection headers
-        # Note: Host header is automatically set by HTTP client based on URL,
-        # so we use x-adcp-tenant header for explicit tenant selection in E2E tests
         headers = {
             "x-adcp-auth": test_auth_token,
-            "x-adcp-tenant": "ci-test",  # Explicit tenant selection for E2E tests
+            "x-adcp-tenant": "ci-test",
         }
         transport = StreamableHttpTransport(url=f"{live_server['mcp']}/mcp/", headers=headers)
 
         async with Client(transport=transport) as client:
-            # ================================================================
-            # PHASE 1: Product Discovery (Synchronous)
-            # ================================================================
             print("\n📦 PHASE 1: Product Discovery")
 
             products_result = await client.call_tool(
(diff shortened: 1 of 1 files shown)


ChrisHuie:
**CI green and verified at `fca2e33e1` — handing to review.** This PR is now test-only: the reference E2E lifecycle (`test_complete_campaign_lifecycle_with_webhooks`) is un-`xfail`'d and passing. The `push_notification_config` serialization fix it originally carried landed independently in #1464, so it was dropped on rebase — the diff is one file.

**What to review** — `tests/e2e/test_adcp_reference_implementation.py`:
- Creative formats are discovered from `list_creative_formats` (the source the creative agent validates against) instead of the product's bare `format_ids`, fixing the `action="failed"` → empty-`list_creatives` cascade.
- Silent-pass hatches removed; an `action != "failed"` assertion lands a sync rejection at sync, not as a downstream empty list.
- Shared `_webhook_capture.run_webhook_capture_server` helper; the delivery test uses the default callback host so the server reaches it in host-mode.

**Known limitation — documented, not skipped.** `update_media_buy`'s task-status webhook isn't delivered: `_send_push_notifications` queries `object_workflow_mapping` in a separate session before the update's UnitOfWork commits that mapping → "No object mappings found" → never POSTs. The create/auto-approve path was fixed in #1461 (mapping linked before the notification fires); `update_media_buy` needs the same treatment. Guarded by `test_update_media_buy_push_webhook_delivery` (`xfail(strict=True)`), which fails the suite the moment the bug is fixed — forcing the marker's removal rather than letting it rot.

---
**Verified at `fca2e33e1`:**
- Ran — full CI: 31/31 checks pass (unit, type, integration ×5, BDD ×2, E2E, admin UI, CodeQL, security audit, schema contract, migration roundtrip, title-lint). Locally: the e2e file on a fresh Docker stack rebased onto current main → `1 passed, 1 xfailed`, repeated 5× (0 xpass — the strict xfail is stable).
- Not run — a fix for the update-path delivery bug (out of scope here; the strict xfail is its tracker).

Earlier review points are all addressed.

mkomorski:
Thanks for the thorough review and the rebased 5× run, @ChrisHuie — all addressed in the force-pushed rebase (now onto `main`, so CI runs).

**Must-dos**
- **Rebased** onto main (#1420 / #1461). The `media_buy_create.py` change applied clean; the test file was re-applied onto main's shared-helper version.
- **Shared helper + host**: dropped the inline fixture. The new delivery test uses `run_webhook_capture_server`'s default (reachable) host, **not** `127.0.0.1`, so it can flip to xpass once the bug is fixed. (Left the lifecycle's `webhook_server` on `127.0.0.1` since delivery is de-scoped there.)
- **xfail reason** narrowed to `update_media_buy` only — #1461 fixed the create path, so the reason now cites the `link_workflow_to_object()`-before-`update_workflow_step()` pattern the update path still needs. Dropped the trailing issue number (nit a).
- **`media_buy_id` comment**: adopted your AdCP 3.1 `oneOf` wording (synchronous-path framing).
- **`schemes` Enum**: documented in the fix commit + both wrapper comments — `mode="json"` repairs that crash (the legacy HMAC/Bearer path) too.

**Nit d (create-webhook race)**: turned out to be necessary after rebasing onto #1461 — the create webhook now delivers, so racing a bare `clear()` against it could false-xpass. I kept the create-time config registration (needed so the test flips to xpass once update is fixed) and added a deterministic drain: wait for the create webhook to land → clear → isolate the update. Verified 3× with `--runxfail` — all fail at the *update* assertion (drain succeeds, update never delivers).

**Deferred** (kept this PR scoped to the rebase + your must-dos): nit b (offline unit test pinning `mode="json"`) and nit c (shared serialization helper across the two wrappers). Happy to fold them in here or as fast-follows — your call.

Verified on the rebased tree: lifecycle passes, `test_update_media_buy_push_webhook_delivery` xfails (strict, stable); `make quality` → 5029 passed, 8 skipped, 26 xfailed.

ChrisHuie (approved):

PR #1473, opened 2026-06-22
feat: add CPA and time-based pricing model support (AdCP 3.1)

torbenbrodt:
Implements two new AdCP 3.1 pricing models in convert_pricing_option_to_adcp():

- CPA (cost-per-action): fixed_price required; event_type defaults to EventType.purchase when not specified; unknown event_type values fall back to purchase with a warning log
- time (time-based): parameters.time_unit required (hour/day/week/month); supports both fixed and auction variants; optional min_duration and max_duration constraints via TimeParameters

Adds 21 unit tests covering both models:
- TestCpaConversion (8 tests): default/explicit event types, unknown fallback, missing rate, currency, min_spend
- TestTimeBasedConversion (13 tests): all four time units, duration constraints, auction variants, all error paths

Satisfies CONSTR-PRICING-OPTION-01 (all 9 AdCP 3.1 pricing models now supported: cpm, vcpm, cpc, cpcv, cpv, cpp, flat_rate, cpa, time).

### DIFF
--- a/src/core/product_conversion.py
+++ b/src/core/product_conversion.py
@@ -257,27 +257,33 @@ def convert_pricing_option_to_adcp(
 
     elif pricing_model == "cpa":
         # CPA (Cost Per Acquisition) - AdCP v3.1 pricing model for affiliate/conversion pricing.
+        # CPA always emits fixed_price; use a stable _fixed suffix regardless of is_fixed flag.
         # event_type is required per cpa-option.json; no default — an unknown value is a mispricing.
+        cpa_pricing_option_id = f"cpa_{currency.lower()}_fixed"
         if not rate:
-            raise ValueError(f"CPA pricing option {pricing_option_id} requires rate")
+            raise ValueError(f"CPA pricing option {cpa_pricing_option_id} requires rate")
         raw_event = parameters.get("event_type") if isinstance(parameters, dict) else None
         if not raw_event:
-            raise ValueError(f"CPA pricing option {pricing_option_id} requires parameters.event_type")
+            raise ValueError(f"CPA pricing option {cpa_pricing_option_id} requires parameters.event_type")
         try:
             event_type_val = EventType(raw_event)
         except ValueError:
             raise ValueError(
-                f"CPA pricing option {pricing_option_id} has unknown event_type '{raw_event}'. "
+                f"CPA pricing option {cpa_pricing_option_id} has unknown event_type '{raw_event}'. "
                 f"Supported values: {[e.value for e in EventType]}"
             )
-        custom_event_name = parameters.get("custom_event_name") if isinstance(parameters, dict) else None
-        if event_type_val == EventType.custom and not custom_event_name:
-            raise ValueError(
-                f"CPA pricing option {pricing_option_id} with event_type 'custom' requires parameters.custom_event_name"
-            )
+        if event_type_val == EventType.custom:
+            custom_event_name = parameters.get("custom_event_name") if isinstance(parameters, dict) else None
+            if not custom_event_name:
+                raise ValueError(
+                    f"CPA pricing option {cpa_pricing_option_id} with event_type 'custom' requires parameters.custom_event_name"
+                )
+        else:
+            custom_event_name = None
         event_source_id = parameters.get("event_source_id") if isinstance(parameters, dict) else None
+        cpa_fields = {**common_fields, "pricing_option_id": cpa_pricing_option_id}
         return CpaPricingOption(
-            **common_fields,
+            **cpa_fields,
             event_type=event_type_val,
             custom_event_name=custom_event_name,
             event_source_id=event_source_id,
@@ -306,6 +312,15 @@ def convert_pricing_option_to_adcp(
             min_duration=parameters.get("min_duration"),
             max_duration=parameters.get("max_duration"),
         )
+        if (
+            time_params.min_duration is not None
+            and time_params.max_duration is not None
+            and time_params.max_duration < time_params.min_duration
+        ):
+            raise ValueError(
+                f"Time-based pricing option {pricing_option_id} has max_duration "
+                f"({time_params.max_duration}) < min_duration ({time_params.min_duration})"
+            )
         time_fields: dict = {**common_fields, "parameters": time_params}
         if is_fixed:
             if rate is None:
--- a/src/core/product_conversion.py
+++ b/src/core/product_conversion.py
@@ -20,13 +20,13 @@ from adcp import (
     CpmPricingOption,
     CppPricingOption,
     CpvPricingOption,
+    EventType,
     FlatRatePricingOption,
     TimeBasedPricingOption,
     TimeUnit,
     VcpmPricingOption,
 )
 from adcp.types._generated import MediaChannel
-from adcp.types.generated_poc.enums.event_type import EventType
 from adcp.types.generated_poc.pricing_options.time_option import Parameters as TimeParameters
 from packaging.version import InvalidVersion, Version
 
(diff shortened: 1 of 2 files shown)


ChrisHuie:
# Review — #1473 (verified at head `86405925b`)

Adds `cpa` and `time` branches to `convert_pricing_option_to_adcp()` (`src/core/product_conversion.py:258-316`), the DB-`PricingOption` → AdCP-library-type serializer on the `get_products` path. Checked against the pinned spec **3.1.0-beta.3** (`dist/schemas/3.1.0-beta.3/`): both models serialize correctly, and `test_time_auction_without_floor_price` is spec-valid — `time-option.json` requires only `[pricing_option_id, pricing_model, currency, parameters]` and defines no `fixed_price`/`floor_price` XOR, so a no-floor open auction is permitted. The mutual exclusivity the schema does *not* enforce is correctly held by the `is_fixed` branch (`:305`), which never emits both prices. 48 tests pass at head (21 new).

No blockers. Substantive items are all in the CPA `event_type` path, plus one `time` zero-rate divergence; the rest are nits/scope.

**1 — CPA coerces an unknown `event_type` to `purchase`; `time` raises on an unknown `time_unit`.** `:266-275` vs `:294-298`. `event_type` is the billing trigger. An unrecognized value (typo, or a newer-spec event) is silently mapped to `purchase` (`:271-275`) with a warning, while the parallel `time_unit` path raises. Mapping one billing event to another is a silent mispricing. Separately, a *missing* `event_type` defaults to `purchase` (`:264`), but `cpa-option.json` marks `event_type` required with no default and `cpp` raises on its missing required params (`:235`). The unknown-value coercion is the clear defect; the missing-value default is defensible but inconsistent with `cpp` and the spec — recommend requiring it. Resolve both:

```diff
-        # Resolve event_type from parameters if provided, otherwise default to "purchase"
-        event_type_val = EventType.purchase
-        if parameters and isinstance(parameters, dict):
-            raw_event = parameters.get("event_type")
-            if raw_event:
-                try:
-                    event_type_val = EventType(raw_event)
-                except ValueError:
-                    logger.warning(
-                        "Unknown event_type '%s' for CPA pricing option %s, defaulting to 'purchase'",
-                        raw_event,
-                        pricing_option_id,
-                    )
+        raw_event = parameters.get("event_type") if isinstance(parameters, dict) else None
+        if not raw_event:
+            raise ValueError(f"CPA pricing option {pricing_option_id} requires parameters.event_type")
+        try:
+            event_type_val = EventType(raw_event)
+        except ValueError:
+            raise ValueError(
+                f"CPA pricing option {pricing_option_id} has unknown event_type '{raw_event}'. "
+                f"Supported values: {[e.value for e in EventType]}"
+            )
```

**2 — CPA drops `custom_event_name` / `event_source_id`.** `:276-280`. `custom` is a valid `EventType` and `cpa-option.json` states `custom_event_name` is "Required when event_type is 'custom'." The constructor never sets it, and the library does not enforce the conditional (`CpaPricingOption(event_type=custom, custom_event_name=None)` constructs), so `event_type=custom` emits a spec-violating object. Read both from `parameters` and reject `custom` without a name:

```diff
+        custom_event_name = parameters.get("custom_event_name") if isinstance(parameters, dict) else None
+        if event_type_val == EventType.custom and not custom_event_name:
+            raise ValueError(
+                f"CPA pricing option {pricing_option_id} with event_type 'custom' requires parameters.custom_event_name"
+            )
         return CpaPricingOption(
             **common_fields,
             event_type=event_type_val,
+            custom_event_name=custom_event_name,
+            event_source_id=parameters.get("event_source_id") if isinstance(parameters, dict) else None,
             fixed_price=float(rate),
         )
```

**3 — `time` rejects a spec-valid zero rate.** `:306`. `if not rate` treats `rate == 0` as missing (`bool(0.0) is False`), but `time-option.json` `fixed_price` is `minimum: 0` (a free/added-value placement is valid; the library accepts `0.0`). CPA differs — its `fixed_price` is `exclusiveMinimum: 0`, so rejecting `0` there (`:261`) is correct and should stay. For `time` only:

```diff
         if is_fixed:
-            if not rate:
+            if rate is None:
                 raise ValueError(f"Fixed time-based pricing option {pricing_option_id} requires rate")
             time_fields["fixed_price"] = float(rate)
```

**4 — Gratuitous deep import.** `:29`. `adcp.EventType` is the same object as the `generated_poc` path, and the sibling pricing types are already imported from `adcp` (`:16-26`). `generated_poc` is a codegen-internal namespace; use the public export:

```diff
 from adcp import (
     CpaPricingOption,
     CpcPricingOption,
     CpcvPricingOption,
     CpmPricingOption,
     CppPricingOption,
     CpvPricingOption,
+    EventType,
     FlatRatePricingOption,
     TimeBasedPricingOption,
     TimeUnit,
     VcpmPricingOption,
 )
 from adcp.types._generated import MediaChannel
-from adcp.types.generated_poc.enums.event_type import EventType
 from adcp.types.generated_poc.pricing_options.time_option import Parameters as TimeParameters
```

**5 — Test gaps.** `tests/unit/test_product_conversion_pricing.py`. Add: `event_type=custom` without `custom_event_name` (→ #2); unknown/missing `event_type` now raising (→ #1); `max_duration < min_duration` (spec says "max ≥ min", enforced by neither library nor code — `Parameters(min=30, max=3)` constructs); `time` `fixed_price=0` accepted (→ #3); CPA negative `rate` (the library's `gt=0` raises `ValidationError`, which the `:261` guard doesn't pre-empt — decide whether to surface a friendly `ValueError`). The four near-identical `time_unit` tests collapse to one parametrized case.

**Nits / informational**
- `time` builds `parameters` via the imported `Parameters` (`:299-303`); the `cpp`/`cpcv`/`flat_rate` branches instead pass `parameters` as a raw dict and let Pydantic coerce (still rejects a bad `time_unit`). The explicit `TimeUnit()` conversion gives a more specific error — a fair reason to keep it; flagging the inconsistency only.
- `pricing_option_id` (`:105`) labels CPA `..._auction` when `is_fixed` is false, though CPA always emits `fixed_price`. Cosmetic.
- `docs/test-obligations/constraints.md` CONSTR-PRICING-OPTION-XOR-01 ("Neither = invalid") contradicts `time-option.json` for this spec version — stale doc, separate fix, not this PR.

**Scope (worth one line in the PR description)**
- GAM cannot transact `cpa`/`time`: `ADCP_TO_GAM_COST_TYPE` (`src/adapters/gam/pricing_compatibility.py:42`) maps only cpm/vcpm/cpc/flat_rate; `is_compatible` returns `False` without crashing. Mock supports all. This PR delivers serialization/discovery; adapter line-item support is a separate step.
- A single unconvertible product fails the whole `get_products` response — `_get_products_impl` catches the `ValueError` per product and re-raises `AdCPAdapterError` rather than skip-and-log (contrast the channels loop). Pre-existing across all 9 models, not introduced here.

**--- Verified at `86405925b`:**
Ran: `pytest tests/unit/test_product_conversion_pricing.py` (48 passed); fetched `cpa-option.json` / `time-option.json` / `core/pricing-option.json` from `dist/schemas/3.1.0-beta.3/`; `git show 86405925b:` for the diff; library checks (`EventType` identity; `TimeBasedPricingOption` accepts neither-price and `fixed_price=0.0`; `CpaPricingOption` accepts `event_type=custom` with `custom_event_name=None`; `Parameters(min=30,max=3)` constructs); duplication baseline `src:36` unchanged; `PricingParameters` dead (1 ref); GAM cost-type map.
Not run: `./run_all_tests.sh`; live GAM `create_media_buy` / admin product-create (those claims inferred from the cost-type map and form-field grep, not executed); proposed diffs are verified against schema + library behavior, not applied-and-tested as a batch.

torbenbrodt:
Thx @ChrisHuie  , I addressed this in f0c7335

ChrisHuie:
# Re-review — #1473 (verified at head `f0c7335f3`)

The five SHOULD-FIX items are resolved — verified against the new head (`pytest tests/unit/test_product_conversion_pricing.py` → 52 passed): CPA `event_type` is now required and raises on missing/unknown (no silent default); `custom_event_name` is read and required when `event_type='custom'`, with `event_source_id` passed through; `from adcp import EventType` (top-level, the `generated_poc` import removed); `time` accepts a `0` `fixed_price` (`if rate is None`, matching the spec `minimum: 0`). The old default/fallback tests were replaced with missing/unknown/custom cases, and the four near-identical `time_unit` tests consolidated into one parametrized test. Against `cpa-option.json` / `time-option.json` the branches are now field-by-field conformant.

One spec MUST is still unenforced; the rest is optional polish.

## `max_duration >= min_duration` is not validated

`time-option.json` parameters: *"max_duration … **Must be >= min_duration** when both are present."* This is prose-only — the schema enforces just `ge: 1` per field and the SDK has no relational constraint, so a misconfigured row (`min=30, max=3`) serializes a spec-non-conformant option silently (confirmed at head). It mirrors the existing `time_unit` validation:

```diff
     time_params = TimeParameters(
         time_unit=time_unit_val,
         min_duration=parameters.get("min_duration"),
         max_duration=parameters.get("max_duration"),
     )
+    if (
+        time_params.min_duration is not None
+        and time_params.max_duration is not None
+        and time_params.max_duration < time_params.min_duration
+    ):
+        raise ValueError(
+            f"Time-based pricing option {pricing_option_id} has max_duration "
+            f"({time_params.max_duration}) < min_duration ({time_params.min_duration})"
+        )
     time_fields: dict = {**common_fields, "parameters": time_params}
```

Add a unit case asserting it raises on `min > max`.

## Optional / cosmetic (non-blocking)

- The CPA branch passes `custom_event_name` even when `event_type != custom`. The spec says it is *"ignored otherwise,"* so this is harmless; optionally set it only when `event_type == custom`.
- `pricing_option_id` labels CPA `…_auction` when `is_fixed` is false, though CPA always emits `fixed_price` — cosmetic.
- Pre-existing (not this PR): a misconfigured pricing option surfaces to the buyer as `SERVICE_UNAVAILABLE` / `recovery=transient` — the per-product `except` in `products.py` re-raises convert failures as `AdCPAdapterError`. Per `error-code.json` a configuration problem is `CONFIGURATION_ERROR` / `terminal` (retrying will not fix a publisher's pricing config). This PR's stricter validation makes that path fire more often, so it is worth a separate follow-up to map convert/config failures to `CONFIGURATION_ERROR`.

**--- Verified at `f0c7335f3`:**
Ran: `pytest tests/unit/test_product_conversion_pricing.py` (52 passed); `git diff 86405925b f0c7335f3`; probed production at head (missing/unknown/custom-without-name raise; custom-with-name, `lead`, and `time` rate=0 succeed); the proposed `max_duration` guard against `TimeParameters` in isolation (raises on `min>max`, passes valid cases); spec checks against `cpa-option.json`, `time-option.json`, `price-guidance.json` (`additionalProperties: true` — `floor` inside `price_guidance` is spec-allowed, so not flagged), and `error-code.json` (`SERVICE_UNAVAILABLE`→transient, `CONFIGURATION_ERROR`→terminal); no compliance storyboard grades pricing-option conversion (these constraints are graded-silent).
Not run: `./run_all_tests.sh` (CI green on the head); the `max_duration` guard verified in isolation, not applied-and-run as a batch.

ChrisHuie (approved):

PR #1474, opened 2026-06-22
test(bdd): wire UC-018 list_creatives-after-sync storyboard (#1405)

### LINKED ISSUE #1405: test: list_creatives with no filters returns library including recent syncs
Part of #1247 (AdCP 3.1 compliance). **Good first issue** — a clean intro to the `list_creatives` path.

## The behavior (v3.1 requirements Test `T-UC-018-storyboard-list-all-creatives-after-sync`)

```gherkin
Given the buyer recently synced three creatives in three different formats via sync_creatives
When the Buyer Agent sends list_creatives with no filters for the same account
Then the response should be schema-valid against list-creatives-response.json
And the creatives array should include each of the synced creatives
And each creative entry should expose creative_id, name, format_id, and status
```

## Links
- **Requirements (web):** https://main.d1tyvbf1at6iao.amplifyapp.com/#T-UC-018-storyboard-list-all-creatives-after-sync
- **Feature file (PR #1384):** https://github.com/prebid/salesagent/blob/c59ebcfeaddae3ffd45b27014f0d06e7111c9130/tests/bdd/features/BR-UC-018-list-creatives.feature#L759
- **Source obligation:** adcp protocols/creative/index.yaml · list_all

## Where to implement
- The `list_creatives` impl under `src/core/tools/`.
- Mostly wiring plus a strong assertion that each entry exposes `creative_id`, `name`, `format_id`, `status`.

## Definition of done
- [ ] `list_creatives` with no filters returns the account library including just-synced creatives, each exposing the four fields.

## Notes
- Good first task in the `list_creatives` area — pairs naturally before the `format_id`-object and `concept_ids` filter issues.
- Light SDK exposure; safe to start before the #1399 upgrade lands.

mkomorski:
Wire T-UC-018-storyboard-list-all-creatives-after-sync to real production across all wire transports (a2a/mcp/rest): after seeding creatives, a no-filter list_creatives returns them, schema-valid against list-creatives-response.json, each entry exposing
creative_id/name/format_id/status.

- Route UC-018 -> CreativeListEnv in the bdd conftest (_detect_uc/_harness_env), gated on @list-after-sync; the remaining UC-018 scenarios xfail fast at the fixture (mirrors UC-002/006/011) until they are individually wired.
- Add tests/helpers/pinned_schema.py: an offline validator against the vendored pinned AdCP schemas (v3.1-04f59d2d5) with $ref resolution.
- Steps are defined inline (module-scoped) so the shared "schema-valid against <file>" / "authenticated as principal" phrasings are NOT registered globally, which would alter the already-wired UC-004/005/006 suites.

### DIFF
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -2739,6 +2739,8 @@ def _detect_uc(request: pytest.FixtureRequest) -> str | None:
         return "UC-004"
     if any(t.startswith("T-UC-011") for t in marker_names):
         return "UC-011"
+    if any(t.startswith("T-UC-018") for t in marker_names):
+        return "UC-018"
     if any(t.startswith(_ADMIN_TAG_PREFIX) for t in marker_names):
         return "ADMIN"
     if "inventory_profile" in marker_names:
@@ -2854,6 +2856,26 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
         with CreativeFormatsEnv(e2e_config=ctx.get("e2e_config")) as env:
             ctx["env"] = env
             yield
+
+    elif uc == "UC-018":
+        # list_creatives — only the @list-after-sync storyboard scenario is wired
+        # (#1405). The remaining UC-018 scenarios (main/partition/boundary/filter
+        # siblings) have no step definitions yet, so xfail fast at the fixture
+        # (mirrors UC-002/006/011) rather than spinning up a DB per scenario only
+        # to auto-xfail at the first missing step.
+        marker_names = {m.name for m in request.node.iter_markers()}
+        if "list-after-sync" in marker_names:
+            # CreativeListEnv mocks only the audit logger; DB, repository, and
+            # query building are real. The Background auth step switches the env
+            # principal; the seed step owns the creatives under it.
+            request.getfixturevalue("integration_db")
+            from tests.harness.creative_list import CreativeListEnv
+
+            with CreativeListEnv(e2e_config=ctx.get("e2e_config")) as env:
+                ctx["env"] = env
+                yield
+        else:
+            pytest.xfail("UC-018 harness wired only for the @list-after-sync storyboard scenario (#1405)")
 
     elif uc == "UC-011":
         marker_names = {m.name for m in request.node.iter_markers()}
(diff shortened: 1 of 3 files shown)


ChrisHuie:
Superseded by #1493. That PR was stacked on this branch and merged first, so it brought this branch's UC-018 scaffolding in verbatim: `tests/helpers/pinned_schema.py` is byte-identical on main, and `tests/bdd/test_uc018_list_creatives.py` on main is a superset (the `@list-after-sync` scenario here plus the `@concept-id` block). The merge-base has neither file, so this branch now conflicts add/add on both, and force-resolving toward it would un-wire the concept-id scenario main already has. Recommend closing this as superseded — the work is in main.

Reviewed the merged result across a2a/mcp/rest against a real DB. Follow-ups worth filing against the merged storyboard (they apply to main, not just here):

1. Schema-validity isn't graded on the wire. The `Then` steps validate `model_dump(mode="json", exclude_none=True)` of the typed payload, not each transport's serialized output. REST and A2A serialize via `model_dump(mode="json")` (drops `None`) and are schema-valid; the MCP wire serializes via fastmcp `to_jsonable_python` (no `exclude_none`, because the MCP `list_creatives` path hands fastmcp the raw model instead of pre-dumping like `get_products`) and emits `null`-typed optional fields (`adcp_version`, `adcp_major_version`, `context`, `errors`, `ext`, `format_summary`, `status_summary`, `sandbox`) that the schema rejects. `exclude_none=True` drops exactly those, so `[mcp]` passes without the wire being checked. `uc005_format_id_shape` asserts on `ctx["wire_response"]` for the wire transports and falls back to `model_dump` only for IMPL — worth adopting here. Grading the real MCP wire will fail until that path stops emitting null optionals (a production fix, or xfail MCP with a tracking issue).

2. `tests/helpers/pinned_schema.py` `_load_by_ref` duplicates `load_json_schema` in `tests/unit/test_pydantic_schema_alignment.py` (same fixture tree, same hard-fail). Consider one shared loader.

3. The `list_all` storyboard grades four checks; the generated feature renders three — the `context.correlation_id` echo isn't asserted. Sending `context` in the request and asserting the echo would close that.

4. The seed override `data={"assets": {}}` is unnecessary — `CreativeFactory`'s default already provides non-null assets that pass the repository filter — and it skips the asset-union part of the schema. The factory default keeps the row realistic and still schema-valid.

Correct as written (no change needed): inline module-scoped steps (step modules are global plugins), the xfail-fast branch (non-regressive), `format_id` as `{agent_url, id}` with no production wire bug, and the scenario passing on all three in-process transports. Checked at b18c8896d against main d441d4c3.

ChrisHuie:
closing as superseded by #1493.

Issue #1376, opened 2026-06-01
UC-004 delivery breakdowns: by_device_type (and truncation flags) not implemented


KonstantinMirin:
## Production gap

`_list_media_buy_delivery` / `_impl` only populates `by_placement` (and now `by_geo`) breakdowns. The `by_device_type` reporting dimension is **not implemented**:

- `PackageDelivery` (`src/core/schemas/delivery.py`) has no `by_device_type` field — only `by_placement` and `by_geo`.
- The delivery builder (`src/core/tools/media_buy_delivery.py`, ~line 425) has no `_build_device_type_breakdown` helper (cf. `_build_geo_breakdown`).
- When a request asks for `device_type` (or multiple dimensions including it), `by_device_type` comes back as `None`/absent instead of a list.

Truncation flags (`by_*_truncated`, e.g. `by_device_type_truncated`) are also unimplemented.

## Behavioral evidence (BDD)

The following UC-004 scenarios are currently **xfailed** in `tests/bdd/conftest.py` pending this work:
- `T-UC-004-dim-supported` — by_device_type breakdown not implemented in _impl
- `T-UC-004-dim-truncated` — truncation flags not implemented
- `T-UC-004-dim-complete` — by_device_type_truncated flag not implemented
- `T-UC-004-dim-multi` — by_device_type breakdown not implemented on PackageDelivery (by_geo works)

Feature: `tests/bdd/features/BR-UC-004-deliver-media-buy-metrics.feature` (scenarios at ~lines 510, 539, 583).

## Suggested implementation

1. Add `by_device_type: list[DeviceTypeBreakdown] | None` (and a `DeviceTypeBreakdown` model) to `PackageDelivery` in `src/core/schemas/delivery.py`.
2. Add a `_build_device_type_breakdown(...)` helper in `src/core/tools/media_buy_delivery.py` mirroring `_build_geo_breakdown`, and pass `by_device_type=...` into the `PackageDelivery(...)` constructor, gated on `req.reporting_dimensions.device_type`.
3. Implement the `*_truncated` flags when a limit truncates entries.
4. Remove the corresponding xfail entries from `tests/bdd/conftest.py` once green.

## History

[...]

PR #1475, opened 2026-06-22
feat: implement by_device_type breakdown and truncation flags (#1376)

torbenbrodt:
- Add DeviceTypeBreakdown schema extending LibraryByDeviceTypeItem
- Add by_device_type, by_geo_truncated, by_device_type_truncated fields to PackageDelivery (AdCP 3.1.0 BR-RULE-091 INV-3/INV-4)
- Extract shared _apply_breakdown_limit() helper (DRY) used by both _build_geo_breakdown and _build_device_type_breakdown
- Add _build_device_type_breakdown(): synthesises mobile/desktop/tablet split from package totals; uses adapter-supplied raw_device_type when available
- Extend AdapterPackageDelivery with by_device_type so real adapters can pass device-type data through the core layer
- Graduate 4 xfailed BDD scenarios: T-UC-004-dim-supported, T-UC-004-dim-truncated, T-UC-004-dim-complete, T-UC-004-dim-multi
- Update TestPackageDeliveryFields and TestAdapterPackageDelivery field assertions; add test_device_type_breakdown.py (76 unit tests)

### DIFF
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -1276,11 +1276,23 @@ class MockAdServer(AdServerAdapter):
                         package_spend = spend / len(packages) if packages else spend
                         package_impressions = int(impressions / len(packages) if packages else impressions)
 
+                    # Simulate a device-type split from package totals.
+                    # Weights are distinct so descending sorts are verifiable in tests.
+                    # Real adapters replace this with actual platform data.
+                    imp = float(package_impressions)
+                    spd = float(package_spend)
+                    simulated_device_type = [
+                        {"device_type": "mobile", "impressions": imp * 0.50, "spend": spd * 0.50},
+                        {"device_type": "desktop", "impressions": imp * 0.35, "spend": spd * 0.35},
+                        {"device_type": "tablet", "impressions": imp * 0.15, "spend": spd * 0.15},
+                    ]
+
                     by_package.append(
                         AdapterPackageDelivery(
                             package_id=package_id,
                             impressions=package_impressions,
                             spend=package_spend,
+                            by_device_type=simulated_device_type,
                         )
                     )
 
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -45,6 +45,34 @@ from src.core.schemas import (
 )
 
 
+def simulate_breakdowns(impressions: float, spend: float) -> tuple[list[dict], list[dict]]:
+    """Return deterministic geo + device_type splits for mock delivery data.
+
+    Weights are distinct and descending so sort/truncation behaviour is
+    verifiable in tests.  Real adapters replace this with actual platform data.
+
+    Returns:
+        (geo, device_type) — lists of raw dicts ready for AdapterPackageDelivery.
+    """
+    device_type = [
+        {"device_type": "mobile", "impressions": impressions * 0.50, "spend": spend * 0.50},
+        {"device_type": "desktop", "impressions": impressions * 0.35, "spend": spend * 0.35},
+        {"device_type": "tablet", "impressions": impressions * 0.15, "spend": spend * 0.15},
+    ]
+    _geo_codes = ["US", "GB", "DE", "FR", "CA", "AU", "JP", "BR", "IN", "MX"]
+    _geo_weights = [0.30, 0.15, 0.12, 0.10, 0.08, 0.07, 0.06, 0.05, 0.04, 0.03]
+    geo = [
+        {
+            "geo_code": geo_code,
+            "geo_level": "country",
+            "impressions": impressions * weight,
+            "spend": spend * weight,
+        }
+        for geo_code, weight in zip(_geo_codes, _geo_weights, strict=False)
+    ]
+    return geo, device_type
+
+
 class MockConnectionConfig(BaseConnectionConfig):
     """Connection config for Mock adapter."""
 
@@ -1276,32 +1304,9 @@ class MockAdServer(AdServerAdapter):
                         package_spend = spend / len(packages) if packages else spend
                         package_impressions = int(impressions / len(packages) if packages else impressions)
 
-                    # Simulate a device-type split from package totals.
-                    # Weights are distinct so descending sorts are verifiable in tests.
-                    # Real adapters replace this with actual platform data.
-                    imp = float(package_impressions)
-                    spd = float(package_spend)
-                    simulated_device_type = [
-                        {"device_type": "mobile", "impressions": imp * 0.50, "spend": spd * 0.50},
-                        {"device_type": "desktop", "impressions": imp * 0.35, "spend": spd * 0.35},
-                        {"device_type": "tablet", "impressions": imp * 0.15, "spend": spd * 0.15},
-                    ]
-
-                    # Simulate a geo split across 10 countries so that a small limit
-                    # (e.g. 5) triggers truncation (BR-RULE-091 INV-3).
-                    # Weights are distinct and descending for verifiable sort order.
-                    # Real adapters replace this with actual platform geo data.
-                    _geo_codes = ["US", "GB", "DE", "FR", "CA", "AU", "JP", "BR", "IN", "MX"]
-                    _geo_weights = [0.30, 0.15, 0.12, 0.10, 0.08, 0.07, 0.06, 0.05, 0.04, 0.03]
-                    simulated_geo = [
-                        {
-                            "geo_code": geo_code,
-                            "geo_level": "country",
-                            "impressions": imp * weight,
-                            "spend": spd * weight,
-                        }
-                        for geo_code, weight in zip(_geo_codes, _geo_weights, strict=False)
-                    ]
+                    simulated_geo, simulated_device_type = simulate_breakdowns(
+                        float(package_impressions), float(package_spend)
+                    )
 
                     by_package.append(
                         AdapterPackageDelivery(
--- a/src/adapters/mock_ad_server.py
+++ b/src/adapters/mock_ad_server.py
@@ -1287,11 +1287,28 @@ class MockAdServer(AdServerAdapter):
                         {"device_type": "tablet", "impressions": imp * 0.15, "spend": spd * 0.15},
                     ]
 
+                    # Simulate a geo split across 10 countries so that a small limit
+                    # (e.g. 5) triggers truncation (BR-RULE-091 INV-3).
+                    # Weights are distinct and descending for verifiable sort order.
+                    # Real adapters replace this with actual platform geo data.
+                    _geo_codes = ["US", "GB", "DE", "FR", "CA", "AU", "JP", "BR", "IN", "MX"]
+                    _geo_weights = [0.30, 0.15, 0.12, 0.10, 0.08, 0.07, 0.06, 0.05, 0.04, 0.03]
+                    simulated_geo = [
+                        {
+                            "geo_code": geo_code,
+                            "geo_level": "country",
+                            "impressions": imp * weight,
+                            "spend": spd * weight,
+                        }
+                        for geo_code, weight in zip(_geo_codes, _geo_weights, strict=False)
+                    ]
+
                     by_package.append(
                         AdapterPackageDelivery(
                             package_id=package_id,
                             impressions=package_impressions,
                             spend=package_spend,
+                            by_geo=simulated_geo,
                             by_device_type=simulated_device_type,
                         )
                     )
--- a/src/core/schemas/delivery.py
+++ b/src/core/schemas/delivery.py
@@ -185,6 +185,12 @@ class PackageDelivery(SalesAgentBaseModel):
         None,
         description="Placement-level delivery breakdown (populated when reporting_dimensions includes 'placement')",
     )
+    by_placement_truncated: bool | None = Field(
+        None,
+        description="True when by_placement was truncated by the requested limit; false when complete. "
+        "MUST be present whenever by_placement is present "
+        "(get-media-buy-delivery-response.json §by_placement_truncated; get_media_buy_delivery.mdx §Truncation).",
+    )
     by_geo: list[GeoBreakdown] | None = Field(
         None,
         description="Geographic delivery breakdown (populated when reporting_dimensions includes 'geo'). "
--- a/src/core/schemas/delivery.py
+++ b/src/core/schemas/delivery.py
@@ -130,8 +130,9 @@ class GeoBreakdown(LibraryByGeoItem):
 
     Library provides geo_level, system, geo_code, geo_name plus the full
     DeliveryMetrics surface. For metro/postal_area levels the ``system``
-    field carries the classification system the seller used (BR-RULE-091
-    INV-5, e.g. 'nielsen_dma', 'us_zip').
+    field carries the classification system the seller used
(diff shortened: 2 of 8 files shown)


ChrisHuie:
# Review — #1475 (verified at head `924b31ab`)

Implements `by_device_type` + `by_geo`/`by_device_type` truncation flags for `get_media_buy_delivery` (issue #1376), extracts a shared `_apply_breakdown_limit` helper, moves device/geo simulation into the mock adapter, and graduates 4 xfailed UC-004 scenarios. Checked against the pinned spec **3.1.0-beta.3** (`dist/schemas/3.1.0-beta.3/`). What holds: `DeviceTypeBreakdown(LibraryByDeviceTypeItem)` correctly extends the library type (Pattern #1); the deep `ByDeviceTypeItem` import is necessary (no top-level alias — verified — consistent with the pre-existing `ByGeoItem` import, the `# TODO` is accurate); the new unit tests call the real `_build_*`/`_apply_breakdown_limit` (not mock-echo), 71 pass at head; the graduated scenarios drive the real `_impl` builders (the harness patches only `get_adapter`); the geo tuple-return change is regression-guarded; the `*_truncated` MUST is satisfied.

No hard blockers. The functional items: truncation ignores the spec's sort order, and the device_type no-data path emits an empty array that contradicts "rows sum to the package total." Plus a real DRY duplication, mislabeled spec citations, one vacuously-graded assertion, and coverage gaps. All of this is **graded-silent** — the conformance storyboard (`delivery_reporting.yaml`) never requests `reporting_dimensions`, so green CI does not validate any of it.

**1 — Truncation drops arbitrary rows; spec requires sort-by-`sort_by`-descending first.** `_apply_breakdown_limit` (`media_buy_delivery.py:954`) does `entries[:limit]` with no sort. Spec (`get_media_buy_delivery.mdx` §Truncation): *"Rows are sorted by the requested `sort_by` metric descending"* (default `spend`). `_build_placement_breakdown` already resolves this correctly (`:935-950`, `_PLACEMENT_SORTABLE_METRICS` at `:887`), but geo/device_type never sort — they truncate in adapter order. Latent today (the mock pre-sorts), wrong once a real adapter supplies unsorted `by_geo`/`by_device_type`. Note `sort_by` is a `SortMetric` enum, so it must be resolved via `.value`. Fold the sort into the shared helper so all three breakdowns sort+truncate identically (this also fixes placement silently ignoring its `limit`):

```diff
+_BREAKDOWN_SORTABLE_METRICS = {"impressions", "spend", "clicks"}
+
 def _apply_breakdown_limit(entries: list[Any], dim: Any) -> tuple[list[Any], bool]:
+    # Spec: rows sorted by sort_by descending; fall back to spend if the metric
+    # is unknown or unreported on these entries (mirrors _build_placement_breakdown).
+    requested = getattr(dim, "sort_by", None)
+    sort_metric = (getattr(requested, "value", None) or str(requested)) if requested else "spend"
+    if not (sort_metric in _BREAKDOWN_SORTABLE_METRICS
+            and any(getattr(e, sort_metric, None) is not None for e in entries)):
+        sort_metric = "spend"
+    entries = sorted(entries, key=lambda e: getattr(e, sort_metric, 0) or 0, reverse=True)
     limit = getattr(dim, "limit", None)
     if limit is not None and len(entries) > limit:
         return entries[:limit], True
     return entries, False
```
Then route `_build_placement_breakdown` through the same helper (it currently sorts inline but never applies `limit`), and add a test that feeds **unsorted** rows + a limit and asserts the survivors are the top-N (no test does today — the mock's rows are pre-sorted, so this is invisible).

**2 — `by_device_type=[]` for a delivering package contradicts the breakdown contract; the deeper cause is that "seller supports it" is never checked.** `_build_device_type_breakdown` (`:1053-1056`) returns `([], False)` when the adapter supplies no device rows — pinned by `test_returns_empty_list_when_no_adapter_data` for a 1000-impression package. The spec says by_X is present *"when the buyer requests … AND the seller supports it"*, unsupported dimensions are *"silently omitted"* (the `dim-unsupported` INV — no empty array), and rows *"should sum to the package total"* (`[]` sums to 0). The no-fabrication instinct is right; the issue is emitting `[]` rather than omitting. Minimal fix:
```diff
     else:
-        # No adapter data available — return an empty breakdown rather than
-        # fabricating a split.  The array presence satisfies BR-RULE-091 INV-1;
-        # truncated=False is correct (nothing was cut).
-        entries = []
+        # No per-device data — omit the dimension (None) rather than emit an
+        # empty array, which asserts a complete zero-row breakdown for a package
+        # that delivered impressions.
+        return None, None
```
(Updates the 3 unit tests that pin `[]`.) Root cause / likely follow-up: the spec gates breakdown presence on `reporting_capabilities.supports_device_type_breakdown` (and `supports_geo_breakdown`), but the builders emit on request presence alone — `reporting_capabilities` is a free-form `JSONType` dict (`models.py:293`) that nothing reads. That is why the `dim-unsupported` scenario only "passes" by testing `audience` (a dimension never modeled). Gating emission on the declared capability is the spec-correct behavior; it is pre-existing for geo, so it is a separate change, but it is the real fix and worth a tracked follow-up. Apply the same omit-vs-`[]` decision to geo's no-data fallback so both dimensions behave identically (today geo returns a single `geo_code="aggregate"` row — see nits).

**3 — DRY: the simulated split is byte-identical in production and the test harness.** The device split (`mobile 0.50 / desktop 0.35 / tablet 0.15`) and the 10-country geo split are duplicated verbatim in `src/adapters/mock_ad_server.py:1284-1303` and `tests/harness/_mixins.py:120-134`. The duplication guard misses it (it scans `src/` and `tests/` in separate pylint runs; baseline unchanged at `src:36`; running R0801 across both files together flags it). Extract one helper — the mock adapter is the natural home, and `_mixins.py` already imports from `src.core.tools.media_buy_delivery`, so it can call it:
```python
def simulate_breakdowns(impressions: float, spend: float) -> tuple[list[dict], list[dict]]:
    """Deterministic geo + device_type splits (distinct descending weights so
    sort/truncation are verifiable in tests)."""
    device_type = [
        {"device_type": "mobile", "impressions": impressions * 0.50, "spend": spend * 0.50},
        {"device_type": "desktop", "impressions": impressions * 0.35, "spend": spend * 0.35},
        {"device_type": "tablet", "impressions": impressions * 0.15, "spend": spend * 0.15},
    ]
    codes = ["US", "GB", "DE", "FR", "CA", "AU", "JP", "BR", "IN", "MX"]
    weights = [0.30, 0.15, 0.12, 0.10, 0.08, 0.07, 0.06, 0.05, 0.04, 0.03]
    geo = [{"geo_code": c, "geo_level": "country", "impressions": impressions * w, "spend": spend * w}
           for c, w in zip(codes, weights, strict=False)]
    return geo, device_type
```

**4 — Mislabeled spec citations.** Comments, docstrings, and schema `Field` descriptions cite "AdCP 3.1.0 BR-RULE-091 INV-1..6" (`delivery.py:192-205`; `media_buy_delivery.py:897/973/1029/1058`). `BR-RULE-091`/`INV-*` are salesagent-internal IDs (`docs/test-obligations/bdd-traceability.yaml`) — 0 matches in the AdCP repo. The genuine basis is `get-media-buy-delivery-response.json` (`by_*_truncated`: *"Sellers MUST return this flag whenever … present"*) and `get_media_buy_delivery.mdx` §Truncation. ("AdCP 3.1.0" is also imprecise — `by_device_type_truncated` exists back to 3.0.1.) Cite the schema/section in wire-facing descriptions; keep the internal rule IDs in internal docs/test tags only.

**5 — A graduated scenario's INV-2 half is a vacuous assertion.** `dim-multi` and `dim-unsupported` assert "should NOT include `by_audience`" via `then_packages_exclude_field`/`then_packages_exclude_breakdown` (`uc004_delivery.py:2043/2164`). `PackageDelivery` has no `by_audience` field and `extra="forbid"`, so `getattr(pkg, "by_audience", None)` is always `None` — the assertion can't fail, independent of production. The inclusion half (`by_geo` + `by_device_type` present) is substantive, so `dim-multi` isn't fully circular, but graduating it grades a no-op INV-2 check. The step bodies are pre-existing; strengthen by asserting on the serialized dict (`"by_audience" not in pkg.model_dump()`) or use a dimension the model can represent.

**Coverage gaps** — (a) no test asserts breakdown rows sum to the package total (the invariant the docstrings invoke); (b) `by_device_type_truncated == True` is only unit-tested via a hand-fed helper call — no wire/BDD scenario exercises device_type truncation (the only wire truncation scenario uses geo); (c) per item 1, no test feeds unsorted rows. The xfail removals are the correct ratchet direction.

**Nits** — PR body/commit say "76 unit tests" (the file has **29**) and "synthesises mobile/desktop/tablet split from package totals" (core no longer synthesizes — it returns `[]`; the split moved to the mock adapter); `_collect_all_packages` docstring advertises evading the count-only assertion guard (pre-existing; the consuming assertions are substantive — reword to state intent, or fix the guard); `geo_code="aggregate"` is a non-spec sentinel (pre-existing; reconcile with the item-2 omit decision); `get_media_buy_delivery` is exempt from the roundtrip-test hook and this PR widens the field surface flowing through `apply_testing_hooks` — add a roundtrip test and drop the exemption; the `then_field_true/false` per-package broadening leaves a dead top-level fallback branch, and `given_geo_exceeds_limit`/`given_device_type_under_limit` set `ctx` flags nothing reads.

**--- Verified at `924b31ab`:**
Ran: `pytest tests/unit/test_device_type_breakdown.py tests/unit/test_delivery_schema_contracts.py` (71 passed; the new file = 29 tests); fetched `get-media-buy-delivery-{request,response}.json`, `get_media_buy_delivery.mdx`, `reporting-capabilities.json`, `delivery_reporting.yaml` from `dist/.../3.1.0-beta.3/`; `git show 924b31ab:` for all 8 changed files; confirmed `BR-RULE-091`/`INV-*` absent from the AdCP repo (`search/code`=0), present in `bdd-traceability.yaml`; confirmed `sort_by` is a `SortMetric` enum and ran the proposed sort helper (default→top-2-by-spend, `sort_by=impressions`→reordered, unreported metric→spend fallback); `PackageDelivery.model_fields` (no `by_audience`, `extra="forbid"`); `ByDeviceTypeItem`/`ByGeoItem` not top-level exported; no `supports_*_breakdown` referenced in `src/`; mock vs `_mixins` sim block byte-identical + R0801 cross-file reproduction; duplication baseline `src:36` unchanged.
Not run: `./run_all_tests.sh`; the UC-004 BDD suite end-to-end and a live MCP/A2A/REST wire capture (graduated scenarios pass per CI; not re-run here); a live real-adapter (GAM) delivery to observe the empty-array/aggregate fallback and the unsorted-truncation path; the proposed diffs are verified against spec + library behavior (the sort helper was executed in isolation), not applied-and-run as a batch.

torbenbrodt:
Thank @ChrisHuie ! I addressed your review in 8ccd995

ChrisHuie:
# Re-review — #1475 (verified at head `8ccd9959c`)

The five SHOULD-FIX items and the coverage gaps are resolved — verified against the new head (`pytest tests/unit/test_device_type_breakdown.py tests/unit/test_delivery_schema_contracts.py` → 74 passed): sort-before-truncate (`SortMetric` `.value` resolution + `spend` fallback, all three builders through `_apply_breakdown_limit`); device_type no-data now omits (`None, None`); `simulate_breakdowns` deduplicated and imported by the harness; spec citations corrected on the docstrings and the wire-facing `Field` descriptions; the `by_audience` exclusion now checks `model_dump()`; the guard-evasion docstring, dead `ctx` flags, and dead fallback branch removed. New tests cover rows-sum-to-total, sort override, unknown-metric fallback, and top-N truncation.

One spec-MUST seam remains, introduced by routing placement through the shared helper.

## `by_placement_truncated` is not surfaced

Placement now truncates at its `limit` (default 25) via `_apply_breakdown_limit`, but `_build_placement_breakdown` discards the flag (`limited, _truncated = …; return limited`) and `PackageDelivery` doesn't expose it. The spec mandates it the same way as the two flags already added — `by_placement_truncated`: *"Sellers MUST return this flag whenever by_placement is present (false means the list is complete)"* (`get-media-buy-delivery-response.json`) — and the library `ByPackageItem` already defines the field. The helper already computes the value; surface it and make placement's return contract match geo/device.

1. `schemas/delivery.py` — add the field after `by_placement`:
```python
by_placement_truncated: bool | None = Field(
    None,
    description="True when by_placement was truncated by the requested limit; false when complete. "
    "MUST be present whenever by_placement is present "
    "(get-media-buy-delivery-response.json §by_placement_truncated; get_media_buy_delivery.mdx §Truncation).",
)
```

2. `media_buy_delivery.py` — return the tuple instead of discarding the flag:
```diff
-) -> list[PlacementBreakdown] | None:
+) -> tuple[list[PlacementBreakdown] | None, bool | None]:
     ...
     if placement_dim is None:
-        return None
+        return None, None
     ...
-    limited, _truncated = _apply_breakdown_limit(placements, placement_dim)
-    return limited
+    return _apply_breakdown_limit(placements, placement_dim)
```

3. `media_buy_delivery.py` — unpack and assign at the caller (now identical to the geo/device lines beside it):
```diff
-                        placement_breakdown = _build_placement_breakdown(
+                        placement_breakdown, placement_truncated = _build_placement_breakdown(
                             placement_dim, raw_placements, package_impressions, package_spend, package_clicks
                         )
                         ...
                                 by_placement=placement_breakdown,
+                                by_placement_truncated=placement_truncated,
```

4. `tests/unit/test_delivery_schema_contracts.py` — add `"by_placement_truncated"` to `EXPECTED_FIELDS` (the strict `==` set assertion fails otherwise).

Add a placement-truncation unit case mirroring `test_adapter_data_with_limit_keeps_top_by_spend`: feed more placements than the limit, assert `by_placement_truncated is True`, and `False` when complete.

## Optional / cosmetic (non-blocking)

- PR description still says "76 unit tests" (`test_device_type_breakdown.py` has 32) and "synthesises mobile/desktop/tablet split from package totals" (core now omits when there is no adapter data; the split lives in the mock's `simulate_breakdowns`).
- `_build_geo_breakdown`'s no-data fallback still returns a single `geo_code="aggregate"` row while device_type now omits — reconciling geo to also omit would make the two consistent (pre-existing).
- `get_media_buy_delivery` remains in the roundtrip hook's `exempt_operations`; this PR widens the field surface flowing through `apply_testing_hooks`, so a delivery roundtrip test plus dropping the exemption is a reasonable follow-up.

**--- Verified at `8ccd9959c`:**
Ran: `pytest tests/unit/test_device_type_breakdown.py tests/unit/test_delivery_schema_contracts.py` (74 passed); `git diff 924b31ab 8ccd9959c` across the 6 changed files; probed production at head — `_apply_breakdown_limit` sorts unsorted entries to top-N by spend, `_build_device_type_breakdown(raw=None)` returns `None, None`; confirmed `simulate_breakdowns` single source + harness import; confirmed `by_placement_truncated` is in the spec response schema and on the library `ByPackageItem`, and absent from `PackageDelivery`/`EXPECTED_FIELDS`; CI all-green on the new head.
Not run: `./run_all_tests.sh`; UC-004 BDD end-to-end / live wire capture (CI green); the proposed placement diff is grounded in the spec + the existing geo/device pattern, not applied-and-run.

ChrisHuie (approved):

Issue #1468, opened 2026-06-19
chore: post-#1400 follow-ups (review backlog + #1234 stack)


mkostromin-sigma:
> **Status (2026-06-27):** #1234 stack (#1400, #1401, #1476) on `main`. Follow-up PRs open and CI-green: [#1486](https://github.com/prebid/salesagent/pull/1486) (digest guard + `_install-uv` docs), [#1490](https://github.com/prebid/salesagent/pull/1490) (Part B helper privatization). ADR-008 py312 bump still deferred.

## Context

Tracking issue for work that should happen **after** [#1400](https://github.com/prebid/salesagent/pull/1400) (PR 5 of #1234 — version anchor consolidation) merges.

Parent epic: [#1234](https://github.com/prebid/salesagent/issues/1234) — **merged**.

Source reviews:
- [Approval review on #1400](https://github.com/prebid/salesagent/pull/1400#pullrequestreview-4528657959) — non-blocking follow-ups
- [#1400 author follow-up comment](https://github.com/prebid/salesagent/pull/1400#issuecomment-4747051809) — Part B helper trim sequencing

---

## Stack sequencing (required)

- [x] **Rebase and merge [#1401](https://github.com/prebid/salesagent/pull/1401) (PR 6 of #1234)** onto `main` after #1400 lands — merged 2026-06-26
- [ ] **ChrisHuie post-merge admin checklist** (after #1401): dependency graph, Security / Dependency Review gating, secret scanning, harden-runner follow-up, OpenSSF badge — see #1401 PR body

---

### Done in #1400 (do not re-open)

- [x] **ADR-008 postgres surface list** — removed nonexistent `compose.yaml`; repo uses `docker-compose*.yml` only ([review item #2](https://github.com/prebid/salesagent/pull/1400#pullrequestreview-4528657959))

### Recommended follow-up PR(s) — [#1486](https://github.com/prebid/salesagent/pull/1486) (awaiting merge)

- [x] **Digest guard honesty** ([review item #1](https://github.com/prebid/salesagent/pull/1400#pullrequestreview-4528657959)) — option (a): rename test/docstring to state presence + shape only

- [x] **`_install-uv` naming / description** ([review item #3](https://github.com/prebid/salesagent/pull/1400#pullrequestreview-4528657959)) — expanded `description:` in `.github/actions/_install-uv/action.yml`

[...]

PR #1401, opened 2026-06-09
ci: image supply-chain hardening (PR 6 of #1234)

mkostromin-sigma:
## Summary

Part of **#1234 — PR 6 of 6 (final)**. Image and workflow supply-chain hardening: egress auditing, signed release artifacts, OS-layer CVE gating, and security workflow expansion. Closes the #1234 rollout code path; post-merge admin steps remain (branch protection, dependency graph).

**Branch tip:** `2bbcc645` · **CI:** green (Summary ✅ on latest run)

### harden-runner (step-security v2.19.0, CVE-2025-32955)
- Added to **all Ubuntu jobs** in `ci.yml`, `security.yml`, `codeql.yml`, `release-please.yml`, `scorecard.yml`
- **Tiered policy:** `disable-sudo-and-containers: true` on jobs without Docker; audit-only on Docker/service-container jobs (quickstart, integration, e2e, admin, BDD, migration, `build-and-push`) — host Docker required for GHA services
- `harden-runner-emergency-revert.yml` — `workflow_dispatch` PR creator to flip `block` → `audit` if egress lockout occurs

### Release pipeline (`release-please.yml`)
Split into **`build-and-push`** + **`sign-and-attest`**:
- `cosign sign --yes --bundle` keyless signing
- `actions/attest-build-provenance` with `provenance: mode=max`, `sbom: true`
- `aquasecurity/trivy-action` — CRITICAL/HIGH OS-layer scan gates release
- **D47 gate:** sign/attest only when CI is green on release commit
- `SOURCE_DATE_EPOCH` from git + BuildKit `rewrite-timestamp=true`

### Security workflows
- **`dependency-review`** job (`Security / Dependency Review`) with `.github/dependency-review-config.yml` (`fail-on-severity: moderate`)
  - `continue-on-error: true` until dependency graph enabled on `prebid/salesagent` (post-merge admin)
- **`scorecard.yml`** — OpenSSF Scorecard self-host, SARIF upload
- **`codeql.yml`** — removed `continue-on-error` (gating per D10 Week 5)
- **`.github/zizmor.yml`** — allowlist entries for intentional postgres rolling tag + Docker Hub secrets
- **zizmor auditor persona** in `security.yml` — catches workflow permission drift

### Governance
- `docs/decisions/adr-007-build-provenance.md` — cosign/SBOM/provenance policy
[...]

PR #1476, opened 2026-06-22
chore(ci): #1234 closure tails (stacked on #1401)

mkostromin-sigma:
Closes #1234.

## Summary
Closure tails for #1234 / #1468, stacked on #1401 (PR 6). **Merge after #1401 lands**, then rebase onto `main`.

- **C5:** uv cache key includes `pyproject.toml` (#1228)
- **E1:** remove phantom `google_ad_manager_original` config (#1228)
- **D9:** document local vs CI Postgres isolation (#1233)
- **#1468 nits:** digest guard honesty, `_install-uv` description, `anchor_kind` comment

Part of closing #1234.

Depends on #1401.

## Test plan
- [x] `make quality-ci`
- [ ] Rebase onto `main` after #1401 merge
- [ ] CI green post-rebase

### DIFF
--- a/.github/actions/_install-uv/action.yml
+++ b/.github/actions/_install-uv/action.yml
@@ -1,5 +1,5 @@
-name: Install uv
-description: Install the uv version from .uv-version (single source; pin setup-uv here only).
+name: Install uv and Python
+description: Install the uv version from .uv-version and provision Python from .python-version (single-source anchors; pin setup-uv SHA here only).
 runs:
   using: composite
   steps:
--- a/.github/actions/_install-uv/action.yml
+++ b/.github/actions/_install-uv/action.yml
@@ -1,5 +1,5 @@
-name: Install uv and Python
-description: Install the uv version from .uv-version and provision Python from .python-version (single-source anchors; pin setup-uv SHA here only).
+name: Install uv
+description: Install the pinned uv from .uv-version (setup-uv SHA pinned here only). Python is resolved downstream by `uv sync` honoring .python-version.
 runs:
   using: composite
   steps:
@@ -12,4 +12,3 @@ runs:
       uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb # v5.4.0
       with:
         version: ${{ steps.uv_version.outputs.version }}
-        python-version-file: .python-version
--- a/.github/actions/_setup-env/action.yml
+++ b/.github/actions/_setup-env/action.yml
@@ -11,7 +11,7 @@ runs:
         path: |
           ~/.cache/uv
           .venv
-        key: ${{ runner.os }}-uv-v4-py${{ hashFiles('.python-version') }}-${{ hashFiles('uv.lock') }}
+        key: ${{ runner.os }}-uv-v5-py${{ hashFiles('.python-version') }}-${{ hashFiles('uv.lock', 'pyproject.toml') }}
 
     - name: Install dependencies
       shell: bash
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -8,7 +8,6 @@ queries:
 # Exclude generated / vendored / migration content from analysis to reduce noise.
 paths-ignore:
   - alembic/versions/
-  - src/adapters/google_ad_manager_original.py
   - tests/
   - htmlcov/
   - .venv/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -153,7 +153,6 @@ repos:
         language: system
         args: [--config-file=mypy.ini, src/]
         files: '^src/'
-        exclude: '^src/adapters/google_ad_manager_original\.py$'
         pass_filenames: false
         require_serial: true
         stages: [pre-push]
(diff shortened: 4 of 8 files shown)


ChrisHuie:
Reviewed at the PR's top commit (`1b5502b`, the "#1234 closure tails"). Note this PR is **stacked on #1401**, so the diff against `main` also carries #1401's 4 commits — the comments below concern only the closure-tails commit (the `_install-uv`/`_setup-env` changes, the codeql/pre-commit/mypy phantom-ref removals, the D9 doc, and the `_architecture_helpers`/digest-guard docstrings).

Net: the four closure items land correctly and E1 is actually more complete than #1228 listed. Two things worth fixing, plus three doc nits.

## Should-fix

**1. `_install-uv` name/description is inaccurate (the #1468 "description" item).**
`.github/actions/_install-uv/action.yml` now reads `name: Install uv and Python` and `…provision Python from .python-version`. But `setup-uv` v5.4.0 (the pinned SHA `f0ec1fc…`) declares only a `python-version` input — there is **no `python-version-file` input** (checked against the action's `action.yml` at that SHA and on its `main`). So `python-version-file: .python-version` (line 15) is an undeclared key on a node action — a silent no-op — and this action does **not** provision Python. Python is actually resolved by `uv sync --group dev` in `_setup-env` (uv reads `.python-version` itself). The new description attributes provisioning to the wrong step.

The dead input predates this PR (it came in with #1400), but since this is the #1468 `_install-uv` description item, suggest fixing both together:

```yaml
name: Install uv
description: Install the pinned uv from .uv-version (setup-uv SHA pinned here only). Python is resolved downstream by `uv sync` honoring .python-version.
```

and drop the no-op `python-version-file: .python-version` line (removing it changes nothing functionally, since `uv sync` already honors `.python-version`).

**2. Stacking.**
The closure-tails commit cherry-picks cleanly onto `main` and shares **zero files** with #1401, so it has no code dependency on it. #1228 also scoped C5 to PR 3 and E1 to PR 2 (or standalone). Retargeting this onto `main` as a standalone PR would unblock it now and shrink the review surface; at minimum, the #1401 commits carried here are an older snapshot than #1401's current head, so CI on this PR is exercising a stale combination.

## Nits (all in `tests/unit/_architecture_helpers.py`)

**3. `anchor_kind` comment over-generalizes.** The new comment says "each kind mirrors its config key verbatim," but only about half of the `_PY_VERSION_PATTERNS` kinds are verbatim config keys (`target-version`, `python_version`, `requires-python`, `basepython`); the others are descriptive labels (`dockerfile-arg`, `dockerfile-from`, `python-version-file`, `workflow-python-version`). The prior `(e.g. target-version)` wording was accurate by not asserting a universal rule.

**4. `src_python_files` docstring.** It says "excluding migrations," but the body (`yield from (repo / "src").rglob("*.py")`) excludes nothing, and there are no migrations under `src/` (they live in `alembic/versions/`). Pre-existing wording, but this commit edits the line — suggest `"""Every .py file under src/."""`.

**5. Issue ref in docstring.** `assert_dockerfile_digest_args_present` now ends with "(see #1468 for version-aware follow-up)" — issue refs in code comments are rare in this repo, and the commit message already carries the link.

## Verified good
- **C5:** single cache-key definition; `hashFiles('uv.lock', 'pyproject.toml')` + the `v4`→`v5` bump correctly invalidate stale caches.
- **E1:** `google_ad_manager_original` is gone from every tracked file (codeql / pre-commit / mypy + `src_python_files`) — more complete than #1228's list. The file is absent and not dynamically loaded (the adapter registry uses `google_ad_manager.py`), so these are dead-config removals with no behavior change.
- **D9:** the doc claims check out — fresh `postgres:17-alpine` service container per integration/admin/BDD job, persistent local Postgres, the `tests/admin/conftest.py` process-wide factory binding, and `test_architecture_postgres_image_anchor`.
- **#1468 digest honesty:** the `assert_dockerfile_digest_args_present` / `test_dockerfile_digest_args_present` docstrings now accurately state presence + `sha256:<64-hex>` shape only, matching the matcher.
- All `arch_guard` tests pass (330) at the closure-tails commit.

## Verification performed
- Ran: full `-m arch_guard` suite (330 pass); cherry-pick of the closure-tails commit onto `main` (clean, zero conflicts); repo-wide `google_ad_manager_original` sweep (0 refs); `setup-uv` `action.yml` input check at the pinned SHA and `main`.
- Not run: CI itself (the cache-key behavior and `uv sync` Python resolution) — verified by config/source inspection rather than a live run.

mkostromin-sigma:
@ChrisHuie All review items addressed on latest head (rebased onto `main` as standalone — no longer stacked on #1401, which is merged).

### Should-fix
1. **`_install-uv`:** Restored accurate `name`/`description`; dropped no-op `python-version-file` (undeclared on setup-uv v5.4.0 — Python resolved by `uv sync` in `_setup-env`).
2. **Stacking:** Branch retargeted onto `main` (1 commit: closure tails only).

### Nits
3. **`anchor_kind` comment:** Reverted over-generalized "mirrors config key verbatim" wording.
4. **`src_python_files` docstring:** `"""Every .py file under src/."""`.
5. **Digest guard docstring:** Removed `#1468` issue ref from code comment.

`make quality` green locally. CI re-running.

ChrisHuie (approved):

PR #1479, opened 2026-06-23
feat: match list_creative_formats format_ids on (agent_url, id) federation pair (#1411)

### LINKED ISSUE #1411: feat: third-party format_id (foreign agent_url) reported as observation, not failure
Part of #1247 (AdCP 3.1 compliance). **Good first issue** (slightly subtle) — best taken after the other two format-id issues.

## The behavior (v3.1 requirements Test `T-UC-005-storyboard-format-id-third-party-agent-out-of-scope`)

```gherkin
Given a product advertises a format_id whose agent_url points at a third-party creative agent
And the seller has no local copy of that format in its own catalog
When the Buyer Agent sends list_creative_formats with that third-party format_id
Then the seller should NOT fabricate a local format entry to satisfy the third-party reference
And the verification result should be reported as an observation rather than a graded failure
```

When `format_ids[].agent_url` points at a creative agent other than this seller, the seller cannot verify it without calling that agent. Such references are **observations**, not graded failures — and the seller must NOT fabricate a local format entry to cover them.

## Links
- **Requirements (web):** https://main.d1tyvbf1at6iao.amplifyapp.com/#T-UC-005-storyboard-format-id-third-party-agent-out-of-scope
- **Feature file (PR #1384):** https://github.com/prebid/salesagent/blob/c59ebcfeaddae3ffd45b27014f0d06e7111c9130/tests/bdd/features/BR-UC-005-discover-creative-formats.feature#L1077
- **Source obligation:** adcp media-buy/index.yaml · list_formats_integrity (`scope.equals=$agent_url`, `on_out_of_scope=warn`)

## Where to implement
- `list_creative_formats` under `src/core/tools/`.
- Ensure a foreign `agent_url` returns no fabricated local entry. The "observation vs failure" semantics live partly in the BDD assertion.

## Definition of done
- [ ] A third-party `format_id` (foreign `agent_url`, no local copy) does not produce a fabricated local format entry and is reported as an observation, not a graded failure.

## Notes
- Slightly subtle — assign to whoever takes the format-id object-shape / roundtrip issues once they understand the format path.
- Cleanest **after PR #1399** (`adcp==5.7.0`) lands.

mkomorski:
Closes #1411.

## What & why
Third-party `format_id`s (foreign `agent_url`) are now reported as observations, not graded failures. `list_creative_formats` matched `format_ids` on `id` alone, mis-resolving a foreign reference to a local format that merely shared an `id` — fabricating a local entry the seller does not host.

## Spec grounding
- Authority: `adcp@04f59d2d5` — `static/compliance/source/protocols/media-buy/index.yaml`, phase `list_formats_integrity`: `refs_resolve.match_keys: [agent_url, id]`, `scope.equals: $agent_url`, `on_out_of_scope: warn`.
- Schema: `core/format-id.json` requires `[agent_url, id]`. Pinned `adcp==5.7.0` (3.1.0-beta.3).
- Graded step: `list_formats_integrity` (third-party refs → observation/`warn`, not failure).

## Changes
- **Production:** add `canonical_agent_url()` + `format_id_identity()` (`schemas/_base.py`); filter on the `(agent_url, id)` pair; reuse the canonicalizer in `CreativeAgentRegistry._cache_key` (DRY).
- **Harness:** remove the stale `CreativeFormatsEnv.build_rest_body` override (returned `{}` behind a false docstring) so REST transmits `format_ids` + filters.
- **BDD:** wire `T-UC-005-storyboard-format-id-third-party-agent-out-of-scope` (`uc005_format_id_third_party.py`) with a collision setup; passes on a2a/mcp/rest.
- **Consistency:** fix a `creatives` vs `creative` `agent_url` typo in `given_config.py` (masked by id-only matching); update `BR-RULE-049 INV-2` ("(agent_url, id) pair") in the feature scenario + obligation doc; add REST to integration filter-test transports.

## Verification
- Full UC-005 BDD: 456 passed, 148 xfailed, 0 failed, 0 XPASS.
- Integration filter tests: 53 passed (incl. new REST coverage).
- `make quality`: 5034 passed (ruff, mypy, structural guards). No new allowlist/xfail entries.
- Falsifiability confirmed: reverting to id-only → scenario fails on all 3 transports.

## Follow-ups (separate PRs)
- Wire the roundtrip sibling `T-UC-005-storyboard-format-id-roundtrip-from-products` (shares this fix).
- Retire the now-unnecessary `_REST_XFAIL_TAGS` to gain REST filter coverage.

### DIFF
--- a/docs/test-obligations/business-rules.md
+++ b/docs/test-obligations/business-rules.md
@@ -813,7 +813,7 @@ Then activation proceeds and deployments are returned
 ### BR-RULE-049: Per-Filter Format Discovery Semantics
 **Obligation ID** BR-RULE-049-01
 **Layer** behavioral
-**Invariant:** type=exact match, format_ids=id match with silent exclusion, asset_types=OR, dimensions=ANY render, is_responsive=bidirectional, name_search=case-insensitive substring.
+**Invariant:** type=exact match, format_ids=(agent_url, id) pair match with silent exclusion, asset_types=OR, dimensions=ANY render, is_responsive=bidirectional, name_search=case-insensitive substring.
 **Scenario:**
 ```gherkin
 Given type_filter="video"
--- a/src/core/creative_agent_registry.py
+++ b/src/core/creative_agent_registry.py
@@ -38,7 +38,6 @@ from adcp.types import AssetContentType as AssetType
 from adcp.types import Error as AdCPResponseError
 from adcp.types import ImageFormatAsset
 from pydantic import ValidationError
-from yarl import URL
 
 from src.core.exceptions import (
     AdCPAdapterError,
@@ -46,7 +45,7 @@ from src.core.exceptions import (
     AdCPRateLimitError,
     AdCPServiceUnavailableError,
 )
-from src.core.schemas import Format, FormatId, url
+from src.core.schemas import Format, FormatId, canonical_agent_url, url
 
 
 def _known_asset_types() -> frozenset[str]:
@@ -297,10 +296,12 @@ class CreativeAgentRegistry:
     def _cache_key(agent_url: str) -> str:
         """Canonicalize agent URL for consistent cache keys (RFC 3986).
 
-        yarl handles: scheme/host lowercase, default port removal, percent-encoding.
-        We additionally strip trailing slash so `/` and empty path are equivalent.
+        Delegates to ``schemas.canonical_agent_url`` so the format cache key and the
+        format_id federation identity (``format_id_identity``) share one
+        canonicalization (DRY) — a reference and its cached catalog can never
+        disagree over trailing-slash/case/default-port noise.
         """
-        return str(URL(str(agent_url))).rstrip("/")
+        return canonical_agent_url(agent_url)
 
     def _build_adcp_client(self, agents: list[CreativeAgent]) -> ADCPMultiAgentClient:
         """Build AdCP client from creative agent configs."""
--- a/src/core/schemas/_base.py
+++ b/src/core/schemas/_base.py
@@ -143,13 +143,17 @@ def url(value: str) -> AnyUrl:
 
 
 def canonical_agent_url(agent_url: object) -> str:
-    """Canonicalize an agent_url for identity comparison (RFC 3986 + trailing-slash strip).
+    """Canonicalize an agent_url for identity comparison (spec MUST canonicalization).
 
-    yarl normalizes scheme/host case, default ports, and percent-encoding; we
-    additionally strip the trailing slash so ``https://x.org`` and ``https://x.org/``
-    compare equal. This is the single canonical form used both to compare two
-    FormatId references for federation identity (see ``format_id_identity``) and to
-    key the creative-agent format cache (``CreativeAgentRegistry._cache_key``).
+    Delegates to the SDK's ``adcp.signing.canonicalize_target_uri`` so federation
+    identity uses the *same* canonical form the spec mandates for target URIs —
+    lowercased scheme/host, dropped default ports, normalized percent-encoding, and
+    stripped userinfo + fragment (a hand-rolled normalizer such as yarl keeps those,
+    diverging from the spec). The SDK preserves a trailing slash, so we additionally
+    strip it to keep ``https://x.org`` and ``https://x.org/`` equal. This is the
+    single canonical form used both to compare two FormatId references for federation
+    identity (see ``format_id_identity``) and to key the creative-agent format cache
+    (``CreativeAgentRegistry._cache_key``).
 
     Args:
         agent_url: A URL string or ``AnyUrl`` (FormatId.agent_url, CreativeAgent.agent_url).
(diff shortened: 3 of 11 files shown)


ChrisHuie:
Verified the mechanism end-to-end at head `13e13fc43` — the `(agent_url, id)` fix is correct and spec-conformant (silent empty result IS the conformant "observation"; `on_out_of_scope: warn` is a grader outcome, not a wire obligation), and I confirmed the failing oracle: reverting to id-only reds the new scenario on all 3 transports with the federation assertion firing. Nice work. A few items, none blocking CI:

**1 (recommended fold-in) — REST xfails are now stale.** Removing the `build_rest_body` override makes REST filter for real, which flips `_REST_XFAIL_TAGS` to XPASS under `BDD_ALL_TRANSPORTS=1`: `14 failed, 45 passed` — including the `inv-2-holds[rest]` scenario this PR renames. They don't fail default CI (the flag is unset there), but the reason strings are now false and strict-xfail is the signal to retire them. Emptying the set → 0 failed (verified). Please delete `_REST_XFAIL_TAGS` and its `if is_rest:` block in `conftest.py` (the 3 non-graduating tags stay xfailed via their type-filter / no-dim markers). The e2e_rest "build_rest_body returns {}" comment + `uc005_filter_e2e_untestable` are stale for the same reason.

**2 — `output_format_ids` / `input_format_ids` still match on id alone** (`creative_formats.py:374-381`), while `format_ids` now matches the pair. All three `$ref` the same format-id schema. Could you confirm whether the `list_formats` storyboard `match_keys` cover output/input refs? If yes, route them through `format_id_identity` too; if not, a one-line note that the asymmetry is intentional.

**3 — `canonical_agent_url` diverges from the spec's MUST canonicalization** (it keeps userinfo + fragments). The SDK ships a conformant helper: `adcp.signing.canonicalize_target_uri(x).rstrip("/")` matches the spec while preserving your `x.org ≡ x.org/` equivalence (the SDK keeps the trailing slash, so the `rstrip` is still needed). Applies to both call sites (federation identity + `_cache_key`).

**4 — IMPL has no foreign-agent coverage.** BDD runs a2a/mcp/rest; integration runs IMPL but only same-agent matches. One integration case seeding `(AGENT_URL, "x")` and requesting `(third-party, "x")` asserting `formats == []` would cover the discrimination on IMPL.

Minor: `REQ_TRANSPORTS` in `test_creative_formats_ordering.py` is unused (no test references it); the spec citation should name the `list_formats` step (not `list_formats_integrity`, which owns different checks) and the pinned `3.1.0-beta.3` (the cited `04f59d2d5` is `3.0.3`); drop the `pre-` issue refs from code comments.

---
**Verified at `13e13fc43`:** ran the new scenario (3 passed — a2a/mcp/rest), the id-only mutation (3 failed, federation assertion fired), and `BDD_ALL_TRANSPORTS=1` (14 XPASS → 0 after emptying `_REST_XFAIL_TAGS`). Not run: full e2e_rest (Docker).

mkomorski:
Thanks for the thorough review — all four items plus the minors are in (commit 41478e5c). Verified each against the spec (`adcontextprotocol/adcp`) and the SDK before changing anything.

**1 — REST xfails retired.** Deleted `_REST_XFAIL_TAGS` and its `if is_rest:` block; in-process REST serializes the body and filters for real now. Verified `BDD_ALL_TRANSPORTS=1` over UC-005 → 478 passed, 0 failed.

The same `build_rest_body` removal also turned the `disclosure_positions` strict-xfail block stale: it marked a2a/mcp/rest, but a2a/rest now pass (only the MCP wrapper rejects the keyword), so I scoped it to `is_mcp`. The e2e_rest comment/reason no longer reference `build_rest_body returns {}`; they now cite the actual transport-independent gaps (`type` removed from the SDK 5.7 request model; `disclosure_positions` `uniqueItems` unvalidated in production).

**2 — `output_format_ids` / `input_format_ids`.** Routed both through `format_id_identity`. Tmat-id.json` schema andcarry the same `(agent_url, id)` identity, so I removed the asymmetry rather than documenting it. Ths only the top-level`format_id`, but the contract is symmetric across every FormatId reference,
and the INV-9/INV-10 scgent_url, id}` objectsso they now grade the pair.

**3 — Canonicalization.** `canonical_agent_url` now delegates to
`adcp.signing.canonical` at both call sites(federation identity + `_cache_key`). Confirmed it applies the spec's MUST
canonicalization (stripeserves the trailingslash, so the `rstrip` still backs the `x.org ≡ x.org/` equivalence.

**4 — IMPL foreign-agent coverage.** Added `test_foreign_agent_urleds `(AGENT_URL,"shared_id")`, requests `(third-party, "shared_id")`, asserts `formats ==  []`, parametrized over e — the id-only mutation reds it on all four including IMPL.

**Minors.** Corrected the citations to the `list_formats` step (it owns    `scope.equals: $agent_u;`list_formats_integrity` owns the roundtrip `refs_resolve`) and the pinned `3.1.0-beta.3` (the `04ed the `pre-#1411`comment refs and removed the unused `REQ_TRANSPORTS`.

**Verification:** `make quality` 5034 passed; integration filter + ordering 47 passed; UC-005 BDD a

ChrisHuie (approved):

PR #1487, opened 2026-06-26
fix: cover A2A permission error boundary

numarasSigmaSoftware:
## Summary
- add focused A2A boundary coverage for the new PermissionError translation path
- assert PermissionError becomes InvalidRequestError (-32600) with the original message
- keep scope limited to the recent boundary change in d0b472cb

## Validation
- ============================= test session starts ==============================
platform darwin -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/nicolas.umaras/.codex/worktrees/7f43/prebid_salesagent
configfile: pytest.ini
plugins: mock-3.15.1, anyio-4.12.0, json-report-1.5.0, Faker-40.8.0, timeout-2.4.0, metadata-3.1.1, asyncio-1.3.0, typeguard-4.4.4, bdd-8.1.0, logfire-4.16.0, cov-7.0.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 42 items

tests/unit/test_error_boundary_translation.py .......................... [ 61%]
................                                                         [100%]

============================== 42 passed in 1.83s ==============================


numarasSigmaSoftware:
Closing: the assertions here pin behavior that #1547 deliberately changes.

This test asserts `error.message == "access denied"` (the raw handler message passed
through) and `error.data is None`. #1547 changes both at the A2A boundary:

- `_internal_error_for` builds `error.message` from the **sanitized** error, not the
  original, so raw exception text no longer reaches the wire.
- Auth rejections stay JSON-RPC errors but now **attach the two-layer AdCP envelope as
  `data`**, so `data is None` stops holding.

It also drives `_handle_explicit_skill` rather than `on_message_send`, so it grades the
handler's re-raise rather than the wire the buyer sees — the same gap #2070 describes.

The underlying coverage is still worth having: main has no `-32600` assertion in
`tests/unit/test_error_boundary_translation.py` at all. It should land against the
post-#1547 contract (sanitized message, envelope in `data`) and on the real boundary,
not be rebased onto the old expectations.

Issue #1442, opened 2026-06-12
Migrate api_v1 REST *Body models to SalesAgentBaseModel (Pattern #7 extra-field policy)


KonstantinMirin:
## Context
Surfaced by the hardened `no_bare_basemodel` guard added in #1399 (commit `683cff60a`).

`src/routes/api_v1.py` declares 11 REST request-body models (`GetProductsBody`, `CreateMediaBuyBody`, `UpdateMediaBuyBody`, `GetMediaBuyDeliveryBody`, `SyncCreativesBody`, `ListCreativesBody`, `UpdatePerformanceIndexBody`, `ListCreativeFormatsBody`, `ListAuthorizedPropertiesBody`, `ListAccountsBody`, `SyncAccountsBody`) that inherit bare `pydantic.BaseModel`. They therefore lack the Pattern #7 environment-based extra-field policy (`extra="forbid"` in dev/CI, `extra="ignore"` in prod).

These parse untrusted buyer REST input and should extend `SalesAgentBaseModel`.

## Why this is a follow-up, not a mechanical swap
Migrating flips extra-field handling on the REST boundary: today unknown fields in a request body are silently dropped in all environments (pydantic default `extra="ignore"`); after migration an unexpected field would raise a 422 in dev/CI. That is a behavioral change to the REST contract and needs REST integration coverage before flipping.

## Scope
- [ ] Extend the 11 `*Body` classes from `SalesAgentBaseModel`
- [ ] Add REST integration tests asserting the dev/CI `extra="forbid"` rejection and prod `extra="ignore"` passthrough
- [ ] Remove the 11 entries from `ALLOWLIST` in `tests/unit/test_architecture_no_bare_basemodel.py` (drain to empty)
- [ ] Drop the `# FIXME` markers in `src/routes/api_v1.py`

pmezzich:
Before anyone picks this up: the work is already done inside #1417. On `feature/media-buy-validation-refactor`, every `*Body` model in `src/routes/api_v1.py` now extends `SalesAgentBaseModel` (GetProductsBody, CreateMediaBuyBody, UpdateMediaBuyBody, GetMediaBuyDeliveryBody, SyncCreativesBody, ListCreativesBody, UpdatePerformanceIndexBody, ListCreativeFormatsBody, ListAuthorizedPropertiesBody, plus the new ListAccountsBody and SyncAccountsBody), while main still has all of them on bare `BaseModel` with the FIXME(#1442) tags.

[...]

PR #1493, opened 2026-06-29
feat: 1407 list creatives concept ids

### LINKED ISSUE #1407: feat: list_creatives filter by concept_ids (v3.1 new filter)
Part of #1247 (AdCP 3.1 compliance). **Good first issue** — adds one new v3.1 filter field plus its predicate.

## The behavior (v3.1 requirements Test `T-UC-018-storyboard-filter-by-concept-id`)

```gherkin
Given the principal has creatives under concept "concept_summer_2026" and others under different concepts
When the Buyer Agent sends list_creatives with filters.concept_ids ["concept_summer_2026"]
Then the response should be schema-valid against list-creatives-response.json
And the creatives array should only include creatives belonging to concept "concept_summer_2026"
And each returned creative should carry concept_id "concept_summer_2026" and a concept_name
```

`filters.concept_ids` is a **v3.1-added** filter (array of concept-id strings, minItems 1). Concepts group related creatives across sizes and formats.

## Links
- **Requirements (web):** https://main.d1tyvbf1at6iao.amplifyapp.com/#T-UC-018-storyboard-filter-by-concept-id
- **Feature file (PR #1384):** https://github.com/prebid/salesagent/blob/c59ebcfeaddae3ffd45b27014f0d06e7111c9130/tests/bdd/features/BR-UC-018-list-creatives.feature#L787
- **Source obligation:** adcp `static/schemas/source/creative/list-creatives-request.json`

## Where to implement
- `list_creatives` request schema (add `concept_ids` filter) + the filter predicate, under `src/core/tools/` and the creatives schema.
- Verify the creative model carries `concept_id` / `concept_name` first; each returned creative must expose both.

## Definition of done
- [ ] `filters.concept_ids` scopes results to the named concept(s); each returned creative carries `concept_id` and `concept_name`.

## Notes
- This touches the list-creatives request schema. **Cleanest after PR #1399** (`adcp==5.7.0`, spec 3.1.0-beta.3) lands, since `concept_ids` is a 3.1 field the upgraded SDK provides. If you start before, you may declare it locally and then reconcile when #1399 merges — coordinate.

mkomorski:
## feat: `list_creatives` filter by `concept_ids` (v3.1)

closes #1407 

Implements AdCP 3.1 `filters.concept_ids` on `list_creatives`: scoping the library to a creative concept and exposing `concept_id` / `concept_name` on each returned creative. Concepts group related creatives across sizes and formats (Flashtalking concepts, Celtra campaign folders, CM360 creative groups).

> **Stacked on #1405 (PR #1474).** This branch cherry-picks #1405's UC-018 BDD scaffolding (`test_uc018_list_creatives.py`, conftest `_detect_uc`/`_harness_env` UC-018 routing, `tests/helpers/pinned_so has a harness to bindto. **Merge #1474 first**, then rebase this — the #1405 commit drops out and
the diff reduces to the, retarget the PR baseto the #1474 branch for a clean review diff.)

### Spec grounding

Authoritative version: **AdCP 3.1.0-beta.3**, pinned via `adcp==5.7.0` (ref`v3.1-04f59d2d5`).
| Contract | Source |
|---|---|
| `filters.concept_ids` (array of strings, `minItems: 1`) | `core/creative-filters.json`, `creative/list-creatives-request.json` |
| `creatives[].concept_id`, `creatives[].concept_name` | `creative/list-creatives-response.json` |
| `filters_applied: ["concept_ids=…"]` | response example in `list-creatives-response.json` |

 The adcp 5.7.0 library already defines all three types —                   `CreativeFilters.concep`Creative.concept_id`/`concept_name`. **No schema change was needed**; the work was wiring them tong: the scenario bindsits `@source` to the request schema (the compliance YAML is not vendored), so this dimension is scifecycle /list_filtered` + the scenario's own concept-membership assertions.         
### What changed                                                           
**Production**                                                             
- **`repositories/creative.py`** — `get_by_principal(…, concept_ids=…)`    filters on the JSONB `d_string()` idiom as`account.py`).                                                             - **`tools/creatives/li_impl`) — threads`concept_ids` from the structured filters into the DB query (previously it fed only `filters_applid** before reaching thequery), populates `concept_id`/`concept_name` on each response creative frothe data blob, and repors_applied`.
- **`routes/api_v1.py`** — `ListCreativesBody` gained a `filters` object anthe handler coerces it eviously carried **nostructured filters at all**; this unblocks the whole `CreativeFilters`     surface over REST, not sses `FIXME(#1442)`).
- **`a2a_server/adcp_a2a_server.py`** — the `list_creatives` skill handler was **dropping `filtersded withdict→`CreativeFilters` coercion.                                           
**Tests**                                                                  
- **`test_uc018_list_creatives.py`** — wires the `@concept-id` storyboard  scenario (`BR-UC-018-lieeds creatives under the target concept plus **falsifiable decoys** (a different concept + a        no-concept creative); trned set is *exactly*the in-concept creatives and that each carries the requested `concept_id` +a non-empty `concept_naface a decoy and fail.
- **`tests/bdd/conftest.py`** — widened the UC-018 harness gate to also wir`@concept-id`.
- **`tests/bdd/steps/generic/when_request.py`** — extended the shared      `_call_via` helper to f. `filters`); existingcallers are unaffected.                                                    - **`tests/harness/crea_body` now carries`filters`.                                                                 
### Storage note                                                           
In adcp 5.7.0 the `sync_creatives` input carries no concept field (conceptsoriginate from external, so`concept_id`/`concept_name` live on the creative's JSON `data` blob and areseeded directly in the ng approach and the same DB row shape sync persists.                                                
### Testing                                                                
- `make quality` — **5138 passed**, 0 failures (incl. structural guards:   boundary completeness, tory pattern).
- `@concept-id` scenario passes on **a2a / mcp / rest** (IMPL is sunsetted in BDD per #1405).
- #1405 after-sync scenario (6 passed) and UC-005 `_call_via` consumers (47passed) verified green
                                                                           ### Out of scope / foll
                                                                           - The structured `filtetuses` are still onlyreported in `filters_applied` and not threaded to the query (same class of gap as concept was) — w `@format-id-object`sibling scenario (`:773`) depends on it.
- `ListCreativesBody` sME(#1442)` (extend`SalesAgentBaseModel`).

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -67,7 +67,7 @@ from src.core.exceptions import (
     normalize_to_adcp_error,
 )
 from src.core.resolved_identity import ResolvedIdentity
-from src.core.schema_helpers import to_account_reference
+from src.core.schema_helpers import coerce_creative_filters, to_account_reference
 from src.core.schemas import CreativeStatusEnum
 from src.core.tool_context import ToolContext
 from src.core.tool_error_logging import record_boundary_error
@@ -1651,11 +1651,9 @@ class AdCPRequestHandler(RequestHandler):
 
         # Structured AdCP CreativeFilters (statuses, concept_ids, format_ids, …)
         # arrive over the wire as a JSON dict; coerce to the typed model the core
-        # function expects so they are honoured rather than dropped.
-        from adcp import CreativeFilters
-
-        filters_param = parameters.get("filters")
-        filters = CreativeFilters.model_validate(filters_param) if isinstance(filters_param, dict) else filters_param
+        # function expects so they are honoured rather than dropped. Invalid filters
+        # raise AdCPValidationError (VALIDATION_ERROR + suggestion) via the shared helper.
+        filters = coerce_creative_filters(parameters.get("filters"))
 
         # Call core function with optional parameters (fixing original validation bug)
         response = core_list_creatives_tool(
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1649,6 +1649,14 @@ class AdCPRequestHandler(RequestHandler):
         # Create ToolContext from A2A auth info and resolve identity
         tool_context = self._make_tool_context(identity, "list_creatives")
 
+        # Structured AdCP CreativeFilters (statuses, concept_ids, format_ids, …)
+        # arrive over the wire as a JSON dict; coerce to the typed model the core
+        # function expects so they are honoured rather than dropped.
+        from adcp import CreativeFilters
+
+        filters_param = parameters.get("filters")
+        filters = CreativeFilters.model_validate(filters_param) if isinstance(filters_param, dict) else filters_param
+
         # Call core function with optional parameters (fixing original validation bug)
         response = core_list_creatives_tool(
             media_buy_id=parameters.get("media_buy_id"),
@@ -1658,6 +1666,7 @@ class AdCPRequestHandler(RequestHandler):
             created_after=parameters.get("created_after"),
             created_before=parameters.get("created_before"),
             search=parameters.get("search"),
+            filters=filters,
             page=parameters.get("page", 1),
             limit=parameters.get("limit", 50),
             sort_by=parameters.get("sort_by", "created_date"),
--- a/src/core/database/repositories/creative.py
+++ b/src/core/database/repositories/creative.py
@@ -87,6 +87,7 @@ class CreativeRepository:
         created_before: datetime | None = None,
         search: str | None = None,
         media_buy_ids: list[str] | None = None,
+        concept_ids: list[str] | None = None,
         sort_by: str = "created_date",
         sort_order: str = "desc",
         offset: int = 0,
@@ -117,6 +118,12 @@ class CreativeRepository:
 
         if format:
             stmt = stmt.where(Creative.format == format)
+
+        # v3.1 concept_ids filter: concepts group related creatives across sizes
+        # and formats (e.g. Flashtalking concepts, Celtra campaign folders). The
+        # concept identifier is stored on the creative's JSON data blob.
+        if concept_ids:
+            stmt = stmt.where(Creative.data["concept_id"].as_string().in_(concept_ids))
 
         if tags:
             for tag in tags:
--- a/src/core/schema_helpers.py
+++ b/src/core/schema_helpers.py
@@ -133,13 +133,13 @@ def coerce_creative_filters(filters: dict[str, Any] | CreativeFilters | None) ->
     A2A coerce identically (the MCP transport coerces via FastMCP's TypeAdapter on
     the tool signature).
 
-    Unlike the ``to_*`` converters above, an *invalid* structured filter is a loud
-    failure, not a silent ``None``: a malformed filter (e.g. ``concept_ids`` with an
-    empty array, violating the schema's ``minItems: 1``) is raised as an
-    ``AdCPValidationError`` carrying a recovery suggestion, so every transport
-    surfaces the spec's two-layer ``VALIDATION_ERROR`` envelope (with a suggestion,
-    per POST-F3) instead of a bare framework ``ValidationError`` that
-    ``normalize_to_adcp_error`` would flatten without a suggestion.
+    A malformed filter (e.g. ``concept_ids`` with an empty array, violating the
+    schema's ``minItems: 1``) is raised as a *typed* ``AdCPValidationError`` carrying
+    a recovery suggestion, so every transport surfaces the spec's two-layer
+    ``VALIDATION_ERROR`` envelope (with a suggestion, per POST-F3). Constructing the
+    model directly instead (as the ``to_*`` converters above do, via ``Model(**dict)``)
+    surfaces a raw pydantic ``ValidationError`` that ``normalize_to_adcp_error``
+    flattens into a suggestion-less envelope.
 
     Args:
         filters: Filters as a wire dict, an already-typed CreativeFilters, or None.
--- a/src/core/schema_helpers.py
+++ b/src/core/schema_helpers.py
@@ -13,7 +13,7 @@ Philosophy:
 from typing import Any
 
 # FIXME(#1388): GetProductsResponse, Product have local subclasses; import from src.core.schemas.
-from adcp import GetProductsResponse, Product
+from adcp import CreativeFilters, GetProductsResponse, Product
 
 # FIXME(#1388): ProductFilters has a local subclass; import from src.core.schemas.
 from adcp.types import (
@@ -24,8 +24,11 @@ from adcp.types import (
     PropertyListReference,
     ReportingWebhook,
 )
+from pydantic import ValidationError
 
+from src.core.exceptions import AdCPValidationError
 from src.core.schemas.product import GetProductsRequest
+from src.core.validation_helpers import format_validation_error
 
 
 def to_context_object(context: dict[str, Any] | ContextObject | None) -> ContextObject | None:
@@ -123,6 +126,41 @@ def to_property_list_reference(
     return None  # Fallback for unexpected types
 
 
+def coerce_creative_filters(filters: dict[str, Any] | CreativeFilters | None) -> CreativeFilters | None:
+    """Coerce a raw list_creatives filters value into a typed CreativeFilters.
+
+    Single source of truth for the dict -> CreativeFilters boundary so REST and
+    A2A coerce identically (the MCP transport coerces via FastMCP's TypeAdapter on
+    the tool signature).
+
+    Unlike the ``to_*`` converters above, an *invalid* structured filter is a loud
+    failure, not a silent ``None``: a malformed filter (e.g. ``concept_ids`` with an
+    empty array, violating the schema's ``minItems: 1``) is raised as an
+    ``AdCPValidationError`` carrying a recovery suggestion, so every transport
+    surfaces the spec's two-layer ``VALIDATION_ERROR`` envelope (with a suggestion,
+    per POST-F3) instead of a bare framework ``ValidationError`` that
+    ``normalize_to_adcp_error`` would flatten without a suggestion.
+
+    Args:
+        filters: Filters as a wire dict, an already-typed CreativeFilters, or None.
+
+    Returns:
+        CreativeFilters or None (when no filter was supplied).
+
+    Raises:
+        AdCPValidationError: when ``filters`` is a dict that fails CreativeFilters validation.
+    """
+    if filters is None or isinstance(filters, CreativeFilters):
+        return filters
+    try:
+        return CreativeFilters.model_validate(filters)
+    except ValidationError as e:
+        raise AdCPValidationError(
+            format_validation_error(e, context="list_creatives filters"),
+            suggestion="Fix the filters object — e.g. concept_ids must contain at least 1 concept id.",
+        ) from e
+
+
 def create_get_products_request(
     brief: str = "",
     brand: dict[str, Any] | BrandReference | str | None = None,
@@ -172,9 +210,11 @@ __all__ = [
     "to_brand_reference",
     "to_context_object",
     "to_reporting_webhook",
+    "coerce_creative_filters",
     "create_get_products_request",
     # Re-export types for type hints
     "BrandReference",
+    "CreativeFilters",
     "GetProductsRequest",
     "GetProductsResponse",
     "Product",
--- a/src/core/tools/creatives/listing.py
+++ b/src/core/tools/creatives/listing.py
@@ -157,6 +157,11 @@ def _list_creatives_impl(
     # Build structured objects
     structured_filters = LibraryCreativeFilters(**filters_dict) if filters_dict else None
 
+    # v3.1 concept_ids filter has no flat equivalent — it arrives only via the
+    # structured filters object and must be threaded into the DB query (not merely
+    # reported in filters_applied), or it would be silently dropped.
+    effective_concept_ids = structured_filters.concept_ids if structured_filters else None
+
     # Build pagination
     offset = (page - 1) * effective_limit
     # 3.6.0: PaginationRequest is cursor-based (max_results, cursor). DB query uses offset/limit internally.
@@ -211,6 +216,7 @@ def _list_creatives_impl(
             created_before=created_before_dt,
             search=search,
             media_buy_ids=effective_media_buy_ids or None,
+            concept_ids=effective_concept_ids,
             sort_by=sort_by,
             sort_order=valid_sort_order,
             offset=offset,
@@ -299,6 +305,10 @@ def _list_creatives_impl(
                 status=status_enum,
                 created_date=created_at_dt,
                 updated_date=updated_at_dt,
+                # v3.1 concept grouping (Flashtalking/Celtra/CM360-style), stored on
+                # the JSON data blob; omitted from the response when absent.
+                concept_id=db_creative.data.get("concept_id") if db_creative.data else None,
+                concept_name=db_creative.data.get("concept_name") if db_creative.data else None,
                 # Internal field (our extension)
                 principal_id=db_creative.principal_id,
             )
@@ -319,6 +329,8 @@ def _list_creatives_impl(
             filters_applied.append(f"format_ids={','.join(str(f) for f in req.filters.format_ids)}")
         if req.filters.tags:
             filters_applied.append(f"tags={','.join(req.filters.tags)}")
+        if req.filters.concept_ids:
+            filters_applied.append(f"concept_ids={','.join(req.filters.concept_ids)}")
         if req.filters.created_after:
             filters_applied.append(f"created_after={req.filters.created_after.isoformat()}")
         if req.filters.created_before:
--- a/src/core/tools/creatives/listing.py
+++ b/src/core/tools/creatives/listing.py
@@ -295,6 +295,24 @@ def _list_creatives_impl(
                 # Default to pending_review if invalid status
                 status_enum = CreativeStatus.pending_review
 
+            # v3.1 concept grouping (Flashtalking/Celtra/CM360-style), stored on the
+            # JSON data blob, omitted from the response when absent.
+            #
+            # Population is out-of-band by design: the AdCP sync_creatives request
+            # carries no concept field (concepts are seller/ad-server-assigned, not
+            # buyer-supplied), so concept_id/concept_name are written into data[] by
+            # the ingesting integration (e.g. an adapter mapping the ad server's
+            # native creative groups). Mapping GAM creative groups -> these fields is
+            # a separate, adapter-scoped follow-up; until then the field is populated
+            # only by external producers writing to the blob.
+            #
+            # The blob is untyped and external producers may write numeric ids (CM360
+            # group ids), so coerce to the spec's string type instead of letting a
+            # non-string value fail Creative validation and crash the whole listing.
+            concept_data = db_creative.data or {}
+            concept_id_raw = concept_data.get("concept_id")
+            concept_name_raw = concept_data.get("concept_name")
+
             creative = Creative(
                 creative_id=db_creative.creative_id,
                 name=db_creative.name,
@@ -305,10 +323,8 @@ def _list_creatives_impl(
                 status=status_enum,
                 created_date=created_at_dt,
                 updated_date=updated_at_dt,
-                # v3.1 concept grouping (Flashtalking/Celtra/CM360-style), stored on
-                # the JSON data blob; omitted from the response when absent.
-                concept_id=db_creative.data.get("concept_id") if db_creative.data else None,
-                concept_name=db_creative.data.get("concept_name") if db_creative.data else None,
+                concept_id=str(concept_id_raw) if concept_id_raw is not None else None,
+                concept_name=str(concept_name_raw) if concept_name_raw is not None else None,
                 # Internal field (our extension)
                 principal_id=db_creative.principal_id,
             )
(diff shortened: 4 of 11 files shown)


ChrisHuie (commented):
Reviewed at `dc185feb4`. The `concept_ids` filter is wired through MCP/A2A/REST into a parameterized JSONB predicate (`creative.py:126`), the response exposes `concept_id`/`concept_name` (`listing.py:310-311`), and the `@concept-id` BDD scenario is genuinely falsifiable — mutating the predicate reddens all three transports. The spec contract matches: `creative-filters.json` defines `concept_ids` (array, `minItems` 1) and `list-creatives-response.json` defines scalar `concept_id`/`concept_name`. Tenant + principal isolation holds (verified: a second principal and a second tenant sharing the same concept are not returned).

Nothing below is a happy-path correctness bug; the SHOULD-FIX items are robustness and reachability.

## SHOULD-FIX

**1. A non-string `concept_id`/`concept_name` in `data` crashes the entire `list_creatives`.**
`listing.py:310-311` pass the raw JSON blob value into the response `Creative`, where both fields are strictly `str | None`. The predicate `CAST(data->>'concept_id' AS VARCHAR)` text-coerces and matches a numeric value, so the row is returned — then response construction raises `ValidationError` and the whole call fails, even with no filter, on all three transports, and under `ENVIRONMENT=production` (`extra="ignore"` does not bypass declared-field typing). Repro: one creative with `data={"concept_id": 12345}` makes `list_creatives` return `VALIDATION_ERROR`/`correctable` instead of the library. Concept data is external-origin (CM360/Flashtalking/Celtra), where numeric IDs are common, so this is in the feature's own data path.

```python
_data = db_creative.data or {}
_cid = _data.get("concept_id")
_cname = _data.get("concept_name")
# ...
concept_id=str(_cid) if _cid is not None else None,
concept_name=str(_cname) if _cname is not None else None,
```

(or normalize the type at the ingestion point.)

**2. No production path writes `concept_id`.**
The filter and the two response fields read a key nothing in `src/` populates — `_build_creative_data` (the `sync_creatives` data builder) has a fixed allowlist with no concept field, and the adcp 5.7.0 `sync_creatives` input carries none. The feature is correct per spec but returns empty / absent in production until an external producer writes the blob. Worth stating in the PR description and filing the producer as a follow-up, so the BDD (which seeds the blob directly) isn't read as end-to-end coverage.

**3. The dict→`CreativeFilters` coercion is duplicated.**
`api_v1.py:341` (`… if body.filters else None`) and `adcp_a2a_server.py:1658` (`… if isinstance(filters_param, dict) else …`) implement the same boundary coercion two slightly different ways; MCP gets it from FastMCP. Extract one helper shared by both wrappers — also the natural home for the error handling in (4).

**4. No error-path test for the coercion.**
A malformed `filters` (e.g. `{"concept_ids": []}`, violating `minItems` 1) surfaces as `VALIDATION_ERROR`/400 on REST and A2A today (pydantic `ValidationError` is a `ValueError`, caught by the global handler) — correct, but unpinned. Add a per-transport wire-envelope assertion so a change to the boundary normalization can't silently regress it.

## NIT

- `_serialized_response` (`test_uc018_list_creatives.py:155`) asserts on `model_dump()` of the typed payload, not the captured `wire_response`. The wire carries both fields on all three transports today, but `model_dump` and the MCP wire (`to_jsonable_python`) can diverge; `uc005_format_id_shape.py` reads `wire_response` with a guard for exactly this. The docstring's "same bytes the wire produces" is accurate for REST/A2A only.
- `from adcp import CreativeFilters` is function-local at `api_v1.py:339` and `adcp_a2a_server.py:1655` but not load-bearing (`adcp` is module-level in both; `listing.py` keeps it module-level). Hoist.
- `creative_list.py:83` `hasattr(filters, "model_dump")` model branch has no caller — every caller passes a dict.

## Out of scope (pre-existing, not this PR)

- MCP returns a raw FastMCP `ToolError` (no AdCP code) when the filter *shape* is rejected at the TypeAdapter layer, where REST/A2A return `VALIDATION_ERROR` (reproduces on `statuses`/`creative_ids` at the parent commit).
- `format_ids`/`statuses` are reported in `filters_applied` but not threaded to the query — the same gap this PR closes for `concept_ids`.
- `concept_ids` has no `MaxLen` (upstream SDK; `creative_ids` caps at 100).
- The cross-principal isolation scenario (`BR-RULE-034 INV-1`) is xfail-stranded, so `list_creatives` principal-scoping has no live regression test (the scoping itself is correct).

## Verification performed (at `dc185feb4`)

- Executed across MCP/A2A/REST: exact in-concept set (decoys excluded); multi-concept = OR; NULL/missing concept excluded; exact (non-prefix) match; `total_count` reflects the filtered set under pagination; `status` + `concept_ids` combine; `concept_id`/`concept_name` present on the wire; `filters_applied` carries `concept_ids=…`.
- Cross-principal and cross-tenant isolation: executed.
- Malformed-filter and non-string-concept error paths: executed on all three transports.
- JSONB predicate compiled (bound parameters, not interpolated).
- Spec contract checked against the pinned vendored schemas; SDK 5.7.0 field parity exact.
- Not run: full `run_all_tests.sh ci` suite (CI is green on 32 checks); list/dict (non-scalar) `concept_id` shapes.

mkomorski:
Thanks Chris — all four SHOULD-FIX addressed in 89c8a449.

**1. Numeric concept_id crash** ✅ Confirmed and fixed. `Creative(concept_id=12345)` raised a `string_type` ValidationError and failed the whole listing. Now `str()`-coerced when building the response (consistent with the filter side — Postgres `->>` already returns the text form, so a buyer filtering `concept_ids=["12345"]` matches).

**3. Duplicated coercion** ✅ Extracted `coerce_creative_filters()` in `schema_helpers.py`; REST and A2A both call it.

**4. Missing error-path test** ✅ Added `tests/integration/test_list_creatives_concept_filter.py`. This also surfaced a real bug: the old raw `model_validate` path produced `VALIDATION_ERROR` *without* a `suggestion` (POST-F3 gap), because `normalize_to_adcp_error` flattens a bare Pydantic `ValidationError`. The shared helper now raises `AdCPValidationError` with a suggestion, so REST/A2A emit the full two-layer envelope. Tests assert rejection on a2a/mcp/rest and the full envelope + suggestion on REST/A2A.

- Scope note on MCP: `filters` is typed `CreativeFilters` (per the wrapper-typed-params guard), so FastMCP's TypeAdapter rejects `concept_ids=[]` *before* the tool body runs — a raw input error, not an `AdCPError` the boundary wraps into the envelope. Translating FastMCP TypeAdapter errors into the two-layer envelope is a pre-existing, tool-agnostic MCP concern (it affects every typed param, not just concept), so I filed it as a follow-up rather than expanding this PR; the test still pins that MCP rejects.

**2. Production population** — This is by design per AdCP 3.1: `sync-creatives-request.json` has no concept field (concepts are seller/ad-server-assigned — CM360/Celtra/Flashtalking groups), so a buyer can't set it via `sync_creatives`. `concept_id`/`concept_name` are populated out-of-band into the creative `data` blob by the ingesting integration. I documented this in `listing.py` and am filing a follow-up to map GAM creative groups → these fields. I'd push back on inventing a sync field here — it would diverge from the spec.

NITs all addressed: concept assertions now read `ctx["wire_response"]` (real wire bytes); the `CreativeFilters` import is hoisted via the helper; the dead `hasattr` branch is removed from `build_rest_body`.

Follow-ups I'll open as separate issues: (1) GAM creative-group → concept_id/name mapping during ingestion; (2) MCP TypeAdapter input-validation → two-layer `VALIDATION_ERROR` envelope at the MCP boundary.

ChrisHuie:
Round-2 review of `89c8a449`. The four SHOULD-FIX and the NITs from the prior review are addressed; specifics confirmed below by running each path on a2a/mcp/rest. Two items remain (both SHOULD-FIX, neither a correctness bug), plus one spec-grounding clarification on the concept model that matters for the planned GAM follow-up.

## Confirmed fixed in `89c8a449`

- **Numeric `concept_id` crash** — `str()`-coercing `concept_id`/`concept_name` when building the response resolves it. Verified by execution on all three wire transports: a creative with `data={"concept_id": 999, "concept_name": 777}` now lists successfully with the wire carrying `concept_id="999"` (string), where before it raised a `string_type` `ValidationError` and failed the whole listing — filtered *or* unfiltered, and under `ENVIRONMENT=production` too. The filter side (`->>` text-coercion) and the response side (`str()`) now normalize consistently, so a numeric blob value both matches `concept_ids=["999"]` and projects as `"999"`.
- **Duplicated coercion** — `coerce_creative_filters()` in `schema_helpers.py` is now the single source; REST and A2A both delegate; MCP correctly does not duplicate (FastMCP's TypeAdapter coerces the typed param). The only `CreativeFilters.model_validate` in `src/` is now the helper, and the duplication baseline is unchanged.
- **Error-path test + POST-F3** — the new `test_list_creatives_concept_filter.py` pins rejection on a2a/mcp/rest and the two-layer `VALIDATION_ERROR`/`correctable` envelope with a `suggestion` on REST/A2A. The POST-F3 point is correct and worth stating precisely: pydantic's `ValidationError` subclasses `ValueError`, so the old bare `model_validate` path was caught by the global `ValueError` handler → `normalize_to_adcp_error` → `AdCPValidationError(str(exc))` with `suggestion=None`. Routing through the helper (which raises `AdCPValidationError(..., suggestion=...)`) closes that gap. Confirmed by mutation: removing the `suggestion=` kwarg reddens the new test on a2a and rest.
- **NITs** — the concept Then-steps now read `ctx["wire_response"]` (real wire bytes); the lazy import is hoisted into the helper; the dead `hasattr` branch is gone from `build_rest_body`.

The empty-`{}` filter case is unchanged on the wire even though the helper now returns an empty `CreativeFilters()` rather than `None`: `_merge_structured_filters` collapses an empty model's `model_dump(exclude_none=True)` to `{}`, so `req.filters` resolves to `None` and `filters_applied` to `[]` either way (verified).

## Spec clarification — the concept-ownership model (matters for the GAM follow-up)

The pushback that a sync-side concept field would diverge from the spec is correct — AdCP 3.1.0-beta.3 has no concept field on `sync-creatives-request.json`, `creative-asset.json`, or `creative-manifest.json`. The prior review did not propose adding one; it recommended an out-of-band producer + documentation, which is what you did.

One framing point in the code comment is worth correcting, because it's the stated basis for the planned GAM mapping. The comment (`listing.py:302`) says *"concepts are seller/ad-server-assigned, not buyer-supplied."* The spec points the other way:

- The creative library is **buyer-populated** — `list_creatives` is "implemented by any agent that hosts a creative library … and sales agents that manage creatives," and key-concepts has "Creative Manifest (by Buyer)" / "Buyers create manifests."
- The three `concept_ids` examples — Flashtalking concepts, Celtra campaign folders, CM360 creative groups — are all **buyer/agency-side** creative platforms.
- The only ad-server reference in the creative model is the seller's **downstream delivery** ("sales agents deliver validated creatives to their ad servers"), which is unrelated to concept grouping.

The spec is genuinely silent on *who assigns* a concept, so this isn't a contradiction of a single line — but the weight of the evidence is that a concept is a **buyer-side** grouping that AdCP simply doesn't yet carry on sync. That makes "map GAM creative groups → `concept_id`" a **seller-side enrichment / fallback**, not the authoritative concept the spec describes. Wording the comment (and the follow-up) that way avoids the producer baking in "GAM groups *are* the concept." The read/filter code in this PR is spec-correct regardless; this is about the rationale and the next step.

## SHOULD-FIX

**1. The `str()` coercion has no regression test.** It's the headline fix of this commit, but reverting `str(concept_id_raw)` to the raw value reddens zero tests — every concept fixture seeds string ids, and the new test covers only the empty-array error path. A future refactor could silently drop the coercion and reintroduce the crash with a green suite. This is a clean fold-in (same scope), not a follow-up. The test below passes on the current fix and fails if the coercion is reverted (verified both directions):

```python
# tests/integration/test_list_creatives_concept_filter.py
class TestNumericConceptCoercion:
    """A numeric concept_id/concept_name in the data blob (CM360-style group ids)
    is coerced to the spec's string type, not crashed on."""

    def test_numeric_concept_id_is_coerced_to_string(self, integration_db):
        with CreativeListEnv() as env:
            tenant = TenantFactory(tenant_id=env._tenant_id)
            principal = PrincipalFactory(tenant=tenant, principal_id=env._principal_id)
            CreativeFactory(
                tenant=tenant, principal=principal, format="display_300x250",
                status="approved", data={"assets": {}, "concept_id": 12345, "concept_name": 678},
            )
            result = env.call_via(Transport.REST)
            assert not result.is_error, f"listing errored on a numeric concept_id: {result.error!r}"
            creative = result.wire_response["creatives"][0]
            assert creative["concept_id"] == "12345"
            assert creative["concept_name"] == "678"
```

**2. Re-ground the concept comment in the spec** (per the section above). Replace the "seller-assigned" causal claim with what the spec actually says:

```python
# v3.1 concept grouping. AdCP exposes concept_id/concept_name on the
# list_creatives RESPONSE (a creative's concept membership, sourced from the
# buyer's creative-management platform — Flashtalking/Celtra/CM360), but
# standardizes no concept INPUT on sync_creatives, so the field is populated
# out-of-band into the data blob. The blob is untyped and an external producer
# may write numeric group ids, so coerce to the spec's string type rather than
# letting a non-string value fail Creative validation and crash the listing.
```

## NIT

- `_wire_creatives` (`test_uc018_list_creatives.py`) falls back to `_serialized_response` (a `model_dump` of the typed payload) when no wire is captured. For the three wired transports this never fires (all stash), but the sibling `uc005_format_id_shape.py` uses a loud guard so a future non-stashing transport trips instead of silently asserting on a re-serialization. Mirroring that guard keeps the docstring's "real wire bytes" claim honest.
- In the new test, `body = envelope.envelope if hasattr(envelope, "envelope") else envelope` (line 83) — `wire_error_envelope` is always a dict here, so the `hasattr` branch is dead; `body = envelope` is enough (the same dead branch you removed from `build_rest_body`). The accessor is load-bearing in `assert_envelope_shape` because that helper also accepts an `AdCPToolError`; here it doesn't.
- `coerce_creative_filters`'s docstring says the `to_*` converters return "silent None" on invalid input — they actually raise a raw pydantic `ValidationError` on a malformed dict and return `None` only on an unexpected type. The commit message states the real distinction correctly (typed `AdCPValidationError` + suggestion vs a flattened framework error); the docstring just needs the same precision.
- `str()` on a non-scalar blob value (a list/dict) would stringify its repr rather than surface corrupt data — borderline against the "No Quiet Failures" pattern. Low-stakes (the only writer is out-of-band ingestion and the field is scalar by spec); coercing only scalars (`str(x) if isinstance(x, (str, int)) else None`) would tighten it if you want.

## Verification

- Executed on a2a/mcp/rest at `89c8a449`: numeric `concept_id` coerced (no crash); empty `concept_ids` rejected with `VALIDATION_ERROR`+suggestion on REST/A2A and raw rejection on MCP; multi-concept OR, cross-principal and cross-tenant isolation, exact-match (non-prefix), and `total_count`/pagination under the filter all hold.
- New error-path test (5) and the `@concept-id` BDD scenario (3, asserting on wire bytes) pass; full unit suite green; all structural guards pass with no allowlist/baseline change; rebase onto current main is clean (disjoint file sets).
- The proposed coercion test above was run both ways: passes on the fix, fails on revert.
- Spec checked against the vendored pinned schemas (adcp `04f59d2d5` / 3.1.0-beta.3) + the key-concepts prose. Not checked: the conformance storyboard (whether concept filtering is graded upstream).

mkomorski:
Thanks for the thorough round-2 pass, Chris — all addressed in b8320a21.

**SHOULD-FIX 1 — str() coercion had no regression test.** Added `TestNumericConceptCoercion` (essentially your snippet): seeds `concept_id=12345`/`concept_name=678`, asserts the wire carries `"12345"`/`"678"`. Falsifiable both ways — reverting the coercion crashes the listing mid-build (`is_error`), so it reddens.

**SHOULD-FIX 2 — re-ground the concept comment.** You're right, and the correction matters for the follow-up framing. I adopted your wording: concept is a **buyer-side** grouping (Flashtalking/Celtra/CM360 are buyer/agency platforms) that AdCP exposes on the `list_creatives` RESPONSE but standardizes no INPUT for on `sync_creatives`; the field is populated out-of-band. I also reframed the GAM mapping in-comment as a **seller-side enrichment/fallback, not the authoritative concept** — and I'll word the follow-up issue the same way so the producer doesn't bake in "GAM groups *are* the concept."

**NITs**
- `_wire_creatives` now carries the same loud guard as `uc005_format_id_shape` — a real-wire transport that didn't stash `wire_response` trips instead of silently falling back to a `model_dump` re-serialization, keeping the "real wire bytes" claim honest.
- Dropped the dead `hasattr(envelope, "envelope")` branch in the new test — `wire_error_envelope` is always a dict there (the `AdCPToolError` accessor is load-bearing only inside `assert_envelope_shape`).
- Tightened the `coerce_creative_filters` docstring to your precision: typed `AdCPValidationError` + suggestion vs a raw pydantic `ValidationError` flattened suggestion-less — not "silent None".
- Non-scalar `str()` — extracted `_coerce_concept_value`: scalars → `str`; a non-scalar (list/dict) blob value is dropped to `None` **with a `logger.warning`** rather than projecting a repr or crashing the listing. That keeps it observable (No Quiet Failures) without nuking the listing on one corrupt row.

Thanks also for confirming the empty-`{}` collapse and re-running each path — the `_merge_structured_filters` `exclude_none` → `{}` → `req.filters=None` behavior is what I was relying on for that case.

Follow-ups I'll open (worded per your spec note): (1) GAM creative-group → `concept_id`/`name` as a **seller-side enrichment**; (2) MCP TypeAdapter input-validation → two-layer `VALIDATION_ERROR` envelope at the boundary.

mkomorski:
Opened the two follow-ups from review (worded per the spec note — concept is buyer-side; GAM is seller-side enrichment, not authoritative):

- #1506 — populate `concept_id`/`concept_name` from GAM creative groups (seller-side enrichment)
- #1507 — MCP TypeAdapter input-validation → two-layer `VALIDATION_ERROR` envelope (pre-existing, tool-agnostic)

Referenced both from the code in 22e83a79 (`listing.py` → #1506; `test_list_creatives_concept_filter.py` → #1507). Nothing else outstanding from the round-2 review.

ChrisHuie:
Round-3 review of `22e83a79` — all six round-2 items confirmed addressed, verified by running each path on a2a/mcp/rest: the numeric-coercion test is a real regression oracle, the concept comment is re-grounded accurately (buyer-side concept; GAM as enrichment, not authoritative), `_wire_creatives` is now a loud guard (reddens when a real-wire transport doesn't stash), the dead `hasattr` is gone, the `coerce_creative_filters` docstring is accurate, and `_coerce_concept_value` drops non-scalars with a warning rather than crashing — the No-Quiet-Failures path. CI green, guards pass. One item remains; it's the symmetric completion of the round-2 ask.

**SHOULD-FIX — the non-scalar drop branch has no test.** `_coerce_concept_value` introduced two behaviors; only the scalar one is pinned. `TestNumericConceptCoercion` covers `int → "12345"`, but the non-scalar branch (`list`/`dict` → `None` + `logger.warning`, whose docstring says it "prevents crashing the whole listing on one bad row") reddens no test if removed — revert `return None` to a passthrough and a malformed row 500s the listing again with a green suite. Same gap we closed for the scalar path one round ago, left half-swept.

This passes on the current code and fails if the drop branch is reverted (verified both directions; isolated — mutating only the non-scalar branch reddens it while the numeric test stays green):

```python
class TestNonScalarConceptValueDropped:
    """A non-scalar concept_id/concept_name in the data blob (corrupt external value)
    is dropped to null, not projected as a repr and not crashed on — regression guard
    for the _coerce_concept_value non-scalar branch."""

    def test_non_scalar_concept_value_is_dropped(self, integration_db):
        from tests.factories import CreativeFactory

        with CreativeListEnv() as env:
            tenant, principal = _seed_authenticated_principal(env)
            CreativeFactory(
                tenant=tenant, principal=principal, format="display_300x250",
                status="approved", data={"assets": {}, "concept_id": ["x"], "concept_name": {"k": "v"}},
            )
            result = env.call_via(Transport.REST)
            assert not result.is_error, f"non-scalar concept value crashed the listing: {result.error!r}"
            creative = result.wire_response["creatives"][0]
            assert "concept_id" not in creative   # dropped → exclude_none omits it
            assert "concept_name" not in creative
```

(A `caplog` assertion on the "Dropping non-scalar concept value" warning would additionally pin the observability claim, but the `is_error` + absence checks are the load-bearing oracle.)

**Pre-existing, not this PR — `tags` has the same hazard.** One line above the concept calls, `tags=db_creative.data.get("tags")` passes a raw blob value into `Creative.tags: list[str] | None` — the identical untyped-blob→typed-field crash the new helper prevents for concept, on a field that predates this PR. A malformed `tags` blob (a string, or `[1, 2]`) would 500 the whole listing the same way a numeric `concept_id` did. Out of scope here; worth a follow-up (the `_coerce_concept_value` pattern generalizes).

Nothing else outstanding. The `#1506`/`#1507` comment refs are consistent with the repo's existing `FIXME(#…)` style — no change needed.

mkomorski:
Thanks Chris — last item closed in 577820d7.

**SHOULD-FIX — non-scalar drop branch untested.** Added `TestNonScalarConceptValueDropped` (your snippet + a `caplog` assertion): seeds `concept_id=["x"]`/`concept_name={"k":"v"}`, asserts the listing succeeds, both keys are omitted from the wire (dropped → `exclude_none`), and the "Dropping non-scalar concept value" warning fires — so the observability half of the No-Quiet-Failures claim is now a regression oracle too, not just the drop. Falsifiable both ways; isolated from the numeric test (mutating only the non-scalar branch reddens it while the numeric test stays green). 7/7 in the concept-filter file, full suite green.

**`tags` hazard (out of scope, agreed).** Confirmed identical: `tags=db_creative.data.get("tags")` → `Creative.tags: list[str]`, so a bare string or `[1, 2]` 500s the whole listing. Filed #1508 and left a `FIXME(#1508)` on the line. Noted there that `tags` is a *list* field, so it needs a list-of-str coercion rather than `_coerce_concept_value` directly — a small shared blob-coercion helper (scalar + list variants) would cover both.

Nothing else outstanding.

ChrisHuie (approved):

Issue #1506, opened 2026-06-30
feat: populate concept_id/concept_name from GAM creative groups (seller-side enrichment) — follow-up to #1407


mkomorski:
## Context

AdCP 3.1 `list_creatives` exposes `concept_id` / `concept_name` on each creative (concepts group related creatives across sizes/formats). PR #1493 (issue #1407) wired the `concept_ids` filter and the response fields, reading them from the creative's JSON `data` blob.

There is **no production writer** for those keys today: the AdCP `sync_creatives` request carries no concept field (verified against `sync-creatives-request.json` / `creative-asset.json` / `creative-manifest.json` in adcp 3.1.0-beta.3), and no adapter populates them. So in production the fields are absent until an external producer writes the blob.

## Spec framing (important)

Per the spec, a concept is a **buyer-side** grouping — the creative library is buyer-populated and the `concept_ids` examples (Flashtalking concepts, Celtra campaign folders, CM360 creative groups) are all buyer/agency-side creative platforms. AdCP simply doesn't yet carry a concept on `sync_creatives`.

A GAM creative-group mapping is therefore a **seller-side enrichment / fallback**, *not* the authoritative buyer-side concept. The implementation must not bake in "GAM creative groups *are* the concept".

## Task

During GAM adapter creative ingestion, map the ad server's native creative grouping → `data["concept_id"]` / `data["concept_name"]` as a seller-side enrichment, so the v3.1 `concept_ids` filter and response fields have a production data source.

- Source: GAM creative groups (or the closest native grouping).
- Write into the creative `data` blob (the read/filter path already consumes it — `src/core/tools/creatives/listing.py`, `repositories/creative.py`).
- Document it as enrichment/fallback, distinct from a (future, spec-defined) buyer-supplied concept.

Follow-up from PR #1493 review (ChrisHuie). Read/filter side is already spec-correct; this only adds a producer.

mkomorski:
Completed via #1509 (merged 2026-07-07, approved by @ChrisHuie).

The issue stayed open only because #1509's body referenced this issue in the title but carried no `closes #1506` keyword, so GitHub never auto-closed it.

Everything in the task is on `main` and verified green:

- **Producer**: the GAM adapter derives a concept from the GAM Order (closest native grouping) during creative push and folds `concept_id`/`concept_name` into the creative `data` blob (`src/adapters/gam/managers/creatives.py`).
[...]

Issue #1508, opened 2026-07-01
fix: apply defensive blob coercion to Creative.tags in list_creatives (untyped-blob crash) — follow-up to #1407


mkomorski:
## Context

`_list_creatives_impl` builds the response `Creative` from the untyped JSON `data` blob. PR #1493 added `_coerce_concept_value` so a corrupt/non-string `concept_id`/`concept_name` in the blob is coerced (scalar → str) or dropped (non-scalar → `None` + warning) instead of crashing the whole `list_creatives` on one bad row.

The `tags` field one line above has the **identical untyped-blob → typed-field hazard** and predates this PR:

```python
# src/core/tools/creatives/listing.py
tags=db_creative.data.get("tags") if db_creative.data else None,
```

`Creative.tags` is typed `list[str] | None`. A malformed blob value — a bare string, or `[1, 2]` (numeric elements) — raises a pydantic `ValidationError` (`tags` / `tags.0`) during response construction and 500s the entire listing (`VALIDATION_ERROR`), filtered or unfiltered, on all transports. Verified against the current schema.

## Task

Apply the same defensive coercion pattern to `tags`. Note `tags` is a **list** field, so it needs a list-of-str coercion, not `_coerce_concept_value` directly:

- Non-list blob → drop to `None` (or coerce a bare string to a single-element list, TBD) with a `logger.warning`.
- List with non-str elements → stringify scalars, drop non-scalar elements, with a warning.
- Add a regression test mirroring `TestNonScalarConceptValueDropped` (a malformed `tags` blob lists without crashing; the drop is logged).

Generalizing `_coerce_concept_value` into a shared "coerce untyped blob value to a typed field, dropping+logging corrupt data" helper (scalar and list variants) would cover both fields.

Follow-up from PR #1493 review (ChrisHuie); pre-existing, out of scope for #1407.

Issue #72, opened 2025-08-17
AEE config


bokelley:
When we set up the ad server we need to set up targeting keys for the AEE fields, defaulting to aee_include, aee_exclude, aee_macro. We could show the user a list of the keys that exist in GAM and give them the option to select existing keys or set up new ones (either manually or automatically).

These should also be configurable when we create a new tenant via the super admin API.

PR #1509, opened 2026-07-01
feat: populate concept_id/concept_name from GAM order (#1506)

mkomorski:
## Summary

Follow-up to #1407 (PR #1493). AdCP 3.1 exposes read-only `concept_id` / `concept_name` on `list_creatives` (concepts group related creatives across sizes/formats), but AdCP `sync_creatives` carries **no** concept field — so #1407 wired the reader/filter with **no production writer** for those keys.

This PR adds that writer as a **seller-side enrichment/fallback**: during GAM creative push, the adapter derives a concept from the GAM Order (its closest native creative grouping) and folds it into the creative `data` blob, giving the `concept_ids` filter a production data source.

Per the spec, a concept is a **buyer-side** grouping and AdCP doesn't yet carry one on `sync_creatives`. A GAM order mapping is therefore explicitly a fallback — **not** the authoritative buyer concept. The implementation keeps that distinction:

- `concept_id` is namespaced `gam-order-<order_id>` (self-evidently seller-derived).
- `concept_source = "gam_order"` records provenance.
- The enrichment is written **only when no concept is already present** — a future buyer-supplied concept always takes precedence and is never overwritten.

> **Base:** `main`. The #1407 reader/filter this producer feeds merged to `main` as a squash (#1493), so this PR carries only the producer delta.

## Spec grounding

- Pin: `adcp==5.7.0` → AdCP **3.1.0-beta.3**.
- `list_creatives_response.Creative` has `concept_id: str | None` / `concept_name: str | None` ("Creative concept this creative belongs to. Concepts group related creatives across sizes and formats.").
- `sync_creatives_request` has no concept field, and `sync_creatives_response.Creative` has none → concept is read-only on the wire, no protocol writer.
- **Conformance:** at `adcp 5.7.0` / spec 3.1.0-beta.3 the only graded `list_creatives` artifact is the creative-ad-server **`browse_library`** scenario, which validates the list-creatives-response **shape** (`response_schema_ref`) — `concept_id`/`concept_name` are optional strings there. Neither the `concept_ids` filter nor the `concept_id` value/source is a graded `check` step. So the seller-derived source is **ungraded**; the response shape it feeds is covered by the response schema + the internal `BR-UC-018` BDD.

## Changes

| File | Change |
|------|--------|
| `src/core/schemas/_base.py` | `AssetStatus` gains optional `concept_id` / `concept_name` / `concept_source` (internal DTO, not a wire type) |
| `src/adapters/gam/managers/creatives.py` | `add_creative_assets` tags approved statuses with `gam-order-<id>` + `concept_source="gam_order"` |
| `src/core/tools/media_buy_create.py` | `_merge_creative_enrichment()` (pure merge) + `_apply_creative_enrichment()` (merge + provenance log + change detection). **All three** GAM push paths — auto-approval create, manual-approval execution (`execute_approved_media_buy`), and retroactive per-creative push — fold in `platform_creative_id` + concept **fill-only-when-absent** and persist through one path (`CreativeRepository.update_data`) |
| `docs/adapters/gam/README.md` | documents the seller-side fallback, precedence, and coverage |

The read/filter path is unchanged — this only adds a producer.

## Limitations

- Only creatives **pushed to GAM** (trafficked into a media buy) are enriched — now uniformly across all three push paths; library-only creatives carry no concept until trafficked.
- The Order is a coarse grouping (creatives reused across orders, or unrelated creatives in one order, won't map cleanly) — it's a best-effort fallback. A semantically-truer source (GAM CreativeSet) is noted as a possible future refinement.

## Tests

- **Producer** (`test_gam_validation_integration.py`): approved creatives carry the order-derived concept; failed creatives carry none.
- **Precedence** (`test_creative_enrichment_merge.py`): fills when absent, never overwrites a buyer concept or existing `platform_creative_id`, defaults `concept_source`, doesn't mutate input.
- **Writeback chain** (`test_execute_approved_platform_ids.py::TestExecuteApprovedEnrichesSellerConcept`): the **real** dry-run GAM producer → **real** `execute_approved_media_buy` writeback → DB → **real** `CreativeRepository` `concept_ids` filter. Proves the manual-approval path enriches (previously it did not) and pins producer↔reader key alignment end-to-end.
- **Writeback (retroactive path)** (`test_push_creative_to_existing_buy.py`): a concept-bearing `AssetStatus` is persisted via `update_data` (concept branch of that writeback).
- **Merge-helper → filter** (`test_list_creatives_concept_filter.py`): the merge helper's output flows through the #1407 `concept_ids` filter to the wire response; pins that `concept_source` never reaches the wire.
- `make quality` green; changed integration suites green.


ChrisHuie:
Reviewed at `ed6fe4f04`. Verdict: the design is sound and spec-correct — the premise holds against `adcp==5.7.0` / spec 3.1.0-beta.3 (`sync_creatives` carries no concept field, so a seller-side fallback is the right shape), and the namespaced id + `concept_source` provenance marker + fill-only-when-absent precedence are the right calls. `make quality` is green locally (5148 passed) and the merge helper's unit coverage is thorough. The items below are completeness plus one packaging step before this lands on `main`.

## Before merge: rebase onto `main` and retarget

The base PR merged to `main` as a **squash**, so its commits aren't in `main`'s history and the `feat/1407-list-creatives-concept-ids` branch this PR targets is still present. As a result the PR still shows a base of `feat/1407`, and its diff against `main` is 7 commits / 18 files (it re-includes the already-merged reader). Rebasing onto `main` drops the redundant commits and leaves just the producer delta:

```bash
git fetch origin
git rebase --onto origin/main origin/feat/1407-list-creatives-concept-ids
git push --force-with-lease
```

Then retarget the PR base to `main`. Verified locally: this replays cleanly to the intended 8-file / +327/−23 changeset. It matters for two reasons — the heavy suite (`ci.yml`) and CodeQL only trigger for PRs based on `main`/`develop`, so no test suite has actually run on this branch yet; and it keeps the merged history clean.

## Findings

### 1. Enrichment is dropped on one of the three GAM push paths (recommended before merge)

There are three call sites for `add_creative_assets`; the writeback is wired into two of them, but not the third:

| Path | Site | Writeback |
|---|---|---|
| Auto-approval create (`_create_media_buy_impl`) | `media_buy_create.py:3830` | ✅ |
| Per-creative retroactive push (`push_creative_to_existing_buy`) | `media_buy_create.py:1236` | ✅ |
| **Manual-approval execution (`execute_approved_media_buy`)** | **`media_buy_create.py:1076`** | ❌ statuses used only for failure logging (1082–1086) |

`execute_approved_media_buy` is a live path — it runs when a pending media buy is approved from the admin UI (`operations.py:437`, `workflows.py:233`, `creatives.py:648`). Creatives trafficked that way get no concept in their `data` blob, so the `concept_ids` filter won't return them. That makes the README line "Only creatives **pushed to GAM** (i.e. trafficked into a media buy) are enriched" an overstatement — this path pushes to GAM but doesn't enrich. (The `platform_creative_id` half of this gap is pre-existing on that path; the concept half is new here, so it lands asymmetrically across the three GAM paths.)

Approach: `creative_map` and the `MediaBuyUoW` (`uow2`) are already open at that loop, so each non-failed status can fold in the same helper and persist. This block handles creatives via the raw session (per its own FIXME), so mirror `CreativeRepository.update_data` (reassign + `flag_modified`):

```python
for status in asset_statuses:
    if status.status == "failed":
        logger.error(...)
        continue
    creative = creative_map.get(status.creative_id)
    if creative is None:
        continue
    merged = _merge_creative_enrichment(creative.data, status)
    if merged != (creative.data or {}):
        creative.data = merged
        attributes.flag_modified(creative, "data")  # raw-session block; JSONType change needs the nudge
# uow2 commits on exit
```

That also closes the pre-existing `platform_creative_id` gap on this path. If leaving it unenriched is intentional scope, the alternative is to narrow the README line to name the covered paths and open a follow-up.

### 2. No test exercises the real producer → writeback → DB → reader chain (recommended)

`test_list_creatives_concept_filter.py::TestSellerConceptEnrichmentIsFilterable` is positioned as the end-to-end proof, and its docstring says it "chains the real producer helper → data blob → filter" / "folded into the data blob by the production writeback path." But `_enriched_data` builds the blob by calling `_merge_creative_enrichment` directly and inserting via `CreativeFactory(data=…)` — it runs neither the GAM `_approved` producer nor either real writeback site. So the seam it actually proves is *merge-helper → reader* (which is real and useful — a mutation probe confirms the filter keys on `concept_id`), but a regression where a writeback site stopped calling the helper, or the GAM producer changed its key, would stay green. The `push_creative_to_existing_buy` happy-path unit test also uses a concept-free `AssetStatus`, so the concept branch of site 1236 is likewise unexercised.

Approach: add one integration test that drives a dry-run GAM adapter returning a concept-bearing `AssetStatus` through the real writeback (`push_creative_to_existing_buy` or `_create_media_buy_impl`), then asserts the *persisted* `creative.data["concept_id"]` and that `list_creatives(filters.concept_ids=[…])` returns it. That proves the chain the current test's docstring describes (and would also cover path 1 once wired).

### 3. Spec-grounding citation (recommended: reword)

The description says the producer's source is "ungraded by the UC-018 storyboard." There's no "UC-018 storyboard" in the AdCP compliance tree at 3.1.0-beta.3 (or any version) — "UC-018" is this repo's internal BDD id (`BR-UC-018-list-creatives.feature`). The graded `list_creatives` artifact is the creative-ad-server `browse_library` scenario, which validates the response *schema/shape* (`response_schema_ref`), not the `concept_id` value or the `concept_ids` filter. The conclusion (the seller source is ungraded) is correct — actually stronger, since neither the value nor the filter is a graded `check` step. Suggested rewording:

```text
Conformance: at adcp 5.7.0 / spec 3.1.0-beta.3 the only graded list_creatives
artifact is the creative-ad-server "browse_library" scenario, which validates the
list-creatives-response SHAPE (response_schema_ref) — concept_id/concept_name are
optional strings there. Neither the concept_ids filter nor the concept_id value/source
is a graded check step. So the seller-derived source is ungraded; the response shape it
feeds is covered by the response schema + the internal BR-UC-018 BDD.
```

## Minor / optional

- **Persistence consistency:** site 1236 uses `uow.creatives.update_data(...)`; site 3832 uses raw `creative.data = merged_data; session.add(...)`. Both persist (reassigning a new dict marks the attribute dirty), but routing 3832 through `update_data` too would give the enrichment one persistence path. (The reassignment is more robust than the old in-place `creative.data["x"] = …` at that site, which — plain `JSONType`, no `flag_modified` — was at risk of not persisting.)
- **`_merge_creative_enrichment` (`media_buy_create.py:212`):** the `concept_source or "adapter_enrichment"` fallback isn't reachable from any current producer (GAM always sets `"gam_order"`). A one-line comment that it's for future non-GAM adapters would help the next reader.
- **Anti-drift test docstring** (see #2): reword to "chains the merge helper's output → data blob → filter" so it matches what runs.
- **Log prefix:** `[GATE-PUSH] Enriched creative …` (1240) vs unprefixed (3835) — same event, two forms.
- **Optional hardening:** add `assert "concept_source" not in creatives[0]` to the wire test to pin that the internal provenance marker never reaches the wire (it can't today — the reader uses explicit kwargs and the subclass `extra` policy — but nothing pins it).

## Verified — no action needed

Spec premise (SDK + published 3.1.0-beta.3 schema/prose); producer↔reader key alignment (`data["concept_id"]` both sides); the `concept_ids` filter is scoped by `tenant_id` **and** `principal_id` before the concept clause, so the `gam-order-<numeric>` namespace can't collide across tenants; failed statuses never carry a concept; the `platform_creative_ids` refactor at 3839 is behavior-equivalent; no new duplication (the helper reduced it); the never-overwrite precedence has a real (mutation-tested) oracle.

**Verification performed:** `make quality` green (5148 passed, 0 failures); ran the merge (7), push (14), GAM-producer (2) and concept-filter integration (1) tests; mutation-probed the filter (drifted key → 0 matches); cross-checked the concept fields against `adcp==5.7.0` and the published 3.1.0-beta.3 schema/prose; verified the rebase onto `main` replays cleanly to the 8-file changeset. Not run: the full integration/e2e/admin/bdd suites end-to-end (they'll trigger in CI once the base is `main`).

mkomorski:
Thanks for the thorough review, Chris — all findings addressed. Summary of what changed:

### Finding 1 — enrichment now wired into the third GAM push path ✅
`execute_approved_media_buy` (manual-approval execution) now enriches trafficked creatives, matching the other two paths. Failed statuses still only log; non-failed ones fold in `platform_creative_id` + concept and persist within `uow2` (commits on block exit). The README "pushed to GAM → enriched" line is now accurate across all three paths.

### Finding 2 — real producer → writeback → DB → reader chain ✅
New integration test `TestExecuteApprovedEnrichesSellerConcept` (`test_execute_approved_platform_ids.py`) drives the **real dry-run GAM producer** → **real `execute_approved_media_buy` writeback** → DB → **real `CreativeRepository.get_by_principal(concept_ids=…)`**. It asserts the *persisted* `creative.data["concept_id"]` and that the filter returns it, so a writeback site dropping the helper, or a producer key drift, reddens it. Reworded the `TestSellerConceptEnrichmentIsFilterable` docstring/comment to say what it actually exercises (merge helper → filter), and added a concept-bearing case to the `push_creative_to_existing_buy` unit test (its concept branch was previously unexercised).

### Finding 3 — spec-grounding reworded ✅
Applied your suggested wording — the PR description no longer references a non-existent "UC-018 storyboard"; it now cites the graded `browse_library` scenario (response-shape only) and states the concept value/source and `concept_ids` filter are ungraded, response shape covered by the response schema + internal `BR-UC-018` BDD.

### Minor / optional
- **One persistence path (#69):** extracted `_apply_creative_enrichment()` (pure merge + provenance log + change-detection); **all three** sites now persist via `CreativeRepository.update_data` (reassign + `flag_modified`). Dropped the redundant `session.add` on the create path.
- **One log form (#72):** the "Enriched creative …" line now lives in the helper (single format) across all paths.
- **concept_source fallback (#70):** commented that it's a safety net for a future non-GAM adapter (GAM always sets `gam_order`).
- **Wire hardening:** added `assert "concept_source" not in creatives[0]` so the internal provenance marker can never leak to the wire.

### Packaging
Done: the branch was merged with `main` (collapsing the diff-vs-`main` to the ~10-file producer delta — the reader is no longer re-included), the PR base is now **`main`** so `ci.yml`/CodeQL run, and the changes above are pushed as `5b36f4be`. The PR description is updated (incl. the spec-grounding rewording). Happy to force-push a clean linear rebase if you'd prefer pristine history over the squash-merge.

All changes verified locally: unit suite 5146 passed, mypy clean, changed + create-path integration suites green, duplication baseline unchanged.

ChrisHuie (approved):

KonstantinMirin:
@mkomorski @ChrisHuie — flagging that the BDD coverage here is incomplete. I think this warrants a follow-up ticket.

**Problem**

The only concept scenario, `@T-UC-018-storyboard-filter-by-concept-id` (`tests/bdd/features/BR-UC-018-list-creatives.feature:787`), doesn't test how a `concept_id` gets onto a creative. Its Given step `given_creatives_grouped_under_concept` (`tests/bdd/test_uc018_list_creatives.py:219`, seed at `:245`) injects the concept straight into the creative `data` blob:

```python
data={"assets": {}, "concept_id": concept_id, "concept_name": _CONCEPT_NAME}
```

That bypasses production. In our harness the Given establishes the precondition **through production code** so the When exercises the real path and the Then asserts real output across all 4 transports. Seeding the blob directly means the scenario only exercises the reader/filter on hand-planted data.

**Why it matters here**

This PR added the real producer — GAM order → `concept_id="gam-order-<media_buy_id>"`, `concept_source="gam_order"` → persisted blob, in all three push paths (auto-approval create, `execute_approved_media_buy`, `push_creative_to_existing_buy`). That producer has **no BDD coverage across the 4 transports**. The integration tests here (`TestExecuteApprovedEnrichesSellerConcept` et al.) are real, but single-transport, so they don't meet the boundary-to-boundary bar.

**The fix (not a sibling scenario)**

Repair the existing Given so the concept precondition is created by the production writer:

- Drive creatives through the real push/enrichment path this PR added, OR seed via `sync_creatives` where the buyer supplies the concept — whichever matches the scenario's intent (seller-derived vs buyer-supplied).
- Keep the decoys (different concept / no concept) so the filter stays falsifiable.
- Scenario flips through the real producer → reader path and passes on all 4 transports.

**Suggested acceptance criteria for the follow-up**

- [ ] `given_creatives_grouped_under_concept` no longer writes `concept_id`/`concept_name` directly into `data`.
- [ ] The concept precondition is established via production code (GAM push enrichment from this PR, and/or `sync_creatives`).
- [ ] `@T-UC-018-storyboard-filter-by-concept-id` passes across MCP / A2A / REST / raw.
- [ ] `concept_source="gam_order"` is asserted where the seller-derived path is used.
- [ ] No new allowlist entries; `make quality` green.

**Origin**

The Given step landed in #1493; the producer it should exercise landed here. Neither connected the two. Related: #1506 (feature request), #1508 (blob-coercion follow-up).

Could someone file a follow-up ticket to fix the scenario properly? Happy to help wire it.

PR #1197, opened 2026-04-15
feat: TMPProvider model, admin UI, and discovery endpoint for Trusted Match Protocol

torbenbrodt:
## Summary

Add Trusted Match Protocol (TMP) provider management to the Sales Agent: a `TMPProvider` model for storing buyer-side provider registrations, a FastAPI admin router for CRUD operations, a `GET /tenant/{tenant_id}/tmp-providers/discovery` REST endpoint that any conforming TMP Router can poll for dynamic provider discovery, and **package sync** that pushes package definitions to TMP Providers on media buy create/update.

The Sales Agent stores registrations as an internal detail and exposes them via a versioned REST endpoint using the `provider-registration.json` wire format, pinned at `adcp/_schemas/3.1/trusted-match/provider-registration.json` (declared once in code as `src.routes.tmp_providers.PROVIDER_REGISTRATION_SCHEMA`; the path is the one that resolves in this tree). The Go TMP Router (or any conforming implementation) polls this endpoint — it never reads from the Sales Agent database directly.

## Motivation

See issue #1205 (TMPProvider registration — motivation and deployment picture) for the full deployment picture and rationale for why the Sales Agent is the right home.

Short version: provider registration is part of the planning lifecycle (media buys → packages → provider endpoints). The Sales Agent already owns this data and the admin surface. A REST endpoint decouples storage from consumption — any router implementation can discover providers without coupling to the DB schema.

**#1205 stays open after this merges.** This PR delivers the three things it asked for that were in scope — per-tenant registrations stored in the Sales Agent, an admin surface to manage them, and one cross-service contract (`GET /tenant/{tenant_id}/tmp-providers/discovery`) with the storage schema kept internal — plus package sync. Two surfaces #1205 also lists are deliberately **not** built here and remain its residual scope:

- an MCP `create_media_buy → registers provider` surface (admin-only CRUD was accepted as the scope for registration in review, 2026-05-25 / 2026-07-10)
- a `get_products` TMP capability

Rather than `Closes #1205`, the issue keeps those two so the remaining goal lives in a linked, closable issue instead of in a merged PR body.

## Architecture

```
Sales Agent (internal)         Sales Agent (external)
┌───────────────────────┐      ┌─────────────────────────────────┐
│ TMPProvider table     │─────▶│ GET /tenant/{tenant_id}/        │
│ (Postgres, internal)  │      │         tmp-providers/discovery │
│                       │      │ → provider-registration.json    │
│ TMPProviderRepository │      │   array per tenant              │
[...]

PR #1550, opened 2026-07-07
fix: change AdCPConfigurationError default recovery from terminal to correctable

torbenbrodt:
## Summary

Changes `AdCPConfigurationError._default_recovery` from the inherited `"terminal"` to `"correctable"`.

## Motivation

`terminal` means "no recovery is possible". That is wrong for configuration errors: an operator can set the missing env var, rotate the encryption key, or fix the adapter config — and the service will recover. The buyer has no lever, but the *operator* does, which is the definition of `correctable`.

This was surfaced during review of #1197 (TMP integration), where the discovery route raises `AdCPConfigurationError` for a missing `TMP_DISCOVERY_API_KEYS` env var. Konstantin noted the change is cross-cutting and should be reviewed independently.

## Affected call sites (~14 on main)

| File | Context |
|------|---------|
| `src/adapters/base.py:249` | Adapter missing-field guard |
| `src/adapters/google_ad_manager.py:155,172,312,321,450` | 5 GAM config errors |
| `src/core/database/models.py:183,648,1312` | Fernet decrypt failures (Gemini key, OIDC secret, GAM JSON) |
| `src/core/tools/creatives/_processing.py:217,535` | Creative processing config errors |
| `src/services/ai/factory.py:154` | AI factory config error |

For every one of these: the operator can fix the configuration and the service recovers. None are permanently unrecoverable.

## Changes

- `src/core/exceptions.py` — adds `_default_recovery = "correctable"` to `AdCPConfigurationError`; updates docstring to reflect the corrected semantics
- `tests/unit/test_adcp_exceptions.py` — adds `test_configuration_error_defaults_to_correctable` to `TestRecoveryClassification`, pinning the new default alongside the existing per-subclass recovery tests

## Cross-references

- Surfaced in: #1197 (TMP integration PR, Round 11 review by @KonstantinMirin)

### DIFF
--- a/src/core/exceptions.py
+++ b/src/core/exceptions.py
@@ -558,13 +558,19 @@ class AdCPConfigurationError(AdCPError):
     """Server-side configuration is broken (500).
 
     Raised when encrypted secrets cannot be decrypted (key rotation,
-    corruption, missing ENCRYPTION_KEY). Callers should NOT silently
-    fall back — the configuration needs admin intervention, so recovery is
-    ``terminal`` (inherited): the buyer has no lever to fix server config.
+    corruption, missing ENCRYPTION_KEY), or when a required environment
+    variable (e.g. ``TMP_DISCOVERY_API_KEYS``) is absent or empty.
+
+    Recovery is ``correctable``: an operator can fix the configuration
+    (set the env var, rotate the key, etc.) and the service will recover.
+    This is distinct from ``terminal`` (no recovery possible) — the
+    configuration *can* be corrected; it just requires operator action
+    rather than buyer action.
     """
 
     _default_status_code: ClassVar[int] = 500
     _default_error_code: ClassVar[str] = "CONFIGURATION_ERROR"
+    _default_recovery: ClassVar[RecoveryHint] = "correctable"
 
 
 class AdCPServiceUnavailableError(AdCPError):
--- a/tests/unit/test_adcp_exceptions.py
+++ b/tests/unit/test_adcp_exceptions.py
@@ -230,6 +230,19 @@ class TestRecoveryClassification:
         exc = AdCPAuthorizationError("forbidden")
         assert exc.recovery == "terminal"
 
+    def test_configuration_error_defaults_to_correctable(self):
+        """AdCPConfigurationError defaults to recovery='correctable'.
+
+        Operator can fix missing/broken config (env var, key rotation,
+        missing ENCRYPTION_KEY) and the service will recover — which is
+        the definition of correctable, not terminal. The buyer has no
+        lever, but the *operator* does.
+        """
+        from src.core.exceptions import AdCPConfigurationError
+
+        exc = AdCPConfigurationError("ENCRYPTION_KEY missing")
+        assert exc.recovery == "correctable"
+
     def test_not_found_error_defaults_to_terminal(self):
         """AdCPNotFoundError (the *base*) defaults to recovery='terminal'.
 


KonstantinMirin (changes requested):
**Requesting changes — this inverts the recovery classification the spec defines for `CONFIGURATION_ERROR`.**

I checked this against our pinned spec (AdCP **3.1.0-beta.3**) rather than the older 3.0.x prose, and the pinned version is decisive here. 3.1.0-beta.3 added `CONFIGURATION_ERROR` as a first-class **System** error code and classifies it **`terminal`**:

> `CONFIGURATION_ERROR` | **terminal** | Seller-side deployment misconfiguration (e.g., missing `mock_upstream_url`, undeclared `upstream_url`, **unset env var**). Distinct from `SERVICE_UNAVAILABLE` (transient) and `INVALID_REQUEST` (**buyer-fixable**) … | **Surface to the seller's operator; do not auto-retry — retries will not resolve a misconfigured deployment**
>
> — `dist/docs/3.1.0-beta.3/building/by-layer/L3/error-handling.mdx`, System codes table

That row names the exact case this PR cites — an unset env var like `TMP_DISCOVERY_API_KEYS` — and grades it `terminal`, not `correctable`.

The PR's rationale is that an operator *can* fix the config, so it's correctable. 3.1 actually considered that pattern and split it deliberately: an out-of-band operator fix is `correctable` for the authorization codes (`SCOPE_INSUFFICIENT`, `READ_ONLY_SCOPE` — see the "About `recovery: correctable` on authorization errors" note), but a deployment misconfiguration is `terminal`, explicitly "distinct from `INVALID_REQUEST` (buyer-fixable)." The recovery field is a signal to the **buyer agent**: `terminal` means "requires human action → escalate to a human operator," which is precisely the correct behavior for a missing env var. `correctable` tells the agent "modify the request and retry" — but there's no request field the buyer can change, so it produces a pointless retry against an unchanged fault. That's the outcome the spec's "do not auto-retry" line is warning against.

So the inherited `terminal` default was already spec-correct, and this change moves a correct value to a spec-violating one. My recommendation is to drop the change.

One thing that likely hid this: `adcp==5.7.0`'s `STANDARD_ERROR_CODES` doesn't include `CONFIGURATION_ERROR` yet, so today we map it to `SERVICE_UNAVAILABLE` on the wire. That SDK lag is a separate, pre-existing issue (the SDK trails the spec; per our source-of-truth order the spec wins) and not something this PR needs to fix — but it's worth a follow-up, because under 3.1 a config fault should ultimately surface as `CONFIGURATION_ERROR`/terminal, not `SERVICE_UNAVAILABLE`/transient.

**On testing — this is the part I want to flag as more than a nit.** The change ships with a single unit test that constructs the exception and asserts `.recovery == "correctable"`. That test can only ever fail if someone edits the same constant — it asserts a ClassVar equals the value it was just set to, in-process, on no transport, never touching the wire envelope. It gave green confidence for a change that's backwards.

The reason the inversion is invisible to the full suite is that **nothing grades config-error recovery as a behavior**. There's no scenario in any feature file for `CONFIGURATION_ERROR`, and no storyboard/compliance step grades it — the only place the invariant is written down is a *comment* in `tests/bdd/steps/domain/uc004_delivery.py:2685` (which says config → terminal, and which this PR would silently falsify). We have a real behavioral gap here, and a unit test papering over it.

The fix that actually protects us is a scenario, not another assertion on the class. Something like:

- **Given** a tenant whose encrypted secret is corrupt / a required env var is unset (real fixture state in the DB, not stashed in context)
- **When** the buyer sends a request that hits that config path
- **Then** the wire error's `recovery` is `terminal` (asserted on `wire_error_envelope` via `assert_envelope_shape`, not the in-process `error.recovery`)

dispatched across MCP/A2A/REST through the `Env` helpers like our other error scenarios, with a `# @source` comment pointing at the 3.1.0-beta.3 `CONFIGURATION_ERROR` row. That exercises real production code end-to-end, locks the spec-correct value against future regressions, and gives the same coverage on every transport for free.

More broadly: for any change to an error's wire contract — code or recovery — the bar should be a connected scenario that asserts the buyer-visible envelope, because that's the thing the buyer agent actually keys off. A per-class unit assertion doesn't tell us what a buyer receives on the wire, and as this PR shows, it can be green while the wire behavior is wrong. Happy to write that scenario if it's useful.

torbenbrodt:
Agreed — your review is correct, this inverts the spec-mandated classification for CONFIGURATION_ERROR (terminal per AdCP 3.1.0-beta.3). Closing without merging, thanks for the thorough spec citations.

Would you file a follow-up issue for the gap you flagged — `adcp==5.7.0`'s `STANDARD_ERROR_CODES` doesn't include `CONFIGURATION_ERROR` yet, so it currently maps to `SERVICE_UNAVAILABLE` on the wire instead of the spec-correct `CONFIGURATION_ERROR`/`terminal`?

PR #1551, opened 2026-07-07
test(bdd): wire UC-018 BR-RULE-034 cross-principal isolation scenario…

### LINKED ISSUE #1503: list_creatives: wire the BR-RULE-034 cross-principal isolation BDD scenarios (builds on #1474)
### Background

`list_creatives` is principal-scoped: a buyer only ever sees creatives under their own `principal_id` within their tenant. The production code is correct — `CreativeRepository.get_by_principal` builds `select(Creative).filter_by(tenant_id=..., principal_id=...)`, and a second principal's creatives (even sharing the same concept) are not returned (verified by execution). But there is no **live test** that goes red if that scoping is ever dropped — e.g. a future refactor of `get_by_principal`. For a tenant/principal-isolation invariant, an unenforced guarantee is the gap worth closing.

### Where it is

`tests/bdd/features/BR-UC-018-list-creatives.feature` defines two `@BR-RULE-034` scenarios:
- `@T-UC-018-inv-034-1-holds` — "BR-RULE-034 INV-1 holds — query always scoped by principal"
- `@T-UC-018-inv-034-1-violated` — "BR-RULE-034 INV-1 counter — cross-principal creatives never visible"

### Dependency / current state (important)

The UC-018 BDD binding (`tests/bdd/test_uc018_list_creatives.py`) and the conftest harness routing are introduced by **#1474 (UC-018 list-creatives-after-sync scaffolding)**, which is not yet merged. On `main` today the feature file exists but is bound to no test module, so these scenarios are not exercised at all. **This issue builds on #1474**: once it lands, the unwired UC-018 scenarios (including the two `@BR-RULE-034` ones) auto-xfail at the conftest harness gate, and this issue wires them. Pick it up after #1474 merges (or coordinate if you want to develop against that branch).

### Contract / why it matters

BR-RULE-034 INV-1: the `list_creatives` query is always scoped by principal; cross-principal creatives are never visible. The code honors it; the test net does not yet.

### Scope

Wire both scenarios against real production across the wire transports:
- Seed two principals in the **same tenant**, each with creatives.
- Authenticate as principal A; call `list_creatives`.
- Assert the result contains exactly A's creatives and **none** of B's (the counter scenario's whole point — a decoy under B that must never appear).

### Suggested approach (a starting point)

1. Widen the UC-018 gate in `tests/bdd/conftest.py` (`_harness_env`, UC-018 branch) to route the `@BR-RULE-034` scenarios into `CreativeListEnv` — mirror how `@list-after-sync` / `@concept-id` were added (a marker-set membership check).
2. Add the Given/When/Then steps in `tests/bdd/test_uc018_list_creatives.py` (module-scoped — see Gotchas), seeding the second principal with `PrincipalFactory` and its creatives with `CreativeFactory`.

Templates already in the repo:
- `tests/bdd/steps/domain/uc006_sync_creatives.py` — working cross-principal step implementations.
- `tests/bdd/test_uc018_list_creatives.py` — the wired UC-018 storyboard scenarios (`@list-after-sync`, `@concept-id`): module-scoped steps, `CreativeListEnv`, factories, assertions on `ctx["wire_response"]`.

### Definition of done

- [ ] Both `@BR-RULE-034` UC-018 scenarios flip xfail → pass across the wired transports (a2a/mcp/rest).
- [ ] The counter scenario is a **real oracle**: temporarily removing the `principal_id` clause from `get_by_principal` makes it fail (confirm locally, then restore).

### How to verify (test harness)

- `CreativeListEnv` + `TenantFactory`/`PrincipalFactory`/`CreativeFactory`; seed principal B in the same tenant as A.
- Authenticate as A (the BDD Background sets the principal; see the existing `given_buyer_authenticated_as_principal` step).
- `eval $(.claude/skills/agent-db/agent-db.sh up)` → migrate → `uv run pytest tests/bdd/test_uc018_list_creatives.py -k "scoped or visible"` (match on the scenario text, not the tag — pytest-bdd names tests from the scenario name).

### Gotchas (from reviewing UC-018)

- Step phrasings shared with other wired feature files (UC-004/005/006) must **not** be registered globally, or they'll alter those suites. UC-018 deliberately defines its steps inline/module-scoped — follow that pattern.
- Assert on `ctx["wire_response"]` (real wire bytes) for field-shape claims, not a `model_dump` re-serialization of the typed payload.
- Both scenarios run only on the wired wire transports; IMPL is sunset in BDD.
- The isolation is enforced in a single `select().filter_by(tenant_id=..., principal_id=...)` — the counter scenario should seed B in the **same tenant** (cross-tenant is a separate, weaker check; the sharp case is same-tenant, different-principal).

mkomorski:
Wire the two @BR-RULE-034 scenarios in BR-UC-018-list-creatives.feature (INV-1 holds / INV-1 counter) to real production across a2a/mcp/rest, building on the UC-018 scaffolding from #1474.

- conftest.py: route @BR-RULE-034 markers to CreativeListEnv in the UC-018 harness gate (unambiguous — the branch only runs for T-UC-018-* scenarios).
- test_uc018_list_creatives.py: extract a shared _authenticate_env_as helper (reused by the Background Given and the new When), add the seed Given (two phrasings, one body), the isolation When (re-auth after seeding so the wire auth token resolves), and three Then steps.

Ownership is verified by matching returned creative_ids against the seeded per-principal id sets: principal_id is Field(exclude=True) and never appears on the buyer-facing wire, and CreativeFactory assigns globally-unique ids so the two principals' sets are disjoint. Assertions read ctx["wire_response"] (real serialized bytes).

Verified: both scenarios pass on a2a/mcp/rest; removing the principal_id filter from CreativeRepository.get_by_principal turns all six red; no regression in the rest of the UC-018 suite; DRY duplication baseline unchanged.

Closes #1503

### DIFF
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -2846,13 +2846,18 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
             yield
 
     elif uc == "UC-018":
-        # list_creatives — the wired storyboard scenarios are @list-after-sync
-        # (#1405) and @concept-id (#1407). The remaining UC-018 scenarios
-        # (main/partition/boundary/other filter siblings) have no step definitions
-        # yet, so xfail fast at the fixture (mirrors UC-002/006/011) rather than
-        # spinning up a DB per scenario only to auto-xfail at the first missing step.
+        # list_creatives — the wired scenarios are @list-after-sync (#1405),
+        # @concept-id (#1407), and the @BR-RULE-034 cross-principal isolation
+        # invariants (#1503). The remaining UC-018 scenarios (main/partition/
+        # boundary/other filter siblings) have no step definitions yet, so xfail
+        # fast at the fixture (mirrors UC-002/006/011) rather than spinning up a
+        # DB per scenario only to auto-xfail at the first missing step.
+        #
+        # BR-RULE-034 is unambiguous here: this branch only runs for T-UC-018-*
+        # scenarios (see _detect_uc), so the tag never collides with the UC-006
+        # BR-RULE-034 scenarios routed elsewhere.
         marker_names = {m.name for m in request.node.iter_markers()}
-        if marker_names & {"list-after-sync", "concept-id"}:
+        if marker_names & {"list-after-sync", "concept-id", "BR-RULE-034"}:
             # CreativeListEnv mocks only the audit logger; DB, repository, and
             # query building are real. The Background auth step switches the env
             # principal; the seed step owns the creatives under it.
@@ -2864,7 +2869,8 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
                 yield
         else:
             pytest.xfail(
-                "UC-018 harness wired only for the @list-after-sync (#1405) and @concept-id (#1407) storyboard scenarios"
+                "UC-018 harness wired only for the @list-after-sync (#1405), @concept-id (#1407), "
+                "and @BR-RULE-034 isolation (#1503) scenarios"
             )
 
     elif uc == "UC-011":
--- a/tests/bdd/steps/domain/uc003_ext_error_scenarios.py
+++ b/tests/bdd/steps/domain/uc003_ext_error_scenarios.py
@@ -44,21 +44,20 @@ def _inject_privilege_error(ctx: dict) -> None:
 
 @given(parsers.parse('the Buyer is authenticated as principal "{principal_id}"'))
 def given_buyer_authenticated_as(ctx: dict, principal_id: str) -> None:
-    """Override the authenticated principal to a specific principal_id.
+    """Authenticate as *principal_id* and mark the auth state.
 
-    Sets ctx['principal_override'] so the When step can build an identity
-    with a different principal_id. Also marks the auth state.
+    Delegates the principal switch, the canonical ``ctx["principal_id"]``, and the
+    identity post-condition to the shared ``authenticate_env_as`` helper; adds only
+    the use-case-specific ``has_auth`` flag.
+
+    NOTE: this module is currently dormant (see ``steps/generic/_auth.py``) — it is
+    not registered in ``tests/bdd/conftest.py`` ``pytest_plugins`` and UC-003 is not
+    wired into the BDD harness, so these steps do not execute (UC-003 update
+    scenarios auto-xfail). Kept on the shared helper so it is correct when UC-003 is
+    activated.
     """
-    ctx["principal_override"] = principal_id
+    authenticate_env_as(ctx, principal_id)
     ctx["has_auth"] = True
-    # Update the env identity to use this principal_id (shared helper).
-    env = authenticate_env_as(ctx, principal_id)
-    # Post-condition: verify the identity mutation took effect
-    actual = env.identity.principal_id
-    assert actual == principal_id, (
-        f"env.identity.principal_id is {actual!r} after setting "
-        f"env._principal_id to {principal_id!r} — cache not rebuilt"
-    )
 
 
 @given(parsers.parse('the principal "{principal_id}" does not exist in the database'))
--- a/tests/bdd/steps/domain/uc003_ext_error_scenarios.py
+++ b/tests/bdd/steps/domain/uc003_ext_error_scenarios.py
@@ -14,6 +14,7 @@ from pytest_bdd import given, parsers, then
 
 from tests.bdd.steps._harness_db import db_session
 from tests.bdd.steps.domain.uc003_update_media_buy import _ensure_update_defaults
+from tests.bdd.steps.generic._auth import authenticate_env_as
 
 
 def _inject_privilege_error(ctx: dict) -> None:
@@ -50,10 +51,8 @@ def given_buyer_authenticated_as(ctx: dict, principal_id: str) -> None:
     """
     ctx["principal_override"] = principal_id
     ctx["has_auth"] = True
-    # Update the env identity to use this principal_id
-    env = ctx["env"]
-    env._identity_cache.clear()
-    env._principal_id = principal_id
+    # Update the env identity to use this principal_id (shared helper).
+    env = authenticate_env_as(ctx, principal_id)
     # Post-condition: verify the identity mutation took effect
     actual = env.identity.principal_id
     assert actual == principal_id, (
--- a/tests/bdd/steps/domain/uc003_update_media_buy.py
+++ b/tests/bdd/steps/domain/uc003_update_media_buy.py
@@ -15,6 +15,7 @@ from typing import Any
 from pytest_bdd import given, parsers, then, when
 
 from tests.bdd.steps._harness_db import db_session
+from tests.bdd.steps.generic._auth import authenticate_env_as
 from tests.bdd.steps.generic._dispatch import dispatch_request
 from tests.bdd.steps.generic.given_media_buy import _resolve_date_token
 
@@ -2054,9 +2055,7 @@ def given_authenticated_principal(ctx: dict, principal: str) -> None:
     Updates the env's identity to use the specified principal_id.
     """
     principal_id = principal.strip()
-    env = ctx["env"]
-    env._identity_cache.clear()
-    env._principal_id = principal_id
+    authenticate_env_as(ctx, principal_id)
     ctx["principal_override"] = principal_id
 
 
--- a/tests/bdd/steps/domain/uc003_update_media_buy.py
+++ b/tests/bdd/steps/domain/uc003_update_media_buy.py
@@ -2050,13 +2050,14 @@ def given_media_buy_owned_by_principal(ctx: dict, owner_id: str) -> None:
 
 @given(parsers.parse("the authenticated principal is {principal}"))
 def given_authenticated_principal(ctx: dict, principal: str) -> None:
-    """Set the authenticated principal for the request.
+    """Set the authenticated principal for the request via the shared helper.
 
-    Updates the env's identity to use the specified principal_id.
+    ``authenticate_env_as`` owns the switch, the canonical ``ctx["principal_id"]``,
+    and the identity post-condition. Currently dormant — see the note on
+    ``given_buyer_authenticated_as`` / ``steps/generic/_auth.py``; retained on the
+    shared helper so the DRY convention holds when UC-003 is activated.
     """
-    principal_id = principal.strip()
-    authenticate_env_as(ctx, principal_id)
-    ctx["principal_override"] = principal_id
+    authenticate_env_as(ctx, principal.strip())
 
 
 # ═══════════════════════════════════════════════════════════════════════
--- a/tests/bdd/steps/generic/_auth.py
+++ b/tests/bdd/steps/generic/_auth.py
@@ -1,0 +1,25 @@
+"""Shared authentication helpers for BDD step definitions.
+
+Plain helper module (no ``@given``/``@when``/``@then`` decorators) — importing it
+never registers a step, mirroring ``_account_resolution.py``. Home for the
+principal-switch mutation shared by every "authenticated as principal" Given/When
+across use cases (UC-003 error + update, UC-018 isolation).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def authenticate_env_as(ctx: dict, principal_id: str) -> Any:
+    """Switch the harness env's authenticated principal to *principal_id*; return the env.
+
+    Clears the identity cache so the next ``env.identity`` access re-resolves identity
+    from scratch — picking up a principal row committed after the env was created — then
+    points the env at *principal_id*. Callers layer their own ctx bookkeeping
+    (``principal_id`` / ``principal_override``) and post-conditions on top.
+    """
+    env = ctx["env"]
+    env._identity_cache.clear()
+    env._principal_id = principal_id
+    return env
--- a/tests/bdd/steps/generic/_auth.py
+++ b/tests/bdd/steps/generic/_auth.py
@@ -1,9 +1,20 @@
-"""Shared authentication helpers for BDD step definitions.
+"""Shared authentication helper for BDD step definitions.
 
 Plain helper module (no ``@given``/``@when``/``@then`` decorators) — importing it
 never registers a step, mirroring ``_account_resolution.py``. Home for the
-principal-switch mutation shared by every "authenticated as principal" Given/When
-across use cases (UC-003 error + update, UC-018 isolation).
+principal-switch shared by every "authenticated as principal" Given/When across
+use cases (UC-018 isolation; UC-003 error + update).
+
+**Live coverage.** Only UC-018 (``test_uc018_list_creatives``) exercises this
+helper at runtime today. The UC-003 callers (``uc003_update_media_buy`` and
+``uc003_ext_error_scenarios``) are currently **dormant**: neither step module is
+registered in ``tests/bdd/conftest.py`` ``pytest_plugins`` and UC-003 is not wired
+into the BDD harness, so every UC-003 update scenario auto-xfails (verified: 0
+passed / 1404 xfailed). Those sites are converted to this shared helper anyway so
+they are correct-by-construction when UC-003 is later activated — but the DRY edit
+there is uncovered until then. Wiring UC-003 into the harness (registering the step
+modules + mapping ``MediaBuyUpdateEnv`` in ``_detect_uc`` / ``_harness_env``) remains
+a follow-up.
 """
 
 from __future__ import annotations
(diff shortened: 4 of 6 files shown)


ChrisHuie:
Reviewed the wiring end-to-end (ran both scenarios across a2a/mcp/rest, and confirmed the counter is a real oracle by dropping the `principal_id` clause from `get_by_principal` — all six go red, then restored). The DoD is met and the invariant is a genuine AdCP L1 isolation MUST that the conformance storyboards only cover single-account, so this is a real addition to the net. No blockers.

One worthwhile hardening on the counter step (`then_none_belong_to`): it asserts only that the returned set doesn't intersect the co-tenant principal's ids, which passes vacuously if the response is ever empty. Its sibling `then_all_creatives_belong_to` already guards this with `assert returned`; mirroring it makes the counter independently falsifiable (the paired holds scenario compensates today, but the counter shouldn't have to rely on it):

```python
returned = _returned_creative_ids(ctx)
assert returned, "isolation counter is vacuous on an empty response (list_creatives returned no creatives)"
leaked = returned & set(ctx[_ISOLATION_CREATIVES_KEY][principal_id])
```

Two optional docstring touch-ups while you're in the file: the inline comment above `scenarios(...)` still says only `@list-after-sync` is wired / "no UC-018 harness yet", which the module docstring now contradicts; and the "across all wire transports (plus e2e_rest in-network)" line is accurate for a2a/mcp/rest but optimistic for e2e_rest, where the E2E REST dispatcher doesn't stash `wire_response` on the success path (pre-existing, not introduced here).

Everything else — factory usage, module-scoped steps, the double-`@given`, the real-token auth chain, the `exclude_none` wire-shape rationale — checks out.

KonstantinMirin:
Following up on Chris's review rather than re-treading it — I verified the one open question (is this an AdCP MUST or a local invariant?) against the pinned spec, and it resolves in his favor, with one consequence worth acting on.

**Grounding — this is a spec MUST, so let's cite it as one.** Confirmed against `v3.1-04f59d2d5`: cross-principal isolation is normative — `dist/docs/2.5.3/media-buy/advanced-topics/principals-and-security.mdx:30-34` ("the server **MUST** verify that the `principal_id` from the request matches… one principal can never view or modify another principal's data") and `reference/security.mdx:714,721`. I also confirmed the other half of Chris's read: no conformance storyboard grades multi-principal isolation (`universal/security.yaml` grades authentication, not authenticated isolation; the test-kits are single-advertiser). So this is a genuine net addition — and it's the *only* executable guard of that MUST. Given that, the docstring should ground in the spec section it satisfies, not the enforcement code: cite `principals-and-security.mdx` §Data Isolation (with an "ungraded-by-storyboard" note, per the Spec-Grounding Gate), keeping the `get_by_principal` pointer as the enforcement site. Right now it grounds in code, the weakest tier.

**+1 on the counter-step floor — and it's load-bearing, not just hardening.** Because this scenario is the sole executable check of the isolation MUST, `then_none_belong_to` being vacuously green on an empty response is the one gap that matters. Concur exactly with Chris's fix (`assert returned` before the intersection). Flagging it as a merge condition rather than optional, for that reason.

**One net-new item the file-scoped view doesn't surface — the helper extraction is incomplete.** `_authenticate_env_as` is a good extraction, but the same `env._identity_cache.clear()` + `env._principal_id = …` + post-condition body it consolidates is still hand-rolled in two sibling modules:

- `tests/bdd/steps/domain/uc003_ext_error_scenarios.py:44` — byte-identical step text (`'the Buyer is authenticated as principal "{principal_id}"'`) and logic
- `tests/bdd/steps/domain/uc003_update_media_buy.py:2056` — same core mutation, different phrasing

Neither guard catches this (the duplicate-step AST guard only scans `tests/bdd/steps/`; R0801's 6-line floor can't see the body), so it won't show up in CI. Since the helper is a plain function (importable like `_call_via`), moving it to a shared `steps/generic/` module and having the uc003 steps call it completes the "extract a shared helper" objective this PR states. Firm on this one — it's the project's zero-duplication invariant, and it's cheap and in-scope.

**Two smaller ones:**

- The seed's `data={"assets": {}}` override is redundant *and* slightly counterproductive: `CreativeFactory` already defaults `data` to a realistic `{"assets": build_assets(...)}`, which satisfies the repo's `data["assets"].isnot(None)` guard. The override replaces a realistic asset with an empty dict and its comment implies it's required, which it isn't. Dropping it (and using the `approved` trait) reads cleaner.
- Seconding Chris's docstring notes: the "full header → token → DB-lookup auth chain" line is accurate for MCP/A2A but not REST (FastAPI dep override) or IMPL, and the stale wiring comment at `:72-75` contradicts the module docstring.

**Suggested disposition:** merge conditions — counter-step floor (Chris's fix) + finishing the `_authenticate_env_as` extraction into the uc003 siblings. Everything else (spec citation, `data` override, docstring touch-ups) is small and can ride along. Nothing here blocks the approach, which is the right one.

mkomorski:
Thanks @ChrisHuie and @KonstantinMirin — pushed 6d9cf154 addressing all of it.

**Merge conditions**
- **Counter-step floor** — `then_none_belong_to` now asserts `returned` before the intersection, so the isolation counter is independently falsifiable rather than vacuously green on an empty response (mirrors `then_all_creatives_belong_to`).
- **Finished the `authenticate_env_as` extraction** — the principal-switch core (identity-cache clear + `_principal_id` set) now lives in a decorator-free shared helper `tests/bdd/steps/generic/_auth.py` (mirroring `_account_resolution.py`, so importing it registers no steps — it couldn't go in `given_auth.py`, which is a global `pytest_plugins` entry and would collide with the inline steps). Both siblings — `uc003_ext_error_scenarios.py:44` and `uc003_update_media_buy.py:2056` — now call it, keeping their own ctx bookkeeping. I swept every `_identity_cache.clear()` in `tests/bdd/`; the three tenant-override sites (`given_media_buy.py`, `uc002`) clear the cache to re-read `_tenant_overrides` — a different operation — so I left them alone.

**Ride-alongs**
- **Spec grounding** — the `@BR-RULE-034` section comment + module docstring now cite the AdCP normative MUST (`principals-and-security.mdx` §Data Isolation, `reference/security.mdx:714,721`, pinned `v3.1-04f59d2d5`), with the ungraded-by-storyboard note per the Spec-Grounding Gate, keeping `get_by_principal` as the enforcement site.
- **Seed `data` override** — dropped `data={"assets": {}}` + `status="approved"`; now `CreativeFactory(..., approved=True)`, which keeps the factory's realistic default assets (already satisfies the repo's non-null guard).
- **Docstrings** — refreshed the stale `scenarios()` comment; qualified the e2e_rest claim (the `wire_response` Then steps cover a2a/mcp/rest, not e2e_rest); corrected the auth-chain note (full header→token→DB is MCP/A2A; REST resolves via a FastAPI dependency override).

**Verified**: UC-018 BDD `12 passed / 546 xfailed` — all six `@BR-RULE-034` parametrizations green across a2a/mcp/rest; falsifiability re-confirmed. uc003 collection clean, no regression. ruff + mypy (changed files) clean; BDD structural guards pass; duplication baseline unchanged (the removed copies were below R0801's 6-line floor — exactly why it needed a manual flag).

KonstantinMirin:
Thanks for the quick turnaround, @mkomorski — the two merge conditions from the last round are genuinely resolved. I re-ran the isolation scenarios (6/6 green across a2a/mcp/rest), confirmed the counter step is now falsifiable (the `assert returned` floor closes the vacuous-empty path), and re-checked the spec grounding verbatim against the pin. The scenarios are the right shape and remain the only executable guard we have of the isolation MUST. Good work.

I'm going to hold the merge, though. The follow-up commit, while extracting the shared helper, introduced drift that has to be resolved first. Each item below looks small on its own — the reason they're blocking rather than nits is that they're the same failure repeated: a shared primitive that was only half-unified, plus a fix applied to the cited site but not its siblings. A half-unified primitive is worse than none, because the next contributor inherits two conventions and picks one at random. That is exactly the drift the zero-duplication invariant exists to stop, so I don't want to wave any of it through.

**1. `authenticate_env_as` must be ctx-complete, not partial.** Today it mutates the env and hands ctx bookkeeping back to each caller, and the three callers immediately diverge:
- Two keys for one concept — `ctx["principal_id"]` (written by the generic steps and UC-018, actually read downstream in uc004/uc006) vs `ctx["principal_override"]` (uc003_update:2059, uc003_ext:52), which is written and read nowhere. The helper docstring (`_auth.py:20`) enshrines both as acceptable, locking the split in.
- The identical Gherkin sentence `the Buyer is authenticated as principal "…"` resolves to two different functions with two different ctx effects.
- The post-condition assert is present at two of three call sites, absent at the third (`uc003_update`).

The model you cited, `_account_resolution.py`, is ctx-complete — it owns its ctx state. Make this one own the canonical `ctx["principal_id"]` and the identity post-condition; callers then add only genuinely use-case-specific state (uc003's `has_auth`). That collapses all three bullets and restores a single convention. While you're in there, the reach into `env._identity_cache` / `env._principal_id` from the step layer should go through a public `switch_principal()` on the env — the harness convention is public accessors (`get_session()` is the precedent), and a shared helper shouldn't institutionalize the private reach.

**2. Finish the seed-idiom migration across the file.** Dropping the `data={"assets": {}}` override on the isolation seeder was the right call — but the same override is still hand-rolled at every other seeder in the file (119/128, 255/260, 265/270, 273/278). You've now shown, in this same file, that `CreativeFactory(..., approved=True)` with the factory's default realistic assets satisfies the non-null guard, which proves those verbose sites are unnecessary. Leaving them means the file teaches two ways to seed an approved creative. Migrate all of them to the trait.

**3. Point the spec citation at the pinned-era doc.** The citation resolves to the superseded 2.5.3 `principals-and-security.mdx`; at the pin the prose moved to `accounts-and-security.mdx` / `building/implementation/security.mdx §Agent and Account Isolation`. The MUST is identical in force, so there's no behavioral risk — but the grounding gate exists so the citation is a reference someone can actually follow, and this one points at a renamed/moved file. Correct and version-prefix the path so it resolves against the pin.

**4. Confirm the UC-003 half of the extraction actually runs — I don't believe it does.** `uc003_update_media_buy.py` and `uc003_ext_error_scenarios.py` are in neither `pytest_plugins` (conftest.py:48-66) nor the UC-003 binding module (`test_uc003_update_media_buy.py` imports nothing and only calls `scenarios(...)`); the sole cross-import is function-local in uc006 and doesn't register steps. So the `@given` bodies you rewired there are undiscoverable, their scenarios auto-xfail, and the "both siblings now call it" claim is true in source but never exercised — the 12-passed run is UC-018 only. Please verify with a real UC-003 BDD run. If they're dormant, either register the modules the way every peer domain module is (so the shared helper is genuinely under test there), or state explicitly that UC-003 is intentionally dormant and that the DRY edit is currently uncovered. Don't let the PR imply coverage it doesn't have.

The approach is right and the isolation scenarios are ready. These are about not landing a half-unified primitive and a split idiom into the shared test system, where the ambiguity propagates. Happy to re-review as soon as the helper owns its state and the UC-003 run is confirmed.

mkomorski:
Thanks @KonstantinMirin — pushed d8b13d8e addressing all four. Verified UC-018 BDD **12 passed / 546 xfailed** at real Postgres; ruff + 356 architecture guards + the DRY duplication hook green.

**1. `authenticate_env_as` is now ctx-complete.** It owns the whole principal-switch contract: re-points the env via a new public `env.switch_principal()` (added on `BaseTestEnv`, mirroring `get_session()` — no more reaching into private `_identity_cache` / `_principal_id`), writes the single canonical `ctx["principal_id"]`, and runs the identity post-condition. The dead `ctx["principal_override"]` key is gone (written at the two uc003 sites, read nowhere). The three callers are now thin — UC-018 is a bare call, uc003_ext adds only its use-case-specific `has_auth`, and uc003_update picks up the post-condition it was missing. One sentence → one function → one ctx effect.

**2. Seed idiom unified.** Every seeder in the file now goes through one local `_seed_creative()` helper — the single place `data` is assembled: the `approved` trait + CreativeFactory's realistic default assets, with concept fields layered on when needed. No `data={"assets": {}}` / `status="approved"` hand-rolls remain (119/255/265/273 + the isolation seeder all migrated).

**3. Spec citation.** You're right it resolved to the superseded 2.5.3 doc. Corrected to the pinned-era, resolvable **source** paths (verified against `v3.1-04f59d2d5`): `docs/media-buy/advanced-topics/accounts-and-security.mdx` §Data Isolation (L33-37), plus the deeper `docs/building/by-layer/L1/security.mdx` §Agent and Account Isolation (L159, incl. §"Client-side isolation: cross-principal tool-call confusion" at L229). Two things I hit while verifying, worth flagging: the deeper ref's repo path is `building/by-layer/L1/security.mdx` — `building/implementation/security.mdx` is the rendered-site URL and doesn't resolve as a file; and the built `dist/docs/3.1.0-beta.3/` tree isn't present at the frozen schema-pin commit, so I cite the source `docs/` paths, which do resolve at the pin. Replaced the old "principal_id … never view" quote with the pin's account-isolation wording and dropped the stale `reference/security.mdx:714,721`.

**4. UC-003 coverage — confirmed it doesn't run.** A real run of `test_uc003_update_media_buy.py` is **1404 xfailed / 0 passed** ("No harness wired for None"). It's deeper than step registration: UC-003 isn't wired into the BDD harness (`_detect_uc` / `_harness_env`) at all. Rather than expand this PR into UC-003 activation, I kept the DRY rewiring (so those sites are correct-by-construction) and documented the dormancy explicitly in `_auth.py` and both call sites — the PR no longer implies coverage it doesn't have. Wiring UC-003 into the harness (register the two step modules + map `MediaBuyUpdateEnv`, then curate the strict xfails) is left as a documented follow-up.

Ready for re-review.

ChrisHuie:
Re-reviewed `d8b13d8e` merged onto current main — the PR's CI ran on an older base, and #1545 independently touched `_base.py` and `conftest.py`, so I checked the merge result. All four round-2 conditions resolve:

**1. `authenticate_env_as` is ctx-complete** — single `ctx["principal_id"]`, post-condition assert, dead `principal_override` removed; the switch goes through the new public `env.switch_principal()`, which reproduces the prior inline mutation exactly.

**2. Seed idiom unified** — every UC-018 seeder routes through `_seed_creative`; no `{"assets": {}}` hand-rolls remain; the duplication baseline shrank (tests 89→88).

**3. Spec citation resolves at the pin** — `accounts-and-security.mdx §Data Isolation` and `building/by-layer/L1/security.mdx §Agent and Account Isolation` both resolve at `04f59d2d5`; no stale path survives.

**4. UC-003 dormancy documented and accurate** — `test_uc003_update_media_buy.py` is 1404 xfailed / 0 passed; the DRY rewiring there is honestly marked uncovered-until-activation.

**Ran on the merge (a2a/mcp/rest):** both isolation scenarios pass (6/6); the oracle still reddens all six when the `principal_id` filter is dropped from `get_by_principal` (the empty→realistic asset change didn't weaken it); the `#1405`/`#1407` storyboard+concept scenarios still pass with the new seed shape; the regression slice (UC-018/003/004/006) reports 0 failed / 0 error. The 21 UC-004 xpassed on the merge trace to #1545's delivery change, not this PR.

**Non-blocking — two optional one-liners** if you're in the files again: narrow the `switch_principal` docstring's "must not reach into private `_identity_cache`/`_principal_id`" line to the switch mutation (steps still legitimately reach `_principal_id`/`_tenant_id` for seeding and tenant-override), and add a clause to `_seed_creative` noting the populated default asset still validates against the pinned asset-union (so `#1405`'s schema check holds). Everything else — `RestE2EDispatcher` not stashing `wire_response` on `e2e_rest`; a `set_tenant_override` accessor mirroring `switch_principal`; widening the BDD guard scan to inline UC-018 steps — is pre-existing and better tracked as separate follow-ups.

---
*Ran: isolation scenarios + oracle mutation + `#1405`/`#1407` + the UC-018/003/004/006 regression slice + the architecture/BDD guard suite (395 tests) + the DRY hook, on the PR merged onto current main. Not run: in-network `e2e_rest` (`BDD_E2E_ENABLED`) and the full `./run_all_tests.sh ci` — the PR's own CI is passing and the merge is conflict-free.*

ChrisHuie (approved):

mkomorski:
Thanks @ChrisHuie — appreciate the merge-result re-review and the approval. Landed both optional one-liners in 61fbc821:

- **`switch_principal` docstring** — narrowed the "don't reach into private state" line to the *switch mutation* itself. Reading `_principal_id`/`_tenant_id` to seed factory rows (as the UC-018 seeders do) or clearing the cache for a tenant override is a distinct, legitimate step-layer operation, and the docstring no longer implies otherwise.
- **`_seed_creative`** — added a clause that the realistic default asset conforms to the pinned `list-creatives-response.json` asset union, so `#1405`'s `schema-valid against` Then step still holds now that the empty `{"assets": {}}` hand-roll is gone.

Docstring-only, no behavior change: UC-018 collects 558 (12 pass / 546 xfail), ruff + the BDD structural guards green. Left the three pre-existing items you flagged (`RestE2EDispatcher` not stashing `wire_response` on `e2e_rest`, a `set_tenant_override` accessor mirroring `switch_principal`, and widening the BDD guard scan to inline UC-018 steps) as separate follow-ups per your note.

Issue #1559, opened 2026-07-07
chore(security): track MAL-2026-4750 fastapi fastar withdrawn-advisory lag


mkostromin-sigma:
## Summary

Track the active `MAL-2026-4750` suppression in [`scripts/security-ignored-vulns.sh`](https://github.com/prebid/salesagent/blob/main/scripts/security-ignored-vulns.sh) (also in `PIP_AUDIT_IGNORED_VULNS`).

## Why suppressed

OSV advisory for undocumented `fastar` dep was **withdrawn** 2026-05-26 (internal tooling artefact). uv-secure and pip-audit databases lag withdrawal by hours/days.

## Remove when

- [ ] Both uv-secure and pip-audit clean on `uv.lock` without this ignore
- [ ] Remove from `UV_SECURE_IGNORED_VULNS` and `PIP_AUDIT_IGNORED_VULNS`; close this issue

Part of #1228 (C7 residual).

Issue #1558, opened 2026-07-07
chore(security): track PYSEC-2025-183 disputed PyJWT suppression


mkostromin-sigma:
## Summary

Track the active `PYSEC-2025-183` suppression in [`scripts/security-ignored-vulns.sh`](https://github.com/prebid/salesagent/blob/main/scripts/security-ignored-vulns.sh).

## Why suppressed

Supplier **disputes** the weak-signing-key framing; no fix version. This codebase only uses PyJWT via `jwt.decode(..., options={"verify_signature": False})` for Google OIDC ID token parsing — no application-configured signing keys.

## Remove when

- [ ] Advisory withdrawn/rescinded in OSV **or** uv-secure stops flagging without ignore
- [ ] Remove ID from `UV_SECURE_IGNORED_VULNS` and close this issue

Part of #1228 (C7 residual).

Issue #1557, opened 2026-07-07
chore(security): track PYSEC-2026-89 Markdown OSV suppression until databases sync


mkostromin-sigma:
## Summary

Track the active `PYSEC-2026-89` suppression in [`scripts/security-ignored-vulns.sh`](https://github.com/prebid/salesagent/blob/main/scripts/security-ignored-vulns.sh) (also documented in [`scripts/security-audit.sh`](https://github.com/prebid/salesagent/blob/main/scripts/security-audit.sh)).

## Why suppressed

Python-Markdown advisory fixed in 3.8.1, but the OSV affected-range events array is malformed (missing `{"fixed": "3.8.1"}`), so uv-secure flags every version. We are on Markdown >= 3.8.1.

## Remove when

- [ ] uv-secure / pip-audit no longer flag this ID against our locked version **without** `--ignore-vulns`
- [ ] Remove ID from `UV_SECURE_IGNORED_VULNS` and close this issue

Part of #1228 (C7 residual).

mkostromin-sigma:
## Suppression removal checklist (canonical)

Use this when re-checking #1557 / #1558 / #1559.

### 1. Local gate (uv-secure — matches `ci.yml` `security-audit` job)
```bash
UV_SECURE_IGNORED_VULNS="" ./scripts/security-audit.sh
```
Must exit 0 with **no** active ignores. Covers all three IDs (PYSEC-2026-89, PYSEC-2025-183, MAL-2026-4750).

### 2. pip-audit gate (MAL-2026-4750 only)
Runs in `.github/workflows/security.yml` on **ubuntu-latest** — not reliably on macOS via `uvx` (ensurepip/SIGABRT on py3.13).

**Do not** use `grep ... || echo clean` — a crashed pip-audit looks "clean".

Verify on the removal PR: `security.yml` → **pip-audit** job green with empty `PIP_AUDIT_IGNORED_VULNS`.

Optional local (Docker):
```bash
docker run --rm -v "$PWD:/repo" -w /repo ghcr.io/astral-sh/uv:python3.12-bookworm bash -lc \
  'uv export --no-hashes --format requirements-txt > /tmp/req.txt && uvx pip-audit -r /tmp/req.txt'
```

### 3. Remove + PR
[...]

PR #1560, opened 2026-07-07
chore(ci): #1228 mechanical residuals (E2, H2)

### LINKED ISSUE #1628: chore(ci): #1228 mechanical residuals (E2, H2)
Leaf tracker for #1228 mechanical residuals landed in PR #1560.

### PR
- #1560

mkostromin-sigma:
Closes #1628.

## Summary
- **E2:** remove orphan `.mypy_baseline` (zero readers since #1234)
- **H2:** stale-entry guard for `_ALLOWED_DUPLICATES` in BDD duplicate-steps arch test

## Context
Mechanical in-repo residuals from #1228 status audit (2026-07-07).
C7 tracking issues filed separately: #1557, #1558, #1559.

## Test plan
- [x] `make quality` — 5155 passed


ChrisHuie (commented):
Reviewed at `b551c1e92`. CI is green (30/30) and `make quality` passes at head — verified locally. **E2 is good to merge as-is.** **H2 works, but I'd hold for one required change (DRY) and one design decision (guard completeness)** before it's at our bar. Mechanism and the checks I ran are below.

## E2 — remove `.mypy_baseline` ✅

Verified zero readers: `git grep -i "mypy.baseline"` is empty, a sweep across `*.yml/yaml/toml/ini/sh/py/Makefile` finds nothing, and there's no `mypy-baseline` tool in `pyproject.toml`/`uv.lock`. It's the one-byte `0` file from #1107. Safe deletion — no changes requested.

## Finding 1 (required) — new DRY duplication in the scan loop

**File:** `tests/unit/test_architecture_bdd_no_duplicate_steps.py`

`_collect_step_function_names()` re-implements the same "walk `bdd/steps/` → `ast.walk` → `_is_step_decorated`" loop that `_scan_bdd_steps()` already runs ~10 lines below it in the same file (and that `_iter_step_functions()` runs in `test_architecture_bdd_no_pass_steps.py:98`).

**Why it matters here:** it's the exact pattern `CLAUDE.md` treats as a non-negotiable DRY defect, in a PR whose purpose is allowlist governance. It slips past the R0801 ratchet because the identical runs are interrupted by the `ast.parse` line (never 6 contiguous) — I ran the hook and `tests/` stays at 89 — which is precisely why it needs a human catch rather than the gate.

**Suggested fix (in-file, minimal):** extract one iterator and have both callers use it:

```python
def _iter_step_functions() -> Iterator[tuple[ast.FunctionDef | ast.AsyncFunctionDef, str]]:
    """Yield (step_func_node, repo_relative_path) for every @given/@when/@then under bdd/steps/."""
    for py_file in sorted(_BDD_STEPS_DIR.rglob("*.py")):
        if py_file.name.startswith("_"):
            continue
        tree = ast.parse(py_file.read_text(), filename=str(py_file))
        relative = str(py_file.relative_to(_BDD_STEPS_DIR.parent.parent))
        for node in ast.walk(tree):
            if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and _is_step_decorated(node):
                yield node, relative
```

Then `_collect_step_function_names` becomes `{node.name for node, _ in _iter_step_functions()}`, and `_scan_bdd_steps` groups over the same iterator. (Promoting this into `tests/unit/_architecture_helpers.py` to also fold in the `no_pass_steps`/`no_silent_env` copies is a reasonable follow-up, not required here.)

## Finding 2 (decision needed) — the guard can't see the staleness it's named for

The guard asserts each entry **names a live function** — and that part works: injecting a bogus name into `_ALLOWED_DUPLICATES` fails the test as intended. But it never checks the entry is **load-bearing** (actually part of a 3+ identical-body cluster). Today **7 of the 16 entries are not**:

```
given_registry_format_named          identical-body group size 2   (< 3 threshold)
given_registry_format_no_dimensions  size 2
given_registry_format_no_disclosure  size 1   (unique body)
given_registry_format_no_input_ids   size 1
given_registry_format_no_output_ids  size 1
then_buyer_refs_for_correlation      size 1
then_package_buyer_ref               size 1
```

Removing all 7 from `_ALLOWED_DUPLICATES` leaves `test_no_excessive_duplicate_step_bodies` green — they suppress nothing today, yet the new guard reports them healthy. So the allowlist it protects already carries entries it's blind to, which is at odds with "allowlists can only shrink."

**Decision:** either

- **(a)** strengthen to a load-bearing assertion — with the trade-off that it would also reject entries kept **deliberately** defensive against future variants (the `registry_format` comment reads that way); or
- **(b)** keep the rename/delete scope, state that in the docstring, and separately confirm whether those 7 are dead or intentional (a tracking issue would fit the C7 pattern).

What's worth avoiding is landing it as "the stale-entry guard" while 7/16 non-load-bearing entries pass unremarked.

## Minor (optional)

- `_ALLOWED_DUPLICATES` FIXMEs cite beads ids (`# FIXME(salesagent-ebb5)`); `CLAUDE.md` asks allowlist FIXMEs to cite a GitHub issue/PR so they resolve for outside contributors. Pre-existing, but in the spirit of this PR.
- `found`/`allowlist` in the new test can only ever fire the "stale" branch (`found ⊆ allowlist` by construction); a plain `missing = {n for n in _ALLOWED_DUPLICATES if n not in step_names}; assert not missing` reads simpler. Harmless.

---

**Ran (@ `b551c1e92`):** both guard tests (pass); mutation — bogus allowlist entry → stale test fails as intended; body-group probe + removing the 7 non-load-bearing entries → primary guard still green; R0801 ratchet → `tests/ 89 unchanged`; `gh pr checks` → 30/30 pass.
**Not run:** full `./run_all_tests.sh` — the change is test-only and CI already ran the matrix green.

mkostromin-sigma:
@ChrisHuie — addressed review at new head:

### Finding 1 (required) — DRY ✅
Extracted `_iter_step_functions()`; `_scan_bdd_steps` and `test_allowed_duplicate_entries_still_exist` share it.

### Finding 2 (decision) — option **(b)** ✅
- Stale guard docstring: **rename/delete scope only** (not load-bearing audit)
- Removed 7 non-load-bearing entries (verified @ `b551c1e92`); allowlist 16 → 9
- Tracking: #1561

### Optional ✅
- FIXME beads → `#1561`
- Simplified stale check to plain `missing` assert (no hand-rolled set-diff)

`make quality` green locally. Ready for re-review.

ChrisHuie (approved):

Issue #1447, opened 2026-06-14
Remove local plain-Enum workarounds once adcp ships StrEnum (blocked: adcp-client-python#913 unreleased / #946 deferred)


KonstantinMirin:
## Summary
Our pinned SDK `adcp==5.7.0` declares its enums as plain `Enum`, not `StrEnum`, so
`"created" == CreativeStatus.created` is `False` and any bare string↔enum comparison
silently misbehaves. We carry several workarounds to stay correct. The upstream fix
exists but is **not released**, so the workarounds must stay until then — this issue
tracks removing them once we can.

## Root cause (upstream — fixed but unreleased)
- **adcontextprotocol/adcp-client-python#913** — "Enums (CreativeAction, CreativeStatus,
  etc.) should be StrEnum, not Enum". Merged on `main` (closed), but in **no released
  version**.
- **adcontextprotocol/adcp-client-python#946** — request for a `v5.7.1` patch carrying
  #913 (bundled with #914, #918). **Deferred**: maintainer triage found the fixes are
  entangled with v6 schema work and "cannot be isolated as clean cherry-picks"; whether
  to open a v5.x maintenance track is a release-policy decision (`ready-for-human`, no
  committed 5.7.1).

Net: we're pinned to `5.7.0` plain `Enum`s with no patched release to move to.

## Workarounds to remove when StrEnum ships (a real 5.7.1, or our v6 move)
1. **Local `CreativeStatusEnum`** — a `StrEnum` stand-in for the SDK's plain
   `CreativeStatus`. Delete; import the SDK enum once it's `StrEnum`.
2. **Duplicated enum→str normalization** — the `x.value if isinstance(x, Enum) else str(x)`
   pattern is inlined ~13× in tests (`then_payload.py` ×5, `uc006_sync_creatives.py` ×6,
   +2) and 2× in src (`media_buy_delivery.py`, `products.py`), plus the `enum_value`
   helper. Most of it only papers over plain `Enum`s; with `StrEnum` it collapses.
3. **No guard enforces `enum_value` / null-guard adoption** — a *new*
   `creative.status == CreativeStatus.approved` would silently be `False`. Worth a guard
   only while we're stuck on plain `Enum`s; moot under `StrEnum`.

## Acceptance
- When a StrEnum-carrying adcp release exists: bump the pin, delete `CreativeStatusEnum`,
  remove the redundant normalizations, re-verify `make quality` + creative/account suites.
[...]

Issue #1450, opened 2026-06-15
Reduce the CodeQL alert backlog: logging hygiene + per-PR differential scanning


ChrisHuie:
## Summary
CodeQL reports 477 open alerts repo-wide (all advisory — every CI check passes, none block merge). They are surfaced on every PR by the github-advanced-security review, which obscures genuinely-new findings behind a large pre-existing backlog. This issue tracks triaging and reducing that backlog and making CodeQL-on-PR differential.

## Current state (open alerts by rule)
- `py/log-injection` — 287 (CWE-117: user-influenced values reach logs unsanitized)
- `py/stack-trace-exposure` — 129 (full exceptions/tracebacks logged)
- `py/clear-text-logging-sensitive-data` — 33 (mostly false-positive — logs *about* credentials, e.g. `"Tenant {id} Kevel adapter missing credentials"`, not the secret)
- `py/url-redirection` — 17
- `py/path-injection` — 6
- `py/partial-ssrf` — 3
- `py/polynomial-redos` / `py/clear-text-storage-sensitive-data` — 2

Concentration: ~94% are in the admin Flask blueprints (`src/admin/blueprints/*`), adapters, and tools that log request data. The logging cluster (log-injection + stack-trace + clear-text-logging ≈ 449) is low-severity and largely false-positive; the non-logging classes (url-redirection, path-injection, partial-ssrf ≈ 26) are fewer but individually warrant real security triage and must NOT be dismissed as noise.

## Gap / contract
1. **Logging hygiene (the ~449 cluster):** adopt one repo-wide sanitization seam so user-influenced data cannot forge log lines, and stop logging full tracebacks / apparent secrets. A logging `Filter` at the root logger (or a shared `loggable()` helper — one already exists at `src/core/log_safety.py`) is the lowest-effort path to clear the bulk at once.
2. **Non-logging classes (~26):** individually triage url-redirection / path-injection / partial-ssrf — confirm real vs false-positive, fix the real ones, dismiss the rest with a recorded rationale.
3. **CodeQL-on-PR signal:** make PR annotations differential (flag only alerts a PR's diff introduces), so the pre-existing backlog stops masking new regressions.

## Acceptance
- A logging-sanitization convention is in place and the log-injection / stack-trace counts are materially reduced (or the remainder dismissed with rationale).
- Every url-redirection / path-injection / partial-ssrf alert is either fixed or dismissed with a recorded reason.
[...]

PR #1567, opened 2026-07-08
chore: bump adcp 5.7.0 → 6.6.0 (spec 3.1.1)

KonstantinMirin:
## Summary

Bumps the `adcp` Python SDK **5.7.0 → 6.6.0** (AdCP spec **3.1.0-beta.3 → 3.1.1**), fixes every failure the bump introduced, and (added since the initial review) **merges `origin/main`** (adcp 6.6 × #1545 status taxonomy) and **`perf/parallelize-test-suite`** (faster test suite), then stabilizes the resulting parallelized suite. Lockfile resolves cleanly.

## Spec-grounding

`adcp==6.6.0` targets spec **3.1.1**. `EXPECTED_SPEC_VERSION`, `docs/adcp-spec-version.md`, and the SDK↔spec mapping are updated; the pin-drift guard passes.

## What broke and how it was fixed

| Area | 6.6 change | Fix |
|---|---|---|
| Geo / signals / delivery types | `GeoPostalArea`→`PostalArea`, `GeoPostalAreas`→`PostalAreaSupport`, inline `ByGeoItem`→`GeoDeliveryMetrics` (codegen renames) | rebind imports (aliased) |
| `SignalId` | codegen ordinal churn (`SignalId5`→`SignalId54`) | construct via the stable `SignalId` RootModel (discriminated on `source`), not the ordinal alias |
| `CreateMediaBuySuccess` / `UpdateMediaBuySuccess` | `status`/`confirmed_at`/`revision` now required | spec-correct subclass defaults (single source of truth); fixed A2A + admin construction paths that bypassed the subclass |
| `Product` | 7 new inherited passthrough fields with no DB columns | declared transport-only (no persistence) |
| `SyncCreativeResult` | 6.6 re-added 6 fields to the library parent | inherit all 6; internal `status`→`internal_status` (wire-excluded); schema-inheritance allowlist shrunk **6→0** |
| `Placement` | `kind`/`mode` now required | default `kind="seller_inline"` (legacy rows carry `name`; schema-valid per the 3.1.1 `allOf`) + cross-transport jsonschema-validation test |
| Creative enums | now `StrEnum` | test expectations updated |

Also fixed (via the A2A investigation): admin approve/reject built the raw `CreateMediaBuySuccessResponse`, so the new required-field `ValidationError` was silently swallowed and the buyer webhook never fired — now routes through the defaulted subclass.

## Review feedback (@ChrisHuie) — all addressed

- **Placement default** → `seller_inline` + schema-validation test (`739fbdff0`)
- **`sync_success` DRY** → dropped redundant `setdefault`s (`57efde5f9`)
- **Vacuous exclusion tests** → added `internal_status` guards, mutation-verified (`57efde5f9`)
- **`update_media_buy` pending → completed** → `UpdateMediaBuySubmitted` + cross-transport grading + upstream scenario reconcile (`b8b7e751b`/`1b2f03bc9`/`26e7c69b2`)
- **Reject webhook completed** → embeds `CreateMediaBuyError` + real-route regression test (`2391e8603`)
- **Stale comments** → corrected incl. the MCP null-serialization nuance (`57efde5f9`)
- Follow-ups filed (out of scope): `paused` honor, `iter_git` split-out PR, `push_notification_config.url` validation, #1447 StrEnum cleanup

## Merges included

- **`origin/main`** — #1545 status taxonomy (`_media_buy_status.py`, read-side) coexists with our write-path defaults; `completed_views` rename adopted; verified against the 6.6 delivery-status enum (superset). soupsieve 2.8.4 (#1573) retained.
- **`perf/parallelize-test-suite`** (#1572) — scenario-level BDD xdist, tmpfs test DB, per-worker e2e stacks, schema-fingerprinted DB template. #1572 has since landed on `main`; reconciled by merging `main` back in, keeping our `--dist loadfile` integration fix.

## Parallelized-suite stabilization

Running the parallel suite surfaced pre-existing issues the serial path never exercised — all fixed: guard env-isolation; UC-019 e2e_rest exclusion (no-REST UCs); **integration `--dist loadfile`** for cross-class-coupled migration modules; 12 pre-existing MAIN e2e_rest scenarios ledgered (owner-approved ceiling 308→315, deferred to the e2e_rest retirement epic); `#1270` date-range tripwire graduated (production now validates).

## Verification — full `./run_all_tests.sh` (in-network, parallelized)

| Suite | Result |
|---|---|
| unit | 5322 passed |
| integration | 2107 passed (`--dist loadfile`) |
| bdd_inprocess | 1341 passed |
| bdd_e2e | 99 passed, 169 xpassed, **0 failed** |
| admin | 86 passed |
| e2e | passed |
| ui | 5 passed |

`ruff` + `mypy` clean. The `security-audit` non-zero on the CI box is the box's *global* `uv 0.11.6` (GHSA-4gg8-gxpx-9rph), not a project dependency — box-tooling, updated separately.


ChrisHuie:
Reviewed the bump end-to-end against the v3.1.1 schemas and by running the suites. The spec grounding is correct, and the scope holds up: I introspected the full 5.7→6.6 model+enum surface and every change not explicitly handled here is either unused in `src/`, value-identical, or absorbed by an existing subclass override — so "fix what broke" didn't miss anything material. CI is green and there are no blockers. Catching the swallowed-`ValidationError` admin webhook (raw `CreateMediaBuySuccessResponse` → `ValidationError` under 6.6 → silent skip) is a genuinely good find. Items below are grouped by how I'd handle them.

## Worth folding into this PR (small, and introduced by the bump)

**1. Placement `kind` default is schema-invalid — `src/core/product_conversion.py:455`**
`_placement_defaults = {"kind": "publisher_ref", "mode": "targetable"}`, but `placement.json` (3.1.1) has `allOf: if kind == "publisher_ref" then required: ["publisher_domain"]`, which legacy placements don't carry. `seller_inline` requires only `name`. The SDK doesn't enforce the conditional, so unit tests pass, but a draft-07-validating consumer rejects these objects. `mode="targetable"` is correct (spec: legacy omission is treated as targetable).
Fix: default `kind="seller_inline"` (and confirm legacy placements carry `name`), or set `publisher_domain` when defaulting `publisher_ref`.

**2. `sync_success` restates the class-field defaults — `src/core/schemas/_base.py:279-281` vs `289-291`**
`status`/`confirmed_at`/`revision` are declared as field defaults and again via `setdefault` in the factory. The setdefaults are runtime-redundant (the field defaults alone construct fine) and can drift — raise `revision: int = 1` → `2` at the field and the factory callers would still emit `1` while the direct constructors (`operations.py:477/572`) emit `2`. Dropping the three `setdefault` lines makes the field defaults the single source; keep the `**kwargs` spread — its purpose (dodging the pydantic-plugin `call-arg` on typed callers) is real, which is also why the Create-has-factory / Update-has-none split is correct.

**3. Three exclusion tests no longer guard the renamed field — `tests/unit/test_creative_response_serialization.py:94`, `tests/unit/test_creative.py:277`, `tests/unit/test_response_shapes.py:293`**
After `status`→`internal_status`, these construct `internal_status=...` but still assert `"status" not in dumped` — a different field (the now-inherited spec `status`, which is `None` and omitted for an unrelated reason). Removing `exclude=True` from `internal_status` keeps all three green. Add `assert "internal_status" not in dumped` to restore the real guard.

**4. `paused` is accepted but not honored — `src/core/tools/media_buy_create.py:4302`**
It's threaded into `CreateMediaBuyRequest` on both transports, but `req.paused` is read nowhere on the create path (`_build_create_success` takes a separate `paused` param; the adapters read the *response* package's `paused`, never the request's), so a `paused=true` create still delivers. The field description promises "delivery starts only once unpaused." Either wire it to the initial `media_buy_status`, or reword the description to note it's accepted for 3.1.1 compatibility but not yet honored.

**5. A few comments now describe the old surface**
- `src/core/schemas/creative.py:373` — "the sync wire is byte-identical this bump" holds for A2A/REST, but on MCP the inherited spec `status` serializes as `null` (`structured_content` goes through `to_jsonable_python`, which bypasses the `model_dump` override, so `exclude_none` doesn't apply). `changes`/`warnings` shift from `[]` to `null` there too.
- `src/core/schemas/delivery.py:129,138` still say "extends library ByGeoItem"; the type is now `GeoDeliveryMetrics`. `src/core/tools/capabilities.py:209` says "Build GeoPostalAreas" while building `PostalAreaSupport`. `src/core/tools/signals.py:171` names `SignalCatalogType`; the field is now `SignalAvailabilityType` (identical values, so behavior is unaffected).

## Needs a design decision

**6. `update_media_buy` manual-approval returns `status:"completed"` for a not-yet-applied update — `src/core/tools/media_buy_update.py:382,410`**
Create routes through `CreateMediaBuyResult._serialize`, which overrides the envelope status to `submitted` on the pending path — but the update path returns the raw `UpdateMediaBuySuccess`, so the new `status="completed"` default reaches the wire for an update pending approval (`model_dump(exclude_none)` → `{..., "status": "completed", "revision": 1, ...}`, on all transports). `media_buy_update.py` isn't in this diff, so this is a downstream effect of the new `UpdateMediaBuySuccess.status` default. Gated behind non-default manual-approval config. 3.1.1 models this as `UpdateMediaBuySubmitted` (`status:"submitted"` + `task_id`); consider mirroring create's wrapper on that branch.

**7. Reject webhook body carries `status:"completed"` — `src/admin/blueprints/operations.py:572`**
The reject payload's outer status is correctly `rejected`, but the embedded `result` is a `CreateMediaBuySuccess` (`status:"completed"`, `confirmed_at=<now>`). Since the Success type can't represent a rejection, consider a non-Success body — or send the stored `response_data` the way the primary async path does — so a buyer reading the embedded envelope doesn't see a rejected buy asserting it completed. The approve path is coherent and needs no change.

## Follow-ups (out of scope for the bump)

- The `iter_git_tracked_files` filesystem-walk fallback (`tests/unit/_architecture_helpers.py`) is unrelated to the bump and makes the version-anchor guards depend on untracked files in non-pruned directories: when `git ls-files` fails in the bind-mounted test container it silently walks the tree, so a checkout with untracked files under (e.g.) `.claude/` yields spurious `postgres_image_anchor` / `uv_version_anchor` failures under `./run_all_tests.sh` (CI is unaffected — clean checkout + working git). `git config --global --add safe.directory /app` in `Dockerfile.test` addresses the root cause (verified: container `git ls-files` then returns the tracked set); a positive/negative self-test would pin the fallback. Worth its own PR so it's reviewed on its own merits.
- The buyer `push_notification_config.url` reaches the outbound HTTP client without the URL validation the other ingestion boundaries apply; this fix re-enables the admin approve/reject trigger for it. Pre-existing, but worth a follow-up to validate at the create-ingestion boundary for consistency.
- 6.6.0 ships `StrEnum`, so the plain-`Enum` workaround tracked in #1447 is unblocked — this PR already uses it for `action`; the remaining `CreativeStatusEnum` / normalization cleanup can now proceed.

---

**Ran:** full `./run_all_tests.sh` (admin/e2e/ui green; unit 5276, integration 2095, bdd 1277 passed) + unit locally with real git (5158 passed, 0 failed) + spec-version / AdCP-contract / schema-inheritance guards (88 passed) + a full 5.7→6.6 model-surface delta + v3.1.1 schema/storyboard grounding. The 8 bdd `[e2e_rest]` failures (missing tenant context / wire not stashed / breakdown dims not populated) and the 1 integration caplog failure are transport-harness/observability, not bump-related — consistent with the pre-existing set noted in the description.
**Not run:** the 5.7 baseline behind the byte-identical claim (verified the failure *reasons* are transport-harness rather than adcp-6.6, which supports it).

KonstantinMirin:
Thanks for the thorough review, @ChrisHuie — every item is addressed. Point-by-point below (commit refs inline).

## Folded into this PR ✅

**1. Placement `kind` default schema-invalid** → fixed in `739fbdff0`. Defaulted `kind="seller_inline"` (confirmed legacy rows carry `name`, which is what `seller_inline` requires per the `allOf` in `placement.json` 3.1.1). Added a **cross-transport integration test** that jsonschema-validates emitted `get_products` placements against the 3.1.1 `placement.json` bundled in adcp 6.6, on all four transports — so the SDK-doesn't-enforce-the-conditional gap is now covered.

**2. `sync_success` restates field defaults** → fixed in `57efde5f9`. Dropped the three `setdefault` lines; the subclass field defaults are the single source now. Kept the `**kwargs` spread (the pydantic-plugin `call-arg` dodge) and added a docstring note not to re-default here.

**3. Three exclusion tests guard the wrong field** → fixed in `57efde5f9`. Added `assert "internal_status" not in dumped` to all three. I kept `Field(exclude=True)` on `internal_status` (rather than removing it as the alt-suggestion) — the new assertion is the real guard and mutation-verified (drop the exclude and all three fail).

**5. Stale comments** → fixed in `57efde5f9` / `75f8c373c`: `GeoPostalAreas`→`PostalAreaSupport`, `ByGeoItem`→`GeoDeliveryMetrics`, `SignalCatalogType`→`SignalAvailabilityType`, and reworded the `creative.py` "byte-identical this bump" note to spell out that MCP's `structured_content`→`to_jsonable_python` path bypasses the `model_dump` override, so the inherited `status`/`changes`/`warnings` serialize as `null` there (not omitted).

## Design-decision items — implemented ✅

**6. `update_media_buy` pending → `status:"completed"`** → fixed in `b8b7e751b`. Added `UpdateMediaBuySubmitted` (extends the SDK `UpdateMediaBuySubmittedResponse`), returned from the manual-approval branch with `status="submitted"` + `task_id=step_id`, extended the union, and taught A2A reconstruction the submitted variant. Then **graded it cross-transport** (`1b2f03bc9`) — the BR-UC-003 manual-approval scenarios were dormant ("No harness wired"), so I wired them on a2a/mcp/rest and reconciled the over-specified assertions upstream (`26e7c69b2`): the submitted envelope carries `task_id`, **not** `media_buy_id`/`implementation_date` (matches storyboard POST-S8). This also exercised the new A2A reconstruction branch.

**7. Reject webhook embeds `status:"completed"`** → fixed in `2391e8603`. The embedded result is now a `CreateMediaBuyError` (spec Error variant) carrying the rejection reason. Added a regression test that drives the **real admin reject route** and asserts the outbound body's embedded result is not a completed success.

## Follow-ups — filed, out of scope for the bump ✅

- **`paused` accepted-but-ignored** (#4): traced it — request-level `paused` is 2.12.0-lineage, honored nowhere on create *before* this bump, so it's a pre-existing gap, not a 6.6 regression. Filed as a bead for a dedicated fix/reword.
- **`iter_git_tracked_files` fallback**: agreed it deserves its own PR; your `git config --global --add safe.directory /app` root-cause is captured in the bead. (Note: the parallelization work below made `./run_all_tests.sh` run in-network on a CI box, which is a cleaner path than the worktree fallback — I'll reconcile the two in that split-out PR.)
- **`push_notification_config.url` ingestion validation**: filed (pre-existing SSRF-surface follow-up).
- **#1447 StrEnum cleanup**: filed; 6.6 unblocks the remaining `CreativeStatusEnum` / normalization work.

See the follow-up comment for the additional merge + test-parallelization work now included in this PR.

KonstantinMirin:
## Additional work since the review — merges + test parallelization

The PR grew beyond the pure bump. Two branches were merged in and the resulting parallelized suite was stabilized. Full board is green.

### Merged `origin/main` (adcp 6.6 × #1545 status taxonomy)
Semantic per-file merge (one reviewer per overlap file, traced to each side's commit intent). Key integration: #1545's new `_media_buy_status.py` (read-side status refinement) coexists cleanly with our write-path status defaults — no overlap on any return path. #1545's `completed_views` rename (was `video_completions`) is adopted in `delivery.py` + `mock_ad_server`; verified against the adcp 6.6 delivery-status enum (a superset — `pending`/`failed`/`reporting_delayed` all present). soupsieve 2.8.4 CVE bump (#1573) retained.

### Merged `perf/parallelize-test-suite` (faster suite)
Scenario-level BDD xdist distribution, tmpfs test DB, per-worker e2e stacks, schema-fingerprinted DB template. Big wall-clock win on the in-network runner. (This is #1572, which has since landed on `main` — reconciled by merging `main` back in and keeping our `--dist loadfile` integration fix; main's version lacks it and still has the cross-class migration-coupling bug.)

### Stabilized the parallelized suite
Running the parallel suite surfaced pre-existing issues the serial path never exercised — all fixed:

- **Unit guard** (`test_bdd_e2e_enabled_xdist_guard`): the runner exports `E2E_PER_WORKER=1` globally, which leaked into the shared-server guard test. Isolated the env + added an explicit per-worker-allowed case.
- **UC-019 e2e_rest 404s** (40): no-REST UCs (`get_media_buys`) dropped plain REST but still got `e2e_rest` appended. Fixed — `e2e_rest` is a REST transport, so excluded for no-REST UCs (`and not no_rest`).
- **Migration xdist coupling**: `test_timestamptz_migration` / `test_delivery_measurement_migration` share a module-scoped `migration_db` across ordered/cross-class tests; under `-n auto` (default `--dist load`) they split across workers. Fixed at the tox level — **`--dist loadfile`** keeps each integration file on one worker in order (verified: both modules green).
- **e2e_rest ledger**: 12 pre-existing MAIN scenarios (UC-004 #1545, UC-005 #1479, UC-018 BR-RULE-034 #1551) are exercised over real HTTP for the first time and fail for transport-inherent reasons (mock-injection invisible, auth/tenant-context, wire-not-stashed) — **none from the 6.6 bump**. Ledgered (owner-approved one-time ceiling recalibration 308→315, deferred to the e2e_rest retirement epic). Separately, the `#1270` date-range strict-xfail tripwire **XPASSed deterministically** (production now validates date range over e2e_rest) → graduated.

### Verification — full `./run_all_tests.sh` (in-network, parallelized)
unit ✅ · integration ✅ · bdd_inprocess ✅ · bdd_e2e ✅ (0 failed) · admin ✅ · e2e ✅ · ui ✅. `ruff` + `mypy` clean.

The `security-audit` non-zero on the CI box is the box's *globally-installed* `uv 0.11.6` (GHSA-4gg8-gxpx-9rph) — not a project dependency (the project's 249 deps are clean); box-tooling, being updated separately.

ChrisHuie:
Re-reviewed the update against the v3.1.1 schemas and by running the full suite. The prior feedback is largely addressed — the `sync_success` dedup, the exclusion-test assertions, the `seller_inline` placement default (with a jsonschema test), and the stale comments all landed; `UpdateMediaBuySubmitted` is spec-correct (`task_id`, no `media_buy_id`/`implementation_date`) and the previously-dormant BR-UC-003 manual-approval scenarios now run and pass on a2a/mcp/rest. Suite confirms no regression (bdd 0-failed; the e2e_rest transport failures are correctly ledgered). One coherent cluster is worth finishing before merge — the new required-field defaults were swept into the update and reject paths but not the create path — plus a few cleanups.

**Worth finishing before merge (same required-field-default issue, across all sibling paths):**

1. **Reject webhook emits an internal error code.** `operations.py:574` builds `Error(code="MEDIA_BUY_REJECTED", ...)`, but that code is in `INTERNAL_CODES` ("never reach the buyer agent") and maps to `POLICY_VIOLATION`; the tool path emits `POLICY_VIOLATION` for this same event, so the webhook diverges and sends the buyer an internal token. Fix: `Error(code=translate_error_code("MEDIA_BUY_REJECTED"), ...)`, or route the rejection through `AdCPMediaBuyRejectedError` so the wire code is derived. And pin it in the reject test — it currently asserts the embedded status isn't `completed` but not the code; add `assert embedded["errors"][0]["code"] == "POLICY_VIOLATION"`.

2. **Create pending-approval still emits `confirmed_at`.** `media_buy_create.py:3011` / `:3166` return `CreateMediaBuySuccess.sync_success(...)` on the awaiting-approval branch, so the wire carries `confirmed_at=<now>` + `revision=1` on a `status:"submitted"` (not-yet-committed) buy — the same shape you fixed for `update_media_buy` (→ `UpdateMediaBuySubmitted`) and reject. Your reject test even asserts `not embedded.confirmed_at`. Mirror the fix here: a `CreateMediaBuySubmitted`, or drop `confirmed_at`/`revision` when the `CreateMediaBuyResult` status is `submitted`.

3. **MCP `sync_creatives` serializes `changes`/`warnings` as `null`.** They're now inherited with a `None` default, and the MCP `structured_content` path (`to_jsonable_python`) bypasses the `model_dump` override, so both emit `null` on the wire (the spec types them `array`). Restoring `Field(default_factory=list)` gives valid `[]` on every transport. (The inherited `status` serializing as `null` on MCP is the broader `structured_content` behavior — a separate follow-up if you want repo-wide None-stripping there.)

4. **`paused` docstring over-claims.** `media_buy_create.py:4318` still reads "delivery starts only once unpaused", but nothing on the create path reads `req.paused`. Reword to something like "accepted for AdCP 3.1.1 compatibility; pause-on-create not yet honored" (the implementation can stay the tracked follow-up).

**Small cleanups (fold in):**
- `SyncCreativesResponse.dry_run` (`creative.py:474`) is redeclared identically to the 6.6 parent with a "SDK 5.7 removed dry_run" comment that's no longer true — inherit it like `context`/`ext`.
- Approve path (`operations.py:477`) constructs `CreateMediaBuySuccess(...)` directly while the other sites route through `.sync_success()` — use the factory for uniformity.
- The new e2e_rest ledger comments reference internal tracker ids; add the GitHub issue for the retirement epic so the reference resolves for outside contributors. The ceiling comment reads "308 → 315" while 12 entries were added (5 consumed prior slack) — wording.
- A couple of docstrings describe the old surface: the placement test's pre-fix attribution (the `publisher_ref` default came from the `Placement` model, not `product_conversion`), and the submitted unit test's "the model dump is the wire on every transport" (not true for A2A, which conveys the submitted state via the Task, not a serialized body).
- `.type-ignore-baseline` lost its trailing newline (churn risk).

**Follow-ups (out of scope for the bump):**
- No MCP-wire jsonschema oracle for `sync_creatives` (the placement + exclusion tests validate the model/payload, not the MCP wire bytes); the `__init__.py` rebuild-loop invariant has no failing oracle (and import-order dependence is worse under xdist); the A2A reject branch has no test; the A2A `UpdateMediaBuySubmitted` reconstruction branch is unreachable (the submitted early-return precedes it) — remove it or add an A2A submitted wire test.
- The `seller_inline` placement default assumes legacy rows carry `name`; a fixture for a row missing `name` would cover the other shape.
- Guard hardening: the e2e_rest ledger ceiling is decrease-only by convention only (nothing stops a silent raise); the schema-inheritance redefinition guard skips subclasses whose name differs from the library base, so `SyncCreativesResponse`'s parent-field overrides go unchecked; and the `iter_git_tracked_files` filesystem fallback isn't hermetic (it descends into `.claude/` — prune it, or set `git safe.directory` in the test container so `git ls-files` works).

---

**Ran:** full `./run_all_tests.sh` on the head — admin / e2e / ui green; bdd 0-failed (the e2e_rest transport artifacts are ledgered as xfail); integration has the one pre-existing `test_non_scalar_concept_value_is_dropped` caplog artifact (passes in CI in-process); unit passes on a real-git checkout (5202 passed, 0 failed; the in-container version-anchor-guard failures are the `iter_git_tracked_files` fallback descending into local untracked dirs, not a code issue). 395 architecture + boundary guards green; the update-submitted path graded live on a2a/mcp/rest; DRY baseline dropped 89→87 (no new clones). **Not run:** the 5.7 baseline behind the byte-identical-BDD claim.

KonstantinMirin:
Thanks for the second pass, @ChrisHuie — the whole round is addressed. Point-by-point below (commit refs inline).

## Blockers ✅

**1. Reject webhook internal error code** → fixed in `4f60cbf4c`. The rejection now routes through the typed `AdCPMediaBuyRejectedError`, so the wire code is derived (`POLICY_VIOLATION`), never hand-picked — same source of truth as the tool path. The reject regression test now pins the code exactly as you suggested (`embedded["errors"][0]["code"] == "POLICY_VIOLATION"`).

**2. Create pending-approval emitting `confirmed_at`/`revision`** → fixed in `20e5b60d8`. Mirrored the update-path fix: both manual-approval branches now return a `CreateMediaBuySubmitted` (extends the SDK submitted variant; `task_id` only — no `media_buy_id`/`confirmed_at`/`revision`, matching the storyboard shape we reconciled for update). Two things this surfaced along the way:
- **Idempotency replay of a submitted create was silently minting a second workflow step** — a retry with the same `idempotency_key` now replays the original submitted envelope instead of opening a duplicate approval. Covered by a wire-level regression test.
- The `T-UC-002-alt-manual` BDD scenario was dormant; it's now wired live across a2a/mcp/rest/e2e_rest, so the create-submitted path is graded end-to-end the same way the update one is.
- Added a structural guard (`test_architecture_submitted_envelope_variant.py`) so the next tool that wraps a pending-approval result in a confirmed Success fails `make quality` immediately — this class of bug is now caught mechanically, not by review.

**3. MCP `changes`/`warnings` as `null`** → fixed in `90df7295f`. Redeclared with `default_factory=list` (plus `errors`, same family), so every transport emits spec-valid arrays. Also added the MCP-wire jsonschema oracle you flagged as missing — the regression test drives the real `structured_content` path and validates the wire bytes, not the model.

**4. `paused` docstring** → fixed in `371309553`. Now reads "accepted for AdCP 3.1.1 compatibility; pause-on-create is NOT yet honored", on both transports. The implementation stays the tracked follow-up.

## Cleanups ✅

- **`dry_run` redeclaration** → removed in `5a8953a46`; inherited from the 6.6 parent (verified identical type + default), comment block rewritten to describe the current surface.
- **Approve path via `.sync_success()`** → done in `4f60cbf4c`, and both `TODO: @yusuf` items at that site are resolved (context is echoed, audit metadata is passed and typed) with a new approve-wire test pinning the body.
- **Test docstrings** → `d8c05cbee`. The submitted test's "model dump is the wire on every transport" claim is now per-transport-precise (A2A conveys state via the Task; MCP goes through `to_jsonable_python`). On the placement test's attribution: I checked the git history and the `publisher_ref` default *did* come from `product_conversion._placement_defaults` (`739fbdff0` shows it), so I kept the attribution and documented the trace rather than changing it.
- **Legacy placement missing `name`** → also `d8c05cbee`. Nameless legacy rows now get a defined `name := placement_id` fallback (fail-loud if neither exists) with a fixture covering that shape, so `seller_inline` can't emit a schema-invalid placement.
- **Dead A2A `UpdateMediaBuySubmitted` reconstruction branch** → removed in `c877d796a` after a full concrete-request trace confirmed the submitted early-return makes it structurally unreachable (A2A conveys submitted via the Task, never a serialized result body). A comment at the site records why the variant is absent, and guard tests pin the reachable set.
- The comments these fixes introduced cite the PR/review items instead of internal tracker ids (`d25b99890`), per the convention you raised for the ledger.

## Deferred (tracked, not in this round)

- The e2e_rest ledger GH-issue reference + ceiling wording, and the `.type-ignore-baseline` trailing newline — bundled into the e2e_rest retirement epic and the split-out infra PR respectively.
- Guard hardening (ledger ceiling ratchet, schema-inheritance guard's name-match blind spot, `iter_git_tracked_files` hermeticity) and the repo-wide MCP None-stripping question — filed as follow-ups.

## Also fixed along the way (infrastructure the new tests exposed)

- A sibling branch's creative-agent image was silently shadowing this branch's pin on the shared CI box → image tags are now pin-keyed (`4a0c4dac2`).
- `./run_all_tests.sh ci <target>` had never worked on the box (docker-compose v1 absent; `pipefail` swallowed the error) → v2 fallback (`b3b15b8eb`).
- An e2e test leaked manual-approval adapter state into the shared `ci-test` tenant, randomly poisoning later tests → `try/finally` restore (`ad87aff05`).

---

**Ran:** full `./run_all_tests.sh` on the head — all 7 suites, 0 failures (unit 5323, integration 2122, bdd_inprocess 1344, bdd_e2e 100, e2e 91, admin 86, ui 5). Four BDD scenarios graduated from dormant to graded. The one pre-existing non-test failure is the uv-secure dependency audit (click/pillow CVEs) — tracked separately, not introduced here.

ChrisHuie:
# Review — #1567 (verified at head `d25b9989`)

Re-reviewed the round-3 changes against the pinned 3.1.1 schemas + the published conformance storyboards, tracing each path end-to-end and dumping the wire on every branch. The three round-2 blockers are genuinely and completely addressed — I confirmed each oracle goes red under the reverting mutation, not green-by-construction: the reject webhook derives `POLICY_VIOLATION` through the typed error (no hand-picked code), the pending create emits a `task_id`-only `CreateMediaBuySubmitted` (with `media_buy_id`/`confirmed_at`/`revision` correctly withheld and revealed on the completion artifact), the idempotency replay discriminates submitted-vs-success so a retry can't mint a second workflow step, and MCP `sync_creatives` now emits arrays via the real `structured_content` path. The submitted→approve lifecycle threads `task_id` coherently, and the new structural guard flags all three inline disease forms.

One 3.1.1 conformance gap is worth closing before merge; the rest are fold-ins and pre-existing follow-ups.

## Close before merge — spec-conformance

**`media_buy_status` is not emitted on the create-success wire — `media_buy_create.py:4002`.** The lifecycle value is computed at `:3570` (`_determine_media_buy_status`) and used only for `valid_actions` at `:4005`, but never passed to `sync_success()`, so the wire omits it. The 3.1.1 conformance storyboard `media-buy/scenarios/pending_creatives_to_start.yaml` (step `create_buy_no_creatives`) grades it explicitly:

    - check: field_value
      path: "media_buy_status"
      value: "pending_creatives"
    - check: field_value
      path: "status"
      value: "completed"

`status: "completed"` (envelope task-status) is emitted correctly; `media_buy_status` (the non-deprecated lifecycle field the spec expects from 3.1 onward) is the missing half. The omission predates this PR, but this is the 3.1.1 migration and `abfa3ab1e` rewrote this exact construction site — the value is already in scope. Fix:

```diff
 adcp_response = CreateMediaBuySuccess.sync_success(
     media_buy_id=response.media_buy_id,
     packages=response_packages,
+    media_buy_status=media_buy_status,
     valid_actions=valid_actions_for_status(media_buy_status),
     creative_deadline=getattr(response, "creative_deadline", None),
     context=req.context,
     errors=property_list_unsupported_advisories(req.packages, adapter),
 )
```

Verified this emits `media_buy_status: "pending_creatives"` alongside `status: "completed"`, both distinct. Mirror on the dry-run branch at `:3508`. The local mirror scenario `@T-UC-002-v31-success-status-values` is dormant (its steps have no binding, so it auto-xfails) — binding it gives this a failing oracle.

## Fold-ins

**`replayed: false` leaks on a fresh submitted response — `_base.py:427-439`.** The adcp 6.6 submitted base declares `replayed: bool = False`; the success base does not. So a fresh submitted create serializes `replayed: false` while a fresh success omits it — contradicting the comment ("omitted when False so fresh responses are byte-identical") and its "Only valid on a successful result" (the replay test asserts `replayed is True` on a submitted replay). Spec-valid (`additionalProperties: true`), so no wire breakage — it's a false load-bearing docstring in a serializer. Single-source it:

```diff
     def _serialize(self, serializer, info):
         result = self.response.model_dump(mode=info.mode, context=info.context)
         result["status"] = self.status
-        if self.replayed:
-            result["replayed"] = True
+        # The SDK submitted base carries replayed=False; keep the marker single-sourced
+        # on the wrapper so fresh responses omit it uniformly across all variants.
+        result.pop("replayed", None)
+        if self.replayed:
+            result["replayed"] = True
         return result
```

(Or `exclude=True` on the submitted subclass, or just correct the comment — any single-sources it.)

**Two byte-identical `CreateMediaBuySubmitted` branches — `media_buy_create.py:3024-3032` and `:3164-3172`.** The manual-approval and config-approval branches build the identical `CreateMediaBuyResult(response=CreateMediaBuySubmitted(task_id=step.step_id, context=req.context, errors=property_list_unsupported_advisories(...)), status=submitted)` + `return _cache_and_return(...)`, differing only by a comment. The R0801 duplication hook stays under threshold, so it won't flag it. Extract `_submitted_approval_result(step, req, adapter)` and call from both.

**Stale `CreateMediaBuySuccess` redeclarations — `_base.py:302-315`.** All five fields (`account`, `sandbox`, `creative_deadline`, `valid_actions`, `context`) are back on the 6.6 parent, so the "SDK 5.7 removed these" comments are stale, and `account: Any | None` weakens the parent's typed `Account` (`creative_deadline: datetime` weakens `AwareDatetime`). This is the cleanup already done for `product.py` — inherit the typed fields (confirm no caller assigns a non-`Account`/naive value first) and drop/fix the comments. Same stale-comment pattern lives in `account.py:148-164` (`SyncAccountsResponse`, untouched file) — worth a separate sweep.

**Context-echo idiom duplicated — `operations.py:494-497` and `creatives.py:180-183`.** This PR added the second copy of the `ContextObject.model_construct(**request_data["context"])` reconstruction. Extract one helper (e.g. in `admin/utils.py`) and call from both blueprints.

**Approve context-echo has no oracle — `operations.py:494-497`.** The approve test's fixture has no `"context"` in `request_data`, so the non-None echo path never executes; reverting to the old `context={}` keeps every test green. Add `"context"` to the fixture and assert `body["result"]["context"]` echoes it.

**MCP-wire oracle omits `errors` — `test_sync_creatives_mcp_wire_shape.py:54`.** The never-null loop checks only `("changes", "warnings")`, but the fix redeclared `errors` with `default_factory=list` too (same MCP null risk). Add `"errors"` to the tuple so a revert reddens.

**New guard needs a self-test and misses two forms — `test_architecture_submitted_envelope_variant.py`.** It has no positive/negative fixture (an empty allowlist + clean tree reads as "caught" even if the matcher silently breaks), and `_status_is_submitted` misses the bare `status=AdcpTaskStatus.submitted` (StrEnum, no `.value`) and a hoisted-variable response. Add a positive fixture (assert a `Success`-under-`submitted` is flagged) and a negative one, and model the bare-enum member.

**A2A reject webhook branch is untested — `operations.py:612-618`.** `create_a2a_webhook_payload` has zero references in `tests/`, and the reject fixture hardcodes `protocol: "mcp"`, so only the MCP branch is exercised. The A2A envelope framing (protobuf `Task`) differs from MCP, so the passing MCP test doesn't cover it — the path is correct at runtime, it's just unpinned. Mirror the MCP reject test with `protocol: "a2a"` and assert the Task artifact carries `POLICY_VIOLATION`.

**A2A submitted `NOT contain` BDD assertions are vacuous — the A2A transport grades a harness-synthesized dict.** Field-absence is architecturally guaranteed on A2A (`del task.artifacts[:]`) and covered for the update path by the shared-code guard, so this is defense-in-depth for the create side: add `assert not env.last_a2a_task.artifacts` to the A2A submitted case (that field was added this round for exactly this).

## Follow-ups (pre-existing, tracked separately)

- `push_notification_config.url` ingestion still bypasses the SSRF validation the admin paths apply (`principals.py`/`settings.py`/`tenants.py` use `WebhookURLValidator`) — the previously-deferred item; confirm the tracking issue exists.
- Separately, `media_buy_create.py:2037` logs the full push-notification config — including `authentication.credentials` (the buyer's webhook bearer token) — at INFO. Pre-existing (from #1107); filing separately as it's a credential-exposure surface, not part of this bump.
- The schema-inheritance redefinition guard keys on the `Library*` alias name, so ~96 field-redeclaring subclasses (incl. `CreateMediaBuySuccess`, `SyncCreativesResponse`) go unchecked — key on the MRO base instead.
- The e2e_rest ledger ceiling is decrease-only by convention only; it's now at cap (315 == 315), which trips any new entry, but a git-history monotonicity assertion (or requiring the bump to cite an issue) would harden it.
- `iter_git_tracked_files` (`_architecture_helpers.py`) is fail-open and non-hermetic — it silences git stderr and its prune set omits `.claude`, so a git-unavailable host silently walks untracked files. Add a visible warning + prune `.claude`, or set `git safe.directory` in the test container.

## Minor nits

- `operations.py:570` — `attributes.flag_modified(media_buy, "status")` is redundant for a scalar `String` column (the approve branch sets `media_buy.status` without it); the append+flag on the JSON `comments` column is the load-bearing one.
- `test_admin_media_buy_reject_webhook.py` — the `async def _capture` closure is defined three times; hoist one to module scope. The 12-mock scaffold could collapse to a class-level `webhook_service` fixture.
- A few `salesagent-*` tracker ids remain in `tests/` docstrings/comments; the `src/` conversion was complete.

## Confirmed solid

The reject regression drives the real admin route and pins the wire code, reason, and no-completed-embed; the submitted-envelope guard flags all three inline disease forms; `SignalId` is constructed via the stable RootModel (avoids the churning ordinal alias); the placement `seller_inline` default has a fail-loud `name := placement_id` fallback with a jsonschema test on all four transports; the dead A2A reconstruction branch was removed with a documented absence + a guard on both create and update sides; the `KNOWN_OVERRIDES` movement (`changes`/`warnings`/`errors`) is sanctioned nested-serialization redeclaration and load-bearing.

--- Verified at `d25b9989`:

Ran: introspected the adcp 6.6 model/enum surface and the pinned 3.1.1 create/sync/placement schemas; fetched and read the `pending_creatives_to_start` + `create_media_buy_async` storyboards at tag `v3.1.1`; traced the reject/approve/submitted/idempotency-replay paths and dumped the serialized wire on each; verified the `media_buy_status` and `replayed` fixes; ran the submitted-envelope + a2a-update-submitted guards (green).

Not run: the full suite (CI is green on this SHA — 32/32) and the upstream conformance-storyboard runner (TS-only). The `media_buy_status` storyboard scenario carries a `requires_capability: media_buy.creative_approval_mode` gate, so whether the runner exercises it against this seller depends on the declared capability set — the field-level grade and the fix stand regardless.

ChrisHuie:
# Review follow-up — provenance pass (head still `d25b9989`)

To lock what lands in this PR versus what rides the tracker — keeping the merge close without dropping anything — I re-graded each finding's provenance against `origin/main` (per-item blame/pickaxe), introspected the 5.7 and 6.6 model surfaces side by side, pulled the `pending_creatives_to_start` storyboard at both pins, and re-dumped the wire on this branch. Every finding reproduces at head. Two change buckets; the rest hold as filed.

## Bucket changes

**`iter_git_tracked_files` moves out of "pre-existing follow-ups" — it's this PR's code.** `main`'s helper fails loud (`raise RuntimeError(f"git ls-files failed in {repo}")`). The `stderr=DEVNULL`, the silent `_iter_files_fallback()` walk, and the prune set missing `.claude` all arrived with `a05a2a628` on this branch (they're what lets the guard suite run inside the in-network container, where the bind-mounted worktree's `.git` back-reference is unreachable). So it can't defer as inherited debt: either fold the fix here (log a visible warning when the fallback engages + add `.claude` to `_FALLBACK_PRUNE_DIRS` — a few lines) or lift the fallback out of this PR into the split-out already on the follow-up list. Folding here is the smaller motion.

**`media_buy_status` keeps the two-line fix, loses the "close before merge" framing.** Deeper trace: the 3.1.0-beta.3 storyboard grades the same step the same way (`field_value: media_buy_status = pending_creatives`, with the legacy `status` slot as `field_value_or_absent`), so the previous pin already failed this grade — the 3.1.1 migration neither created nor widened the gap, and `main` never carried the lifecycle on this wire either (its `status=media_buy_status` at `media_buy_create.py:3592` feeds the DB record, not the response). It's also already tracked: **#1326** (open; documented at #1247 gap #12), whose first done-checkbox is exactly this wire emission. So: not a bump regression, no merge gate. I'd still take the emission here as a partial #1326 — it's verified working, and it's total: `_determine_media_buy_status` returns only `completed` / `pending_creatives` / `pending_start` / `active`, all members of the 6.6 `MediaBuyStatus` enum, so no path can fail construction. The rest of #1326 (the `sync_creatives` transition, webhook, `get_media_buys` visibility) stays on the issue. One correction to my own suggestion there: don't bind `@T-UC-002-v31-success-status-values` as-is — its steps assert the beta.3 deprecation-window shape (lifecycle in the legacy `status` slot), so bound unmodified it would grade the wrong field under 3.1.1. It needs the same spec-reconcile UC-003 got in `26e7c69b2`; the binding goes to #1326 with the rest.

## Fold-in provenance — confirming the in-PR set

- **`replayed: false`** — attribution fix on my side: the 5.7 submitted base declares `replayed = False` too; what's new is emitting the Submitted variant at all (`20e5b60d8` — `CreateMediaBuySubmitted` has zero references on `main`), which is when the leak became reachable. Re-dumped: fresh submitted serializes `{"replayed": false, "status": "submitted", "task_id": …}`; fresh success omits it. Introduced here → fold in here.
- **Duplicate Submitted branches** (`:3024`/`:3164`) — both sides new on this branch. Stands.
- **Context-echo duplication** — blame: the `operations.py:494-497` copy is `4f60cbf4c` (this branch); the `creatives.py:180-183` twin is `main` history (`1ba810775`). This PR added the second copy, so the extraction belongs here. Stands.
- **Approve-echo oracle, MCP `errors` tuple, guard self-test + bare-enum matcher, A2A artifact-absence assert** — all anchor to code and tests added on this branch. Stand as filed.
- **A2A reject-webhook coverage** — the `protocol == "a2a"` branch predates this PR (`main` has it), but what it embeds is exactly what this PR changed (the typed `CreateMediaBuyError` carrying `POLICY_VIOLATION`), and that change has no A2A pin: `create_a2a_webhook_payload` has zero references under `tests/`, and the reject fixture exercises only the MCP branch. The mirrored `protocol: "a2a"` test lands here.
- **`_base.py:302-315` stale redeclarations** — the redeclarations predate the branch; the *staleness* is the bump's doing (6.6 re-added all five, typed: `account: Account`, `creative_deadline: AwareDatetime`, typed `valid_actions`/`context`, `sandbox`). Same cleanup this PR already shipped for `Product` and `SyncCreativeResult`, so finish it here: correct the "SDK 5.7 removed these" comments now, take the typed re-inheritance behind the caller audit. The `account.py:148-164` twin is the same bump-created staleness (6.6 re-added `accounts`/`dry_run`/`context`/`ext` there too — introspected), but in a file this PR never touched — that one goes to the sweep issue below.

## Deferral ledger — a resolvable number for everything that doesn't land here

Confirmed existing:

- SSRF on `push_notification_config.url` ingestion → **#1578** (open; #1589/#1095 are the broader outbound-fetch umbrella). The "confirm the tracking issue exists" box is checked.
- `create_media_buy` status lifecycle → **#1326** (open), as above.
- StrEnum cleanup → **#1447** (open).

To file — nothing on this list resolves on GitHub today, and local tracker ids don't resolve for outside contributors:

1. Credential exposure: `media_buy_create.py:2037` logs the full push-notification config — `authentication` block included — at INFO. Pickaxe confirms it's #1107-era, untouched since.
2. Schema-inheritance guard keys on the `Library*` alias, so `AdCP*`-aliased parents (`CreateMediaBuySuccess`, `SyncCreativesResponse`, …) escape the redeclaration check — key on the MRO base.
3. e2e_rest ledger hardening: count and ceiling both sit at 315 (verified at cap), and decrease-only is comment-convention — add the monotonicity assertion, and give the retirement epic a GitHub number.
4. `paused`-on-create honor (currently documented-unsupported in the param description only).
5. 5.7-era redeclaration sweep: `account.py:148-164` plus any remaining "SDK 5.7 removed/dropped" comments.

Scoped correction on the tracker-id nit: `src/` is fully converted (zero ids added by this change set remain), while `tests/` still carries them on 21 added lines — more than "a few", still cosmetic; sweep with the same treatment `d25b9989` gave `src/`.

## Merge alignment

In this PR: the fold-in set (replayed single-sourcing, Submitted-branch extraction, context-echo helper, approve-echo oracle + fixture `context`, MCP `errors` tuple, guard fixtures + bare-enum matcher, A2A artifact-absence assert, A2A reject mirror test, `_base.py` redeclaration cleanup, iter_git warning + `.claude` prune) plus the two-line `media_buy_status` emission citing #1326. The nits ride the same edits: the `_capture` hoist with the webhook-test changes, the redundant `flag_modified(media_buy, "status")` with the `operations.py` changes, the test-side tracker-id sweep with the docs pass. Everything else is the ledger above. Largest single item is one mirrored test — nothing here moves the timeline.

--- Verified at `d25b9989` (head unchanged):

Ran: per-item blame/pickaxe against `origin/main` for every provenance call above; adcp 5.7.0 vs 6.6.0 model/enum introspection (create Success/Submitted, `SyncAccountsSuccessResponse`, `MediaBuyStatus`); `pending_creatives_to_start` at 3.1.0-beta.3 and 3.1.1; branch-env wire dumps for Submitted/Success with and without the `media_buy_status` kwarg; `_determine_media_buy_status` return-value totality against the 6.6 enum; tracker sweep for the deferral set.

Not run: the full suite (this pass changes no code) and the TS storyboard runner — both unchanged from the prior footer.

KonstantinMirin:
Thanks Chris — both passes were spot on, especially the provenance follow-up; everything below landed on this branch. Point by point, in your order:

**`replayed: false` leak on fresh submitted** — fixed with your diff (the `pop` in `CreateMediaBuyResult._serialize`), and the docstring you called out as doubly wrong is rewritten: the marker is wrapper-owned, valid on success AND submitted replays, omitted when False on every variant (178395dcf). The wire oracle now asserts the key is absent on both approval branches and is mutation-verified (neutering the pop turns 3 integration tests red). While tracing similar cases I found `UpdateMediaBuySubmitted` leaks the same `replayed: false` (returned bare at media_buy_update.py, no replay machinery at all) — filed separately rather than folding in here.

**media_buy_status emission** — agreed on the provenance: not a bump regression, landed here as a partial #1326 (74bd7f53d). Both sites emit it from the same value that feeds `valid_actions` (the dry-run branch through a single coupled local so the two can't diverge). Commit cites the 3.1.1 response schema + the `pending_creatives_to_start.yaml` step `create_buy_no_creatives` and notes the `media_buy.creative_approval_mode` capability gate. The dormant `@T-UC-002-v31-success-status-values` scenario is NOT bound — as you said it asserts the beta.3 deprecation-window shape; its binding goes with the rest of #1326 after a spec-reconcile. Plain wire regressions (TDD-red first) cover both branches instead.

**A2A reject branch unpinned** — mirrored the MCP reject regression with `protocol: "a2a"` through the real admin route (2f3bd71a6). It asserts on the actual protobuf Task `create_a2a_webhook_payload` emits: rejected state, POLICY_VIOLATION + reason in the artifact DataPart, no completed-Success embed. Mutation-verified by repointing the branch at a raw Success.

**Approve context-echo had no oracle** — the fixture now carries a buyer context and the test asserts the outbound body echoes it verbatim; the absent branch is pinned too (deffdc3ef). Reverting the echo to `context={}` goes red, which is exactly the hole you spotted.

**sync_creatives never-null oracle omitted errors** — added to the tuple (8016751b2); reverting only the errors redeclaration in creative.py now turns the test red (that revert used to keep everything green).

**Submitted-envelope guard self-test + missed forms** — the guard got known-bad/known-good inline fixtures via the shared detector-snippet helper, plus the two forms you named: bare `AdcpTaskStatus.submitted` StrEnum member and the hoisted-response-variable form (cd19fc3a0). Lobotomizing the matcher turns the positive self-test red.

**iter_git_tracked_files fallback** — taken into this PR per your provenance correction (the fail-open walk arrived with a05a2a628 here; main fails loud). The fallback now warns loudly naming the repo and the git failure, `.claude` is pruned, and positive/negative self-tests pin both paths (ad460179a). The root-cause alternative (`safe.directory` in Dockerfile.test) stays with the split-out infra PR as discussed.

**Stale "SDK 5.7 removed these" redeclarations** — all five deleted; the 6.6 typed parent is inherited (6a12db782). The caller audit you gated the typed step on passed (no site assigns `account` on the Success; every `creative_deadline` producer is aware-or-None), so both steps collapsed into one. A parametrized contract test pins our annotations to the installed parent's, which also catches future SDK drift. The identical staleness on `SyncAccountsResponse` is untouched here — it belongs to the 5.7-era sweep follow-up.

**DRY extractions** — `_submitted_approval_result()` now the single construction site for both approval branches (38d615235; the guard still sees the construction inside the helper), and `echo_context()` in src/admin/utils shared by the approve and creative webhooks (797ea8f7e), landed after the echo oracle so it's covered by a real assertion.

**Nits** — the redundant `flag_modified(media_buy, "status")` is gone (rode along with the echo-oracle commit); the thrice-defined `_capture` and the per-test mock scaffold collapsed into one `webhook_capture` fixture (a81a3aa2d); and the 21 added tests/ lines carrying internal tracker ids now cite PR #1567 / #1418, same treatment d25b99890 gave src/ (78ad1bdff).

**A2A submitted BDD vacuity** — the shared NOT-contain step now also grades the real Task on A2A (`assert not env.last_a2a_task.artifacts`, gated on submitted), mirroring the update path's guard (c4014a5fc). One honest note from mutation-testing it: removing the `del task.artifacts[:]` in the submitted early-return is a no-op (artifacts are only appended after that return), so the real mutation is emitting one — that's what the assertion catches.

**Since then: two semantic merges with `main`.** While the round-3 work was in flight, #1417 (UC-002/003 validation wiring + canonical error codes) landed on `main` and the PR went CONFLICTING — which, it turns out, is why `pull_request` CI silently stopped firing on our pushes. Resolved as a full semantic merge (`28b575341`): 19 conflicts resolved per-file for union-of-intent, all 30 auto-merged `src/` files audited, and the composition is exactly what you'd hope — your `TaskResultEnvelope`/validation architecture from #1417 carries our 3.1.1 Submitted variants inside it, and both sides had independently implemented the `media_buy_status` emission (deduped). Post-merge triage: main's 20 new scenarios failing over e2e_rest all traced to one harness gap (the create-env's seeded tenant left `human_review_required` at the model default `True`, which only the live server reads) — fixed in the harness with **no ledger growth**. One pre-existing manual-approval state leak (a second inline setter the earlier `try/finally` fix missed) surfaced and is fixed in both the e2e and admin suites. The 3 uc003 e2e_rest strict xfails XPASSed deterministically (your dispatcher update-endpoint support) and graduated after per-scenario inspection recorded in conftest.

Then #1430 (the e2e_rest ledger retirement) landed mid-merge, so a second round (`2cf2d74bd`) adopted the transport-aware harness wholesale — ledger now at its 7-entry retired state, with one load-bearing composition: the idempotent tenant get-or-create from #1430 now applies our `tenant_kwargs` on the get path too, otherwise the e2e auto-seed would have silently resurrected the manual-approval default. Our branch's scenarios all pass under your new harness (bdd_e2e 399 passed, 0 failed — the dropped ledger candidates included). One last find from the Quality Gate: the tests/ duplication baseline number committed earlier was macOS-skewed (pylint R0801 grouping differs by platform — measured on the CI box: the branch had added zero pairs on Linux); rather than touch the baseline we removed a genuine legacy 5-pair cluster (the same AST raw-`select` walker copy-pasted across four guard tests, now one shared helper), landing Linux at 83 vs the 86 ratchet.

Full board green on the final head: `./run_all_tests.sh` innet_140726_1848 (all 7 suites, 0 failures; unit 5583, integration 2223, bdd 1738+399, e2e 91, admin 86, ui 5), uv-secure clean (#1599's pillow/click pins came through the merge), and all 32 PR checks pass including your new BDD In-Network (e2e_rest) job. The follow-ups from your deferral ledger are filed — after a duplicate sweep of the tracker: #1617 (the credential logging; kept separate from #1450 since it's a true positive inside the class that issue writes off as false-positive), #1618 (guard MRO keying), #1619 (paused-on-create), #1620 (the 5.7-era redeclaration sweep, `SyncAccountsResponse` included). The ledger monotonicity hardening went as a checklist comment on #1418 rather than a new issue — that epic already existed with a public number, which also answers your "give the retirement epic a GitHub number": the ledger refs now cite #1418. Your three confirmations (#1578, #1326, #1447) all check out.

ChrisHuie (approved):

PR #1568, opened 2026-07-08
test(bdd): wire UC-019 get_media_buys BDD scenarios across transports

### LINKED ISSUE #1555: Wire UC-019 (get_media_buys) BDD scenarios across transports
## Context

Follow-up from #1545. The UC-019 (get_media_buys) BDD scenarios in
\`tests/bdd/features/BR-UC-019-query-media-buys.feature\` are all xfailed
(729 parametrized cases) because the harness is not wired: \`_detect_uc\` has no
UC-019 branch, so every scenario hits \`pytest.xfail("No harness wired for None")\`
in \`_harness_env\`. #1545 corrected the INV-8 (draft -> pending_creatives) and
INV-1 (pending_activation -> pending_start) scenarios to AdCP 3.1 but left the
wiring as a focused follow-up (it is large and independent).

Everything needed already exists: \`MediaBuyListEnv\` (tests/harness/media_buy_list.py)
with all transport wrappers + REST endpoint, and the full step module
\`tests/bdd/steps/domain/uc019_query_media_buys.py\`.

## Work

1. **Register the plugin** \`tests.bdd.steps.domain.uc019_query_media_buys\` in
   \`tests/bdd/conftest.py\` \`pytest_plugins\` and shrink \`_ALLOWED_UNREGISTERED\`
   in \`test_architecture_bdd_step_module_reachability.py\`.

2. **Resolve 8 step-text collisions** (the no-shadowed-steps guard blocks
   registration until these are gone). For each, verify the generic body subsumes
   the uc019-local one and delete the duplicate (or rename if genuinely distinct):
   - \`the request targets a production account\` / \`... sandbox account\` (vs given_auth)
   - \`the error should be a real validation error, not simulated\` (vs then_error)
   - \`the error should include a "suggestion" field\` (vs then_error)
   - \`the error should include a suggestion for how to fix the issue\` (vs then_error)
   - \`the response should indicate a validation error\` (vs then_error)
   - \`the response should include sandbox equals true\` (vs then_success)
   - \`the response should not include a sandbox field\` (vs then_success)
   Note the sandbox steps assert on \`resp.media_buys\` / a \`sandbox\` field — check
   they are equivalent to the generic get_products-oriented versions before deleting.

3. **Add the conftest wiring**: a \`T-UC-019 -> "UC-019"\` branch in \`_detect_uc\`
   and an \`elif uc == "UC-019":\` branch in \`_harness_env\` mirroring the
   UC-002-idempotency branch — pull \`integration_db\`, enter
   \`MediaBuyListEnv(e2e_config=...)\`, call \`env.setup_default_data()\`, set
   \`ctx["tenant"]\`/\`ctx["principal"]\`. CRITICAL: wrap the yield in try/finally that
   stops \`ctx["_patchers"]\` — the \`today is\` step starts a datetime patch nothing
   currently stops (session leak).

4. **Add ~4 missing givens** for INV-150-6..11:
   \`... with persisted status "X" [and is_paused true]\`,
   \`... has start_date "..." and end_date "..."\` (MediaBuyFactory accepts is_paused).

5. **Empirically re-baseline** (\`tox -e bdd -- -k uc019\`): delete now-passing tags
   from \`_UC019_XFAIL_TAGS\` / \`_UC019_PARAM_XFAIL\` (strict=False -> silent xpass
   otherwise); keep named strict xfails for real gaps (inv-151-4, sandbox, ext-b..e);
   add the "today is"-dependent tags (inv-150-6, -11) to \`_UC019_E2E_DATETIME_TAGS\`
   (the datetime patch has no effect on e2e_rest/Docker).

6. Verify: the two BDD architecture guards, then full \`tox -e bdd\`.

gagliathebuilder:
## Summary

Closes #1555.

Wires the UC-019 get_media_buys BDD scenarios into the existing multi-transport BDD harness.

This PR is intentionally scoped to harness reachability, UC-019 step registration, collision cleanup, missing Given setup, and empirical xfail re-baselining. It does not intentionally change production behavior.

## What changed
- Registered `tests.bdd.steps.domain.uc019_query_media_buys` in `pytest_plugins`.
- Removed the UC-019 step module from the BDD reachability guard allowlist.
- Added `T-UC-019 -> "UC-019"` detection in `_detect_uc`.
- Added a UC-019 branch in `_harness_env` using the existing `MediaBuyListEnv`.
- Added teardown for any `ctx["_patchers"]` started by UC-019 datetime steps.
- Added missing UC-019 Given steps for persisted status, persisted status plus `is_paused true`, and start/end date updates for already-seeded media buys.
- Resolved the eight step-text collisions in favor of established generic steps after verifying they preserve the reachable UC-019 assertions.
- Fixed one test-data determinism bug in `given_principal_owns_active_and_completed`: when no `today is "..."` step is present, the scenario now seeds dates relative to the current UTC date instead of a hardcoded `2026-03-15` fallback.
- Re-baselined UC-019 xfails based on the focused empirical run.

## Remaining xfails / known gaps

The newly active UC-019 scenarios now expose several real gaps that remain xfailed rather than being fixed in this PR:

- REST UC-019 query scenarios: no `/media-buys/query` production route exists.
- `T-UC-019-inv-150-11`: generated feature literal `some_unmapped_internal_state` exceeds `media_buys.status VARCHAR(20)`, causing insert failure before the scenario can exercise behavior.
- `T-UC-019-ext-a` on A2A/MCP: actual error is `AUTH_TOKEN_INVALID`; spec expectation is `AUTH_REQUIRED`.
- Sandbox boundary/production scenarios: same known gap as the existing sandbox-happy xfail; sandbox mode is not implemented.
- Boundary status-filter "removed enum value": same underlying production gap as the existing invalid-status-filter xfail family.

## Generated feature-file note

#1545 carries separate UC-019 generated-feature corrections:

- INV-1: `pending_activation` -> `pending_start`
- INV-8: `draft` -> `pending_creatives`

`tests/bdd/features/BR-UC-019-query-media-buys.feature` is machine-generated from an external `adcp-req` source and should not be hand-edited in this repo. This PR intentionally wires against the generated feature file as it currently exists on main and accounts for not-yet-applied corrections through the UC-019 xfail baseline.

## Verification

- `tests/unit/test_architecture_bdd_step_module_reachability.py` — pass
- `tests/unit/test_architecture_bdd_no_shadowed_steps.py` — pass
- Full BDD architecture guard suite — 31 passed
- Focused UC-019 BDD run — 108 passed, 621 xfailed, 0 failed, 0 xpassed, 729 total
- `ruff format` / `ruff check` — clean
- `mypy` — no new errors; existing unrelated harness errors remain
- Full `tests/unit/` — 5155 passed, 8 skipped, 26 xfailed, 0 failed
- Duplication guard — net improvement; `.duplication-baseline` auto-shrunk

## Notes

No `src/` files are changed. All changes are limited to BDD/test harness files, the reachability guard, and the duplication baseline.

## Environment note
This local worktree path contains a space (`Antigravity Builds`), which breaks `tox -e bdd -- -k ...` under xdist because `{toxworkdir}` is not quoted. Focused BDD verification was therefore run through `.tox/bdd/bin/pytest` directly with the same DB environment sourced from the test stack.

### DIFF
--- a/.duplication-baseline
+++ b/.duplication-baseline
@@ -1,5 +1,5 @@
 {
   "src": 36,
-  "tests": 89,
+  "tests": 88,
   "scripts": 0
 }
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -63,6 +63,7 @@ pytest_plugins = [
     "tests.bdd.steps.domain.admin_accounts",
     "tests.bdd.steps.domain.uc_get_products_inventory",
     "tests.bdd.steps.domain.compat_normalization",
+    "tests.bdd.steps.domain.uc019_query_media_buys",
 ]
 
 # ---------------------------------------------------------------------------
@@ -1865,6 +1866,34 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
                 )
             )
 
+        # --- UC-019: REST transport has no query route — blanket gap ---
+        # src/routes/api_v1.py registers POST /media-buys (create), PUT
+        # /media-buys/{id} (update), and POST /media-buys/delivery (metrics) —
+        # but no query route for get_media_buys. MediaBuyListEnv.REST_ENDPOINT
+        # ("/api/v1/media-buys/query") targets a route that was never built
+        # server-side, so every REST-transport UC-019 scenario 405s
+        # (Method Not Allowed) regardless of what it's otherwise testing. #1555.
+        #
+        # Excluded: T-UC-019-partition-status-invalid. Its Given step
+        # (missing start_date/end_date) hits a DB NOT NULL constraint and sets
+        # ctx["error"] directly during setup — _dispatch_query short-circuits
+        # before any transport call, so these never reach the network layer
+        # and pass on REST regardless of the missing route (empirically
+        # verified: xpassed under strict=False before this exclusion). #1555.
+        _UC019_REST_ROUTE_INDEPENDENT_TAGS = {"T-UC-019-partition-status-invalid"}
+        if (
+            is_rest
+            and any(t.startswith("T-UC-019") for t in marker_names)
+            and not (marker_names & _UC019_REST_ROUTE_INDEPENDENT_TAGS)
+        ):
+            item.add_marker(
+                pytest.mark.xfail(
+                    reason="UC-019: REST transport has no get_media_buys query route "
+                    "(405 Method Not Allowed) — spec-production gap",
+                    strict=False,
+                )
+            )
+
         # --- UC-019: xfails for spec-production gaps ---
         # Graduated (k31s): status_computation active variants, default_status_filter
         # simple variants, status_filter boundary simple variants, inv-150-2/4,
@@ -1885,12 +1914,22 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
             "T-UC-019-inv-153-3",
             "T-UC-019-inv-153-4",
             "T-UC-019-inv-153-5",
+            # Unmapped persisted status: the feature's literal test value
+            # "some_unmapped_internal_state" (29 chars) exceeds
+            # media_buys.status VARCHAR(20) — DataError on insert, all
+            # transports. Not fixable without a schema migration (no
+            # intentional production changes in this PR). #1555.
+            "T-UC-019-inv-150-11",
             # Sandbox mode — not implemented
             "T-UC-019-sandbox-happy",
             "T-UC-019-sandbox-validation",
             # Graduated: T-UC-019-partition-principal-invalid identity_missing (impl/a2a/mcp pass)
             # — moved to _UC019_PARAM_XFAIL for selective identity_missing exclusion.
-            # Graduated: T-UC-019-ext-a (impl/a2a/mcp pass) — moved to selective block below.
+            # T-UC-019-ext-a: REST/e2e_rest xfailed below (suggestion wording).
+            # a2a/mcp xfailed separately below (AUTH_TOKEN_INVALID vs AUTH_REQUIRED,
+            # empirically verified — #1555; the prior "Graduated (impl/a2a/mcp pass)"
+            # comment here predated the harness ever being wired and was never
+            # actually run against this module).
             # Extension errors — error code mismatches / not implemented
             "T-UC-019-ext-b",
             "T-UC-019-ext-c",
@@ -1934,12 +1973,27 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
                 )
 
         # --- UC-019: HTTP transport xfails for auth suggestion mismatch ---
-        # impl/a2a/mcp graduated (kb7y); REST/e2e_rest suggestion string differs
-        # from spec ("authenticate" vs "authentication").
+        # REST/e2e_rest suggestion string differs from spec ("authenticate" vs
+        # "authentication"). a2a/mcp are xfailed separately below for a
+        # different, empirically-verified reason (AUTH_TOKEN_INVALID vs
+        # AUTH_REQUIRED) — #1555.
         if (is_rest or is_e2e_rest) and "T-UC-019-ext-a" in marker_names:
             item.add_marker(
                 pytest.mark.xfail(
                     reason="HTTP transport: auth error suggestion says 'authenticate' not 'authentication' — spec-production gap",
+                    strict=False,
+                )
+            )
+
+        # --- UC-019: a2a/mcp raise the wrong error code for missing credentials ---
+        # T-UC-019-ext-a expects AUTH_REQUIRED; a2a/mcp actually raise
+        # AUTH_TOKEN_INVALID for a no-credentials get_media_buys request.
+        # Empirically verified — #1555.
+        if (is_a2a or is_mcp) and "T-UC-019-ext-a" in marker_names:
+            item.add_marker(
+                pytest.mark.xfail(
+                    reason="UC-019: a2a/mcp raise AUTH_TOKEN_INVALID instead of AUTH_REQUIRED "
+                    "for a missing-credentials get_media_buys request — spec-production gap",
                     strict=False,
                 )
             )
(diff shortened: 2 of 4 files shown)


ChrisHuie:
**Recommend closing this PR as superseded — the UC-019 harness wiring it targets is already on `main` via the merged #1545 lineage (`19735ea7a`, `0adc53ee7`).**

This branch was cut from `c30e86e42`, the same base as #1545, before #1545's follow-up commits merged. #1555 was filed when #1545 had only corrected the feature values and deferred the harness wiring; #1545 then absorbed that wiring during review (`19735ea7a`, "wire get_media_buys status behavior on the wire"), so the follow-up is already in `main`. The wiring approach and the determinism reasoning here are both sound — the only gap is the stale base.

**Already on `main`**
- `_detect_uc` + `_harness_env` UC-019 branches — `tests/bdd/conftest.py:2775` and `:2995` (added by `19735ea7a`).
- Step module made live — `main` uses `from ...uc019_query_media_buys import *` in `tests/bdd/test_uc019_query_media_buys.py`, deliberately kept **out** of `pytest_plugins` so the module's 8 domain step overrides stay UC-019-scoped (reachability-guard comment at `tests/unit/test_architecture_bdd_step_module_reachability.py:58`). This PR takes the opposite approach — global `pytest_plugins` registration + deleting those overrides — which is the source of the `conftest.py` merge conflict.
- Datetime-patch teardown (the session-leak item #1555 flagged as critical) — LIFO patcher stop at `tests/bdd/conftest.py:2752`.
- Determinism of `given_principal_owns_active_and_completed` — `main` pins the clock in the feature (`And today is "2026-03-15"`, `tests/bdd/features/BR-UC-019-query-media-buys.feature:420`), which is more deterministic than seeding off the wall clock. This scenario passes non-vacuously on a2a/mcp on `main`.

**Two issues in the current diff, both stemming from the stale base**

**1. The xfail re-baseline would mask behavior `main` proves passing.** In `tests/bdd/conftest.py`, the UC-019 xfail sets re-xfail `T-UC-019-inv-150-1`, `inv-150-3`, `inv-150-11`, and the `partition-status-filter` multi-status cases. Those pass on `main` today (a2a/mcp): `19735ea7a` graduated inv-150-1/-3 (pre-flight → `pending_start`, post-flight → `completed`) and `0adc53ee7` graduated multi-status. These are `strict=False` xfails, so adopting this baseline turns gating PASSes into silent XPASSes and removes the gate on the wire status-taxonomy. Verify by running the UC-019 subset on current `main` — those tags pass with 0 xfailed.

**2. inv-150-11 is not blocked on a schema migration.** The PR xfails `T-UC-019-inv-150-11` as "exceeds `status VARCHAR(20)`, not fixable without a migration." On `main` the persisted value renders as `unmapped_state`, which fits `VARCHAR(20)`, and the scenario grades passing — `tests/bdd/features/BR-UC-019-query-media-buys.feature:560`. No migration required.

**REST transport.** `get_media_buys` has no REST route, so `main` excludes UC-019 from REST parametrization via `_NO_REST_UC_TAG_PREFIXES` (`tests/bdd/conftest.py:2594`), giving 488 tests (244 × a2a, mcp). This PR parametrizes over REST and blanket-xfails the ~240 REST nodes for a 405 — baseline noise for a route that cannot exist. `main`'s exclusion is the cleaner mechanism.

**Suggested resolution.** Close this PR and #1555 as completed by the #1545 lineage. A rebase does not recover net value — the wiring, teardown, determinism handling, and baseline are all already on `main`, and the global-registration + REST-parametrization changes here reverse deliberate choices. One net-new idea worth a separate issue if desired: `main` keeps 8 module-local step overrides that the generic steps now cover, so those could be DRY'd — but that reverses a commented scoping decision and needs its own review.

**Verified by:** tracing `conftest.py`, the step module, the feature file, and the reachability guard on `main`; running the inv-150-1/-3/-11 and multi-status UC-019 scenarios on current `main` (pass, 0 xfailed for those tags); `git merge-tree` (confirms the `conftest.py` conflict). **Not run:** the full 488-test UC-019 suite; the PR branch's own reported totals (108 passed / 621 xfailed were computed on the pre-#1545 base and do not reflect a merge with current `main`).

gagliathebuilder:
Thanks Chris — very helpful context.

We were already mid-pass on the rebase/conflict resolution when I saw this, so I’m pausing before any force-push and using your notes as the new source of truth for the review.

I’m going to verify the specific points you called out against current `main` — especially the #1545 lineage, local UC-019 step registration, REST exclusion, patcher teardown, persisted-status handling, and the graduated xfail cases. If those all check out, I’ll course-correct and treat this PR as superseded rather than pushing a stale-base cleanup.

Appreciate the detailed trace.

gagliathebuilder:
Verified the key points against current `main` (`5618240a`), and Chris’s read checks out.

UC-019 `_detect_uc` / `_harness_env` wiring, local step import, REST exclusion via `_NO_REST_UC_TAG_PREFIXES`, shared LIFO patcher teardown, persisted-status/date helpers, and the graduated `inv-150-1` / `inv-150-3` / `inv-150-11` / multi-status cases are already on main.

Agreed — I’m treating this PR as superseded by #1545 rather than pushing a rebase.

*** One additional thing we found while tracing this: a rebase/auto-merge can silently create two `elif uc == "UC-019":` branches in `_harness_env`. Since `elif` short-circuits, the PR branch’s older version would win and make main’s LIFO-safe branch dead code, reintroducing the exact patcher teardown risk this work was meant to avoid.

Thanks again for catching this before the force-push.

gagliathebuilder:
Confirmed against a clean checkout of `origin/main` (`5618240a`) — #1545 already wired UC-019 completely: harness detection/env wiring, local-import step architecture, REST exclusion, shared LIFO patcher teardown, and persisted-status/date helpers are all already on main.

I also ran the full UC-019 BDD subset against pure main: `488 total, 124 passed, 364 xfailed, 0 failed, 0 xpassed`.

The specific gaps this PR was trying to address have also been fixed on main at the better layer:

- `inv-150-11` now uses `"unmapped_state"` in the regenerated feature file, so it fits the status column and passes without a schema migration.
- `inv-151-1` now pins `And today is "2026-03-15"` in the feature file, so the datetime determinism issue is fixed at the source rather than by defensive Python fallback logic.

The remaining rebased diff is stale or regressive: global UC-019 `pytest_plugins` registration contradicts the intentional local-import architecture, the old xfail baseline would mask passing behavior, and auto-merge can create a duplicate `elif uc == "UC-019":` branch that makes main's correct branch unreachable.

I did not push anything. Agreed with the recommendation here — treating #1568 as superseded by #1545. Thanks for catching this before the force-push.

gagliathebuilder:
Closing as superseded by #1545.

PR #1572, opened 2026-07-09
perf: parallelize the test suite — scenario-level BDD distribution, tmpfs test DB, opt-in xdist

KonstantinMirin:
## What
Parallelizes the test suite. **All new parallelism is opt-in via env vars with conservative defaults**, so CI and local runs are unaffected unless explicitly enabled. The one default change is behavior-preserving (below).

## Why — the CI hook
CI already shards BDD into 2 jobs, each running `-n auto --dist=loadfile`. The reason it shards is that **`--dist=loadfile` cannot split a single feature file across workers** — one large file (`test_uc004_deliver_media_buy_metrics.py`, ~900 scenarios) is a hard wall-time floor per shard no matter how many workers. This PR switches the shards to **`--dist=load`** (scenario-level), which distributes scenarios and balances each shard. Verified to produce **identical pass/xfail counts**.

## Changes
- **`.github/workflows/ci.yml`** — BDD shards `--dist=loadfile` → `--dist=load` (the one default change; behavior-preserving).
- **`tox.ini`** — `--dist` mode + xdist worker counts made env-overridable (`BDD_DIST`, `BDD_XDIST_N`, `UNIT_XDIST_N`, `INTEGRATION_XDIST_N`); `unit`/`integration` default **serial** (opt-in parallel). Adds `bdd_inprocess`/`bdd_e2e` envs for a two-pass BDD split (in-process transports vs `e2e_rest`).
- **`docker-compose.e2e.yml`** — throwaway test Postgres on **tmpfs** (RAM): `createdb` + DDL dominate DB-heavy suites (~2×); `max_connections` raised for xdist.
- **`tests/conftest_db.py`** — opt-in per-test DB via `CREATE DATABASE … TEMPLATE` clone + skip-drop (`TEST_DB_TEMPLATE`), avoiding a full `create_all()` per test.
- **`tests/bdd/conftest.py` + `run_all_tests.sh`** — opt-in per-xdist-worker `e2e_rest` server stacks (`E2E_WORKERS`): each worker gets its own server + database, so the real-HTTP transport can run in parallel.

## Safety / no regression
Off by default: `UNIT_XDIST_N`/`INTEGRATION_XDIST_N`=0, `TEST_DB_TEMPLATE`/`E2E_WORKERS` unset, `bdd_inprocess`/`bdd_e2e` not in the default suite list. Only the CI BDD `--dist` mode changes, and it preserves results.

## Measured (16-core host)
- in-process BDD (a2a/mcp/rest): ~23 min → ~3.5 min
- `e2e_rest` (per-worker servers): ~8 min → ~1.5 min

Draft while CI validates.


ChrisHuie:
# Notes — #1572 (verified at head `f90d2704`)

FYI, not a blocking review. The one default change — CI BDD shards `--dist=loadfile` → `--dist=load` (`.github/workflows/ci.yml:389`) — is race-safe on its own: every in-process BDD scenario runs against a per-test uuid database (each `_harness_env` branch calls `integration_db`, which `reset_engine()`s the global engine at `tests/conftest_db.py:580` before any harness — including `AdminAccountEnv`, `tests/bdd/conftest.py:2961` — issues `get_db_session()`), and `e2e_rest` (the only shared-DB transport) stays off in CI, so distribution mode can't change isolation. Other defaults are preserved: every `${VAR:-…}` / `{env:VAR:0}` falls back to current behavior, and `.duplication-baseline` is unchanged.

**Nothing below blocks the `--dist` change.** The rest is a heads-up on the opt-in acceleration paths (`TEST_DB_TEMPLATE`, `E2E_WORKERS`): as committed they don't fully activate, so the "Measured" speedups aren't reproducible through the committed runner yet.

**`TEST_DB_TEMPLATE` is filtered out by tox, so the template fast path never runs.** `_use_db_template()` reads `os.environ` at runtime (`tests/conftest_db.py:380`), but the var isn't in `[testenv] pass_env` (the sibling runtime-read `E2E_PER_WORKER` is) and there's no `{env:TEST_DB_TEMPLATE}` substitution, so tox strips it before pytest and the entire clone path is inert via `tox`/`run_all_tests.sh`.
```diff
 pass_env =
     ...
     E2E_PER_WORKER
+    TEST_DB_TEMPLATE
```

**A half-built template is never rebuilt.** `CREATE DATABASE "<template>"` commits under autocommit (`tests/conftest_db.py:414`) before `_build_schema()` (`:415`); if the build raises, the next call sees the DB already exists and skips the rebuild permanently → every clone is schema-less until the Postgres container restarts. Build into a temp name and rename on success, or drop the template on build failure.

**Skip-drop leaks databases (and clones a stale schema) on any non-tmpfs Postgres.** The teardown drop is skipped whenever `_use_db_template()` (`tests/conftest_db.py:619`); correctness silently depends on the tmpfs pairing. Pointed at a persistent Postgres (agent-db, local dev) — the stated "speed up local iteration" use — `test_*` clones accumulate unbounded, and a model/migration change is masked because the template is existence-gated, not freshness-gated. Gate `TEST_DB_TEMPLATE=1` on tmpfs, or gate rebuild on a schema hash.

**`E2E_WORKERS>0` doesn't parallelize the in-process bulk.** The fast path swaps `bdd` → `bdd_inprocess,bdd_e2e` and provisions per-worker servers, but never sets `BDD_XDIST_N`; the container pins `BDD_XDIST_N=0` (`docker-compose.e2e.yml:152`), so `bdd_inprocess`'s `-n {env:BDD_XDIST_N:auto}` (`tox.ini:149`) resolves to `-n 0` (serial). The ~23m→~3.5m in-process win needs `BDD_XDIST_N` exported in that branch as well.

**`uv.lock` has unrelated churn.** No `pyproject.toml` dep changed, but the lock bumps `soupsieve 2.8.1→2.8.4` (already on `main` — vanishes on rebase) and adds resolver markers to `aiologic` / `culsans` / `google-ads`. `git checkout origin/main -- uv.lock`; the PR needs no lock delta.

**Nits**
- Orphaned `postgres_data:` volume (`docker-compose.e2e.yml:262`) — dead after the tmpfs switch; drop the declaration.
- Template migrate uses `|| echo "  (template migrate warning)"` (`run_all_tests.sh:146`) then clones per-worker DBs from the possibly-un-migrated template — fail fast instead of cloning empties.
- The guard-relaxation comment "The worker env inherits BDD_E2E_ENABLED (verified)" (`tests/bdd/conftest.py:113`) — swap "(verified)" for an execution-count assertion (`bdd_e2e.json` `e2e_rest` count > 0). The shipped `-k e2e_rest` path already fails loud (pytest exit 5 on empty selection), so this is documentation, not a live silent-drop.
- Concurrent per-test `CREATE DATABASE … TEMPLATE` clones run unserialized (`tests/conftest_db.py:485`); under heavy xdist they may hit "source database is being accessed by other users" — worth checking at the intended `INTEGRATION_XDIST_N`, with a bounded retry if it surfaces.
- Lower-priority: `PGDATA_TMPFS_SIZE` default `10g` can exceed a hosted runner's RAM (low risk; e2e writes little); the per-worker `postgres_url` swap (`tests/bdd/conftest.py:2687`) is correct for the real URL but `sqlalchemy.make_url().set(database=…)` is sturdier for atypical shapes; an unhealthy per-worker server prints "NOT ready (continuing)" and runs anyway (`run_all_tests.sh:160`).

**What's right** — the `command: [postgres, -c, max_connections=300]` override preserves the image entrypoint's DB init; `-n 0` is behavior-identical to the prior no-`-n`; `{[testenv]setenv}` inheritance and the new `pass_env` additions are correct; the base `bdd` env keeps `--dist loadfile` while CI passes `load` directly (no contradiction); tmpfs is exercised by the CI e2e job and is green.

--- Verified at `f90d2704`:
Ran: full diff (7 files); traced the `_harness_env` isolation path across all UC branches incl. ADMIN → `integration_db`/`reset_engine`/`get_db_session`; `tox.ini` `pass_env` + `{env:…}` substitution; compose env-indirection defaults; `.duplication-baseline` (unchanged on the branch).
Not run: the BDD suite under `--dist=load` — the `load` assessment rests on the isolation trace above, not a fresh multi-run; the opt-in `TEST_DB_TEMPLATE` / `E2E_WORKERS` paths (reasoned statically). CI rollup is green on head `f90d2704` (observed, not a self-run gate).

ChrisHuie (approved):

Issue #1423, opened 2026-06-11
test(e2e): consolidate duplicated e2e fixtures (pre-existing debt; R0801 gate false-flags test-adding PRs)


KonstantinMirin:
## Summary

The `check_code_duplication.py` Quality Gate (pylint R0801, ratcheting baseline introduced in #1372) flags a duplication increase on **any** PR that adds test files — even when the PR introduces **zero** new duplication. PR #1420 (in-network test runner) hit this: tests count went 98 → 99, and the bump is a measurement artifact, not new copy-paste.

## Why this happens (the real diagnosis)

`count = stdout.count("R0801")` counts pylint **R0801 groups**, and R0801 is an *emergent clustering* metric over the whole scanned tree. Adding files perturbs the global clustering, so a **pre-existing, byte-identical** duplicate pair gets recounted as its own group.

Verified for #1420:
- Its changed/added files (`tests/harness/dispatchers.py`, `tests/bdd/conftest.py`, `tests/harness/_base.py`, `tests/e2e/conftest.py`, `tests/ui/conftest.py`) appear in **0** of the 99 groups.
- `origin/main` = 98 and #1420's tree = 99 in the **same** environment; the delta is the pre-existing pair `test_a2a_webhook_payload_types` + `test_delivery_webhooks_e2e` (identical in both trees).
- "Fixing" it by removing a real duplicate made the count **worse** (99 → 101), because removing one cluster re-exposed three other pre-existing near-dupes. The metric is unstable under edits.

**How could #1372 pass at 98?** Because #1372 added no test files to perturb the clustering. The baseline is only stable for PRs that don't touch `tests/` membership — i.e. it's brittle by construction.

## The underlying debt (the actual thing to fix)

There are ~10 R0801 groups entirely within `tests/e2e/`, from a missing-abstraction pattern: every e2e file re-implements the same building blocks instead of sharing them —

1. **Webhook-capture HTTP server** (`BaseHTTPRequestHandler.do_POST` + `log_message` + port-finding fixture) — copied across `test_adcp_reference_implementation`, `test_a2a_webhook_payload_types`, `test_delivery_webhooks_e2e`.
2. **Authenticated MCP client setup** (`x-adcp-auth` / `x-adcp-tenant: ci-test` headers + `StreamableHttpTransport`) — copied everywhere.
3. **Live-server DB mutation via psycopg2** (adapter config, force-approve, tenant lookup) — copied.

[...]

PR #1585, opened 2026-07-11
feat: validation-harness base — adcp 6.6 consolidation, e2e fixture dedup, CI e2e_rest hardening (stacked on #1417/#1430/#1567)

KonstantinMirin:
## Stacked on

This PR is the consolidation base **stacked on top of #1417 (UC-002/003 validation wiring), #1430 (transport-aware BDD harness), and #1567 (adcp 5.7 → 6.6, spec 3.1.1)**. It contains all three plus the work below; land the parents first and this diff shrinks to just the delta. The delta against #1430's head is viewable at:
https://github.com/KonstantinMirin/prebid-salesagent/compare/feature/e2e-harness-wiring...feature/validation-harness-base

### Semantic merge of adcp 6.6 into the #1417+#1430 consolidation
- 19 conflicted files resolved by union-of-intent with a per-file resolution ledger (commit `9401f3f05`); silent auto-merge defects found and fixed beyond the conflicts themselves (dropped step decorator, duplicated helper definitions, a Body field the route already forwarded, obsolete SDK-5.7 `buyer_ref` redeclarations).
- `update_media_buy` manual-approval branch returns the spec-3.1.1 `UpdateMediaBuySubmitted` variant bare (it is itself a protocol envelope: `status="submitted"` const + required `task_id`); Success/Error stay wrapped in `UpdateMediaBuyResult`. Guards updated with justification.

### Pinned creative agent — deterministic, spec-matched
- `preview_creative` sends `format_id` as the federation-identity object (the live public host's leniency had masked the mismatch).
- Pinned reference agent un-gated (every e2e up starts it) and **pin bumped to the v3.1.1 tag** (`467fd93d7711`), so the reference agent matches the spec target; three real divergences triaged in the bump (DEV_USER/NODE_ENV, SDK 6.6 fallback-marker phrasing, pin-keyed local image tag). Formats fixture refreshed (57 formats).
- CI `e2e-tests` and `bdd-in-network` jobs pull the pin-keyed ghcr image instead of a ~5-min source build; job names untouched (frozen required-checks guard).

### BDD graduations (per the xpass-graduation workflow, spec-cited)
- UC-003 manual-approval submitted-envelope scenarios: graduated on e2e_rest — the merged harness dispatches updates over real HTTP PUT, so the parent branch's xfail premise no longer holds. Verified 4-point (scenario vs spec 3.1.1 schema, Given seeding, When dispatch, wire-level Then).
- UC-019: 4 stale "not implemented" xpasses graduated (`status_filter` validation is implemented); a vacuous sandbox Given replaced with real `Account(sandbox=True)` seeding; UC-019 error assertions moved to wire-first (`assert_envelope_shape` on the two-layer envelope).

### e2e fixture consolidation (GH #1423)
- `make_mcp_client` factory: single home for the authed-MCP-client construction, 22 hand-rolled sites across 8 files migrated; intentional variations (no-dry-run, cross-tenant tokens, Host-header routing) are explicit kwargs.
- Five copy-pasted psycopg2 adapter-config upserts replaced by the factory helper `set_adapter_test_behavior` through a live-DB session bridge; fails loud on missing tenant.
- The last hand-rolled webhook-capture handler folded onto the shared module via a record() hook; A2A envelope builder, default-campaign builder, and compliance-report writer deduplicated.
- Duplication ratchet moved by genuine removal: tests 87 → 80, src 36 → 35.

### CI e2e_rest (GH #1419 residuals)
- `E2E_WORKERS` parallelization measured on a 4-core runner rather than assumed: 5–6 min slower than serial with the ledger green — recorded, serial retained (box-only optimization).
- Flaky lifecycle test root-caused (pytest-randomly ordering leaked a2a manual-approval adapter state across the shared DB) and pinned with an `auto_approval_adapter` fixture in all five vulnerable tests.

## Verification

All verified green.


ChrisHuie:
# Review — validation-harness base (verified at head `73ce27d25`)

Reviewed this PR's **own delta** (the new work + the two consolidation merges), not the inherited #1417/#1430/#1567 content. The substance holds up well: all four adcp-6.6 spec claims check out against spec 3.1.1 — the bare `UpdateMediaBuySubmitted` variant (`status='submitted'` const + required `task_id`, `media_buy_update.py:585,614`, serialized flat on all three transports via `_base.py:442`), `preview_creative` sending `format_id` as the `{agent_url, id}` object, the union-of-intent resolutions (`paused: bool`, `buyer_ref` removal, `warnings`-defaults-None sweep), and `CREATIVE_NOT_FOUND` passing through untranslated at `correctable`. The consolidation is honest — `.duplication-baseline` (35/80) and `.type-ignore-baseline` (64) are real reductions from main's 36/87 and 69, not masked, and 502 architecture guards pass on the branch. The UC-003/UC-019 graduations execute against the real wire (UC-019: 124 passed / 360 xfailed / **0 xpassed**).

**Blocking state:** the PR is `CONFLICTING` against current main, so the green CI is stale — it predates main's last two commits (the get_products→`BrandReference` coercion and the pillow/click bump), which both conflict here in `products.py`, `api_v1.py`, `adcp_a2a_server.py`, and `media_buy_create.py`. GitHub won't re-run `pull_request` CI on a conflicted PR, so the rollup isn't merge-with-main. **Please rebase — the conflict resolution (especially `products.py`) needs a fresh look.** Findings below are all on this PR's own delta; nothing here is a hard blocker beyond the rebase, but A1 is a real functional gap.

## SHOULD-FIX

**REST `create_media_buy` silently drops `body.paused`.** The adcp-6.6 merge added `paused` to `CreateMediaBuyBody` (`api_v1.py:86`), but the REST create route never forwards it (`api_v1.py:315`) — while the MCP wrapper (`media_buy_create.py:4446`), `create_media_buy_raw` (`:4533` → `_build_create_media_buy_request`), and the REST **update** route (`api_v1.py:346`) all do. `paused` is a real AdCP 3.1.1 "create in paused state" field (`create_media_buy_raw:4490`), so a REST buyer sending `paused: true` gets an **active** media buy — a silent drop with spend implications. The boundary-completeness guard only models the `_impl`/MCP/`_raw` layers, not the REST Body→raw hop, so it can't catch this.

```diff
         idempotency_key=body.idempotency_key,
+        paused=body.paused,
         identity=identity,
```
Add a REST wire test asserting a paused create round-trips.

**Two duplicate definitions survived the merge** (both below the R0801 `min-similarity-lines=6` threshold, so `.duplication-baseline` doesn't flag them):
- `tests/harness/media_buy_create.py:166,184` — `_get_or_create_context` is defined twice (byte-identical) and its `side_effect` assigned twice (`:201,:203`); delete one def + one assignment.
- `tests/harness/_base.py:402,409` — `self.clock = _TestClock()` appears twice in `__init__`; delete one.

**The flaky-lifecycle fix masks rather than closes the leak** (`tests/e2e/conftest.py:556-569`). `auto_approval_adapter` pins the five named victims, but the leakers set `manual_approval_required=True` with no reset (`test_a2a_webhook_payload_types.py:285,550`) and there's no autouse teardown — a future sync-success e2e test that forgets the fixture will silently re-flake. An autouse reset-to-default fixture (leakers still opt into `True`), or a guard requiring the fixture on sync-success tests, closes it generally.

**The e2e_rest ledger grew +10 under the "genuine gaps" bucket** (`test_e2e_rest_ledger_state.py`, 11→21). The exact-set lock forces this in-change and it's owner-approved, but these are parallel-e2e mock-injection artifacts (transport-incompatibility) — the category the design routes to route-3 `E2EUnsupportedSetup`, not the ledger. Reclassify them to route-3, or gate behind the fault-injection control, and update the "shrinking work-list" framing.

## NIT
- **`preview_creative` stale comment** (`creative_agent_registry.py:917`) still says "not in AdCP spec" — it is in 3.1.1, which mandates the object shape; cite the spec schema instead.
- **`test_creative_agent_fallback.py:364,379`** — `pytest.raises(Exception, match="(?i)valid")` passes on any error containing "valid"; pin the specific exception type + field. The `caplog` WARNING the class docstring promises (`:325-344,:382-390`) is captured but never asserted — assert it or drop the capture.
- **`tests/e2e/utils.py:107`** — `set_live_adapter_behavior` returns a detached ORM row out of a closed session; the 7 callers discard it, but return `None`/copied scalars to remove the trap.
- **New comments carry local task ids** (`creative_agent_registry.py:200,925`; guard docstrings in `test_architecture_impl_result_envelope.py:13`, `..._harness_mcp_with_error_logging.py:24`) — these don't resolve for outside contributors; prefer a GitHub reference or plain prose.
- **UC-019 `[mcp]` graduation grades the typed exception, not the MCP wire** (`tests/harness/media_buy_list.py:56-60`, `call_mcp` uses the deprecated `_run_mcp_wrapper`); the `[a2a]` variant covers the wire. Upgrading `call_mcp` to `_run_mcp_client` (the UC-003 pattern) grades the real MCP envelope.

## Out of scope (inherited from the stacked parents — flagging, not for this delta)
Traced to #1417/#1430/#1567 via `--cc` + merge-base, so not this PR's to fix: the recovery/error-code taxonomy inconsistencies (`AdCPConfigurationError` recovery vs spec 3.1.1; `_sync.py:338-355` dropping `recovery`/`code`; strict-vs-lenient creative-not-found at `_assignments.py:334`) — the `exceptions.py:41` "a follow-up" comment for `CONFIGURATION_ERROR`/`BILLING_NOT_SUPPORTED` isn't filed as an issue; the creative-format check implemented three divergent ways (create id-only vs update pair-based vs the canonical FormatId-aware helper); the A2A submitted-detection gating on `res["success"]` (`adcp_a2a_server.py:693`), which interacts with the new bare submitted variant when advisory `errors` are non-empty; and the creative-agent outbound path lacking `url_validator` (tenant-admin-gated).

---
Verified at `73ce27d25`:
- **Ran:** 502 architecture guards (pass); targeted units `test_architecture_impl_result_envelope` 7/7, `test_adcp_exceptions` 64/64, `test_a2a_transport_contract`+`test_all_response_str_methods` 80/80, `test_update_media_buy_behavioral` 7/7; UC-019 BDD serial (0 xpassed); `.duplication-baseline`/`.type-ignore-baseline` re-derivation; live spec-3.1.1 schema grounding; full provenance on every finding (both merge parents' blobs for A1).
- **Not run:** `./run_all_tests.sh ci` and Docker e2e/bdd-in-network (the in-network graduations rest on the cited run id, not observed; a conflicted PR would skip that gate). The post-rebase conflict resolution isn't reviewable yet — re-review after rebase.

ChrisHuie (approved):

Issue #1595, opened 2026-07-13
get_products: emit brief_relevance per v3.1.1 product schema


KonstantinMirin:
`brief_relevance` is a pinned v3.1.1 product field (core/product.json; also a member of the get-products-request `fields` enum), but the wire never carries it — the impl ranks internally when AI ranking is enabled and never serializes an explanation.

Graded by the strict-xfail assert at the end of `T-UC-001-main` (BR-UC-001-discover-available-inventory.feature) — emitting it forces graduation.

KonstantinMirin:
## Also wire the BDD scenario in this change

Closing this gap unblocks a scenario that is currently strict-xfailed, so the fix and its grading should land together — otherwise the fix ships unverified and the scenario stays dormant.

**Wire when this lands:** `T-UC-001-main`

The scenario asserts `brief_relevance` on the wire — a pinned v3.1.1 field production never serializes. That is the whole of what keeps it masked.

The scenario, step definitions and harness env already exist and are green on `refactor/transport-boundary`; #1822 ships the rest of that use case's wiring and deliberately leaves this tag out, because wiring it while the gap is open would add a strict-xfail that grades nothing.

Source: the `_SPEC_GAP_XFAILS` entry in `tests/bdd/conftest.py` naming this gap.

gagliathebuilder:
Hi - Thank you for documenting this so clearly, especially the connection to `T-UC-001-main`.

I took a pass at the issue and have a small, focused change ready that maps the existing internal ranking explanation to the AdCP 3.1.1 `brief_relevance` field. I also added unit coverage and wired the BDD scenario, which is passing across A2A, MCP, and REST in my local checks.

I’ve kept the implementation limited to this production gap and avoided changing the existing ranking, threshold, or filtering behavior. If this direction looks right, I’d be happy to open a draft PR for review. I’d also appreciate any guidance or adjustments you’d like before it moves forward.

Issue #1596, opened 2026-07-13
REST GetProductsBody: declare buying_mode (declared v3.1.1 request property)


KonstantinMirin:
`buying_mode` is a declared property of the v3.1.1 get-products-request, but `GetProductsBody` (src/routes/api_v1.py) does not declare it — on the dev/CI `extra=forbid` arm a spec-valid request carrying it is rejected; in production it is silently dropped. Same class as the GetSignalsBody gap fixed for the signals route (four missing declared fields).

The UC-001 REST step currently drops buying_mode before dispatch with a spec-gap note (tests/bdd/steps/domain/uc001_discover_inventory.py) — remove that workaround when fixed.

KonstantinMirin:
Another instance of this same gap, found while working on #1868: `GetProductsBody` (`src/routes/api_v1.py:68`) also doesn't declare `property_list`. The MCP wrapper (`src/core/tools/products.py:792`) and the A2A raw wrapper (`:842`) both accept it and forward it into `create_get_products_request`, but REST structurally can't receive it — so `_get_products_impl`'s property-list filtering branch is dead on that transport.

Concretely: a REST `get_products` call passing `property_list` today returns `status: completed, products: []` — a cheerful empty success with the filter silently never applied. There's a wire test documenting the gap at `tests/integration/test_get_products_property_list_error_wire.py`.

Also worth checking while this is being fixed: `context`, which the MCP wrapper accepts and `GetProductsBody` likewise omits.

One more thing worth checking: `tests/unit/test_architecture_boundary_completeness.py` is supposed to guard exactly this class of gap (per CLAUDE.md), and it didn't catch either `buying_mode` or `property_list` missing from the REST body. Worth confirming whether that guard covers REST body models at all.

Not fixing this in #1868 — folding the evidence in here rather than opening a competing issue, since it's the same class of gap this issue already names for `buying_mode` (and the `GetSignalsBody` precedent mentioned in the issue body).

KonstantinMirin:
Related, from the #1941 round-3 remediation: this is not a one-off. The same pattern — a pinned 3.1.1 request property missing from the hand-rolled REST body model — measures at roughly **66 fields across all 11 `*Body` models** in `src/routes/api_v1.py`. Filed the systemic version as #2018 with a per-transport table for update_media_buy.

Two structural notes that explain why `buying_mode` (and the rest) went unnoticed:

[...]

Issue #1594, opened 2026-07-13
test(bdd): wire the remaining dormant UC-001/UC-008/UC-009/UC-010 scenarios onto harness envs


KonstantinMirin:
The BDD harness wiring for UC-001 (ProductEnv), UC-008 (SignalsEnv), UC-009 (PerformanceEnv) and UC-010 (CapabilitiesEnv) covers the main-flow tag sets only; every other scenario in those use cases xfails at the fixture ("harness not yet wired for this scenario", tests/bdd/conftest.py `_wire_simple_env` call sites). Wire the remaining sets — wiring never forces graduation, so genuine spec-production gaps stay xfailed via `_SPEC_GAP_XFAILS`.

Sibling precedent: #1555 (UC-019 wiring).

KonstantinMirin:
Follow-up from branch `refactor/transport-boundary` (PR #1600), specifically about the UC-010 half of this issue — the `CapabilitiesEnv` call site at `tests/bdd/conftest.py:3705-3724`.

Posting here rather than opening a new issue because the finding *is* "the remaining dormant UC-010 scenarios", plus a merge-loss warning for whoever picks this up. Related: #1709 (the narrower predecessor — wire the two main-flow scenarios; PR #1600 has now done that part), #1210 / #1592 (the capabilities+accounts work that owns the POST route and, as it turns out, already owns the fix below).

## Spec check first

**The spec is SILENT on the HTTP verb.** AdCP 3.1.1 binds protocol tasks to MCP tools and A2A skills only — `dist/docs/3.1.0/protocol/get_adcp_capabilities.mdx` ("Tool-Based Discovery": *"AdCP uses native MCP/A2A tool discovery… Browse agent's tool list (MCP) or skills (A2A)"*), and the only REST surface named anywhere in `dist/docs/3.1.0/protocol/architecture.mdx:23` is the separate Registry API. There is no HTTP/JSON binding for `get_adcp_capabilities`, so nothing in the pin mandates GET or POST for `/api/v1/capabilities`. That route is our own transport; **production is authoritative on the verb**, and the owner decision of 2026-07-24 (POST alongside the surviving GET, matching the RPC-over-REST convention elsewhere) stands unchallenged by the spec.

What the spec *is* explicit about is the payload and what gets graded:
- `dist/docs/3.1.0/protocol/get_adcp_capabilities.mdx`, "Request Parameters": `protocols` (optional protocol-domain filter) and `adcp_major_version` are request fields.
[...]

Issue #1591, opened 2026-07-13
get_products anonymous discovery: token-less requests get AUTH_REQUIRED despite public brand_manifest_policy


KonstantinMirin:
The AdCP spec expects anonymous discovery to succeed with pricing suppressed when `brand_manifest_policy` is public, but every `_impl` runs the `require_identity` gate (src/core/auth.py) — a token-less wire request gets AUTH_REQUIRED on all transports.

Graded by the strict-xfail BDD scenario `T-UC-001-alt-anonymous` (BR-UC-001-discover-available-inventory.feature); the moment production allows anonymous discovery the scenario xpasses and must graduate.

Related: #1088 (transport-boundary identity resolution).

KonstantinMirin:
## Also wire the BDD scenario in this change

Closing this gap unblocks a scenario that is currently strict-xfailed, so the fix and its grading should land together — otherwise the fix ships unverified and the scenario stays dormant.

**Wire when this lands:** `T-UC-001-alt-anonymous`

The scenario expects token-less discovery to succeed under a public `brand_manifest_policy`; production returns `AUTH_REQUIRED`. The conftest comment attributes the gap to the #1088 boundary work.

The scenario, step definitions and harness env already exist and are green on `refactor/transport-boundary`; #1822 ships the rest of that use case's wiring and deliberately leaves this tag out, because wiring it while the gap is open would add a strict-xfail that grades nothing.

Source: the `_SPEC_GAP_XFAILS` entry in `tests/bdd/conftest.py` naming this gap.

PR #1600, opened 2026-07-13
refactor: transport-boundary cluster — Principal at the boundary, typed format_ids, AdapterConfig repository, Pattern #7 grounding (#1088, #1169, #1172, #1442; stacked on #1585)

### LINKED ISSUE #1088: Resolve Principal at transport boundary, remove get_principal_object from _impl
## Problem

`get_principal_object()` is called inside every `_impl` function to look up the Principal Pydantic model (with `platform_mappings`) from the DB. This violates the transport boundary pattern — Principal should be resolved once at the boundary and included in `ResolvedIdentity`, same lazy-loading pattern we use for tenant.

## Current state

Every `_impl` that needs adapter selection does:
```python
principal = get_principal_object(identity.principal_id, tenant_id=identity.tenant_id)
adapter = get_adapter(principal, ...)
```

There are also two duplicate implementations:
- `auth.py:291` — the one everyone imports (working)
- `auth_utils.py:84` — dead code with a missing import (removed in PR #1071)

## Proposed fix

1. Add `principal` (lazy-loaded) to `ResolvedIdentity`, resolved at the transport boundary
2. Remove all `get_principal_object()` calls from `_impl` functions
3. Remove `get_principal_object` from `auth.py` once no callers remain

## Affected files

- `src/core/resolved_identity.py` — add principal field
- `src/core/tools/products.py`
- `src/core/tools/media_buy_create.py`
- `src/core/tools/media_buy_delivery.py`
- `src/core/tools/media_buy_list.py`
- `src/core/tools/performance.py`
- `src/core/tools/capabilities.py`
- `src/core/tools/signals.py`

## Context

Identified via PR #1071 code review (Chris's comment on `auth_utils.py:103`).

### LINKED ISSUE #1169: refactor: migrate remaining raw select(AdapterConfig) to AdapterConfigRepository

### LINKED ISSUE #1172: refactor: coerce Product.format_ids to list[FormatId] at the DB boundary
## Summary

`Product.format_ids` is a `JSONType` column that returns raw dicts from PostgreSQL JSONB. All downstream code must either dict-pluck (`fmt.get("id")`, `fmt.get("agent_url")`) or call `FormatId.model_validate(fmt)` at every consumption site. This is error-prone and violates the Pydantic-everywhere principle.

## Current state

```python
# models.py — returns raw dicts
format_ids: Mapped[list[dict[str, str]]] = mapped_column(JSONType, nullable=False)

# Every consumer must coerce manually
for fmt in product.format_ids:
    validated = FormatId.model_validate(fmt)  # or worse: fmt.get("id")
```

## Desired state

```python
# models.py — coerces to Pydantic on read
format_ids: Mapped[list[FormatId]] = mapped_column(JSONType(model=FormatId, is_list=True), nullable=False)

# Consumers get typed objects directly
for fmt in product.format_ids:
    print(fmt.id, fmt.agent_url)  # already FormatId, no coercion needed
```

## Implementation

1. Extend `JSONType` (`src/core/database/json_type.py`) with `model` and `is_list` parameters — `process_result_value` calls `model.model_validate()` on each entry when reading from DB
2. Update `Product.format_ids` column definition in `models.py`
3. Remove all `FormatId.model_validate(fmt)` and `fmt.get("id")` call sites — they become direct attribute access
4. Same for `InventoryProfile.format_ids` (line 1275 in models.py)

## Related

- PR #1167 replaced `isinstance(fmt, dict)` dispatch with `FormatId.model_validate()` — this issue eliminates the need for even that by coercing at the DB boundary
- The account management feature (`JSONType(model=BrandReference)`) was planned to use this pattern but `JSONType` doesn't support `model` yet

## Also address

- `AnyUrl` trailing-slash normalization: `str(AnyUrl("https://example.com"))` → `"https://example.com/"`. Currently worked around with `.rstrip("/")` at call sites. Should be handled once in `normalize_agent_url` or by using `HttpUrl` instead of `AnyUrl` on `FormatId.agent_url`.

### LINKED ISSUE #1442: Migrate api_v1 REST *Body models to SalesAgentBaseModel (Pattern #7 extra-field policy)

KonstantinMirin:
## Stacked on

Stacked on **#1585** (validation-harness base: adcp 6.6 consolidation), which itself stacks on #1417/#1430/#1567. Land those first and this diff shrinks to the transport-boundary work alone. Delta vs the base branch: https://github.com/KonstantinMirin/prebid-salesagent/compare/feature/validation-harness-base...refactor/transport-boundary

## What this resolves

Closes #1169, closes #1172, closes #1442 (residual scope), closes #1088. Also settles the `get_signals` half of #1353 (registered on MCP/A2A/REST with a flat spec-shaped wire; `activate_signal` remains open there, tracked separately).

- **#1088** — `Principal` is resolved once at the transport boundary: `ResolvedIdentity.principal` is populated eagerly in `resolve_identity()` (zero extra queries — the boundary already loaded the row), `require_principal`/`find_principal` helpers serve the `_impl`s, background workers get a `PrincipalRepository` (they resolve *other* principals, not the caller), the dead lookup chain is deleted, and a new empty-allowlist AST ratchet bans principal DB lookups in `src/core/tools/` (recursive scan).
- **#1169** — all 19 remaining raw `select(AdapterConfig)` sites route through `AdapterConfigRepository`; `get_or_create` is pinned to PK-only find. `no_raw_select` allowlist −9.
- **#1172** — `Product.format_ids`/`InventoryProfile.format_ids` are `list[FormatId]` from the DB boundary out: `JSONType(model=…)` validates raw dicts at bind (every writer covered), ~40 shape-reparsing branches deleted, and format-identity comparison now has ONE canonical key (transport-suffix stripping folded into `canonical_agent_url`; shared `supported_format_keys`/`format_key` helpers replace four divergent sites; a guard keeps `normalize_agent_url` at zero callers).
- **#1442** — the Body-model migration had already landed; this grounds the Pattern #7 extra-field policy in evidence: dev-forbid INVALID_REQUEST integration test, prod-ignore subprocess pin, and a cross-transport BDD scenario (spec 3.1.1 sets `additionalProperties: true` on request schemas — prod-ignore is the compliant contract).
- **BDD**: previously dormant BR-UC-001/008/009/010 bindings wired and graded cross-transport (UC-009 now genuinely executes on REST; UC-010's capability asserts run live instead of hiding behind an xfail). Production gaps are strict xfails with GH refs — never papered over.

## Review

A 31-finding multi-agent review round (8 reviewers + synthesis, every finding independently re-verified against the head tree) ran before this PR opened; all 17 Should-fix items plus 4 zero-tolerance DRY/hygiene items are fixed in the branch. Details in the first comment.

Real bugs caught along the way, worth naming: a format-identity check that accepted a buyer's agent_url on update/sync but rejected the same input at create; a dead adapter-config write path behind a tuple-truthiness bug whose pin test couldn't fail; an unguarded typed-column write that could persist an unreadable row; a nested DB session per webhook report under scheduler fan-out; and — exposed by the new bind-validation itself — `sync_accounts` stringifying a `BrandId` RootModel into stored data.

## Follow-ups

Every item found during this branch's review rounds that is *not* fixed here has a GitHub issue — either an existing one it was added to, or a new one linking its relatives. Each was checked against the pinned spec (adcp==6.6.0 / AdCP 3.1.1) before filing; where the spec has no bearing, the issue says so explicitly rather than leaving a reader hunting for a citation.

**Connected to an existing issue**

| Issue | What was added |
|---|---|
| #1602 | Error-code/recovery taxonomy: `_sync.py:344-352` discards `code` and `recovery` when degrading a typed `AdCPError` to a per-item failure; plus the strict-vs-lenient creative-not-found split in `_assignments.py` |
| #1589 | Repo-wide SSRF egress seam: `creative_agent_registry.py:534` dials a registered agent through a raw `httpx.AsyncClient` with no validation. Tenant-admin-gated rather than buyer-supplied, which affects the seam's API — it needs an operator-configured mode distinct from the buyer-facing one |
| #1711 | `get_media_buy_delivery` silently ignores request fields: `include_package_daily_breakdown` is the third instance, accepted and documented at `media_buy_delivery.py:733/751/763/781` then discarded at `:549`. Unlike the other two it is spec-graded (`dist/compliance/.../delivery_reporting.yaml`) |
| #1594 | Wiring dormant UC-010 scenarios: the capabilities harness pins `REST_METHOD = "get"`, so a POST route cannot be graded. Also records that this branch and `feature/spec-gaps-1210` have each independently rewritten `tests/harness/capabilities.py` and the UC-010 feature since `8cb8d0ed6` — neither a superset, so that merge must be semantic |

**Filed new**

| Issue | Summary |
|---|---|
| #1756 | UC-004 scenarios partition over `sampling_method` / `ownership` — fields absent from the entire pinned v3.1.1 tree, so production rejecting them is correct and the scenario is the defect. Retiring them is a ledger **shrink** |
| #1775 | A2A submitted-detection is gated on `res["success"]`, which is derived as `not bool(errors)` — so a submitted response carrying *advisory* errors is never classified as submitted (no `TaskStatusUpdateEvent`, no webhook, no `SUBMITTED` state) |
| #1782 | The `e2e_rest` ledger's prose counts are unverified free text; all three ledger guards strip comments before parsing |
| #1783 | Natural-language `signal_spec` returns zero signals — whole-phrase substring matching is not the inference the spec's `discovery_mode: "brief"` requires. Graded by `dist/compliance/3.1.1/domains/signals/index.yaml` |
| #1784 | Dead `.format-checkbox` / `.format-card` JS and CSS in `add_product_gam.html`, orphaned by the `FormatTemplatePicker` migration |
| #1785 | Removable `cast(list[LibraryProduct], …)` at `src/core/tools/products.py:723` — a true widening, unlike the five casts already retired on this branch |
| #1786 | Orphaned `render_template` kwargs — which turned out to be three classes: **4 admin routes whose template does not exist** (two hard 500s), `error.html` kwarg drift silently discarding every GAM error diagnostic, and ~60 genuinely dead producers including a synchronous fan-out to every creative agent on each GAM product page load |

Earlier rounds also filed #1591–#1596, and `activate_signal` conformance remains the other half of #1353.

The `e2e_rest` ledger no longer cites any local tracker id — every entry now resolves to #1600, #1711, #1739, #1751, #1756 or #1418.

## Verification

All verified green.


KonstantinMirin:
## Closing: this needs to be a series of smaller PRs, on re-architected premises

@ChrisHuie — apologies for the long wait on the review request. Closing this rather than asking you to open 252 files, and the reasons are worth recording.

### The bundle had no architectural coherence

This PR carried #1088, #1169, #1172 and #1442 under a "transport boundary" banner. Examined individually, not one of them is a transport-boundary ticket:

| Issue | Actually touches |
|---|---|
| #1088 | identity plumbing — resolve `Principal` once, delete a dead duplicate |
| #1169 | **admin-layer data access** — 13 of ~20 sites are admin blueprints |
| #1172 | **persistence typing** — coerce `format_ids` at the DB boundary |
| #1442 | **schema policy** — Pattern #7 extra-field handling |

Four layers, one name. There was no single invariant to hold, so each ticket's fallout ran in a different direction and "fix the fallout at the architecture level" was impossible — there were four architectures. The sprawl was structural, not a discipline failure.

And the premise was wrong: `src/core/tools/` — the transport-boundary surface — has **zero** `get_db_session()` calls and an empty `IMPL_SESSION_ALLOWLIST`. That layer was already clean. The 326 session sites are in `admin/blueprints` (23 files), `services` (16) and `adapters`. #1169 is really the first slice of persistence-ownership work, which starts somewhere else entirely.

### What the branch actually produced

Not a refactor — a survey. Production is ~1,000 lines across 51 files; tests are ~8,800 across 180. The findings are the asset, and several invalidate assumptions this repo was relying on:

- **Guards that never guarded.** `test_architecture_bdd_no_duplicate_steps.py` matches **zero** groups under its own key — its empty allowlist read as "clean" and meant "nothing checked". Two more were inert (one PEP 604-blind, so 0 of 14 fields registered), and the `ast-grep` pre-push hook had two never-match placeholder rules. Every real discovery this month came from mutation runs, not from AST guards.
- **`extra="allow"` on ~24 schema classes.** They inherit it from the SDK `Library*` bases, so `SalesAgentBaseModel` — and Pattern #7 with it — is not in their MRO at all. CLAUDE.md's Pattern #1 example (`class Product(LibraryProduct)`) *is* the form that produces this, and the schema-inheritance guard enforces that example. Two mandated patterns that are mutually exclusive as written.
- **`get_products` never used the canonical format key** — `products.py:492` compares bare `id`, discarding `agent_url`, with a comment citing #1172 directly above it. The invariant this PR claims was false at the most-called discovery tool.
- **`estimated_exposures` is never written** by production but is read by the `min_exposures` filter, which therefore excludes every guaranteed product and logs a plausible-looking reason.

### Why smaller PRs, and why re-architected

For each of the four, the structural form is smaller than what is in this branch, and in each case the guard exists only because the wrong thing is still expressible:

- **#1442** — a module re-exporting the SDK types with the policy stamped, plus one `TID251` banned-api line. Pattern #1's documented form then works unchanged, ~24 per-class declarations evaporate, no new guard, and 122 scattered `adcp.types` imports collapse to one boundary.
- **#1172** — canonical identity on `FormatId` itself, so a bare-`id` comparison is not something you can write by accident. Retires the guard and fixes `products.py:492`.
- **#1088** — `_impl` does not receive a session, so the lookup is unreachable.
- **#1169** — reframed and retitled as persistence-ownership, admin blueprints first.

### Where the work goes

The most valuable part is independent of all of the above: six use cases were wired to grading against AdCP 3.1.1, and four of them have no dependency on the refactor. Those ship on their own, each de-scoped to the scenarios that genuinely assert — a scenario wired straight into a strict-xfail grades nothing:

- #1822 BR-UC-001 · #1823 BR-UC-008 · #1824 BR-UC-009 · #1825 BR-UC-010 (under umbrella #1594)
- masked scenarios move to the gap that blocks them, to be wired in the same change as the fix: #1595, #1591, #1783, #1592

Also extracting separately: the `paused` status-precedence work (#1619), and the small independent fixes — `get_signals` request typing, the A2A validation-envelope operation label, the admin format-picker `agent_url` fix, the DRY consolidations, the four *executable* guards, and the removal of three dead pre-commit hook scripts.

Follow-ups filed or widened from this branch: #1756, #1775, #1782, #1783, #1784, #1785, #1786, #1789, #1796, #1797, #1798, plus additions to #1602, #1589, #1711 and #1594.

### One item owed directly to you

From your #1585 review: the "creative-format check implemented three divergent ways" item **is fixed** — the canonicalizer consolidation landed, with a guard keeping `normalize_agent_url` at zero callers in `src/`. It went out with this branch, so it will arrive via the #1172 slice rather than here.

The branch stays on `refactor/transport-boundary` as the reference for all of the above.

PR #1579, opened 2026-07-10
docs(adr): honest #1228 F1/F2/G2 quality-amnesty policy (ADR-009)

mkostromin-sigma:
Closes #1626. Relates to #1228, #1609, #1610, #1611.

## Summary

@ChrisHuie — **decision PR** for [#1228](https://github.com/prebid/salesagent/issues/1228) Cluster **F1 / F2 / G2**. Revised per your review (option 3: ratchet, not permanent amnesty). Aligns with @KonstantinMirin preference for ratchet-style.

### Decisions (revised)

| ID | Surface | Decision | Companion |
|---|---|---|---|
| **F1** | Ruff ignores | **Ratchet** `C901`, `PLR0912`, `PLR0915`. **Accept** `PLR0911`, `PLR0913`, `PLR2004`. | [#1613](https://github.com/prebid/salesagent/pull/1613) / [#1610](https://github.com/prebid/salesagent/issues/1610) |
| **F2** | Mypy lenient flags | **Not** permanently deferred — `check_untyped_defs` count ratchet as follow-up | [#1611](https://github.com/prebid/salesagent/issues/1611) |
| **G2** | Obligation allowlist | Docstring correction + reconcile vs shrink-only rule | [#1612](https://github.com/prebid/salesagent/pull/1612) / [#1609](https://github.com/prebid/salesagent/issues/1609) |

This PR is **ADR-009 + config comment honesty only** (G2 docstring stripped out per your sequencing suggest).

Full write-up: [`docs/decisions/adr-009-quality-amnesty-honesty.md`](docs/decisions/adr-009-quality-amnesty-honesty.md) (status: **Proposed** until Accept).

### Test plan
- [x] Docs/config only — no production behavior change
- [ ] CI green after ADR-only split
- [ ] Chris Accept on revised ADR-009

Issue #1228, opened 2026-04-22
ci: restore quality gate — systemic CI, baseline drift, and config/runtime divergence


ChrisHuie:
> **Status:** Tier 1 (restore the gate) + most of Tiers 2 / 4 / 5 are **absorbed by #1234**. Tier 3 (permanent-amnesty cleanup) and a handful of follow-ups remain. Close this issue when #1234 merges + the residual items are resolved (see "Closure plan" below).
> **Read order:** Status table → Residual scope → expand `<details>` for the original 30-finding catalog and historical context.

## TL;DR

CI looks strict; it isn't. Three independent leaks ("Soft Gate") allow stated quality invariants (DRY, No Quiet Failures, Test Integrity, type ratcheting, structural guards) to drift without enforcement. Downstream of that softness, ratcheting baselines have drifted, hook environments diverged from the runtime, and dead config is being cargo-culted through files.

**#1234's CI/pre-commit refactor closes the soft-gate root cause and ~14 of the 30 findings here.** The remaining ~16 are mostly:
- **Tier 3 architectural-debate items** — the "fix incrementally / never gets fixed" ruff ignores + mypy lenient flags + 301-entry obligation allowlist
- **A few misc cleanup items** that fell outside #1234's scope (orphan `.mypy_baseline`, `google_ad_manager_original` phantom refs, `.type-ignore-baseline` rebaseline, uv cache key)
- **One critical residual P0** that #1234 partially addresses but doesn't fully close: `release-please publish-docker` still doesn't `needs: [test-summary]` — red main can ship images.

### Cluster A — Soft Gate (meta root cause)

| # | Finding | Status | Closed by |
|---|---|---|---|
| A1 | `test-summary` omits `smoke-tests` from failure aggregation | ✅ Absorbed | #1234 PR 3 commit 3 — new `Summary` job's `needs:` covers all 14 frozen checks including `Smoke Tests` |
| A2 | Ruff lint CI-disabled (`\|\| true` + `continue-on-error: true`) | ✅ Absorbed | #1234 PR 3 commit 7 (also closes #1233 D6) |
| A3 | Zero of 27 pre-commit hooks run in CI | ✅ Absorbed | #1234 PR 3 commit 3 — `Quality Gate` job runs `pre-commit run --all-files` (also closes #1233 D5) |
[...]

PR #1612, opened 2026-07-13
docs: correct G2 obligation-allowlist docstring (#1609)

mkostromin-sigma:
## Summary

Standalone #1228 **G2** docstring fix per Chris on #1579 (soft suggest: land G2 separate from ADR).

- Allowlist **may grow** for new untested obligations (reviewable)
- **Must shrink** when `Covers:` lands (existing CI)
- Reconcile vs "allowlists can only shrink" (violation allowlists ≠ coverage backlog)

Closes #1609. Complements #1579 (ADR-009).

## Test plan
- [x] Docs-only docstring change
- [ ] CI green

PR #1534, opened 2026-07-05
fix: wrap MCP TypeAdapter validation errors

jstar0:
## Summary

Fixes #1504.

FastMCP TypeAdapter validation can fail before an MCP tool body runs. Those failures previously escaped as raw Pydantic errors instead of the two-layer AdCP error envelope used at the transport boundaries.

## Changes

- Enrich Pydantic `ValidationError` normalization in the shared `normalize_to_adcp_error` path with a structured field and per-error details.
- Build the buyer-facing message from Pydantic's first safe error message instead of `str(exc)`, avoiding input-value and versioned documentation URL leakage.
- Route terminal MCP TypeAdapter failures through the existing `_translate_to_tool_error` boundary while preserving the production deep-strip retry.
- Record TypeAdapter rejection through the normal MCP boundary audit/activity path, using the identity already resolved by middleware.
- Reuse the structured validation-details builder at the A2A/in-body boundary so all three validation-envelope paths carry the same safe per-error projection.
- Cover real indexed field paths, sanitized messages, retry and no-op-strip behavior, the exact audit identity key and typed record, the in-memory MCP wire envelope, and the shared A2A validation boundary.
- Track the existing `create_media_buy` `INVALID_REQUEST` storyboard mismatch as a strict xfail instead of asserting the current cross-transport gap as intended behavior.
- Link the two dormant validation-error BDD gates to follow-up #1652, which is separate from the error-code classification work in #1604.

## Risk

The behavior change is limited to Pydantic validation failures translated at transport boundaries. The A2A/in-body boundary now adds the same buyer-safe `details.validation_errors` projection already used by MCP and REST. Existing typed `AdCPError` values pass through unchanged, ordinary `ValueError` handling is unchanged, and non-TypeAdapter MCP exceptions still propagate normally.

## Verification

```sh
uv run pytest -q tests/unit/test_exception_normalization.py tests/unit/test_mcp_compat_middleware.py tests/integration/test_mcp_typeadapter_validation_envelope.py tests/unit/test_validation_errors.py tests/unit/test_architecture_error_envelope_two_layer.py tests/unit/test_error_format_consistency.py
uv run pytest tests/unit/ tests/harness/ -q --tb=line
[...]

PR #1613, opened 2026-07-13
ci: ratchet ruff C901/PLR0912/PLR0915 via .ruff-complexity-baseline (#1610)

### LINKED ISSUE #1610: ci: ratchet ruff C901/PLR0912/PLR0915 via .ruff-complexity-baseline
## Summary

#1228 **F1** follow-up per [@ChrisHuie](https://github.com/ChrisHuie) on [#1579](https://github.com/prebid/salesagent/pull/1579) / ADR-009:

> Ratchet the three pure-complexity rules — C901, PLR0912, PLR0915 — via a count baseline.
> A `.ruff-complexity-baseline` is a near-copy of `check_type_ignore_count.py`.

### What

- Add `.pre-commit-hooks/check_ruff_complexity_count.py` (count ratchet)
- Add `.ruff-complexity-baseline` (per-rule JSON: C901 / PLR0912 / PLR0915)
- Wire into `Makefile` `quality-ci` + `.pre-commit-config.yaml` `pre-push` (not commit-stage — D27 ceiling)
- Auto-lower on decrease; fail on increase; `--update-baseline` requires review justification (tracked-file diff)
- Leave `PLR0911` / `PLR0913` / `PLR2004` permanently accepted (ADR-009)

### Measured baseline (src/, ruff 0.15.x)

| Rule | Count |
|------|------:|
| C901 | 186 |
| PLR0912 | 132 |
| PLR0915 | 108 |

### Relationships

- Parent: #1228 (Cluster F1)
- Blocked by: #1579 (ADR-009 Accept — policy record)
- Sibling: #1579 does **not** implement this hook

### Guardrail

Baseline going **up** must be contested in review (`--update-baseline` rewrites a tracked file).

mkostromin-sigma:
## Summary

#1228 **F1** follow-up per Chris on #1579 / ADR-009:

- Hook: `.pre-commit-hooks/check_ruff_complexity_count.py`
- Baseline: `.ruff-complexity-baseline` — C901=186, PLR0912=132, PLR0915=108
- Wired into `make quality-ci` + pre-push (not commit-stage — D27)
- Auto-lower on decrease; fail on increase; `--update-baseline` is a reviewable tracked-file rewrite

Leaves `PLR0911` / `PLR0913` / `PLR2004` permanently accepted (ADR-009).

Closes #1610. Complements #1579.

## Test plan
- [x] Hook matches baseline (exit 0)
- [x] Artificial baseline↓ fails (exit 1)
- [ ] CI green (`make quality-ci` includes new step)


mkostromin-sigma:
CI green on the F1 complexity ratchet.

Implements ADR-009 / Chris review: count-ratchet `C901` / `PLR0912` / `PLR0915` via `.ruff-complexity-baseline` (186 / 132 / 108). Baseline ↑ must be justified in review (`--update-baseline` rewrites a tracked file).

Ready for review. Complements #1579 (policy Accept) and #1612 (G2 docstring).

ChrisHuie:
Re-measured against `origin/main` (ruff 0.15.14, the locked version): `.ruff-complexity-baseline` `{C901:186, PLR0912:132, PLR0915:108}` equals main's current counts exactly (same at the merge-base and this branch head), and running the hook against current main exits 0. The downward ratchet fires correctly — dropping the baseline to 185 makes the hook exit 1 on main's 186 — and `quality-ci` runs it in CI, so an unjustified increase fails unless someone deliberately `--update-baseline`s the tracked file. Right design.

Three things before merge:

**1. [should-fix] DRY.** `check_ruff_complexity_count.py` is a third copy of the count-baseline ratchet skeleton — `check_type_ignore_count.py` and `check_code_duplication.py` are the other two (same create / `--update-baseline` / compare / auto-lower control flow, differing only in the count method, baseline serialization, and labels). The duplication guard can't catch it: `check_code_duplication.py` scans only `src/`, `tests/`, `scripts/` — not `.pre-commit-hooks/`. The three already diverge in robustness (different subprocess return-code guards). Either extract a shared ratchet driver (`count_fn` + `baseline_path` + labels) now, or file a tracked follow-up to consolidate all three — it won't surface on its own.

**2. [nit] The baseline depends on ruff's DEFAULT complexity thresholds.** `[tool.ruff.lint.mccabe]` and `[tool.ruff.lint.pylint]` are absent from `pyproject.toml`, so a future ruff bump that changes a default (the pin is `>=0.8.0`, a floor) would shift the live count while the baseline stays frozen → a spurious trip or a silent auto-lower. The ADR flags this exact drift risk for the mypy ratchet but not this one. Pin to the current defaults so the ratchet is version-independent:

```toml
[tool.ruff.lint.mccabe]
max-complexity = 10
[tool.ruff.lint.pylint]
max-branches = 12
max-statements = 50
```

**3. [nit] The inline counts in the pyproject comments drift** the moment anyone fixes one function (the baseline auto-lowers; the comment won't). The baseline file is the source of truth now — drop the numbers from the C901/PLR0912/PLR0915 comments, e.g.:

```
"C901",    # count-ratcheted via .ruff-complexity-baseline (ADR-009); not a hard gate
```

and drop the bare issue number from those comments (keep `ADR-009`).

Sequencing: this conflicts with #1579 on the same ignore-block lines (both branch from the same base; neither contains the other). Land #1579 first, then rebase this onto its reorganized block so there's one canonical comment per rule with the correct count (186/132/108).

mkostromin-sigma:
## Review reply (CI green)

All three items addressed on the tip (rebased onto updated #1579):

1. **[should-fix] DRY** — extracted `.pre-commit-hooks/count_ratchet.py`; `check_type_ignore_count` / `check_code_duplication` / `check_ruff_complexity_count` share create / `--update-baseline` / compare / auto-lower. Also hardened ruff empty-stdout-on-exit-1 (no silent zeroing).
2. **[nit] Pin defaults** — `[tool.ruff.lint.mccabe] max-complexity = 10` and `[tool.ruff.lint.pylint] max-branches = 12` / `max-statements = 50`.
3. **[nit] Drop inline counts** — comments point at baseline + ADR-009 only.

Baseline remeasured after rebase onto latest main: `{C901:187, PLR0912:134, PLR0915:108}` (↑ vs earlier 186/132 is main tip drift, not this PR’s code).

**CI:** all green. Please land **after** #1579 (ignore-block sections owned there).

mkostromin-sigma:
## Status (CI green)

Rebased onto `upstream/main` (incl. adcp 6.6 / `#1567`). Remeasured `.ruff-complexity-baseline`: **C901 187→186** (auto-lower), PLR0912/PLR0915 unchanged.

No open review threads on this PR — ready when `#1579` / maintainer order allows. Further `count_ratchet` DRY follow-ups landed on `#1634`.

ChrisHuie:
Re-reviewed at `e75812f3` — ran all three hooks at the head, not just read the diff. The three items from my last pass are verified addressed:

- **DRY** — `count_ratchet.py` now owns the create / `--update-baseline` / compare / auto-lower skeleton; `check_type_ignore_count`, `check_code_duplication`, and `check_ruff_complexity_count` each call `run_count_ratchet` and keep only their count method + baseline codec. Verified live: ruff exits 0 on the committed baseline, exits 1 with no auto-write when C901 drops 186→185, and auto-lowers 187→186 on a decrease; the type-ignore `origin/main` raise-guard survives with the `min(baseline, current)` probe intact. The silent-zero robustness gap is also closed on both count methods (ruff aborts on exit-1-with-empty-stdout; duplication aborts on `returncode & 33 and count == 0`).
- **Pin defaults** — `[tool.ruff.lint.mccabe]` / `[tool.ruff.lint.pylint]` set to 10 / 12 / 50; the 186/134/108 baseline reflects them.
- **Inline counts** — dropped; comments carry rule meaning only.

Two things still block a clean merge:

**1. [should-fix] The #1579 overlap is still here.** This branch carries #1579's whole policy surface verbatim — `adr-009-quality-amnesty-honesty.md` is byte-identical, `docs/decisions/README.md` and `mypy.ini` are the same edits, and `pyproject.toml` here is #1579's ignore-block reorg with the mccabe/pylint block added on top. #1579 is still open, so these are parallel branches, not stacked. Either land #1579 first and rebase this down to mechanism-only (hooks + baseline + config wiring + pin block), or supersede #1579 with this PR and close it — but merging both independently double-applies the ADR and lands it as "Proposed" through the mechanism PR.

**2. [should-fix] `count_ratchet.py` has no test.** It now drives all three ratchets, so a silent regression there disables every ratchet — the "looks-enforced but isn't" failure ADR-009 exists to prevent. Add a unit test for `run_count_ratchet` (create / increase→1 / decrease→auto-lower / `--update-baseline`), or fold onto #1634. Heads-up for whoever writes it: `out` / `err` default to `sys.stdout` / `sys.stderr` bound at import, so `capsys` won't capture — use `capfd` or resolve the streams inside the function.

**[nit]** `check_ruff_complexity_count.py` carries a `#!/usr/bin/env python3` shebang but is mode `100644`, while `check_type_ignore_count.py` is `100755`. It's invoked via `python`, so it's cosmetic — drop the shebang or `chmod +x`.

The mechanism is correct and verified end-to-end; the remaining gate is the #1579 sequencing plus an approving review.

mkostromin-sigma:
## Sequencing update — #1579 merged

[#1579](https://github.com/prebid/salesagent/pull/1579) is on `main` (`06f89876`, ADR-009 **Accepted**).

Per @ChrisHuie should-fix #1: this PR can now rebase onto `upstream/main` and drop the overlapping ADR / `docs/decisions/README.md` / shared config comment surface so it is **mechanism-only** (hooks + baseline + pin block + wiring).

Remaining gate from Chris: unit tests for `count_ratchet.py` (or fold onto #1634).

Closes #1610 when this lands.

ChrisHuie:
# Re-review — PR #1613 (ruff complexity ratchet)

Re-reviewed at head `e75812f3`, base `origin/main` `06f89876` (pulled 2026-07-15). Ran all three ratchet hooks and the architecture suite against the head, and re-measured the baseline — not a diff read.

**Bottom line:** The mechanism is correct and verified end-to-end (baseline exact, pins == ruff defaults, ratchet enforces on the wire). It is **not mergeable at this head** — the branch is `CONFLICTING` because #1579 landed on `main`, and the two standing blockers from the last pass (rebase to mechanism-only, `count_ratchet` test) are both still open. Everything green in the checks tab ran *before* #1579 merged, so it does not reflect the merged result. Below: what cleared, two blockers, four should-fix, and the nits.

## Prior feedback status

| Round | Item | State |
|---|---|---|
| 1 | [should-fix] extract shared ratchet driver (DRY) | Addressed — `count_ratchet.py` owns the create/compare/auto-lower skeleton; both siblings call it; the old inline copies are deleted (F401/mypy clean) |
| 1 | [nit] pin ruff default thresholds | Addressed — `[tool.ruff.lint.mccabe]`/`[pylint]` = 10/12/50 |
| 1 | [nit] drop inline counts from pyproject comments | Addressed — the ignore-block already matches `main`; only the pin block is net-new |
| 2 | [should-fix] rebase to mechanism-only (#1579 overlap) | **Open** — see B1 |
| 2 | [should-fix] `count_ratchet.py` has no test | **Open** — see B2 (author's own "remaining gate") |
| 2 | [nit] shebang/mode mismatch | **Open** — see N2 |

## Blockers

**B1 — Not rebased: `CONFLICTING` with `main`, and a stale ADR draft on the branch.**
- **Where:** `docs/decisions/adr-009-quality-amnesty-honesty.md`, `docs/decisions/README.md`, `mypy.ini`; 6 of the 9 branch commits (`50c81449`..`dceb7919`) are #1579's ADR work.
- **Problem:** #1579 is now on `main` (`06f89876`, ADR-009 **Accepted**). The ADR doc did not exist at the merge-base (`c7acbbd6`) — both branches added it — so updating this branch produces an **add/add conflict**, not a clean fast-forward. The branch's copy is the older **"Proposed / Awaiting re-Accept"** draft and drops #1579's residual-honesty edits (the `mypy.ini` two-part-form block, the boy-scout paragraph); `README.md` carries the ADR-009 row as `Proposed`. `mypy.ini` is now byte-identical to `main` (redundant). If the conflict is resolved by keeping the branch side, it silently reverts an accepted ADR.
- **Fix:** rebase onto current `origin/main`; drop/squash the 6 ADR-doc commits (superseded by #1579) and the redundant `mypy.ini` edit; fold this PR's `.ruff-complexity-baseline` rows into #1579's landed ADR rather than re-adding a parallel draft. Result is mechanism-only: the 3 hooks + `count_ratchet.py` + `.ruff-complexity-baseline` + the pyproject pin block + Makefile/pre-commit wiring + the `contributing.md` baseline-list line + the CODEOWNERS/coverage-map entries. The pyproject `--- Ratchet targets ---` split is already on `main` byte-for-byte — do not re-author it.

**B2 — The one enforcer of three debt categories ships untested.**
- **Where:** `.pre-commit-hooks/count_ratchet.py` (`run_count_ratchet`); no test references it (`git grep run_count_ratchet -- tests/` → 0).
- **Problem:** `C901`/`PLR0912`/`PLR0915` sit in ruff's `ignore` list (`pyproject.toml:148-150`), so `ruff check` does not flag them — this hook is now their *only* enforcement, and it also drives the type-ignore and duplication ratchets. A silent regression in the driver (an inverted `cur > base`, or an auto-lower that leaks into the regression branch) disables all three at once — the exact "looks-enforced but isn't" failure ADR-009 exists to prevent. Author's latest comment names this as the remaining gate and offers to fold it onto #1634 — but #1634 *depends on* this driver, so the driver's own test belongs in this PR.
- **Fix:** add a unit test for `run_count_ratchet` where each case reddens under a named mutation: create (no baseline → write + rc 0); `--update-baseline` (rewrite + rc 0); **increase** (`cur > base` → rc 1 **and** baseline bytes unchanged — the highest-value case, catches a no-auto-lower-on-regression break); decrease (rc 0 + baseline lowered); all-equal (rc 0 + no write); mixed multi-key (one up/one down → rc 1 **and** the down key is *not* auto-lowered). Write it with injected streams (`out=io.StringIO()`, `err=io.StringIO()`) or `capfd` — `capsys` will not capture this driver (see N1) and a `capsys`-based assertion passes vacuously. `tests/unit/test_architecture_pre_commit_coverage_map.py` (`_BROKEN_ENTRY_PROBES`) is an in-repo template for the shape. Whoever writes it: the hooks import `from count_ratchet import …`, so the test needs `.pre-commit-hooks/` on `sys.path` (conftest or import-by-path).

## Should-fix

**S1 — `RULES` is a hand-copied subset of a canonical list, unpinned.**
- **Where:** `.pre-commit-hooks/check_ruff_complexity_count.py:28` — `RULES = ("C901","PLR0912","PLR0915")`, duplicating the `--- Ratchet targets ---` bucket at `pyproject.toml:147-150`; no test ties them.
- **Problem:** add a 4th complexity rule to the pyproject ratchet bucket and forget `RULES`, and the new rule is both ignored by ruff and uncounted by the hook — zero enforcement, nothing red.
- **Fix:** add a membership-pin test asserting `RULES` equals the rules under the pyproject "Ratchet targets" header, or derive `RULES` from that header. Cheapest, highest-value item here.

**S2 — The complexity ratchet is soft where the sibling is hard, without saying so.**
- **Where:** `.pre-commit-hooks/check_ruff_complexity_count.py` (no baseline-raise guard) vs `.pre-commit-hooks/check_type_ignore_count.py:62-70` (`check_baseline_not_raised`, which hard-fails when the local baseline exceeds `origin/main`).
- **Problem:** a committed baseline **raise** rides CI green through the new hook (mutation-confirmed: baseline `{C901:187}` + current `{C901:187}` → exit 0). Type-ignore blocks that on the wire; complexity relies on review catching the tracked-file diff. That's a defensible model — ADR-009 documents soft-ratchet-with-review, and CODEOWNERS gates `.ruff-complexity-baseline` — but the asymmetry is undocumented and the stronger guard was kept for type-ignore *in this same PR*.
- **Fix:** port `check_baseline_not_raised` to the new hook (defense-in-depth, consistent with the sibling), or state in ADR-009 that complexity/duplication intentionally stay soft.

**S3 — The DRY win the PR just made is itself outside the DRY guard, and one clone remains.**
- **Where:** `.pre-commit-hooks/check_code_duplication.py` — `SCOPES=("src","tests","scripts")` excludes `.pre-commit-hooks/`. `pylint --enable=R0801 .pre-commit-hooks/` reports one residual clone: `check_ruff_complexity_count.py` `main()` boilerplate ⟷ `check_type_ignore_count.py` `main()` (argparse + repo_root/src_path/baseline_file resolution + `src_path.exists()` guard).
- **Problem:** the consolidation is real, but nothing protects it from re-duplicating, and one clone survived the extraction. Related: the "a crashed subprocess must not report 0 and auto-lower to zero" guard is implemented two different ways — `check_ruff_complexity_count.py` (`returncode not in (0,1) or (==1 and not stdout)`) vs `check_code_duplication.py:51` (`returncode & 33 and count == 0`); the pylint variant does not fire on exit-2-with-count-0 (a scanned-file syntax error). That hole is **pre-existing** (byte-identical on `main`) and self-limiting (an ephemeral CI auto-lower is discarded), so it is not introduced here — but the PR consolidates exactly this family.
- **Fix:** hoist the shared `main()` boilerplate into `count_ratchet.py` (removes the residual clone); consider a parameterized guarded-subprocess helper in the driver so the crash-guard has one home (fixes the pylint exit-2 hole as a byproduct). If the boilerplate stays per-hook, widen the duplication scan set to include `.pre-commit-hooks/` with a baseline so future clones are caught.

## Nits

- **N1 — `out`/`err` default args are bound at import** (`count_ratchet.py:68-69`): `out: TextIO = sys.stdout` / `err: TextIO = sys.stderr` evaluate once at definition, and no caller passes them, so the injection seam is dead and `capsys` can't capture the driver. Default them to `None` and resolve at call-time (`out = sys.stdout if out is None else out`), or drop the params until the B2 test needs them. No production impact (each hook is a fresh process).
- **N2 — shebang/mode mismatch** (prior nit, unaddressed): `check_ruff_complexity_count.py` is mode `100644` with a `#!/usr/bin/env python3` shebang; `check_type_ignore_count.py` is `100755`. All hooks run via `uv run python`, so pick one: `chmod +x` the new hook, or drop the shebang. (`check_code_duplication.py` has the same 644+shebang, but that is pre-existing on `main`.)
- **N3 — thresholds unpinned by a test** (follow-up): nothing asserts `max-complexity=10`/`max-branches=12`/`max-statements=50`. A future pyproject threshold bump lowers the live count, and the driver auto-lowers the baseline silently. Lower priority than S1; a follow-up is fine.
- Also: a double blank line precedes the pin block in `pyproject.toml` (cosmetic); commit `e75812f3` message reads "remmeasure" (moot once the branch is rebased/squashed).

## Verified good (ran at this head)

- Baseline `{C901:186, PLR0912:134, PLR0915:108}` equals current `main`'s exact counts — since the PR touches no `src/`, it stays correct across the required rebase.
- Pins `10/12/50` == ruff 0.15.14 built-in defaults (checked with `--isolated`), so the pin changes nothing flagged today and locks against a future ruff default drift; no duplicate TOML tables.
- Ratchet enforces on the wire: at baseline → exit 0 no change; fake new debt (`current > baseline`) → **exit 1 with no auto-write**; fake fixed debt (`current < baseline`) → exit 0 with auto-lower.
- DRY extraction is honest — the old inline skeletons are deleted, both siblings share the driver + JSON/int codecs, `json` was dropped from the duplication hook; the type-ignore `origin/main` guard correctly stays (baseline-file integrity, not a count method).
- Wiring complete: `make quality-ci` (Makefile) + pre-push hook (`.pre-commit-config.yaml`); D27 intact — the new hook is `pre-push`, commit-stage count unchanged at 12, full architecture suite (506) passes; `.ruff-complexity-baseline` registered in CODEOWNERS and the pre-commit coverage map; `contributing.md` lists it.

## Ran / not run

- **Ran:** re-measured ruff counts on `origin/main`; all three hooks at the head; ratchet mutation tests (regression / decrease / baseline-raise); `git rebase origin/main` (to observe the conflict); architecture suite (506), D27 and coverage-map guards; `ruff --show-settings`/`--isolated` for the pins; `pylint R0801` on `.pre-commit-hooks/`.
- **Not run:** a real complexity-increase through the full hook binary and a real `--update-baseline` write (mutation-tested the driver path instead); a full green run of the duplication hook (pylint over the whole tree timed out — proved equivalence via the byte-identical merge-base `count_duplications` + the shared driver green elsewhere); a forced pylint exit-2 to confirm the S3 hole triggers in practice (the guard divergence is certain; the real-world trigger is inferred).

## Actions

- **B1 — rebase to mechanism-only:** FIX-NOW. Blocks merge — the branch is `CONFLICTING` and cannot merge until it rebases past #1579.
- **B2 — `count_ratchet` test:** FIX-NOW in this PR. The #1634 fold is a FOLLOW-UP alternative, but #1634 depends on this driver, so its test belongs here.
- **S1 — pin `RULES` to the pyproject bucket:** FIX-NOW. Cheapest, highest-value — a membership test or a derivation.
- **S2 — soft-vs-hard ratchet asymmetry:** FOLD-IN the `check_baseline_not_raised` port, or document in ADR-009 that complexity stays soft.
- **S3 — residual duplication + guard blind spot:** FOLD-IN the `main()` clone removal; FOLLOW-UP for widening the duplication scan set and the pre-existing pylint exit-2 hole (file an issue).
- **N1 — `out`/`err` import-bound defaults:** FIX-NOW. Small; also unblocks a non-`capfd` test.
- **N2 — shebang/mode mismatch:** FOLD-IN (`chmod +x` or drop the shebang).
- **N3 — pin ruff thresholds by a test:** FOLLOW-UP (file an issue).

mkostromin-sigma:
Addressing the 2026-07-15 re-review (posted; fuller reply after CI green):

- **B1:** Rebased onto current `main` after #1579 — ADR/docs/mypy.ini overlap dropped; PR is mechanism-only.
- **B2 / N1:** `tests/unit/test_count_ratchet.py` covers create / update / increase-no-write / decrease-auto-lower / all-equal / mixed; `out`/`err` default to `None` and resolve at call time.
- **S1:** `test_ruff_complexity_rules_match_pyproject_ratchet_bucket` pins `RULES` to the pyproject Ratchet targets header.
- **S2:** Ported `check_baseline_not_raised` for the JSON complexity baseline (skips only while the file is absent on `origin/main` — first land; hard after merge).
- **S3:** Shared `parse_ratchet_args` / `resolve_ratchet_paths` / `json_baseline_io` / `run_counting_tool` remove the residual `main()` clone.
- **N2:** `chmod +x` on `check_ruff_complexity_count.py`.
- **N3:** Filed follow-up #1657.

Will confirm on a CI-green tip and mark threads.

ChrisHuie:
Re-reviewed at `be6ae1b7` — ran the hooks and the new tests at the head and mutation-tested them, not a diff read. Every item from the 2026-07-15 pass is addressed in the code:

- **B1 (rebase):** merge-base is current `main`; the ADR / `docs/decisions/README.md` / `mypy.ini` overlap is gone; PR is mechanism-only and mergeable.
- **B2 (driver untested):** `tests/unit/test_count_ratchet.py` (21 tests) covers create / update / increase-no-write / decrease-auto-lower / all-equal / mixed. It reddens under the mutations it targets — inverting `cur > base` and swallowing the regression `return 1` both fail the increase/mixed cases. Streams are injected (`io.StringIO`), so the assertions capture truthfully.
- **S1 (`RULES` partial copy):** `test_ruff_complexity_rules_match_pyproject_ratchet_bucket` pins `RULES` to the pyproject "Ratchet targets" header; dropping a rule from `RULES` fails it.
- **S2 (soft vs hard):** `check_baseline_not_raised` + `raise_probe_counts` ported to the ruff hook; a probe failure short-circuits before any write (two anti-tamper tests). Soft only until the baseline exists on `main` (first land), hard after merge — as intended.
- **S3 (residual DRY):** `main()` boilerplate is shared via `parse_ratchet_args` / `resolve_ratchet_paths` / `json_baseline_io`; `pylint R0801` on `.pre-commit-hooks/` is clean. The shared `run_counting_tool` is ruff-only; the duplication hook keeps its own pylint bitmask guard, so pylint's exit semantics aren't broken by the unification.
- **N1:** `out`/`err` default to `None`, resolved at call time.
- **N2:** `check_ruff_complexity_count.py` is now `0755`.
- **N3:** deferred to #1657.

Also checked at the head: committed baseline `{C901:186, PLR0912:134, PLR0915:108}` equals current main's counts (and the PR touches no `src/`, so it stays valid); D27 intact (commit-stage hooks unchanged at 12; the new hook is pre-push); CI green on the mergeable result; no unresolved review threads.

One pre-existing item, out of scope for this PR: `check_code_duplication.py:49` (`returncode & 33`) does not catch a pylint bitmask-2 crash — it predates #1613 and folding it into `run_counting_tool` would be a separate change.

Nothing here blocks merge. The only open thread is the #1657 follow-up (pinning the mccabe/pylint thresholds).

mkostromin-sigma:
Thanks for the re-review confirmation — CI remains green (32/32). #1657 tracks the N3 threshold-pin follow-up. Ready to merge whenever you are; #1634 should land after this.

ChrisHuie (approved):

mkostromin-sigma:
## Rebase note

Rebased onto current `upstream/main` locally (includes #1534). Tip is ready to force-push; CI will re-run after push.

No open review threads. Still mergeable once head is synced — #1634 remains stacked behind this.

mkostromin-sigma:
**Post-merge note:** this PR briefly reddened Quality Gate on `main` (`PLR0912` live 135 vs baseline 134). Corrected in #1634 via `_optional_format_parameters` (count back to 134) + C901 auto-lower.

Remaining red on `main` after #1634 is **BDD In-Network ENOSPC** — tracked/fixed in #1662.

PR #1648, opened 2026-07-14
fix(a2a): return TaskNotFoundError (-32001) for an unknown task id

pmezzich:
### Problem

`on_get_task` returns `self.tasks.get(task_id)` — i.e. `None` for an unknown task id:

```python
# src/a2a_server/adcp_a2a_server.py
task_id = params.id
return self.tasks.get(task_id)
```

A `None` handler result is neither a success type nor an `A2AError`, so the SDK yields a generic internal error (`-32603`). The A2A spec defines `TaskNotFoundError` (JSON-RPC `-32001`) for an unknown task id, and the class is already imported and used in the push-notification-config handlers — just not here. An A2A client can react precisely to `-32001` but not to a generic internal error.

### Fix

```python
task = self.tasks.get(task_id)
if task is None:
    raise TaskNotFoundError(message=f"Task not found: {task_id}")
return task
```

### Test

`test_on_get_task_unknown_id_raises_task_not_found` asserts the handler raises `TaskNotFoundError` for an unknown id. Deletion oracle: reverting to the bare `return` fails the test.

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1034,13 +1034,27 @@ class AdCPRequestHandler(RequestHandler):
         yield result
 
     def _get_task_or_raise(self, task_id: str) -> Task:
-        """Return the in-memory task, or raise ``TaskNotFoundError`` (JSON-RPC -32001).
+        """Return the in-memory task, or raise ``TaskNotFoundError``.
 
-        A bare ``None`` return makes the SDK emit a generic internal error; the
-        A2A spec defines ``TaskNotFoundError`` for an unknown task id, which A2A
-        clients can react to precisely. The requested id rides both the message
-        and structured ``data`` so clients can read it programmatically. Shared
-        by ``on_get_task`` and ``on_cancel_task`` so both surface the same code.
+        A bare ``None`` return makes the SDK synthesize a generic internal error;
+        the A2A spec defines ``TaskNotFoundError`` for an unknown task id, so
+        raising it is the correct thing to do here and is what an A2A client
+        should be able to react to precisely.
+
+        What a client sees TODAY is still ``-32603``, not the spec's ``-32001``:
+        this app builds its A2A routes with ``enable_v0_3_compat=True``
+        (``src/app.py:302``), so requests dispatch through
+        ``a2a.compat.v0_3.jsonrpc_adapter``, whose ``handle_request`` ends in a
+        bare ``except Exception -> CoreInternalError`` with no ``A2AError -> code``
+        mapping — the mapping the SDK's own main dispatcher performs. Returning
+        ``None`` produces the same ``-32603`` there, so the code cannot be fixed
+        at this layer (#1670). Raising the right type is still correct and is what
+        will surface ``-32001`` the moment that gap closes; the xfail'd
+        live-server test pins the current reality.
+
+        The requested id rides both the message and structured ``data`` so clients
+        can read it programmatically. Shared by ``on_get_task`` and
+        ``on_cancel_task`` so both surface the same error.
         """
         task = self.tasks.get(task_id)
         if task is None:
@@ -1054,8 +1068,8 @@ class AdCPRequestHandler(RequestHandler):
     ) -> Task:
         """Handle 'tasks/get' method to retrieve task status.
 
-        Raises ``TaskNotFoundError`` (-32001) for an unknown task id — see
-        ``_get_task_or_raise``.
+        Raises ``TaskNotFoundError`` for an unknown task id — see
+        ``_get_task_or_raise`` (and #1670 for why the wire code is still -32603).
         """
         return self._get_task_or_raise(params.id)
 
@@ -1066,13 +1080,15 @@ class AdCPRequestHandler(RequestHandler):
     ) -> Task:
         """Handle 'tasks/cancel' method to cancel a task.
 
-        Raises ``TaskNotFoundError`` (-32001) for an unknown task id — cancelling
-        a task that does not exist is the same not-found condition as get, not a
-        silent no-op.
+        Raises ``TaskNotFoundError`` for an unknown task id — cancelling a task
+        that does not exist is the same not-found condition as get, not a silent
+        no-op. See ``_get_task_or_raise`` (and #1670 for why the wire code is
+        still -32603).
         """
         task = self._get_task_or_raise(params.id)
+        # CopyFrom mutates the stored Task in place — self.tasks already holds
+        # this exact reference, so re-storing it would rebind the same object.
         task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
-        self.tasks[params.id] = task
         return task
 
     async def on_list_tasks(
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1033,51 +1033,46 @@ class AdCPRequestHandler(RequestHandler):
         # result is already Task | Message — yield it directly
         yield result
 
+    def _get_task_or_raise(self, task_id: str) -> Task:
+        """Return the in-memory task, or raise ``TaskNotFoundError`` (JSON-RPC -32001).
+
+        A bare ``None`` return makes the SDK emit a generic internal error; the
+        A2A spec defines ``TaskNotFoundError`` for an unknown task id, which A2A
+        clients can react to precisely. The requested id rides both the message
+        and structured ``data`` so clients can read it programmatically. Shared
+        by ``on_get_task`` and ``on_cancel_task`` so both surface the same code.
+        """
+        task = self.tasks.get(task_id)
+        if task is None:
+            raise TaskNotFoundError(message=f"Task not found: {task_id}", data={"task_id": task_id})
+        return task
+
     async def on_get_task(
         self,
         params: GetTaskRequest,
         context: ServerCallContext,
-    ) -> Task | None:
+    ) -> Task:
         """Handle 'tasks/get' method to retrieve task status.
 
-        Args:
-            params: Parameters specifying the task ID
-            context: Server call context
-
-        Returns:
-            Task object if found.
-
-        Raises:
-            TaskNotFoundError: if no task with ``task_id`` exists. Returning
-                ``None`` here makes the SDK emit a generic internal error; the
-                A2A spec defines ``TaskNotFoundError`` (JSON-RPC -32001) for an
-                unknown task id, which A2A clients can react to precisely.
+        Raises ``TaskNotFoundError`` (-32001) for an unknown task id — see
+        ``_get_task_or_raise``.
         """
-        task_id = params.id
-        task = self.tasks.get(task_id)
-        if task is None:
-            raise TaskNotFoundError(message=f"Task not found: {task_id}")
-        return task
+        return self._get_task_or_raise(params.id)
 
     async def on_cancel_task(
         self,
         params: CancelTaskRequest,
         context: ServerCallContext,
-    ) -> Task | None:
+    ) -> Task:
         """Handle 'tasks/cancel' method to cancel a task.
 
-        Args:
-            params: Parameters specifying the task ID
-            context: Server call context
-
-        Returns:
-            Task object with canceled status, or None if not found
+        Raises ``TaskNotFoundError`` (-32001) for an unknown task id — cancelling
+        a task that does not exist is the same not-found condition as get, not a
+        silent no-op.
         """
-        task_id = params.id
-        task = self.tasks.get(task_id)
-        if task:
-            task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
-            self.tasks[task_id] = task
+        task = self._get_task_or_raise(params.id)
+        task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
+        self.tasks[params.id] = task
         return task
 
     async def on_list_tasks(
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1045,10 +1045,19 @@ class AdCPRequestHandler(RequestHandler):
             context: Server call context
 
         Returns:
-            Task object if found, otherwise None
+            Task object if found.
+
+        Raises:
+            TaskNotFoundError: if no task with ``task_id`` exists. Returning
+                ``None`` here makes the SDK emit a generic internal error; the
+                A2A spec defines ``TaskNotFoundError`` (JSON-RPC -32001) for an
+                unknown task id, which A2A clients can react to precisely.
         """
         task_id = params.id
-        return self.tasks.get(task_id)
+        task = self.tasks.get(task_id)
+        if task is None:
+            raise TaskNotFoundError(message=f"Task not found: {task_id}")
+        return task
 
     async def on_cancel_task(
         self,
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1043,7 +1043,7 @@ class AdCPRequestHandler(RequestHandler):
 
         What a client sees TODAY is still ``-32603``, not the spec's ``-32001``:
         this app builds its A2A routes with ``enable_v0_3_compat=True``
-        (``src/app.py:302``), so requests dispatch through
+        (``src/app.py:306``), so requests dispatch through
         ``a2a.compat.v0_3.jsonrpc_adapter``, whose ``handle_request`` ends in a
         bare ``except Exception -> CoreInternalError`` with no ``A2AError -> code``
         mapping — the mapping the SDK's own main dispatcher performs. Returning
@@ -1052,9 +1052,15 @@ class AdCPRequestHandler(RequestHandler):
         will surface ``-32001`` the moment that gap closes; the xfail'd
         live-server test pins the current reality.
 
-        The requested id rides both the message and structured ``data`` so clients
-        can read it programmatically. Shared by ``on_get_task`` and
-        ``on_cancel_task`` so both surface the same error.
+        The requested id is put on both the message and structured ``data``.
+        Only the message reaches a client today: the same compat adapter that
+        flattens the code to ``-32603`` rebuilds the error as
+        ``CoreInternalError(message=str(e))``, which drops ``data`` — driving
+        the real route returns ``data: null``. Populating it is still correct
+        and becomes readable when #1670 closes, the same as the code.
+
+        Shared by ``on_get_task`` and ``on_cancel_task`` so both surface the
+        same error.
         """
         task = self.tasks.get(task_id)
         if task is None:
--- a/tests/e2e/test_a2a_endpoints_working.py
+++ b/tests/e2e/test_a2a_endpoints_working.py
@@ -11,6 +11,7 @@ This test validates the actual HTTP endpoints that our A2A server exposes.
 import json
 import os
 import sys
+from unittest.mock import MagicMock
 
 import pytest
 import requests
@@ -312,9 +313,11 @@ class TestA2ARequestHandler:
         as get, and both route through the shared ``_get_task_or_raise``.
 
         Fast smoke check on the raise only. It does NOT prove the wire code: the
-        exception carries no code, and on this app's v0.3-compat dispatch path the
-        client actually sees -32603 (see #1670 and the xfail'd live-server test in
-        TestA2AServerIntegration). Assert on str(exc), not exc.code — there is none.
+        exception carries no code, and the client actually sees -32603 — see
+        ``_get_task_or_raise`` (src/a2a_server/adcp_a2a_server.py) and #1670 for
+        why, plus the xfail'd live-server test in TestA2AServerIntegration that
+        grades the code on the wire. Assert on str(exc), not exc.code — there is
+        none.
 
         Parametrized over both entry points so the shared assertion cannot drift
         between two byte-identical copies.
@@ -324,8 +327,6 @@ class TestA2ARequestHandler:
         alone would let ``data={"task_id": ...}`` be deleted with the suite still
         green, since the id appears in the message either way.
         """
-        from unittest.mock import MagicMock
-
         with pytest.raises(TaskNotFoundError) as exc:
             await getattr(self.handler, method_name)(request_cls(id="task_does_not_exist"), MagicMock())
         assert "task_does_not_exist" in str(exc.value)  # the requested id is surfaced
@@ -369,26 +370,29 @@ class TestA2AServerIntegration:
         strict=True,
     )
     @pytest.mark.parametrize("method", ["tasks/get", "tasks/cancel"])
-    def test_unknown_task_id_returns_task_not_found_code_on_the_wire(self, method):
+    def test_unknown_task_id_returns_task_not_found_code_on_the_wire(self, method, live_server):
         """The deliverable of the TaskNotFoundError change is what an A2A client
         SEES: JSON-RPC error code -32001. That code is not carried by the
         exception — it is synthesized downstream — so the direct-call test in
         TestA2ARequestHandler cannot prove it. This POSTs the real request to the
         running /a2a endpoint and grades the code on the wire.
 
+        Uses the ``live_server`` fixture so the transport is guaranteed up: the
+        base URL comes from ``live_server['a2a']`` and the assertion runs
+        deterministically instead of skipping when nothing happens to be listening
+        on the ad-hoc port — a skip under strict xfail is neither XFAIL nor XPASS,
+        so the sole on-the-wire grade must never be allowed to no-op.
+
         Parametrized over both methods this PR changed. `tasks/cancel` of an
         unknown id went from a silent None to an error in this PR, so it needs the
         same wire tripwire as `tasks/get` — otherwise only half the contract gets
         locked in when #1670 lands.
 
-        STRICT xfail against #1670: this app builds its A2A routes with
-        enable_v0_3_compat=True (src/app.py:306), so requests dispatch through
-        a2a.compat.v0_3.jsonrpc_adapter, whose handle_request ends in a bare
-        `except Exception -> CoreInternalError` with no A2AError->code mapping —
-        the mapping the SDK's own main dispatcher performs. Both `tasks/get` and
-        `tasks/cancel` reach it through the same _process_non_streaming_request,
-        so both are -32603 today, and both raising TaskNotFoundError and returning
-        None yield that code — this is not fixable at the handler.
+        STRICT xfail against #1670: the code is -32603 today, not the spec's
+        -32001 — see ``_get_task_or_raise`` (src/a2a_server/adcp_a2a_server.py) and
+        #1670 for the enable_v0_3_compat dispatch path that flattens it. Both
+        `tasks/get` and `tasks/cancel` reach that path, so both are -32603 today
+        whether they raise TaskNotFoundError or return None.
 
         Strict on purpose: an a2a-sdk bump that closes the gap makes this XPASS,
         which strict turns into a loud failure so the xfail is removed and -32001
@@ -396,21 +400,18 @@ class TestA2AServerIntegration:
         the marker — the same "green suite lies" shape the tripwire exists to
         prevent.
         """
-        try:
-            response = requests.post(
-                f"{_a2a_base_url()}/a2a",
-                json={"jsonrpc": "2.0", "id": 1, "method": method, "params": {"id": "task_does_not_exist"}},
-                timeout=5,
-            )
-        except (requests.ConnectionError, requests.Timeout):
-            pytest.skip(f"A2A server not running at {_a2a_base_url()}")
+        response = requests.post(
+            f"{live_server['a2a']}/a2a",
+            json={"jsonrpc": "2.0", "id": 1, "method": method, "params": {"id": "task_does_not_exist"}},
+            timeout=5,
+        )
 
         assert response.status_code == 200, f"JSON-RPC errors ride a 200 envelope: {response.status_code}"
-        body = response.json()
-        assert "error" in body, f"unknown task id must produce a JSON-RPC error: {body}"
-        assert body["error"]["code"] == -32001, (
+        data = response.json()
+        assert "error" in data, f"unknown task id must produce a JSON-RPC error: {data}"
+        assert data["error"]["code"] == -32001, (
             f"A2A spec defines -32001 (TaskNotFoundError) for an unknown task id, got "
-            f"{body['error']['code']} — see #1670"
+            f"{data['error']['code']} — see #1670"
         )
 
     @pytest.mark.integration
--- a/tests/e2e/test_a2a_endpoints_working.py
+++ b/tests/e2e/test_a2a_endpoints_working.py
@@ -301,29 +301,32 @@ class TestA2ARequestHandler:
             assert callable(method), f"Method {method_name} is not callable"
 
     @pytest.mark.asyncio
-    async def test_on_get_task_unknown_id_raises_task_not_found(self):
-        """tasks/get for an unknown task id raises TaskNotFoundError (A2A -32001),
-        not the generic internal error a bare None return produces. (Assert on
-        str(exc) / the wire error code, not exc.code — the exception has none.)"""
+    @pytest.mark.parametrize(
+        "request_cls_name, method_name",
+        [("GetTaskRequest", "on_get_task"), ("CancelTaskRequest", "on_cancel_task")],
+    )
+    async def test_unknown_task_id_raises_task_not_found(self, request_cls_name, method_name):
+        """An unknown task id raises TaskNotFoundError, not the generic internal
+        error a bare None return produces — cancel is the same not-found condition
+        as get, and both route through the shared ``_get_task_or_raise``.
+
+        Fast smoke check on the raise only. It does NOT prove the wire code: the
+        exception carries no code, and on this app's v0.3-compat dispatch path the
+        client actually sees -32603 (see #1670 and the xfail'd live-server test in
+        TestA2AServerIntegration). Assert on str(exc), not exc.code — there is none.
+
+        Parametrized over both entry points so the shared assertion cannot drift
+        between two byte-identical copies.
+        """
         from unittest.mock import MagicMock
 
-        from a2a.types import GetTaskRequest, TaskNotFoundError
+        import a2a.types as a2a_types
+        from a2a.types import TaskNotFoundError
 
+        request_cls = getattr(a2a_types, request_cls_name)
         with pytest.raises(TaskNotFoundError) as exc:
-            await self.handler.on_get_task(GetTaskRequest(id="task_does_not_exist"), MagicMock())
+            await getattr(self.handler, method_name)(request_cls(id="task_does_not_exist"), MagicMock())
         assert "task_does_not_exist" in str(exc.value)  # the requested id is surfaced
-
-    @pytest.mark.asyncio
-    async def test_on_cancel_task_unknown_id_raises_task_not_found(self):
-        """tasks/cancel for an unknown task id is the same not-found condition as
-        get — it must raise TaskNotFoundError, not silently no-op to None."""
-        from unittest.mock import MagicMock
-
-        from a2a.types import CancelTaskRequest, TaskNotFoundError
-
-        with pytest.raises(TaskNotFoundError) as exc:
-            await self.handler.on_cancel_task(CancelTaskRequest(id="task_does_not_exist"), MagicMock())
-        assert "task_does_not_exist" in str(exc.value)
 
     def test_handler_has_skill_methods(self):
         """Test that handler has skill-specific methods."""
@@ -356,6 +359,49 @@ class TestA2ARequestHandler:
 
 class TestA2AServerIntegration:
     """Integration tests for complete A2A server setup."""
+
+    @pytest.mark.integration
+    @pytest.mark.xfail(
+        reason="v0.3 compat adapter maps A2AError to -32603; see #1670",
+        strict=True,
+    )
+    def test_tasks_get_unknown_id_returns_task_not_found_code_on_the_wire(self):
+        """The deliverable of the TaskNotFoundError change is what an A2A client
+        SEES: JSON-RPC error code -32001. That code is not carried by the
+        exception — it is synthesized downstream — so the direct-call test in
+        TestA2ARequestHandler cannot prove it. This POSTs the real request to the
+        running /a2a endpoint and grades the code on the wire.
+
+        STRICT xfail against #1670: this app builds its A2A routes with
+        enable_v0_3_compat=True (src/app.py:302), so requests dispatch through
+        a2a.compat.v0_3.jsonrpc_adapter, whose handle_request ends in a bare
+        `except Exception -> CoreInternalError` with no A2AError->code mapping —
+        the mapping the SDK's own main dispatcher performs. Both raising
+        TaskNotFoundError and returning None yield -32603 there, so this is not
+        fixable at the handler.
+
+        Strict on purpose: an a2a-sdk bump that closes the gap makes this XPASS,
+        which strict turns into a loud failure so the xfail is removed and -32001
+        is locked in. A non-strict xfail would let the fix land silently and rot
+        the marker — the same "green suite lies" shape the tripwire exists to
+        prevent.
+        """
+        try:
+            response = requests.post(
+                f"{_a2a_base_url()}/a2a",
+                json={"jsonrpc": "2.0", "id": 1, "method": "tasks/get", "params": {"id": "task_does_not_exist"}},
+                timeout=5,
+            )
+        except requests.exceptions.ConnectionError:
+            pytest.skip(f"A2A server not running at {_a2a_base_url()}")
+
+        assert response.status_code == 200, f"JSON-RPC errors ride a 200 envelope: {response.status_code}"
+        body = response.json()
+        assert "error" in body, f"unknown task id must produce a JSON-RPC error: {body}"
+        assert body["error"]["code"] == -32001, (
+            f"A2A spec defines -32001 (TaskNotFoundError) for an unknown task id, got "
+            f"{body['error']['code']} — see #1670"
+        )
 
     @pytest.mark.integration
     def test_server_discovery_flow(self):
--- a/tests/e2e/test_a2a_endpoints_working.py
+++ b/tests/e2e/test_a2a_endpoints_working.py
@@ -303,13 +303,27 @@ class TestA2ARequestHandler:
     @pytest.mark.asyncio
     async def test_on_get_task_unknown_id_raises_task_not_found(self):
         """tasks/get for an unknown task id raises TaskNotFoundError (A2A -32001),
-        not the generic internal error a bare None return produces."""
+        not the generic internal error a bare None return produces. (Assert on
+        str(exc) / the wire error code, not exc.code — the exception has none.)"""
         from unittest.mock import MagicMock
 
         from a2a.types import GetTaskRequest, TaskNotFoundError
 
-        with pytest.raises(TaskNotFoundError):
+        with pytest.raises(TaskNotFoundError) as exc:
             await self.handler.on_get_task(GetTaskRequest(id="task_does_not_exist"), MagicMock())
+        assert "task_does_not_exist" in str(exc.value)  # the requested id is surfaced
+
+    @pytest.mark.asyncio
+    async def test_on_cancel_task_unknown_id_raises_task_not_found(self):
+        """tasks/cancel for an unknown task id is the same not-found condition as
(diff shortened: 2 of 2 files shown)


src/a2a_server/adcp_a2a_server.py
```diff
@@ -1041,10 +1041,19 @@ async def on_get_task(
...
+            TaskNotFoundError: if no task with ``task_id`` exists. Returning
+                ``None`` here makes the SDK emit a generic internal error; the
+                A2A spec defines ``TaskNotFoundError`` (JSON-RPC -32001) for an
+                unknown task id, which A2A clients can react to precisely.
         """
         task_id = params.id
-        return self.tasks.get(task_id)
+        task = self.tasks.get(task_id)
+        if task is None:
+            raise TaskNotFoundError(message=f"Task not found: {task_id}")
```
KonstantinMirin: Fix applied here but not to the identical sibling `on_cancel_task` two methods below, which does the same `self.tasks.get(task_id)` lookup and still `return task` (i.e. `None`) on a miss — so `tasks/cancel` on an unknown id keeps emitting the generic internal error this PR just declared wrong for `tasks/get`. Extract a `_require_task(self, task_id) -> Task` helper (lookup + `None` check + this raise) and route both handlers through it, so the unknown-id contract is single-sourced. If cancel-of-unknown is meant to stay an idempotent no-op, note that in the PR body.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +300,17 @@ def test_handler_has_required_methods(self):
...
 
+    @pytest.mark.asyncio
+    async def test_on_get_task_unknown_id_raises_task_not_found(self):
+        """tasks/get for an unknown task id raises TaskNotFoundError (A2A -32001),
+        not the generic internal error a bare None return produces."""
+        from unittest.mock import MagicMock
+
+        from a2a.types import GetTaskRequest, TaskNotFoundError
+
+        with pytest.raises(TaskNotFoundError):
```
KonstantinMirin: This pins that the handler *raises*, but the PR's actual deliverable is the wire code `-32001`, which `TaskNotFoundError` does not carry — it's synthesized downstream by the SDK's `JSON_RPC_ERROR_CODE_MAP` / `build_error_response`. A future a2a-sdk remap or a transport layer swallowing the exception would regress the client-visible code while this stays green. Add a test using the live-server JSON-RPC pattern already in `TestA2AServerIntegration`: POST `tasks/get` for a nonexistent id and assert `response.json()["error"]["code"] == -32001`. Keep this direct call as a fast smoke check. While here, assert the message carries the id (`assert "task_does_not_exist" in str(exc.value)`) — but via `str(exc.value)`/the envelope, not `exc.value.code` (the exception has no `code` attribute).

KonstantinMirin (commented):
## Review — PR #1648

**Overview** — Solid, well-motivated change: `on_get_task` now raises `TaskNotFoundError` instead of returning `None`, the docstring explains why (a bare `None` makes the SDK emit a generic internal error rather than the A2A-defined `-32001`), the message text and keyword-arg raise style match the file's existing conventions, and the layering is correct (protocol error framing at the transport handler, no business logic leaked down). Two things should land before merge: the fix was applied to `tasks/get` but not to its identical sibling `tasks/cancel`, and the new behavior — a `-32001` code on the wire — is pinned only by a direct in-process handler call that never exercises the SDK's exception-to-wire translation.

### Should fix

**Fix-the-cited-site, skip-the-sibling — `on_cancel_task` still returns `None` for an unknown id.**
The docstring you added argues that returning `None` for a missing task makes the SDK emit a generic internal error and that `TaskNotFoundError` (`-32001`) is the correct client-reactable signal. That reasoning applies verbatim to `on_cancel_task`, two methods below, which does the identical lookup and still returns `None`:

```python
# on_get_task (fixed)
task = self.tasks.get(task_id)
if task is None:
    raise TaskNotFoundError(message=f"Task not found: {task_id}")
return task

# on_cancel_task (unchanged — same lookup, still returns None on miss)
task = self.tasks.get(task_id)
if task:
    task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
    self.tasks[task_id] = task
return task
```

A client cancelling an unknown task still gets exactly the generic-internal-error behavior this PR set out to eliminate, and the A2A SDK's own reference handler raises `TaskNotFoundError` for cancel-of-unknown too. Single-source the contract: extract a `_require_task(self, task_id) -> Task` helper (the `self.tasks.get` + `None` check + `raise TaskNotFoundError(message=f"Task not found: {task_id}")`) and route both `on_get_task` and `on_cancel_task` through it. If cancel-of-unknown is meant to stay an idempotent no-op, state that in the PR body so it's a documented decision rather than silent drift.

**Wire contract `-32001` is not tested — the added test stops one layer short.**
The deliverable is what A2A clients see on the wire: JSON-RPC error `code: -32001`. That code is not carried by the exception — `TaskNotFoundError` is a plain `A2AError` with only `message`/`data`; the `-32001` is synthesized downstream by the framework's `JSON_RPC_ERROR_CODE_MAP` / `build_error_response`. The new test calls `self.handler.on_get_task(GetTaskRequest(id=...), MagicMock())` directly and asserts only `pytest.raises(TaskNotFoundError)`. It proves the method raises; it does not prove the SDK translates that raise into a `-32001` envelope. If a future a2a-sdk bump remapped the code, or a transport layer swallowed the exception into a generic `-32603`, this test stays green while the exact regression the change exists to prevent ships. This same file already has the live-server JSON-RPC pattern (`TestA2AServerIntegration` POSTs to the running `/a2a` endpoint via `requests`); add a test there that POSTs `{"jsonrpc":"2.0","id":1,"method":"tasks/get","params":{"id":"task_does_not_exist"}}` and asserts `response.json()["error"]["code"] == -32001`. Keep the direct-call test as a fast smoke check — it just isn't sufficient proof on its own.

### Nice to have

- **Assert the payload, not just the class.** `pytest.raises(TaskNotFoundError)` passes even if the message dropped the id. Capture it and assert the requested id is present: `with pytest.raises(TaskNotFoundError) as exc: ...` then `assert "task_does_not_exist" in str(exc.value)`. (Note: assert on `str(exc.value)` / the wire `error["code"]`, not `exc.value.code` — the exception has no `code` attribute.)
- **Return annotation `-> Task | None` now over-admits `None`.** The method can only return a `Task` or raise. Narrowing to `-> Task` documents the new contract more honestly; keeping `| None` for parity with the abstract base `RequestHandler.on_get_task` (also `Task | None`) is a defensible reason to leave it.
- **Put the id in structured `data` too.** `TaskNotFoundError(message=..., data={"task_id": task_id})` lets clients read the id programmatically instead of parsing the message — in the spirit of "clients can react to precisely." Optional; the two sibling raises in this file are message-only as well.

- No prior human review rounds on this PR — nothing to re-raise.
- Clean on the guard dimensions: no structural-guard allowlist grew, the duplication baseline did not move, and no debt was deferred behind a FIXME. Layering is correct — the change is protocol error framing at the transport handler with no business logic pushed down.

pmezzich:
Following up on the wire-contract point (`-32001` is not tested) — I stood up the A2A server locally and POSTed the exact `tasks/get` request against the live `/a2a`. It returns **-32603, not -32001**:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Task not found: task_does_not_exist","data":null}}
```

So the handler raising `TaskNotFoundError` is necessary but not sufficient here — the code never becomes `-32001` on this endpoint. Root cause traced:

- The app builds its A2A routes with `enable_v0_3_compat=True` (`src/app.py:298`), so requests are dispatched by `a2a.compat.v0_3.jsonrpc_adapter`, **not** the SDK's main `JsonRpcDispatcher`.
- The main dispatcher maps `A2AError` subclasses to their JSON-RPC codes via `build_error_response` — `build_error_response(TaskNotFoundError(...))` returns `code: -32001` in isolation.
- The v0.3 compat adapter has no equivalent mapping: its `handle_request` is a bare `except Exception -> CoreInternalError(message=str(e))` (-32603). The server log confirms it — `a2a.compat.v0_3.jsonrpc_adapter - ERROR - Unhandled exception in v0.3 JSONRPCAdapter`, traceback ending at `a2a/compat/v0_3/request_handler.py:124  raise TaskNotFoundError`.

It's not a handler-level choice either: **both** raising `TaskNotFoundError` and returning `None` produce -32603 (return-None routes through the compat layer's own `raise TaskNotFoundError` at `request_handler.py:124`, swallowed the same way). The `data: null` in the response — vs `build_error_response`'s `exclude_none` — is a second tell that `build_error_response` isn't on this path.

Net: the handler change is still correct at the code level (explicit raise, shared `_get_task_or_raise`, cancel parity), but the `-32001` wire contract can't land until the `A2AError -> code` mapping exists on the v0.3 path. Two options I can see:

1. Map `A2AError -> JSON-RPC code` at our `/a2a` boundary (we already have a custom `/a2a` POST handler in `src/app.py`), so `tasks/get` returns `-32001` and I add the live-server test asserting it.
2. Treat it as an upstream a2a-sdk gap (v0.3 `jsonrpc_adapter` missing the mapping the core dispatcher has), file it, and land this with the handler fix plus an xfail live test documenting the `-32603` reality until the SDK/app path is fixed.

Which layer do you think this belongs at? Happy to take the app-level route if mapping at our boundary is right.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1029,42 +1029,46 @@ async def on_message_send_stream(
...
+
+        A bare ``None`` return makes the SDK emit a generic internal error; the
+        A2A spec defines ``TaskNotFoundError`` for an unknown task id, which A2A
+        clients can react to precisely. The requested id rides both the message
+        and structured ``data`` so clients can read it programmatically. Shared
+        by ``on_get_task`` and ``on_cancel_task`` so both surface the same code.
+        """
+        task = self.tasks.get(task_id)
+        if task is None:
+            raise TaskNotFoundError(message=f"Task not found: {task_id}", data={"task_id": task_id})
```
KonstantinMirin, line 1067: This raise advertises `-32001`, but on this app's active path it never reaches the wire. `src/app.py:302` sets `enable_v0_3_compat=True`, so dispatch goes through `a2a.compat.v0_3.jsonrpc_adapter`, whose `handle_request` ends in a bare `except Exception -> CoreInternalError(message=str(e))` (`-32603`) with no `A2AError -> code` mapping. Live `tasks/get` for an unknown id returns `-32603`, not `-32001` (raising vs returning `None` are indistinguishable here). Land the `A2AError -> code` mapping at the `/a2a` boundary, or drop the `-32001` claim and xfail-document the `-32603` reality.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +300,31 @@ def test_handler_has_required_methods(self):
...
+    async def test_on_get_task_unknown_id_raises_task_not_found(self):
+        """tasks/get for an unknown task id raises TaskNotFoundError (A2A -32001),
+        not the generic internal error a bare None return produces. (Assert on
+        str(exc) / the wire error code, not exc.code — the exception has none.)"""
+        from unittest.mock import MagicMock
+
+        from a2a.types import GetTaskRequest, TaskNotFoundError
+
+        with pytest.raises(TaskNotFoundError) as exc:
+            await self.handler.on_get_task(GetTaskRequest(id="task_does_not_exist"), MagicMock())
```
KonstantinMirin: This direct in-process call proves the handler *raises* but never exercises the exception→wire translation, so the PR's deliverable — `error.code == -32001` — is unverified (and empirically it's `-32603` on the live `/a2a`). Add a live-server test in `TestA2AServerIntegration` (already POSTs via `requests`) that sends `{"jsonrpc":"2.0","id":1,"method":"tasks/get","params":{"id":"task_does_not_exist"}}` and asserts `response.json()["error"]["code"]` — `== -32001` once the boundary mapping exists, or an xfail pinning `-32603` until then. Keep this as a fast smoke check.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +300,31 @@ def test_handler_has_required_methods(self):
...
+        assert "task_does_not_exist" in str(exc.value)  # the requested id is surfaced
+
+    @pytest.mark.asyncio
+    async def test_on_cancel_task_unknown_id_raises_task_not_found(self):
```
KonstantinMirin: Byte-for-byte identical to `test_on_get_task_unknown_id_raises_task_not_found` above except the request class and method name — both pin the same `_get_task_or_raise` contract. Collapse into one `@pytest.mark.parametrize("request_cls, method_name", [(GetTaskRequest, "on_get_task"), (CancelTaskRequest, "on_cancel_task")])` case that calls `getattr(self.handler, method_name)(...)`, so a future change to the shared assertion can't drift between the two.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1029,42 +1029,46 @@ async def on_message_send_stream(
...
+        silent no-op.
         """
-        task_id = params.id
-        task = self.tasks.get(task_id)
-        if task:
-            task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
-            self.tasks[task_id] = task
+        task = self._get_task_or_raise(params.id)
+        task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_CANCELED))
+        self.tasks[params.id] = task
```
KonstantinMirin: `task` is the same object already held in `self.tasks` and `task.status.CopyFrom(...)` mutated it in place, so this re-store rebinds the identical reference and is a no-op. Drop it, or add a one-line comment if the explicit re-store is deliberate.

KonstantinMirin (commented):
## Review — PR #1648

**Overview** — The handler-level change is right and the prior round's asks landed: `on_get_task` and `on_cancel_task` now share a single `_get_task_or_raise` helper so the not-found contract is single-sourced (sibling parity fixed), the return type is narrowed to `-> Task`, the id rides both the message and structured `data={"task_id": ...}`, and the tests assert the id is surfaced. The one thing that still has not landed is the whole point of the PR: the `-32001` code never reaches the wire, and nothing in the diff would fail if it never did.

### Should fix

**↩ The `-32001` wire contract does not land, and is still pinned only by a direct in-process raise — raised in the 2026-07-15 review, still open.**
The deliverable (and the PR title) is the client-visible JSON-RPC code `-32001`. `TaskNotFoundError` does not carry that code; it is synthesized downstream. On this app's active path that synthesis never happens: `src/app.py:302` builds the A2A routes with `enable_v0_3_compat=True`, so requests dispatch through `a2a.compat.v0_3.jsonrpc_adapter`, whose `handle_request` ends in a bare `except Exception -> CoreInternalError(message=str(e))` (JSON-RPC `-32603`) with no `A2AError -> code` mapping. POSTing `tasks/get` for an unknown id against the live `/a2a` returns:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Task not found: task_does_not_exist","data":null}}
```

Both raising `TaskNotFoundError` and returning `None` produce `-32603` here, so the code hygiene is correct but the advertised contract is unmet. The two added tests call `self.handler.on_get_task(...)` / `on_cancel_task(...)` directly with a `MagicMock()` context and assert only `pytest.raises(TaskNotFoundError)` + the id substring — they prove the method raises, never exercise the exception→wire translation, and stay green while the wire code is wrong. Their docstrings literally assert "raises TaskNotFoundError (A2A -32001)", which is false confidence about a contract that does not hold.

Two ways to close it, pick one:
1. Add the `A2AError -> JSON-RPC code` mapping at the `/a2a` boundary (the custom POST handler in `src/app.py`) so `tasks/get` / `tasks/cancel` on an unknown id actually emit `-32001`, then add a live-server test in `TestA2AServerIntegration` (it already POSTs to the running `/a2a` via `requests`) asserting `response.json()["error"]["code"] == -32001`.
2. If the boundary mapping is deferred as an upstream a2a-sdk gap, land the handler fix with that same live-server test present but `@pytest.mark.xfail(reason="v0.3 compat adapter maps A2AError to -32603; see #<issue>")` pinning the current `-32603` reality, and do not leave a docstring/PR asserting `-32001` as delivered.

Keep the two direct-call tests as fast smoke checks — they just are not sufficient proof of the wire contract on their own.

**Parametrize the two copy-paste unknown-id tests.**
`test_on_get_task_unknown_id_raises_task_not_found` and `test_on_cancel_task_unknown_id_raises_task_not_found` are byte-for-byte identical except the request class and the handler method name; they pin the same `_get_task_or_raise` behavior through two entry points. Any change to the shared assertion (e.g. also checking `exc.value.data["task_id"]`) has to be mirrored by hand in both — the mirror-or-drift shape. Fold them:

```python
@pytest.mark.parametrize("request_cls, method_name", [
    (GetTaskRequest, "on_get_task"),
    (CancelTaskRequest, "on_cancel_task"),
])
async def test_unknown_task_id_raises_task_not_found(self, request_cls, method_name):
    with pytest.raises(TaskNotFoundError) as exc:
        await getattr(self.handler, method_name)(request_cls(id="task_does_not_exist"), MagicMock())
    assert "task_does_not_exist" in str(exc.value)
```

### Nice to have

- **`on_cancel_task` re-store is a no-op.** `task` is the same object already in `self.tasks` and `task.status.CopyFrom(...)` mutates it in place, so `self.tasks[params.id] = task` rebinds the identical reference. Drop it, or add a one-line comment if the explicit re-store is intentional.
- **`data={"task_id": ...}` is on the task path only.** The two config-not-found `TaskNotFoundError` raises (`adcp_a2a_server.py:1125`, `:1319`) stay message-only. Fine as a deliberate one-off; if structured `data` is becoming the house style for not-found errors, align them too. Not required for merge.

- Resolved on this push (not re-raised): the `on_cancel_task` sibling parity — both handlers now route through the shared `_get_task_or_raise`; the `-> Task` narrowing; `data={"task_id": ...}` on the raise; and asserting the id via `str(exc.value)`. All four earlier points are addressed.
- Clean on guards: no structural-guard allowlist grew, the duplication baseline did not move, no FIXME/deferred debt. Layering is correct — protocol error framing stays at the transport handler.
- The `A2AError -> code` mapping that would turn the raise into a wire `-32001` is not touched by this diff; whichever placement you choose, it is the gating item above.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +300,34 @@ def test_handler_has_required_methods(self):
...
+        """
+        from unittest.mock import MagicMock
+
+        import a2a.types as a2a_types
+        from a2a.types import TaskNotFoundError
+
+        request_cls = getattr(a2a_types, request_cls_name)
+        with pytest.raises(TaskNotFoundError) as exc:
+            await getattr(self.handler, method_name)(request_cls(id="task_does_not_exist"), MagicMock())
+        assert "task_does_not_exist" in str(exc.value)  # the requested id is surfaced
```
KonstantinMirin, line 332: This push added `data={"task_id": task_id}` to the raise as a deliverable, but this is the only assertion and it passes on the message alone: the message is `f"Task not found: {task_id}"`, so `"task_does_not_exist" in str(exc.value)` is true independent of the `data` channel. Delete the `data=` kwarg from `_get_task_or_raise` and the suite stays green. Pin the structured payload here (the parametrization makes this cover both entry points at once): `assert exc.value.data == {"task_id": "task_does_not_exist"}`.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -332,6 +360,49 @@ def test_auth_methods_exist(self):
...
+        the marker — the same "green suite lies" shape the tripwire exists to
+        prevent.
+        """
+        try:
+            response = requests.post(
+                f"{_a2a_base_url()}/a2a",
+                json={"jsonrpc": "2.0", "id": 1, "method": "tasks/get", "params": {"id": "task_does_not_exist"}},
+                timeout=5,
+            )
+        except requests.exceptions.ConnectionError:
```
KonstantinMirin: Diverges twice from the file's own 7-site skip guard (`:85, :103, :117, :139, :152, :374, :396` all use `except (requests.ConnectionError, requests.Timeout):`): it drops `requests.Timeout` and uses the long `requests.exceptions.ConnectionError` path. With `timeout=5` set above, a slow/hung `/a2a` raises `requests.Timeout` — every sibling treats that as skip, this one lets it escape. Change to `except (requests.ConnectionError, requests.Timeout): pytest.skip(...)`.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +300,34 @@ def test_handler_has_required_methods(self):
...
+
+        Parametrized over both entry points so the shared assertion cannot drift
+        between two byte-identical copies.
+        """
+        from unittest.mock import MagicMock
+
+        import a2a.types as a2a_types
+        from a2a.types import TaskNotFoundError
+
+        request_cls = getattr(a2a_types, request_cls_name)
```
KonstantinMirin: Stringly-typed parametrize: the request classes are passed as names and resolved with `getattr(a2a_types, request_cls_name)`, so a rename fails at runtime inside the body rather than at import, and static tooling can't follow the reference. Pass the class objects directly in the parametrize tuple (`[(GetTaskRequest, "on_get_task"), (CancelTaskRequest, "on_cancel_task")]`) and drop the `import a2a.types as a2a_types` alias and the `getattr` lookup.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -332,6 +360,49 @@ def test_auth_methods_exist(self):
 class TestA2AServerIntegration:
     """Integration tests for complete A2A server setup."""
 
+    @pytest.mark.integration
+    @pytest.mark.xfail(
+        reason="v0.3 compat adapter maps A2AError to -32603; see #1670",
+        strict=True,
+    )
+    def test_tasks_get_unknown_id_returns_task_not_found_code_on_the_wire(self):
```
KonstantinMirin: `tasks/cancel`-of-unknown changed materially this PR (silent `None` → error) but only `tasks/get` gets a wire test. They share the v0.3-compat dispatch path so the envelope is identical — optionally parametrize this wire test over `("tasks/get", "tasks/cancel")` so both codes are locked in by the same tripwire when #1670 lands.

KonstantinMirin (commented):
## Review — PR #1648

**Overview** — The handler change is right and the two prior rounds' asks all landed: `on_get_task`/`on_cancel_task` now single-source the not-found contract through `_get_task_or_raise`, the return type is narrowed to `-> Task`, the id rides both the message and structured `data={"task_id": ...}`, the two direct-call tests are folded into one parametrized case, the redundant `self.tasks[...] = task` re-store is gone, and the `-32001`-on-the-wire gap is handled exactly as agreed — a strict-xfail live-server test that POSTs the real request and pins today's `-32603`, tracked by #1670. Two things to tighten before merge, both in the new tests: the newly-added `data` payload is never asserted, and the new live-server test's skip guard diverges from the file's own 7-site convention.

### Should fix

**Changed behavior with no test that pins it — the structured `data` channel is unasserted.**
This push adds `data={"task_id": task_id}` to the raise and the docstring sells it as a deliverable ("clients can read it programmatically instead of parsing the message"). But the parametrized test asserts only `"task_does_not_exist" in str(exc.value)`, and the message is `f"Task not found: {task_id}"` — so the id is in the message and that substring passes on the message alone, entirely independent of `data`. Delete the `data=` kwarg from `_get_task_or_raise` and the whole suite stays green; the new behavior isn't covered. Add an assertion on the structured channel — this is exactly where the parametrization pays off, pin it once and both entry points inherit it:

```python
assert exc.value.data == {"task_id": "task_does_not_exist"}
```

**Convention introduced by the file, violated by the new site — the live-server skip guard.**
Every other live-server test in this file (7 of them: `:85, :103, :117, :139, :152, :374, :396`) guards a down server with `except (requests.ConnectionError, requests.Timeout): pytest.skip(...)`. The new `test_tasks_get_unknown_id_returns_task_not_found_code_on_the_wire` diverges twice in one block: it drops `requests.Timeout` and reaches through the long `requests.exceptions.ConnectionError` path. The dropped `Timeout` isn't cosmetic — the test sets `timeout=5`, so a slow/hung `/a2a` raises `requests.Timeout`, which every sibling treats as "server not up → skip" but this one lets escape. Match the file:

```python
except (requests.ConnectionError, requests.Timeout):
    pytest.skip(f"A2A server not running at {_a2a_base_url()}")
```

### Nice to have

- **Parametrize by class object, not string name.** The case passes `"GetTaskRequest"`/`"CancelTaskRequest"` as strings and resolves them with `getattr(a2a_types, request_cls_name)`. A rename surfaces as a runtime `AttributeError` inside the body instead of an import-time failure, and static tooling can't see the reference. Pass the classes directly (`[(GetTaskRequest, "on_get_task"), (CancelTaskRequest, "on_cancel_task")]`) and drop the alias + `getattr`.
- **Cancel-of-unknown is pinned only in-process on the wire side.** `tasks/cancel` changed materially (silent `None` → error), but only `tasks/get` has a wire test. They share the v0.3-compat dispatch path so the envelope is identical — optionally parametrize the wire test over `("tasks/get", "tasks/cancel")` so both codes flip together when #1670 lands.
- **Confirm the integration lane stands up `/a2a`.** The strict-xfail's value is turning a future `-32001` into a loud XPASS; but the body `pytest.skip`s on a down server, and a skip neither xfails nor xpasses. Worth confirming the CI job running `-m integration` here actually has the server up, otherwise the tripwire never arms.

- Resolved on this push (not re-raised): sibling parity via the shared `_get_task_or_raise`; the `-> Task` narrowing; `data={"task_id": ...}` on the raise; the two byte-identical unknown-id tests folded into one parametrized case; the `self.tasks[...] = task` no-op dropped; and the false-confidence docstrings corrected to state plainly that the client sees `-32603` today.
- The `-32001`-on-the-wire contract is deliberately not re-raised: it landed the way we agreed (option 2) — the strict-xfail live-server test pins the current `-32603` and #1670 tracks the v0.3-compat `A2AError -> code` mapping gap. The handler-side hygiene is done now; only the pre-existing boundary gap is deferred, and it's tracked, not hidden.

pmezzich:
Pushed `860517e9`. Branch head is now `2572e7ce` — a merge of `upstream/main` that moves only `pyproject.toml`; neither file in this PR changes across it. Round-3 items, in order:

**`data={"task_id": ...}` was unpinned.** Right — the id is in the message either way, so the payload could be deleted with the suite green. `tests/e2e/test_a2a_endpoints_working.py:332` now asserts `exc.value.data == {"task_id": "task_does_not_exist"}` alongside the existing message check at `:331`. Deletion oracle: removing the `data=` kwarg from the raise at `src/a2a_server/adcp_a2a_server.py:1067` fails both parametrized cases with

```
E   AssertionError: assert None == {'task_id': 'task_does_not_exist'}
E    +  where None = TaskNotFoundError('Task not found: task_does_not_exist').data
```

**Wire test only covered `tasks/get`.** Taken, not optional — `tasks/cancel` is the leg that changed behaviour in this PR. `test_unknown_task_id_returns_task_not_found_code_on_the_wire` is now parametrized over `["tasks/get", "tasks/cancel"]` (`:371`), so #1670 landing unlocks both codes through the same tripwire instead of half the contract.

**Stringly-typed parametrize.** Gone. The classes go in directly — `[(GetTaskRequest, "on_get_task"), (CancelTaskRequest, "on_cancel_task")]` (`:305-308`) — with a module-level `from a2a.types import CancelTaskRequest, GetTaskRequest, TaskNotFoundError` at `:17`; the `import a2a.types as a2a_types` alias and the `getattr(a2a_types, ...)` class lookup are both gone. The `getattr(self.handler, method_name)` dispatch stays, as in your sketch. `a2a-sdk[http-server]>=1.0.1` is a hard dependency (`pyproject.toml:35`), so the top-level import can't break collection.

**Skip guard divergence.** Fixed to `except (requests.ConnectionError, requests.Timeout)`; all eight guards in the file now match (`:86, :104, :118, :140, :153, :405, :444, :466`), and `requests.exceptions.` appears nowhere in the file. Worth naming why this was worse than a normal inconsistency: under `@pytest.mark.xfail(strict=True)` any escaping exception is reported XFAIL, so a hung server's `requests.Timeout` would have been rendered indistinguishable from the genuine `-32603` this test exists to document. Confirmed with a standalone strict-xfail case that raises `requests.Timeout` — pytest reports `1 xfailed`, no error.

Earlier rounds, for completeness: the shared helper is `_get_task_or_raise` (`src/a2a_server/adcp_a2a_server.py:1036`, raise at `:1067`), routed by both `on_get_task` (`:1080`) and `on_cancel_task` (`:1094`); the no-op `self.tasks[task_id] = task` re-store is deleted, replaced by the comment at `:1095-1096` explaining that `CopyFrom` mutates the stored reference in place; the direct-call tests are one parametrized case rather than two byte-identical copies; assertions go through `str(exc.value)` and `exc.value.data`, never `exc.value.code`. This push also corrected the `src/app.py` line reference in both docstrings from 302 to 306, which is where `enable_v0_3_compat=True` actually sits.

**Still deferred: the `-32001` wire code.** Not fixed here. `src/app.py:306` sets `enable_v0_3_compat=True`, so dispatch goes through `a2a.compat.v0_3.jsonrpc_adapter`, whose `handle_request` ends in a bare `except Exception -> CoreInternalError(message=str(e))` with no `A2AError -> code` mapping. Both raising and returning `None` yield `-32603` there, so it is not fixable at the handler. Filed as #1670; the wire test is `xfail(strict=True)` against it so an SDK bump that closes the gap XPASSes loudly instead of letting the marker rot.

One correction I owe you, and it was not caught before push. The `_get_task_or_raise` docstring in `98d949d1` — the commit you reviewed — claimed the id rides both the message and structured `data` "so clients can read it programmatically." Clients can't. Driving the real route returns `data: null`, because the same compat adapter that flattens the code rebuilds the error as `CoreInternalError(message=str(e))` and drops `data` with it. `860517e9` replaces that sentence with the accurate one; the wrong claim was live on the branch for the whole of round 2.

Test status: `tests/e2e/test_a2a_endpoints_working.py` re-run on `2572e7ce` is 11 passed, 9 skipped locally. Both direct-call cases pass. The two new wire cases are in the skipped set along with 7 pre-existing live-server tests — `A2A server not running at http://localhost:8080`. So the `-32001`/`-32603` assertion is not graded on my machine. It gets graded by CI's `E2E Tests` job, which runs `tests/e2e/` against `docker-compose.e2e.yml` with `ADCP_SALES_PORT=8080`; that job is still pending on the head commit, so I can't yet show you the marker reporting XFAIL rather than SKIP in the pipeline. The only live `-32603` body in this thread is the `tasks/get` one I posted on 07-15.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -332,6 +363,56 @@ def test_auth_methods_exist(self):
...
+        prevent.
+        """
+        try:
+            response = requests.post(
+                f"{_a2a_base_url()}/a2a",
+                json={"jsonrpc": "2.0", "id": 1, "method": method, "params": {"id": "task_does_not_exist"}},
+                timeout=5,
+            )
+        except (requests.ConnectionError, requests.Timeout):
+            pytest.skip(f"A2A server not running at {_a2a_base_url()}")
```
KonstantinMirin: Should fix: this is the ONLY test that grades the PR's headline deliverable on the wire (`body["error"]["code"] == -32001` below) — the direct-call test at :309 asserts on the raised Python exception, not the JSON-RPC envelope. But it's `xfail(strict=True)` and `pytest.skip`s here when `/a2a` is down, and it reaches the server via the ad-hoc `_a2a_base_url()` (`ADCP_SALES_PORT`, default 8080) — it never requests the `live_server`/`docker_services_e2e` fixture (`tests/e2e/conftest.py:501`/`:143`) that actually stands the stack up. A skip under strict-xfail is neither XFAIL nor XPASS: in any lane where the server isn't already up on that port, the sole wire grade silently no-ops — today's -32603 is never observed and the tripwire never fires the loud XPASS when #1670 closes. Converge on `live_server` (use `live_server["a2a"]` as the base URL) so the transport is guaranteed up and this assertion executes deterministically, instead of skipping. "No A2A in this lane" is a reason to stand it up, not to skip — the round-3 "confirm from a green pipeline log" disposition is the no-infra deferral the standard forbids.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +301,36 @@ def test_handler_has_required_methods(self):
...
+
+        Parametrized over both entry points so the shared assertion cannot drift
+        between two byte-identical copies.
+
+        Both halves of the raise are pinned: the human-readable message AND the
+        structured ``data`` payload clients actually parse. Asserting the message
+        alone would let ``data={"task_id": ...}`` be deleted with the suite still
+        green, since the id appears in the message either way.
+        """
+        from unittest.mock import MagicMock
```
KonstantinMirin: Nice to have: the same push hoisted `from a2a.types import ...` to module top (:17) but left this `from unittest.mock import MagicMock` function-local, so one test splits its imports across two scopes. Hoist it to the module import block alongside `a2a.types`.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -332,6 +363,56 @@ def test_auth_methods_exist(self):
...
+            response = requests.post(
+                f"{_a2a_base_url()}/a2a",
+                json={"jsonrpc": "2.0", "id": 1, "method": method, "params": {"id": "task_does_not_exist"}},
+                timeout=5,
+            )
+        except (requests.ConnectionError, requests.Timeout):
+            pytest.skip(f"A2A server not running at {_a2a_base_url()}")
+
+        assert response.status_code == 200, f"JSON-RPC errors ride a 200 envelope: {response.status_code}"
+        body = response.json()
```
KonstantinMirin: Nice to have: the file's other live-server tests bind `data = response.json()`; this one uses `body`. Rename to `data` to match the module's convention (purely cosmetic).

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -300,6 +301,36 @@ def test_handler_has_required_methods(self):
             method = getattr(self.handler, method_name)
             assert callable(method), f"Method {method_name} is not callable"
 
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "request_cls, method_name",
+        [(GetTaskRequest, "on_get_task"), (CancelTaskRequest, "on_cancel_task")],
+    )
+    async def test_unknown_task_id_raises_task_not_found(self, request_cls, method_name):
+        """An unknown task id raises TaskNotFoundError, not the generic internal
```
KonstantinMirin, line 311: Nice to have (DRY prose): this docstring re-derives the full `enable_v0_3_compat` / `-32603` / #1670 mechanism already spelled out in `_get_task_or_raise` (`src/a2a_server/adcp_a2a_server.py:1036`), and re-cites `src/app.py:306` a second time. When #1670 closes all three copies must change together. Point to `_get_task_or_raise` + #1670 instead of restating the mechanism.

tests/e2e/test_a2a_endpoints_working.py
```diff
@@ -332,6 +363,56 @@ def test_auth_methods_exist(self):
...
     """Integration tests for complete A2A server setup."""
 
+    @pytest.mark.integration
+    @pytest.mark.xfail(
+        reason="v0.3 compat adapter maps A2AError to -32603; see #1670",
+        strict=True,
+    )
+    @pytest.mark.parametrize("method", ["tasks/get", "tasks/cancel"])
+    def test_unknown_task_id_returns_task_not_found_code_on_the_wire(self, method):
+        """The deliverable of the TaskNotFoundError change is what an A2A client
```
KonstantinMirin, line 374: Nice to have (DRY prose): third copy of the `-32603` / v0.3-compat / #1670 rationale, with `src/app.py:306` hard-coded a third time. Same parallel-update burden as the sibling docstrings — point back to `_get_task_or_raise` rather than re-deriving it.

KonstantinMirin (commented):
## Review — PR #1648

**Overview** — The handler change is right: `on_get_task`/`on_cancel_task` single-source the not-found contract through `_get_task_or_raise` (`src/a2a_server/adcp_a2a_server.py:1036`), the return type narrows to `-> Task`, the id rides both the message and structured `data={"task_id": ...}`, the two direct-call tests collapse into one parametrized case, and the redundant `self.tasks[...] = task` re-store is gone. Every round-3 hygiene ask has landed. But the maintainer's premise for this round — *does the new behavior have a wire/e2e test that actually executes across the transports it can run on?* — is not met, and correcting an earlier misclassification, that is a **Should fix**, not the Nice-to-have it was filed as.

**Transport-coverage truth.** The new behavior is: an unknown A2A task id on `tasks/get` / `tasks/cancel` raises `TaskNotFoundError` (spec `-32001`; still `-32603` today behind the v0.3-compat adapter, tracked by #1670). `tasks/get` and `tasks/cancel` are A2A JSON-RPC **lifecycle** methods on the a2a-sdk `RequestHandler`, operating on the server's in-memory `self.tasks` dict, and `-32001` is an A2A wire code. **MCP and REST have no such lifecycle method** — the AdCP `get_task`/`list_tasks` *skill* they expose (`src/core/tools/task_management.py`, registered at `src/core/main.py:364-365`) is a different subsystem that reads the DB workflow layer via `uow.workflows.get_by_step_id_or_raise` and already has its own not-found path; this PR does not touch it. So the absence of MCP and REST grades here is a **valid protocol reason** — the operation genuinely does not exist on those transports, and this behavior cannot ride the harness's transport-agnostic BDD parametrization (which dispatches AdCP skills, not raw JSON-RPC lifecycle methods). That leaves **A2A** as the only transport where this is gradable, on the wire and e2e.

### Should fix

**The headline `-32001`-on-the-wire deliverable has no grade that actually executes.** Two tests cover this change, and neither delivers an executing wire grade:

- `test_unknown_task_id_raises_task_not_found` (direct-call, `:309`) executes and passes, but asserts on the raised Python exception (`str(exc.value)`, `exc.value.data`) — **not** the JSON-RPC envelope. It cannot see `-32001`; the code is synthesized downstream of the raise.
- `test_unknown_task_id_returns_task_not_found_code_on_the_wire` (`:372`) is the *only* test that asserts on the wire (`body["error"]["code"] == -32001`, `:411`). It is `@pytest.mark.xfail(strict=True)` but `pytest.skip`s on `ConnectionError`/`Timeout` (`:406`), and it reaches `/a2a` through the ad-hoc `_a2a_base_url()` (reads `ADCP_SALES_PORT`, default `8080`). It **never requests `live_server` / `docker_services_e2e`** (`tests/e2e/conftest.py:501` / `:143`) — the session fixture that actually stands the stack up and health-checks it. No test in this whole file does.

A skip under `xfail(strict=True)` is neither XFAIL nor XPASS: it short-circuits before the strict machinery evaluates. So in any lane where the server is not already listening on that exact port, the sole wire grade **silently no-ops** — today's `-32603` is never observed, and the tripwire never fires the loud XPASS when #1670 closes. The round-3 disposition ("confirm from a green pipeline log; no code change needed if the lane stands the server up") is exactly the no-infra deferral the standard forbids: the tripwire's arming must be *guaranteed by the test*, not confirmed by reading a log after the fact.

Fix: converge the wire test onto the e2e harness so the transport is stood up and the assertion executes deterministically — request the `live_server` fixture (`tests/e2e/conftest.py:501`), which resolves the A2A base URL from the stood-up stack (`live_server["a2a"]`) and guarantees the server is up (or errors), instead of the ad-hoc `_a2a_base_url()` + `pytest.skip`-on-`ConnectionError`. With the stack guaranteed up, the strict-xfail then genuinely reports `xfailed` (arming the tripwire) today and `XPASS` (loud failure) the moment #1670 lands, in every lane — no ambient-server assumption, no silent skip. "The lane has no A2A up" is not a reason to skip; it is a reason to make the test bring the lane up.

### Nice to have

- **`#1670` / v0.3-compat rationale re-derived across three docstrings.** The `enable_v0_3_compat=True` (`src/app.py:306`) → bare `except Exception -> CoreInternalError` → `-32603` → #1670 narrative is spelled out in `_get_task_or_raise` (`src/a2a_server/adcp_a2a_server.py:1036`) and re-derived in both test docstrings (`:310`, `:373`), with `src/app.py:306` hard-coded three times. Prose, not logic — but it carries a real parallel-update burden when #1670 closes. Keep the mechanism in `_get_task_or_raise` and have the two test docstrings point to it plus #1670.

- **`body = response.json()` (`:409`) vs `data = response.json()` in the file's other live-server tests.** Cosmetic naming drift; rename to `data` to match.

- **`from unittest.mock import MagicMock` left function-local (`:327`) while `a2a.types` was hoisted to module top (`:17`).** The same push that hoisted `a2a.types` left its sibling `MagicMock` import inside the test body. Hoist it alongside.

- Resolved on this push (not re-raised): the structured `data={"task_id": ...}` payload is now pinned by `assert exc.value.data == {"task_id": "task_does_not_exist"}`; both handlers route through `_get_task_or_raise`; the two byte-identical direct-call tests are one parametrized case; the wire test is parametrized over both `tasks/get` and `tasks/cancel`; the parametrize passes class objects directly (module-level `from a2a.types import ...` at `:17`), not `getattr`; and the skip guard is `except (requests.ConnectionError, requests.Timeout)`.
- The `-32001` wire *code* staying `-32603` today is correctly deferred to #1670 (the v0.3-compat adapter's missing `A2AError -> code` mapping — not fixable at this handler). That deferral is fine; what is not fine is that the test meant to lock it in doesn't reliably run.
- Clean on layering (`_get_task_or_raise` calls only the in-memory `self.tasks.get` and the `TaskNotFoundError` constructor — transport-layer error framing, nothing pushed into business logic), Python practices (typed raise, no bare except, sync helper correctly called without `await`), AdCP grounding (A2A JSON-RPC lifecycle, not an AdCP tool contract; the spec pin and `adcp` version are untouched), and ratchets (no structural-guard allowlist grew, `.duplication-baseline` did not move; the added `xfail(strict=True)` is a self-removing tripwire, the opposite of allowlist growth).

pmezzich:
## Round 3 @ `a07cfdde`

All five items in.

**SHOULD-FIX (:406)** — the wire test now takes the `live_server` fixture and reads its base URL from `live_server['a2a']`, so the transport is guaranteed up and the `-32001` assertion executes deterministically instead of skipping under strict-xfail. Runs XFAIL today against the live -32603, and would flip to the loud XPASS(strict) when #1670 closes.

**NITs** — `from unittest.mock import MagicMock` hoisted to the module top beside the `a2a.types` import; the wire-test local renamed `body` → `data` to match the file's other live-server tests; the two docstrings that re-derived the -32603 / v0.3-compat / #1670 mechanism now point to `_get_task_or_raise` + #1670 instead of restating it (and the `src/app.py:306` re-cites are gone).

The wire assertion itself needs the docker-compose.e2e stack (CI's E2E job) to execute green; verified here by inspection + collect (both tasks/get and tasks/cancel variants present, direct-call tests pass).

KonstantinMirin:
## Review — PR #1648

**Overview** — Right and complete. `on_get_task`/`on_cancel_task` single-source the unknown-task-id contract through `_get_task_or_raise`, the return type narrows from `Task | None` to `-> Task`, the id rides both the message and structured `data={"task_id": ...}`, and every Should-fix item from the four prior rounds has landed. Recommend merge.

### Should fix

None.

- **`-32001` on the wire is deferred to #1670, and that deferral is honest.** The client still sees `-32603` today because the app builds its A2A routes with `enable_v0_3_compat=True`, and the v0.3 adapter's `handle_request` ends in a bare `except Exception -> CoreInternalError(message=str(e))` with no `A2AError -> code` mapping (it also drops `data`). Raising `TaskNotFoundError` and returning `None` both yield `-32603` there, so the code cannot be fixed at this handler layer. The strict-xfail live-server test flips to a loud XPASS the moment #1670 closes the gap.
- **A2A-only coverage is the raw-JSON-RPC-mechanic exception.** `tasks/get`/`tasks/cancel` act on the in-memory `self.tasks` dict; MCP and REST expose no equivalent, so the behavior cannot ride the transport-agnostic harness. The reason is stated at the scenario.

KonstantinMirin:
## Review — PR #1648

**Overview** — Right and complete. `on_get_task`/`on_cancel_task` single-source the unknown-task-id contract through `_get_task_or_raise`, the return type narrows from `Task | None` to `-> Task`, the id rides both the message and structured `data={"task_id": ...}`, and every Should-fix item from the four prior rounds has landed. Recommend merge.

### Should fix

None.

- **`-32001` on the wire is deferred to #1670, and that deferral is honest.** The client still sees `-32603` today because the app builds its A2A routes with `enable_v0_3_compat=True`, and the v0.3 adapter's `handle_request` ends in a bare `except Exception -> CoreInternalError(message=str(e))` with no `A2AError -> code` mapping (it also drops `data`). Raising `TaskNotFoundError` and returning `None` both yield `-32603` there, so the code cannot be fixed at this handler layer. The strict-xfail live-server test flips to a loud XPASS the moment #1670 closes the gap.
- **A2A-only coverage is the raw-JSON-RPC-mechanic exception.** `tasks/get`/`tasks/cancel` act on the in-memory `self.tasks` dict; MCP and REST expose no equivalent, so the behavior cannot ride the transport-agnostic harness. The reason is stated at the scenario.

ChrisHuie (approved):

Issue #1652, opened 2026-07-15
test: wire dormant validation-error BDD scenarios


jstar0:
## Summary

Two validation-error storyboard scenarios are currently stopped by broad harness
gates before they can exercise the production transport boundaries:

- `T-UC-018-ext-c` (`list_creatives` validation failures) is outside the small
  set of UC-018 scenarios admitted by the `CreativeListEnv` fixture branch.
- `T-UC-002-inv-015-6` (`create_media_buy` invalid asset discriminator) falls
  through the UC-002 non-extension catch-all.

Both gates use non-strict `xfail`, so these scenarios neither fail on a wire
regression nor xpass when the underlying behavior becomes available.

References #1534.

This is separate from #1604: that issue tracks per-operation error-code
classification, while this issue tracks dormant BDD transport coverage. It was
identified while hardening the validation-envelope coverage in #1534.

## Changes

- Route the raw malformed request payloads through the appropriate full harness.
- Execute the scenarios across `a2a`, `mcp`, `rest`, and `e2e_rest` where
  supported.
- Assert the real wire error envelope with the shared envelope assertion helper,
  including the expected code, recovery classification, field, and suggestion.
- Remove the scenario-specific gate markers once the scenarios execute.

## Verification

Completion should be verified by running both named scenarios on each supported
transport and confirming they execute rather than xfail. Their error assertions
must inspect the real wire envelope rather than a reconstructed exception.

```sh
uv run pytest -q tests/bdd
```

mkomorski:
Second blocker behind the UC-018 statuses validation-error rows (`["deleted"]`, `[]`, `["unknown"]`), surfaced in the PR #1621 review:

After PR #1621's wire-accessor work, the UC-018 dispatch now goes through the canonical `dispatch_request`, which stashes `ctx["result"]` / `wire_error_envelope` — so that half of the blocker is cleared. The remaining blocker is that `tests/bdd/steps/domain/uc018_list_creatives.py::_dispatch_structured_filter` builds the `CreativeFilters` client-side and raises a Pydantic error BEFORE dispatch, so the malformed filter never reaches the wire (no wire `VALIDATION_ERROR` to assert on).

Wiring these rows needs a dict-passthrough dispatch that skips the client-side `CreativeFilters` build — send the raw filters dict so production's `coerce_creative_filters` emits the two-layer `VALIDATION_ERROR` on the wire.

KonstantinMirin:
[...]

Issue #1604, opened 2026-07-13
Align request-validation error codes with AdCP storyboard semantics


jstar0:
# Align request-validation error codes with AdCP storyboard semantics

## Problem

Request-schema failures are currently normalized to `VALIDATION_ERROR` at the MCP TypeAdapter boundary. AdCP 3.1.0-beta.3 grades structural request failures as `INVALID_REQUEST`, including `BR-UC-002-create-media-buy.feature` scenario `@T-UC-002-inv-015-6` for a package missing `product_id`.

PR #1534 preserves the current error code while adding a strict xfail tripwire for this mismatch. A separate change is needed because the classification must remain consistent across MCP, A2A, and REST rather than special-casing one test or transport.

## Expected behavior

- Structural failures such as missing, malformed, or unknown request fields emit `INVALID_REQUEST`.
- Value-constraint failures on structurally valid fields emit `VALIDATION_ERROR`.
- MCP, A2A, and REST expose the same classification for equivalent failures.
- Buyer-visible messages remain sanitized and do not include raw Pydantic input values or documentation URLs.

## Acceptance criteria

- Add table-driven coverage for representative structural and value-constraint failures across the relevant transport boundaries.
- Implement a shared classification rule instead of transport-specific branches.
- Remove the strict xfail added in PR #1534 by making its storyboard case pass.
- Preserve the existing two-layer AdCP error envelope and structured `field`/`details` metadata.

numarasSigmaSoftware:
Spec-adjudication note for the classification decision here (and the #1694 regeneration): the pinned 3.1.1 `enums/error-code.json` descriptions place "violates schema constraints" under **INVALID_REQUEST** and reserve VALIDATION_ERROR for failures "beyond schema validation" — so regenerating the UC-003 revision outlines to grade VALIDATION_ERROR would move the storyboard *away* from the enum's own taxonomy, while the current storyboard grade (INVALID_REQUEST) matches it better. Production currently emits VALIDATION_ERROR for schema-invalid `revision` via the shared validation boundary (pinned per transport in `tests/integration/test_update_media_buy_revision_validation_wire.py`). Both codes are `correctable`, so buyer recovery is unaffected either way — but the reconciliation direction should be decided from the enum text (and any error-handling prose), not anchored on current production behavior.

PR #1673, opened 2026-07-17
fix: uniform validation-error suggestions + dormant-scenario guard (remediate #1534)

ChrisHuie:
Remediates the #1534 validation-envelope approach and adds guardrails against recurrence. The prior change handled the wire shape in a way that weakened the tests/repo: presence-only suggestion checks, a self-referential detection assumption, cross-transport suggestion inconsistency, and dormant BDD scenarios.

## Behavior
- **Uniform canonical suggestions.** The validation boundary emits the spec-canonical per-code suggestion (`VALIDATION_ERROR_SUGGESTION`) uniformly across MCP/A2A/REST and removes the field-specific `suggest_validation_fix`. Cross-transport suggestion parity is now uniform; the offending field is still surfaced on `field`.
- **#1417 kept, grounded honestly.** Structural failures → `VALIDATION_ERROR` is a deliberate, storyboard-permitted choice (the conformance storyboards accept both `INVALID_REQUEST` and `VALIDATION_ERROR`, with identical `correctable` recovery). The boundary/test/feature now state this plainly and no longer assert it as the spec-canonical code for a schema failure.
- **#1604 reconciled.** The in-body missing-field grading is reconciled to `VALIDATION_ERROR` (integration test + feature), consistent with the boundary's uniform treatment.

## Guardrails (recurrence prevention)
- **New structural guard** (`test_architecture_bdd_no_dormant_scenario_xfail.py`): forbids scenario-specific imperative `pytest.xfail()`/`pytest.skip()` in the BDD `_harness_env` — the exact shape that introduced the dormant scenarios. Tag-keyed allowlist; meta-tested across inline / set-intersection / equality / skip / nested-yield forms.
- **inv-015-6 migrated** from an imperative dormant xfail to a declarative strict-xfail + real `MediaBuyCreateEnv` wiring + a wire-first `Then`. The real production body runs; the underlying gap (untyped `Creative.assets`) is a genuine production gap tracked in #1652, and the strict-xfail flips to a failure when it lands.
- **Detection assumption-pin**: a real FastMCP `TypeAdapter` title pins the compat-middleware detection so it reddens on an upstream title drift.
- **Actionability oracle**: pins that the canonical per-code suggestions satisfy the BDD actionable-verb check, closing a latent trap where spec-correct output would fail a scenario once it graduates.

## Security (folded in)
- Bumps `mcp` 1.27.2 → 1.28.1 (GHSA-vj7q-gjh5-988w) via a `pyproject` security floor; the `uv.lock` change is surgical (mcp only). Both `fastmcp<3.3.0` and `adcp==6.6.0` allow it. Resolves the dependabot alert on the default branch.

## Verification
Full `./run_all_tests.sh ci` green on mcp 1.28.1: unit 5642 / integration 2227 / bdd 2127 (+7368 xfail) / admin 86 / e2e 91 / ui 5 — 0 failed; security audit clean.

## Follow-ups (tracked, not folded)
- Wire-value suggestion oracle (serialization-layer) — belongs on the in-flight `pin-validation-suggestion-wire-oracle` branch, which edits the same integration test.
- Full named-set resolution in the dormant-scenario guard (currently covers inline-literal + skip/xfail forms; named-set membership is a documented limitation).
- Pre-existing `AUTH_REQUIRED_SUGGESTION` enum divergence.
- Value-constraint (`ge`/`min_length`/`pattern`/enum) code ambiguity → raise upstream in adcp.
- Land the feature-file grading/comment upstream in adcp-req (the file is generated).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dependencies = [
     "pyasn1>=0.6.3",  # GHSA-jr27-m4p2-rc6r
     "mako>=1.3.12",  # GHSA-2h4p-vjrc-8xpq (transitive via alembic)
     "click>=8.3.3",  # PYSEC-2026-2132 (transitive via flask/uvicorn)
+    "mcp>=1.28.1",  # GHSA-vj7q-gjh5-988w (transitive via fastmcp/adcp; both allow <2.0)
 ]
 
 
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -183,6 +183,11 @@ _XFAIL_TAGS: dict[str, str] = {
     "T-UC-002-inv-087-5": "duplicate optimization_goals priority: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
     "T-UC-002-inv-087-6": "empty optimization_goals array: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
     "T-UC-002-inv-087-7": "per_ad_spend without value_field: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
+    # FIXME(#1652): BR-RULE-015 INV-6 — Creative.assets is dict[str, Any] (untyped), so
+    # production does not validate the asset_type discriminator; an inline asset lacking
+    # asset_type is accepted (no error). Scenario is wired through the real harness and
+    # strict-xfailed until asset validation lands, at which point it flips to VALIDATION_ERROR.
+    "T-UC-002-inv-015-6": "asset_type discriminator not validated (Creative.assets untyped) — production gap",
     # FIXME(beads-dul): disclosure_positions filter not implemented in production
     # Note: violated/nofield pass vacuously (field rejected at schema level)
     "T-UC-005-inv-049-8-holds": "disclosure_positions filter not implemented",
@@ -286,10 +291,11 @@ _XFAIL_TAGS: dict[str, str] = {
     "T-UC-002-ext-j": "adapter failure raises exception, no failed result envelope or suggestion — spec-production gap",
     "T-UC-002-inv-026-2": "INVALID_CREATIVES error lacks suggestion field",
     "T-UC-002-inv-026-4": "INVALID_CREATIVES error lacks suggestion field",
-    # Graduated (#1417/gh8p.10): the request-construction boundary now derives a
-    # field-aware suggestion (suggest_validation_fix) and attaches it to the
-    # AdCPValidationError, so a missing idempotency_key rejects with a non-empty
-    # wire suggestion. T-UC-002-v31-idempotency-missing passes.
+    # Graduated (#1417/gh8p.10): the request-construction boundary attaches the
+    # canonical VALIDATION_ERROR suggestion ("review error details and fix field
+    # values") to the AdCPValidationError (the offending field is surfaced
+    # separately on the ``field`` path), so a missing idempotency_key rejects with
+    # a non-empty wire suggestion. T-UC-002-v31-idempotency-missing passes.
     # FIXME(salesagent-9vgz.17): optimization_goals not in adcp v3.6.0 or production schemas
     # PackageRequest(extra='forbid') rejects the field with generic validation error,
     # not spec-expected UNSUPPORTED_FEATURE / INVALID_REQUEST with structured codes.
@@ -3225,6 +3231,7 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
             any(t.startswith("T-UC-002-ext-") for t in marker_names)
             or "nfr-highvalue" in marker_names
             or "T-UC-002-nfr-001-enforcement" in marker_names
+            or "T-UC-002-inv-015-6" in marker_names
         ):
             # Extension/error scenarios: budget validation, pricing errors, etc.
             # Plus the nfr-highvalue >$10k Seller-alert scenario (#1417),
@@ -3270,8 +3277,6 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
                 ctx["default_product"] = product
                 ctx["default_pricing_option"] = pricing_option
                 yield
-        elif "T-UC-002-inv-015-6" in marker_names:
-            pytest.xfail("T-UC-002-inv-015-6 create_media_buy harness wiring is tracked in #1652")
         else:
             # Restore the xfail guard every other use case keeps on its catch-all:
             # non-account / non-extension UC-002 scenarios are NOT yet wired (no
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -180,14 +180,26 @@ _XFAIL_TAGS: dict[str, str] = {
     # FIXME(salesagent-12nd): UC-002 ASAP — response doesn't expose resolved start_time
     "T-UC-002-alt-asap": "response lacks resolved start_time field — spec-production gap",
     # FIXME(salesagent-fie): UC-002 error code mismatch — Pydantic VALIDATION_ERROR vs spec INVALID_REQUEST
-    "T-UC-002-inv-087-5": "duplicate optimization_goals priority: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
-    "T-UC-002-inv-087-6": "empty optimization_goals array: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
-    "T-UC-002-inv-087-7": "per_ad_spend without value_field: VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap",
-    # FIXME(#1652): BR-RULE-015 INV-6 — Creative.assets is dict[str, Any] (untyped), so
-    # production does not validate the asset_type discriminator; an inline asset lacking
-    # asset_type is accepted (no error). Scenario is wired through the real harness and
-    # strict-xfailed until asset validation lands, at which point it flips to VALIDATION_ERROR.
-    "T-UC-002-inv-015-6": "asset_type discriminator not validated (Creative.assets untyped) — production gap",
+    # Deliberate uniform divergence (NOT a gap): the boundary emits VALIDATION_ERROR for these
+    # schema-validation failures while the storyboard grades INVALID_REQUEST — see the CODE-choice
+    # note in adcp_validation_boundary. Scenario asserts the storyboard code, so it stays xfailed.
+    "T-UC-002-inv-087-5": "duplicate optimization_goals priority: boundary emits VALIDATION_ERROR, storyboard grades INVALID_REQUEST — deliberate divergence",
+    "T-UC-002-inv-087-6": "empty optimization_goals array: boundary emits VALIDATION_ERROR, storyboard grades INVALID_REQUEST — deliberate divergence",
+    "T-UC-002-inv-087-7": "per_ad_spend without value_field: boundary emits VALIDATION_ERROR, storyboard grades INVALID_REQUEST — deliberate divergence",
+    # FIXME(#1652): BR-RULE-015 INV-6 — production DOES validate asset_type (the CreativeAsset
+    # discriminated union raises union_tag_not_found), but the WIRE outcome differs per transport
+    # and none matches the scenario's graded "VALIDATION_ERROR referencing asset_type" shape:
+    #   a2a / rest: creative processing rejects with CREATIVE_REJECTED (not VALIDATION_ERROR) — fails
+    #               the earlier validation-error Then.
+    #   mcp: an earlier crash in the idempotency canonical hash (media_buy_create.py
+    #        canonical_payload_hash(raw_wire_payload) -> rfc8785 chokes on a live FormatId ->
+    #        CanonicalizationError) is normalized to VALIDATION_ERROR "unsupported type: FormatId";
+    #        passes Thens 1-2 off that crash envelope, fails Then#3 (message is FormatId, not asset_type).
+    #        (a2a/rest use canonical_request_hash(req), which model_dumps the FormatId, so they don't hit
+    #        it.) That FormatId crash is a separate production bug (tracked in #1679).
+    # Strict-xfailed until the code reconciles (fix the mcp crash AND surface the asset_type rejection
+    # as VALIDATION_ERROR, or reconcile the scenario to CREATIVE_REJECTED); flips XPASS->FAILED then.
+    "T-UC-002-inv-015-6": "asset_type IS validated but wire code differs per transport (a2a/rest CREATIVE_REJECTED; mcp FormatId CanonicalizationError) — not yet the graded VALIDATION_ERROR",
     # FIXME(beads-dul): disclosure_positions filter not implemented in production
     # Note: violated/nofield pass vacuously (field rejected at schema level)
     "T-UC-005-inv-049-8-holds": "disclosure_positions filter not implemented",
@@ -3417,6 +3429,8 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
             with _db_scope_for(request, e2e_config), CreativeListEnv(e2e_config=e2e_config) as env:
                 ctx["env"] = env
                 yield
+        # FIXME(#1652): dormant list_creatives validation scenario, pending real-harness wiring
+        # (allowlisted in test_architecture_bdd_no_dormant_scenario_xfail._KNOWN_DORMANT).
         elif "T-UC-018-ext-c" in marker_names:
             pytest.xfail("T-UC-018-ext-c list_creatives validation harness wiring is tracked in #1652")
         else:
--- a/tests/bdd/features/BR-UC-002-create-media-buy.feature
+++ b/tests/bdd/features/BR-UC-002-create-media-buy.feature
@@ -2578,11 +2578,13 @@ Feature: BR-UC-002 Create Media Buy
     When the Buyer Agent sends the create_media_buy request
     Then the result should be <outcome>
     # @source repo=adcp ref=v3.1-04f59d2d5 commit=04f59d2d5 path=static/schemas/source/media-buy/create-media-buy-request.json
+    # BR-RULE-015 asset-discriminator: prose maps this schema failure to INVALID_REQUEST; the boundary
+    # deliberately emits VALIDATION_ERROR uniformly (matches @T-UC-002-inv-015-6; see adcp_validation_boundary).
 
     Examples: Boundary values
-      | boundary_point                                          | outcome                               |
-      | asset slot value with missing asset_type discriminator  | error INVALID_REQUEST with suggestion |
-      | asset slot value with unknown asset_type value          | error INVALID_REQUEST with suggestion |
+      | boundary_point                                          | outcome                                |
+      | asset slot value with missing asset_type discriminator  | error VALIDATION_ERROR with suggestion |
+      | asset slot value with unknown asset_type value          | error VALIDATION_ERROR with suggestion |
 
   @T-UC-002-boundary-targeting-collection-list @boundary @targeting-overlay @collection-list @br-rule-014
   Scenario Outline: v3.1 targeting_overlay collection-list boundary validation - <boundary_point>
--- a/tests/bdd/features/BR-UC-002-create-media-buy.feature
+++ b/tests/bdd/features/BR-UC-002-create-media-buy.feature
@@ -289,10 +289,10 @@ Feature: BR-UC-002 Create Media Buy
     And the account "acc-001" exists and is active
     When the Buyer Agent sends the create_media_buy request
     Then the response should indicate a validation error
-    And the error code should be "INVALID_REQUEST"
+    And the error code should be "VALIDATION_ERROR"
     And the error should reference the unresolvable asset_type discriminator
     And the error should include "suggestion" field
-    # BR-RULE-015 INV-6: assets value lacking asset_type or carrying an unregistered value (not one of the 14 AssetVariant types) -> INVALID_REQUEST
+    # BR-RULE-015 INV-6: assets value lacking asset_type or carrying an unregistered value (not one of the 14 AssetVariant types). Prose maps this schema failure to INVALID_REQUEST; the boundary deliberately emits VALIDATION_ERROR uniformly (storyboards accept either -- see adcp_validation_boundary).
     # --- ext-h: Format ID Validation Failure ---
     # @source repo=adcp ref=v3.1-04f59d2d5 commit=04f59d2d5 path=static/schemas/source/media-buy/create-media-buy-request.json
 
--- a/tests/bdd/steps/domain/uc002_create_media_buy.py
+++ b/tests/bdd/steps/domain/uc002_create_media_buy.py
@@ -607,6 +607,44 @@ def _first_package(ctx: dict) -> dict:
     return packages[0]
 
 
+@given('a create_media_buy request with an inline creative whose assets map value lacks an "asset_type" field')
+def given_inline_creative_asset_missing_asset_type(ctx: dict) -> None:
+    """Attach an inline creative whose assets-map value omits the asset_type discriminator.
+
+    BR-RULE-015 INV-6: an assets-map value lacking asset_type (or carrying an
+    unregistered value) is unresolvable against the 14 AssetVariant types and
+    should be rejected. Production types ``Creative.assets`` as ``dict[str, Any]``
+    (untyped) and does NOT yet validate the discriminator, so the request is
+    accepted — a production gap. The scenario is strict-xfailed (``_XFAIL_TAGS``)
+    until BR-RULE-015 validation lands; dispatching RAW exercises the future
+    boundary validation on the wire so the xfail flips the moment it ships.
+    """
+    from tests.bdd.steps.generic.given_media_buy import _add_inline_creatives, _ensure_request_defaults
+
+    _ensure_request_defaults(ctx)
+    _add_inline_creatives(ctx, count=1)  # builds an assets-map value with no asset_type
+    ctx["dispatch_mode"] = "create_raw"
+
+
+@then("the error should reference the unresolvable asset_type discriminator")
+def then_error_references_asset_type(ctx: dict) -> None:
+    """The wire error names the unresolvable asset_type discriminator/field.
+
+    Wire-first (reads the real two-layer envelope). Reached only once production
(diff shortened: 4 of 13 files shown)


ChrisHuie:
## Review — `ec7ef3205`

0 BLOCKER / 6 SHOULD-FIX / 14 NIT. CI 31/31 green on this head; no runtime behavior changes needed — all SHOULD-FIX items are wording/honesty and small test-hardening edits.

**Verified (empirical):** the canonical suggestion is emitted uniformly on all three transports (real-wire probes: MCP both sub-paths, A2A failed-Task artifact, REST body); mutating `VALIDATION_ERROR_SUGGESTION` reddens exactly 4 tests via 4 independent mechanisms (pinned-enum conformance guard, 2 literal pins, verb-alignment oracle); `suggest_validation_fix` has 0 remaining references; inv-015-6 runs the real production body and strict-xfails on all 3 transports (0 xpass/0 failed); the new guard fails on a synthetic dormant branch injected into the real conftest; 528 architecture/boundary guard tests pass; both canonical suggestion texts and `allowed_values: [INVALID_REQUEST, VALIDATION_ERROR]` verified verbatim against the published 3.1.1 artifacts (`universal/idempotency.yaml:252`, `universal/error-compliance.yaml:267`); recovery is graded in no storyboard and suggestion content in none — and `transport-errors.mdx:672` prefers generic correction guidance, which supports the uniform-canonical choice. mcp 1.28.1 is exactly the GHSA-vj7q-gjh5-988w patch version; the uv.lock change is package-surgical.

**SHOULD-FIX**
1. `src/core/validation_helpers.py:48-52` (+ test docstring :77-80, feature :295, PR body): "the conformance storyboards accept either code … graded-equivalent" is overgeneralized — `refine_finalize_exclusivity.yaml:255` at 3.1.1 grades a `schema_invalid` failure with exact `INVALID_REQUEST`, and that payload raises a Pydantic ValidationError through `adcp_validation_boundary` (via `get_products` request construction). Latent today (refine unsupported; wire behavior pre-existing) — scope the sentence to the shapes actually graded both-ways and note the known exception.
2. Same-rule siblings not reconciled: `BR-UC-002-create-media-buy.feature:2584-2585` still grades the asset-discriminator shape `INVALID_REQUEST`, and `tests/bdd/conftest.py:183-185` (inv-087-5/6/7) keep the "VALIDATION_ERROR instead of INVALID_REQUEST — spec-production gap" framing the new boundary docstring forecloses. Reconcile the 2 outline rows + 3 reasons to the deliberate-divergence framing.
3. New guard: three probed escape shapes are unmodeled AND undocumented (`if "T-UC-x" not in marker_names: <yield> else: pytest.xfail(...)`; local-var indirection; ternary dispatch) — 0 live occurrences; add them to the KNOWN LIMITATION docstring and pin each with an `assert _violations(...) == []` meta-test so future modeling flips them.
4. inv-015-6 story is inaccurate on every transport: docstrings say the invalid request "is accepted" — observed: a2a/rest fail earlier with CREATIVE_REJECTED; mcp passes Thens 1-2 off a crash-normalized envelope and fails Then#3. Rewrite the Given/Then docstrings + the conftest FIXME block (:186-189) to the per-transport reality so the strict-xfail's graduation conditions are explicit.
5. The Given borrows its invalidity from `_add_inline_creatives`' default asset shape (`given_media_buy.py:2055-2061`) — the same shape valid-creative scenarios rely on. When asset validation lands and the helper gains `asset_type`, the probe is silently defused and the xfail keeps passing on CREATIVE_REJECTED. Make the probe self-sufficient (explicit asset value or `pop("asset_type", None)` with the coupling named).
6. Surfaced pre-existing production bug (follow-up issue, not this PR): on MCP, an inline-creative create hits `canonical_payload_hash(raw_wire_payload)` with a `FormatId` object → `rfc8785.CanonicalizationError` (ValueError subclass) → the wire carries buyer-correctable VALIDATION_ERROR for an internal crash (`media_buy_create.py:2080`, `exceptions.py:1093-1094`). a2a/rest don't hit it.

**NIT** (fold-in candidates)
- Guard prose vs matcher: ":20 ~10x" → 8; ":86 prevailing idiom" describes zero existing dormant named-set branches (all named-set intersections gate yielding branches); `_violations` docstring (:159) + fix_hint (:200) omit `skip`.
- `# FIXME(#1652)` missing at the allowlisted source branch (`conftest.py:3420`).
- `#1604` pointer now absent repo-wide while the issue is open — cite it in the renamed integration test's docstring.
- `test_mcp_typeadapter_validation_envelope.py:102`: the now-live test should also assert `errors[0].field == "packages[0].product_id"` and call `assert_no_raw_validation_leak` like its two siblings; `message_substr="Field required"` is newly load-bearing raw-Pydantic text.
- The two touched unit tests pin suggestion/field but not `error_code` — a subclass swap would pass at unit level (wire layer would catch it).
- Actionability predicate `set(s.lower().split()) & VERBS` is written twice (harness + oracle) — extract a shared helper.
- `uv.lock` `revision = 3` → `2`: re-locked with an older uv than wrote main's lock; package-surgical otherwise; either re-lock with a newer uv or accept one line of churn later.
- "Follow-ups (tracked)": the AUTH_REQUIRED_SUGGESTION divergence has no tracking issue yet; relatedly the pinned enum fixture is frozen at the 2026-05-13 cut while 3.1.1 was re-cut 2026-06-30 (8 codes drifted / 28 absent — this PR's two codes identical in both). One issue can cover re-freeze + AUTH_REQUIRED.
- Pre-existing, untracked: A2A pre-boundary missing-params checks emit non-canonical `"Required: [...]"` suggestions (5 sites in `adcp_a2a_server.py`) vs canonical text on MCP/REST for the same mistake.
- Pre-existing: `then_error_references_missing_field` (uc002:1900) grades a live scenario off the reconstructed exception, not the wire — upgrade alongside the new wire-first step.

Notes: CI executes only the `[a2a]` variant of the strict-xfail scenario until graduation (`BDD_ALL_TRANSPORTS` unset everywhere — by design, worth knowing). The deferred wire-value suggestion oracle is confirmatory for constant drift (the 4-mechanism chain covers that) but load-bearing for a serializer-layer rewrite — the in-flight branch's scope is correct.

pmezzich:
## Review, head `6854f6fc`

Bottom line: no blockers. The core mechanisms hold up under independent re-derivation, and I confirmed your own load-bearing claims by running them rather than taking the PR body's word. What I have is three should-fixes (one in the remediation commit itself, one executed bypass of the new guard, one staleness item from #1674 landing after you opened this) and three nits.

**Verified (empirical, run locally on this head, postgres up)**

- Your "mutating `VALIDATION_ERROR_SUGGESTION` reddens exactly 4 tests" claim: reproduced exactly. Baseline unit run vs mutated run diffed, 133 to 137 failures, and the 4 new ones are precisely the 4 you named, by nodeid.
- The dormant-scenario guard is not vacuous: injecting a synthetic tag-keyed `pytest.xfail()` into the real `_harness_env` reddens `test_no_new_dormant_scenario_specific_xfail` with `new violations (1)`. 14/14 in the guard file pass unmutated.
- inv-015-6 runs the real production body and strict-xfails on all three transports: with `BDD_ALL_TRANSPORTS=1` it collects 3, xfails 3, zero xpass, zero failed. Without that env var only `[a2a]` collects, so anyone spot-checking the per-transport story without it is looking at one transport, worth a line in the reason string or the conftest comment.
- `suggest_validation_fix` has 0 remaining references repo-wide.
- `test_mcp_typeadapter_validation_envelope.py` 3/3 pass locally.

**Not verified**

- The PR body's `run_all_tests.sh ci` counts. My Windows env carries ~133 pre-existing unit failures from cp1252 decode in source-scanning guards (`read_text()` with no encoding), so I cannot reproduce full-suite numbers here. Everything I verified above was diffed against a baseline run to exclude that noise.

### SHOULD-FIX

1. **The remediation's per-transport comment misstates the idempotency hash path.** `tests/bdd/conftest.py:198` (added in `6854f6fc` as the corrected story) says "(a2a/rest use canonical_request_hash(req), which model_dumps the FormatId, so they don't hit it.)". The code disagrees: the single hash site is `media_buy_create.py:2080`, `canonical_payload_hash(raw_wire_payload) if raw_wire_payload is not None else canonical_request_hash(req)`, and `canonical_request_hash` has no other src caller. Both a2a and rest thread a non-None `raw_wire_payload` (a2a: `adcp_a2a_server.py:1403` deepcopies parameters, `:1639` falls back to `params`, never None; rest: `routes/api_v1.py:292-296` returns `json.loads(raw)` or `await request.json()`, always a dict). So all three transports take the `canonical_payload_hash(raw_wire_payload)` branch, and the real asymmetry is the content: a2a/rest carry a plain wire JSON dict whose format_id is a nested object literal, mcp's payload carries a live `FormatId` object that rfc8785 rejects. Why it matters: whoever picks up #1679 will read this comment, conclude a2a/rest are safe because they hash via a different function, and look for the fix at a seam that does not execute. Suggested rewording: "All transports compute canonical_payload_hash(raw_wire_payload); canonical_request_hash(req) is only the raw_wire_payload-is-None fallback for impl-direct callers. a2a/rest thread a plain wire dict (format_id as a nested object literal, rfc8785-canonicalizable); mcp's payload carries a live FormatId object, which rfc8785 rejects."

2. **The guard's family-gate exclusion is defeatable by a one-token rewrite, executed.** `_prefix_match_constants` excludes every string constant appearing as a `.startswith()`/`.endswith()` argument, by value, across the whole if-test. Two consequences, both run against the real conftest in a scratch tree, guard re-run after each, reverted between:
   - `elif any(t.startswith("T-UC-002-esc-999") for t in marker_names): pytest.xfail(...)` with a full exact tag: **not flagged** (the control `elif "T-UC-002-esc-996" in marker_names:` form is flagged, `new violations (1)`).
   - Literal poisoning: `elif "T-UC-002-esc-998" in marker_names or any(t.startswith("T-UC-002-esc-998") ...):` **not flagged** either, because the same-literal `startswith` suppresses the `in` match in the same test.

   The likely path to recurrence is someone copying the neighboring live idiom (`startswith("T-UC-002-ext-")` appears twice inside the scanned function) with a full tag instead of a family prefix, and the guard stays green. Both live family-gate args end in `-`, so a strict-prefix discriminator (`arg.value.endswith("-")`, or cross-checking the arg against real `@T-UC-*` tags from the feature files) plus computing the exclusion per call site instead of per test closes both shapes. Fair caveat: your KNOWN LIMITATION docstring already owns four escape shapes, so you may grade this a documented-limitation nit; I rate it should-fix because it is a one-token rewrite of the guard's own positive fixture using the idiom sitting next door, which is the copy-paste a future dev will actually make.

3. **The mcp hunk is now obsolete and slightly regressive.** #1674 (`324e0c14`) and #1656 landed the identical `mcp>=1.28.1` fix on main on 07-20, after you opened this. A merge simulation (`git merge-tree --write-tree upstream/main pr-1673`) shows the merged pyproject carries three identical `mcp>=1.28.1` constraints and the merged uv.lock differs from main by exactly one line: `revision = 3` downgraded to `revision = 2` (this branch re-locked with an older uv). Nothing breaks, but it is dead cruft plus a lock-format downgrade for a CVE main already fixes twice. Suggest dropping the pyproject line and the uv.lock hunk entirely at rebase.

### NIT

- Carry-over from your own should-fix #1, one site missed: `BR-UC-002-create-media-buy.feature:295` still carries the unqualified "storyboards accept either" claim, while `validation_helpers.py:46` and the integration-test docstring got the refine_finalize_exclusivity scoping. The feature comment is the first place a reader lands, and it now contradicts the note it points at.
- The new guard reads `tests/bdd/conftest.py` via `read_text()` with no `encoding=`. On Windows that decodes as cp1252 and hard-fails on the first non-ASCII character; 8 existing integration files are already undecodable that way and their scanners fail locally today. One `encoding="utf-8"` keeps the guard out of that pile.
- The KNOWN LIMITATION list covers test-side escapes but not body-side indirection (a helper function that performs the xfail). Not worth new detection, but worth a line in the docstring so the boundary of the guard is stated rather than discovered.

### Scope note, no action needed here

The guard covers the setup-fixture shape in one function of one file, which is the right target for the #1534 recurrence. Stating the boundary plainly for anyone who reads it as broader: `tests/bdd/steps/` holds 135 imperative `pytest.xfail(` calls and 1 `pytest.skip(` in the same never-XPASS dormancy class, none visible to this guard. That population is what my #1622 informational scan surfaces, so the two are complementary rather than overlapping, and landing both covers the fixture shape with enforcement and the step-body shape with visibility.

tests/bdd/conftest.py
```diff
@@ -180,9 +180,26 @@ def pytest_configure(config: pytest.Config) -> None:
...
+    # FIXME(#1652): BR-RULE-015 INV-6 — production DOES validate asset_type (the CreativeAsset
+    # discriminated union raises union_tag_not_found), but the WIRE outcome differs per transport
+    # and none matches the scenario's graded "VALIDATION_ERROR referencing asset_type" shape:
+    #   a2a / rest: creative processing rejects with CREATIVE_REJECTED (not VALIDATION_ERROR) — fails
+    #               the earlier validation-error Then.
+    #   mcp: an earlier crash in the idempotency canonical hash (media_buy_create.py
+    #        canonical_payload_hash(raw_wire_payload) -> rfc8785 chokes on a live FormatId ->
+    #        CanonicalizationError) is normalized to VALIDATION_ERROR "unsupported type: FormatId";
+    #        passes Thens 1-2 off that crash envelope, fails Then#3 (message is FormatId, not asset_type).
+    #        (a2a/rest use canonical_request_hash(req), which model_dumps the FormatId, so they don't hit
```
KonstantinMirin, line 198: This says a2a/rest use `canonical_request_hash(req)`, but the single hash site (`media_buy_create.py:2080`) is `canonical_payload_hash(raw_wire_payload) if raw_wire_payload is not None else canonical_request_hash(req)` — a2a and rest both thread a non-None `raw_wire_payload`, so all three transports take the `canonical_payload_hash` branch; `canonical_request_hash(req)` is only the impl-direct fallback. This is the graduation condition for the strict-xfail, so as written it points whoever picks up #1679 at a seam that never runs. The real asymmetry is content (mcp carries a live `FormatId` rfc8785 rejects; a2a/rest carry a plain wire dict). ↩ raised by @pmezzich 2026-07-21, still open.

tests/unit/test_architecture_bdd_no_dormant_scenario_xfail.py
```diff
@@ -0,0 +1,434 @@
...
+
+def _get_harness_env(source: str) -> ast.FunctionDef:
+    tree = ast.parse(source)
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef) and node.name == _TARGET_FUNC:
+            return node
+    raise AssertionError(f"{_TARGET_FUNC} not found")
+
+
+def _prefix_match_constants(test: ast.expr) -> set[str]:
```
KonstantinMirin, line 62: `_prefix_match_constants` excludes every `.startswith()`/`.endswith()` arg by value across the whole if-test, which lets two dormant shapes stay green: a branch keyed on a *complete* tag (`startswith("T-UC-002-esc-999")`) and literal-poisoning (same literal in a `startswith` suppresses the `in`-form match). The scanned function already uses `startswith("T-UC-002-ext-")` twice, so the full-tag copy-paste is the idiom a future dev reaches for. Compute the exclusion per call site and restrict the family gate to strict prefixes (`arg.value.endswith("-")` or cross-check against real `@T-UC-*` tags), and pin both shapes with `assert _violations(...) == []` meta-tests. ↩ raised by @pmezzich 2026-07-21, still open.

tests/bdd/features/BR-UC-002-create-media-buy.feature
```diff
@@ -289,10 +289,10 @@ Feature: BR-UC-002 Create Media Buy
     And the account "acc-001" exists and is active
     When the Buyer Agent sends the create_media_buy request
     Then the response should indicate a validation error
-    And the error code should be "INVALID_REQUEST"
+    And the error code should be "VALIDATION_ERROR"
     And the error should reference the unresolvable asset_type discriminator
     And the error should include "suggestion" field
-    # BR-RULE-015 INV-6: assets value lacking asset_type or carrying an unregistered value (not one of the 14 AssetVariant types) -> INVALID_REQUEST
+    # BR-RULE-015 INV-6: assets value lacking asset_type or carrying an unregistered value (not one of the 14 AssetVariant types). Prose maps this schema failure to INVALID_REQUEST; the boundary deliberately emits VALIDATION_ERROR uniformly (storyboards accept either -- see adcp_validation_boundary).
```
KonstantinMirin, line 295: "(storyboards accept either -- see adcp_validation_boundary)" is the unqualified claim that `validation_helpers.py:46` and the integration-test docstring were already scoped away from: `refine_finalize_exclusivity.yaml` grades a schema-invalid path strict `INVALID_REQUEST`, so the uniform `VALIDATION_ERROR` is a deliberate divergence, not a graded equivalence. This is the first place a reader lands and it now contradicts the note it points at — mirror the scoped wording. ↩ raised by @ChrisHuie 2026-07-18 and @pmezzich 2026-07-21, still open.

pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "pyasn1>=0.6.3",  # GHSA-jr27-m4p2-rc6r
     "mako>=1.3.12",  # GHSA-2h4p-vjrc-8xpq (transitive via alembic)
     "click>=8.3.3",  # PYSEC-2026-2132 (transitive via flask/uvicorn)
+    "mcp>=1.28.1",  # GHSA-vj7q-gjh5-988w (transitive via fastmcp/adcp; both allow <2.0)
```
KonstantinMirin, line 56: #1674 / #1656 landed this identical `mcp>=1.28.1` constraint on main after this PR opened (`origin/main` pyproject already carries it), so this line is now dead cruft and the accompanying `uv.lock` `revision = 3 → 2` downgrade re-locks with an older uv for a CVE main already fixes. Drop this line and the uv.lock hunk at rebase. ↩ raised by @pmezzich 2026-07-21, still open.

tests/bdd/steps/domain/uc002_create_media_buy.py
```diff
@@ -607,6 +607,59 @@ def _first_package(ctx: dict) -> dict:
...
+    not the graded VALIDATION_ERROR); mcp reaches THIS Then but the wire message is
+    the FormatId CanonicalizationError crash (see #1679), not the
+    asset_type discriminator. Graduates when the code reconciles to a
+    VALIDATION_ERROR that references asset_type.
+    """
+    from tests.bdd.steps.generic.then_error import _wire_error_object
+
+    err = _wire_error_object(ctx)
+    assert err is not None, "Expected a wire error envelope referencing asset_type"
+    haystack = f"{err.get('field', '')} {err.get('message', '')}".lower()
```
KonstantinMirin, line 656: This "read `_wire_error_object` → build `field + message` haystack → assert token present" block is now the third copy (also `:1928` in this PR, `uc019_query_media_buys.py:1695` pre-existing) and they've already drifted: this one case-folds, the `:1928` copy does not, uc019 reverses field/message order. Extract one helper next to `_wire_error_object` in `then_error.py` (e.g. `wire_error_references(ctx, token, *, casefold=True)`) and route all three through it.

tests/bdd/steps/domain/uc002_create_media_buy.py
```diff
@@ -1864,9 +1916,22 @@ def then_response_includes_previously_created(ctx: dict, field: str) -> None:
...
+    Wire-first: when the scenario dispatched through a wire transport the buyer
+    contract is the wire envelope's ``field``/``message`` (read via
+    ``_wire_error_object``), not the lossy reconstructed ``ctx['error']``. Falls
+    back to the reconstructed exception on IMPL / no-wire scenarios.
     """
+    from tests.bdd.steps.generic.then_error import _wire_error_object
+
+    wire = _wire_error_object(ctx)
+    if wire is not None:
+        haystack = f"{wire.get('field', '')} {wire.get('message', '')}"
```
KonstantinMirin, line 1928: Same wire-error haystack operation as `then_error_references_asset_type` (`:656`) and `uc019:1695`, but this copy does NOT `.lower()` the haystack while `:656` does — silent drift already present. Route this wire branch through the shared `wire_error_references` helper extracted into `then_error.py`; keep the `ctx['error']` fallback at the call site.

KonstantinMirin (changes requested):
## Review — PR #1673

**Overview** — The core remediation is clean and well-grounded: `suggest_validation_fix` is fully removed, the boundary emits the single canonical `VALIDATION_ERROR_SUGGESTION`, the actionability predicate is now one shared `suggestion_has_action_verb` helper, and the suggestion behavior is graded on the wire by a live idempotency scenario across a2a/mcp/rest. The inv-015-6 strict-xfail correctly runs the real production body instead of an imperative setup-xfail, and the new dormant-scenario guard is a genuine addition. What still needs fixing: four items maintainers raised in the last two rounds are still open on this head (a per-transport hash-path comment that points readers at a seam that never runs, the guard's own bypass hole, the one un-scoped divergence comment, and the now-obsolete mcp dependency hunk), plus one new duplication this push introduces.

### Should fix

**↩ raised by @pmezzich 2026-07-21 (SHOULD-FIX #1), still open — per-transport hash comment points at a seam that never runs.**
`tests/bdd/conftest.py:198` says a2a/rest "use `canonical_request_hash(req)`, which model_dumps the FormatId, so they don't hit it." The single hash site is `media_buy_create.py:2080`, `canonical_payload_hash(raw_wire_payload) if raw_wire_payload is not None else canonical_request_hash(req)` — a2a and rest both thread a non-None `raw_wire_payload`, so all three transports take the `canonical_payload_hash` branch; `canonical_request_hash(req)` is only the impl-direct fallback. The real asymmetry is payload *content* (a2a/rest carry a plain wire JSON dict; mcp carries a live `FormatId` object rfc8785 rejects). This comment is the documented graduation condition for the strict-xfail — whoever picks up #1679 will look for the fix at a function that doesn't execute. Adopt @pmezzich's suggested rewording.

**↩ raised by @pmezzich 2026-07-21 (SHOULD-FIX #2), still open — the dormant-scenario guard is defeatable by the copy-paste sitting next door.**
`_prefix_match_constants` (`test_architecture_bdd_no_dormant_scenario_xfail.py:62`) excludes every string used as a `.startswith()`/`.endswith()` argument, by value, across the whole if-test. Two shapes stay green: a dormant branch keyed on a *complete* tag via `startswith("T-UC-002-esc-999")` (the scanned function already uses `startswith("T-UC-002-ext-")` twice, so this is the idiom a future dev copies), and literal-poisoning where the same literal in a `startswith` call suppresses the `in`-form match in the same test. A guard that stays green on the exact recurrence it exists to catch doesn't deliver the prevention that is this PR's purpose. Compute the exclusion per call site and restrict the family-gate to strict prefixes (`arg.value.endswith("-")`, or cross-check against real `@T-UC-*` tags), and pin both shapes with `assert _violations(...) == []` meta-tests alongside the existing KNOWN LIMITATION pins.

**↩ raised by @ChrisHuie 2026-07-18 (SHOULD-FIX #1, names `feature:295`) and again as a carry-over by @pmezzich 2026-07-21, still open — one divergence comment still un-scoped.**
`src/core/validation_helpers.py:46` and the integration-test docstring were correctly scoped to say some storyboards grade a rejection either-way (`idempotency.yaml` / `error-compliance.yaml`) while `refine_finalize_exclusivity.yaml` grades a schema-invalid path strict `INVALID_REQUEST` — so the uniform `VALIDATION_ERROR` is a deliberate divergence. `tests/bdd/features/BR-UC-002-create-media-buy.feature:295` still reads the blanket "(storyboards accept either -- see adcp_validation_boundary)". That is the first place a reader lands, and it contradicts the very note it points at. Mirror the scoped wording (name `refine_finalize_exclusivity.yaml` as the strict-graded exception).

**↩ raised by @pmezzich 2026-07-21 (SHOULD-FIX #3), still open — mcp dependency hunk is now obsolete and slightly regressive.**
#1674 / #1656 landed the identical `mcp>=1.28.1` constraint on main after this PR opened (`origin/main` pyproject already carries it). This PR's added `pyproject.toml:56` line is now dead cruft, and the `uv.lock` `revision = 3 → 2` downgrade re-locks with an older uv for a CVE main already fixes. Drop the pyproject line and the uv.lock hunk at rebase.

**Wire-error field+message assertion copy-pasted and already drifting (DRY).**
This push introduces a second copy of "read `_wire_error_object(ctx)`, build a `field + message` haystack, assert a token is present" and modifies a sibling — without extracting the operation. The two in-scope copies already diverge silently: the new `then_error_references_asset_type` (`uc002_create_media_buy.py:656`) case-folds the haystack; the modified `then_error_references_missing_field` wire branch (`:1928`) does not; the pre-existing `uc019_query_media_buys.py:1695` reverses field/message order. A future token needing case-insensitive matching will pass on one and silently miss on the other — the canonical silent-drift shape.

```python
# uc002:656  — case-folded, field-then-message
haystack = f"{err.get('field','')} {err.get('message','')}".lower()
# uc002:1928 — NOT case-folded, field-then-message
haystack = f"{wire.get('field','')} {wire.get('message','')}"
```

Extract one helper next to `_wire_error_object` in `tests/bdd/steps/generic/then_error.py` (e.g. `wire_error_references(ctx, token, *, casefold=True) -> bool`) that owns the read, haystack construction, case rule, and `ctx['error']` fallback; route the two in-scope sites through it and fold in the uc019 sibling in the same change.

### Nice to have

- The new guard reads `tests/bdd/conftest.py` via `read_text()` with no `encoding=` (`test_architecture_bdd_no_dormant_scenario_xfail.py`). On a non-UTF-8 default (Windows cp1252) it hard-fails decoding the em-dashes in conftest before it can scan. `read_text(encoding="utf-8")` pins it to the encoding the file is authored in. (↩ raised by @pmezzich 2026-07-21 as a NIT.)

Resolved since the last rounds, credited: `suggest_validation_fix` fully removed (0 refs); the actionability predicate extracted into the shared `suggestion_has_action_verb` (ChrisHuie NIT); `inv-087-5/6/7` reasons and the asset-discriminator outline rows reconciled to the deliberate-divergence framing (ChrisHuie SHOULD-FIX #2); the `# FIXME(#1652)` markers now present at the allowlisted branches; the integration test upgraded to assert `field == "packages[0].product_id"` + `assert_no_raw_validation_leak`; `then_error_references_missing_field` upgraded to grade the real wire envelope. The new guard's allowlist seeds only one pre-existing dormant (`T-UC-018-ext-c`, `# FIXME(#1652)`) — no allowlist growth. The `AUTH_REQUIRED_SUGGESTION` capitalization divergence is pre-existing and already tracked — not re-raised.

PR #1575, opened 2026-07-09
fix: media buy delivery correctness — scheduler serving statuses + webhook-only fields

numarasSigmaSoftware:
Two delivery-path correctness fixes raised in the #1545 review, plus verification that #1548 was already resolved by #1545.

Closes #1556
Closes #1570

## Concern map

This PR is larger than its two issues. Review raised splitting it; the split was measured
and is not reachable (reasons on the thread — the delivery scheduler went 355 → 935 lines
with hunks that mix concerns internally, 23 of 92 source and test files are multi-concern,
and two guard ratchets are bidirectionally fatal so source and test must ship together).
What follows is the per-subsystem surface a split would have produced, so each concern can
be reviewed on its own without four branches.

| # | concern | principal files | follow-up |
|---|---|---|---|
| 1 | **Serving-status vocabulary** (#1556) | `core/tools/_media_buy_status.py`, `services/media_buy_status_scheduler.py`, `services/delivery_webhook_scheduler.py` | [#1713](https://github.com/prebid/salesagent/issues/1713) (sibling literals) |
| 2 | **Webhook-only field gating** (#1570) | `core/tools/media_buy_delivery.py`, `core/schemas/delivery.py`, `core/tools/_media_buy_status.py` | [#1624](https://github.com/prebid/salesagent/issues/1624) (second emitter) |
| 3 | ~~Final/periodic claim + lease~~ — **carved out, see below** | — | [#1606](https://github.com/prebid/salesagent/issues/1606) (durable exactly-once) |
| 4 | ~~Reporting-cadence validation~~ — **carved out**, see below | — | branch `feat/reporting-cadence-validation`, PR opens when this merges |
| 5 | **Production signal + URL redaction** | `core/config.py`, `core/webhook_validator.py`, `admin/*`, `landing/*`, migration `168914d7ca05` | [#1819](https://github.com/prebid/salesagent/issues/1819) (remaining sites) |
| 6 | **Idempotency / durable log** | `services/protocol_webhook_service.py` (`_DeliveryLogContext`, `DeliveryLogPersistenceError`), migration `b7c9d2e4f6a8` | — |
| 7 | **Send-path correctness** | `services/protocol_webhook_service.py`, `services/delivery_webhook_scheduler.py`, `services/media_buy_status_scheduler.py` | — |
[...]

PR #1697, opened 2026-07-22
fix(security): SSRF-validate protocol push and reporting webhook URLs

mkostromin-sigma:
## Summary
- Apply the same `WebhookURLValidator` / `check_url_ssrf` gate used by application webhook delivery to **protocol** push notifications before any outbound POST (`ProtocolWebhookService.send_notification`).
- Reject unsafe URLs at **registration**: `create_media_buy` (`reporting_webhook.url` as `AnyUrl` via `str(...)` + `push_notification_config.url`), A2A `message/send` push-config stash, A2A `set_push_notification_config`, and `PushNotificationConfigRepository.upsert`.
- Create-path PNC gate runs **before** workflow metadata writes; persist via repo `upsert`.
- Callers treat `send_notification` `False` as failure (no success log / A2A warning) — aligns with #1578.

Closes #1695. Closes #1578.

### Spec-grounding
Authoritative: AdCP **3.1.1** [`docs/building/by-layer/L1/security.mdx`](https://github.com/adcontextprotocol/adcp/blob/v3.1.1/docs/building/by-layer/L1/security.mdx) § **Webhook URL validation (SSRF)** — normative "fetchers MUST" list covering `push_notification_config.url` (and sibling buyer/seller-supplied fetch URLs). SDK pin: `docs/adcp-spec-version.md` / `adcp==6.6.0`. Storyboard: **ungraded** (no graded SSRF scenario).

This PR implements:
1. Reject non-HTTPS in production (`_strict_mode` / registration + outbound gates)
2. Resolve hostname and reject reserved ranges (IPv4 RFC1918/CGNAT/loopback/link-local/multicast; IPv6 loopback/ULA/link-local/multicast `ff00::/8`; NAT64 well-known `64:ff9b::/96`)
4. Refuse redirects on outbound webhook POST (`follow_redirects=False`)
5. Cap response size and timeouts — **timeouts** implemented on both outbound paths; **body-size cap** deferred (repo-wide, same bucket as step 3)
6. Do not echo raw fetch topology in buyer-facing errors (structured `VALIDATION_ERROR` + suggestion)

Deferred (repo-wide follow-up, same gap on application delivery):
3. Pin the connection to the validated IP (DNS rebinding TOCTOU) — re-resolving without pinning is not sufficient per the same section.
5. *(partial)* Response body-size cap (timeouts already set).

## Test plan
[...]

Issue #1619, opened 2026-07-14
feat: honor paused=true on create_media_buy


KonstantinMirin:
## Summary

`CreateMediaBuyRequest.paused` (AdCP 3.1.1 surface, exposed on both transports in PR #1567) is accepted but read nowhere on the create path — a `paused=true` create delivers immediately. `_build_create_success` takes a separate `paused` param, and adapters read the *response* package's `paused`, never the request's. The parameter description currently states this honestly ("accepted for AdCP 3.1.1 compatibility; pause-on-create is NOT yet honored").

Provenance: pre-existing gap (request-level `paused` is 2.12.0-lineage, honored nowhere before the 6.6 bump either) — not a 6.6 regression. Surfaced in PR #1567 review round 1.

## Fix

1. **Spec-grounding first** (mandatory per repo policy): cite the 3.1.1 create-media-buy request schema section for `paused` semantics + the conformance storyboard step that grades it (or note "ungraded") before implementation.
2. Thread `req.paused` into the initial media-buy status / adapter package state so delivery is gated until unpaused.
3. Mirror in the mock adapter's metric simulation (paused buy simulates zero delivery until unpaused).
4. Revert the parameter description wording once honored.
5. Grade with the UC-002 paused partition across all four transports.

Related: #1305 (async/submitted response envelopes) — both touch create-path lifecycle semantics; coordinate the status values.

## Acceptance

- `paused=true` create produces a buy that does not deliver until unpaused, on all transports.
- BDD scenario(s) graded, not dormant; mock adapter simulates the pause.
- Parameter description matches behavior again.

PR #1721, opened 2026-07-26
feat: one code table for every buyer-facing error, plus the AdCP 3.1.1 capabilities/accounts contract

### LINKED ISSUE #1210: fix: list_accounts not advertised in capabilities tool output
## Summary

`list_accounts` is fully implemented and registered with MCP and A2A transports, but `src/core/tools/capabilities.py` does not mention accounts in its output. The AdCP spec supports capability advertisement — buyer agents querying capabilities should see that this seller supports account management.

**Follow-up from:** #1011

## Acceptance criteria

- [ ] `capabilities` tool output includes `list_accounts` (and `sync_accounts` if applicable)
- [ ] Verify the capability advertisement aligns with AdCP spec format
- [ ] Unit test covers accounts appearing in capabilities output

### LINKED ISSUE #1521: fix: accounts.billing CHECK constraint rejects spec-valid 'advertiser' → 500 on sync_accounts
## Summary
The `accounts.billing` CHECK constraint allows only `('operator', 'agent')`, but the AdCP `billing-party` enum (pinned spec 3.1.0-beta.3) has three values: `operator`, `agent`, **`advertiser`**. A `sync_accounts` entry with `billing: "advertiser"` passes request validation and the billing-policy check, then violates the DB constraint on INSERT — a 500 instead of a clean success/typed error.

## Failing path (verified @ `a55db82d`)
1. `billing="advertiser"` validates — `adcp.types.Account.billing` is the `BillingParty` enum, which includes `advertiser`.
2. `_check_billing_policy` (`src/core/tools/accounts.py:389-416`) returns accept when `supported_billing` is unconfigured (`:404-405`, accept-all) or when `advertiser` is in the configured set.
3. Sync passes `billing=billing_val` into `Account(...)` (`src/core/tools/accounts.py:626`) → `repo.create` → INSERT.
4. `CheckConstraint("billing IS NULL OR billing IN ('operator', 'agent')", name="ck_accounts_billing")` — `src/core/database/models.py:852` and `alembic/versions/51d4f9009db4_create_accounts_and_agent_account_.py:60` — raises IntegrityError → 500.

## Why it matters
`advertiser` is a first-class billing party in the spec (the seller invoices the advertiser directly while an operator places orders). The constraint makes the spec-legal three-value billing posture unreachable and turns a valid request into a 500. This is semantic single-source-of-truth drift: the value-set lives in the library enum, the policy layer, and the DB constraint — the constraint is the one out of sync.

## Fix
1. New Alembic migration: recreate `ck_accounts_billing` as `billing IS NULL OR billing IN ('operator', 'agent', 'advertiser')`, with a matching `downgrade()`.
2. Update the ORM constraint at `src/core/database/models.py:852`.
3. Confirm no downstream code assumes only two billing values.

## Acceptance criteria
- [ ] `sync_accounts` with `billing: "advertiser"` persists (no 500) when the tenant permits it.
- [ ] Migration has non-empty `upgrade()`/`downgrade()` and keeps a single head.
- [ ] Integration test creates an account with `billing='advertiser'`.
- [ ] Existing `operator`/`agent` behavior unchanged.

## Test plan
- Add an `advertiser`-billing case to `tests/integration/test_sync_accounts.py` (current tests only exercise `agent`/`operator`).

## References
- Spec: AdCP 3.1.0-beta.3 `enums/billing-party.json` = `["operator","agent","advertiser"]`; `core/account.json` example uses `billing: "advertiser"`.
- Parent: #1012; sibling: #1029.

### LINKED ISSUE #1592: capabilities: populate account section + spec-optional supported_pricing_models / reporting_delivery_methods
get_adcp_capabilities never populates three sections buyers use for pre-flight filtering (all declared in v3.1.1 `get-adcp-capabilities-response.json`):

- `account` (sandbox flag + billing models) — pending the account-management work (#1012 follow-up)
- `media_buy.supported_pricing_models` — `adapter.get_supported_pricing_models()` exists (src/adapters/base.py) but capabilities.py never wires it
- `media_buy.reporting_delivery_methods`

Graded by the strict-xfail gap asserts at the END of both UC-010 main scenarios (BR-UC-010-discover-seller-capabilities.feature) — populating them forces graduation.

### LINKED ISSUE #1598: Missing-creative references on create/update media-buy surfaces emit CREATIVE_REJECTED, but the pinned enum says CREATIVE_NOT_FOUND MUST be uniform
AdCP 3.1.0-beta.3 `enums/error-code.json` defines CREATIVE_NOT_FOUND with: "Sellers MUST return this code uniformly for any creative_id not owned by the calling account".

Today the surfaces disagree:
- **sync_creatives** (strict assignment validation) emits `CREATIVE_NOT_FOUND` ✔
- **create_media_buy** pre-validation (`src/core/tools/media_buy_create.py` ~:374 and the workflow-approval path ~:3775) emits `CREATIVE_REJECTED`
- **update_media_buy** (`src/core/tools/media_buy_update.py` ~:210) emits `CREATIVE_REJECTED` — this cell is storyboard-graded (BR-UC-003 ext-i expects CREATIVE_REJECTED)

Either the enum MUST governs all surfaces (then the ext-i storyboard cell needs an upstream adcp reconciliation before we change behavior), or CREATIVE_REJECTED is intended for media-buy-attachment contexts and the enum wording needs an upstream clarification. Until reconciled upstream, the create/update behavior is deliberately left as-is; the sites carry FIXME(#) comments referencing this issue.

### LINKED ISSUE #1880: list_accounts BDD steps bypass wire dispatch via allowlisted TRANSPORT-BYPASS
## Deferred from PR #1721 review round 2

Must be addressed as part of **PR #1728** (`fix/uc004-delivery-breakdowns`) before merge — that branch is already migrating `tests/bdd/steps/domain/uc011_accounts.py`, `tests/bdd/steps/generic/_dispatch.py`, `tests/bdd/conftest.py`, and `tests/unit/test_architecture_bdd_wire_discipline.py` wholesale. Closely related to that branch's own `salesagent-oyiv.8` (migrate uc011_accounts.py's typed-payload oracles to the wire accessor) — worth doing together, same file/step family.

## Finding

`tests/bdd/steps/domain/uc011_accounts.py:410,507` (`when_list_sandbox_filter` and a sibling) call `_impl` directly instead of `dispatch_request`, allowlisted in `test_architecture_bdd_no_direct_call_impl.py:31-36`. The paired Then-step at `:3130-3151` was strengthened to a real assertion, but it reads through this in-process bypass — so the scenario reports green across all 4 transport IDs (a2a/mcp/rest/e2e_rest) while only ever running the same in-process call once.

## Architecturally correct fix

Wire `list_accounts` through `dispatch_request` the same way sibling UC-011 steps already do — add the route in `tests/bdd/steps/generic/_dispatch.py` if not already dispatched there, add whatever env method is missing. Once wired, delete the `TRANSPORT-BYPASS` comments and the allowlist entries in `test_architecture_bdd_no_direct_call_impl.py:31-36`.

## Acceptance

`when_list_sandbox_filter` and its sibling dispatch through `dispatch_request`/`call_via`, not `_impl` directly; the allowlist entries are removed; the scenario genuinely runs on all 4 transport IDs (verify via `-rxX` per-transport).

Full triage evidence: `.claude/notes/pr1721-review-triage-and-bdd-rootcause.md` on `feature/spec-gaps-1210`.

### LINKED ISSUE #1882: properties.py::list_authorized_properties MCP wrapper skips adcp_validation_boundary
Follow-up from PR #1721 review round 2 (F2, salesagent-2bpt.2), scoped to the one site F2 didn't fix.

\`src/core/tools/properties.py::list_authorized_properties\` constructs \`ListAuthorizedPropertiesRequest(...)\` directly, without wrapping it in \`adcp_validation_boundary(context=...)\`. This is the same disease as the accounts.py \`list_accounts\`/\`sync_accounts\` gap fixed in salesagent-2bpt.2 — a raw pydantic \`ValidationError\` from the construction leaks past the MCP transport boundary untranslated instead of surfacing as a typed \`AdCPValidationError\` with code/recovery/suggestion.

Reference pattern (already correct for this exact tool, elsewhere): \`src/core/tools/creative_formats.py\` MCP wrapper, and \`src/routes/api_v1.py\`'s REST route for this same tool.

Currently tracked by an allowlist entry in \`tests/unit/test_architecture_request_construction_boundary.py::ALLOWLIST_DEFERRED\` citing this issue.

Fix: wrap the \`ListAuthorizedPropertiesRequest(...)\` construction in \`with adcp_validation_boundary(context="list_authorized_properties request"):\`, then remove the allowlist entry.

### LINKED ISSUE #1884: admin/blueprints/operations.py:247 relies on the ContextVar tenant fallback instead of threading tenant= explicitly
Follow-up from PR #1721 review round 2 (F5, salesagent-2bpt.5), surfaced while giving `src/core/helpers/adapter_helpers.py` real tenant types.

`src/admin/blueprints/operations.py:247` calls `get_adapter(principal_schema, dry_run=False)` without a `tenant=` argument -- the only call site across `src/` that omits it. This falls through to `adapter_helpers.py::_resolved_tenant()`'s `get_current_tenant()` ContextVar fallback, which by design (`set_current_tenant()`, `src/core/config_loader.py:185-197`) always normalizes `TenantContext` down to a plain `dict[str, Any]` before storing.

This is the one remaining reason `adapter_helpers.py`'s tenant-accepting functions can't be typed as simply `TenantContext | Tenant(ORM) | None` -- the `dict[str, object]` union member exists only for this call site (plus the ORM-`Tenant` member exists for `media_buy_create.py`'s 3 direct-session-query call sites, which are a separate, more defensible pattern).

The admin blueprint already has a DB session live at that point in the request (it constructs `principal_schema` from an ORM `principal` fetched via a session a few lines above) -- resolving/threading the tenant explicitly there instead of relying on the ContextVar fallback should be straightforward.

Fix: thread `tenant=` explicitly through `operations.py`'s `get_adapter()` call (resolve it the same way the surrounding code already resolves `principal`), removing the last caller of the ContextVar-dict fallback path. Once done, `adapter_helpers.py`'s `TenantLike`/`IdentityTenant` aliases can drop the `dict[str, object]` member and narrow to `TenantContext | Tenant | None`.

### LINKED ISSUE #1885: A2A/REST skill handlers bypass existing _build_X_request()/build_X_request() helpers (4 sites)
Follow-up from PR #1721 review round 2 (F2/F9, salesagent-2bpt.2 / salesagent-2bpt.9), consolidating the remaining DEFER instances from the disease-scan disposition table backing \`tests/unit/test_architecture_request_construction_boundary.py\` (originally scanned under the now-closed salesagent-5yik).

Disease: several A2A skill handlers / REST routes bypass their tool's existing, correctly-designed \`_build_X_request()\`/\`build_X_request()\` helper (whose own docstrings state they exist so construction runs inside the ONE validation boundary, #1417) in favor of \`RequestModel.model_validate(params)\` or a hand-rolled inline reconstruction. This is the "reimplements the builder instead of calling it" half of the disease the structural guard's own docstring says a structural scan can't catch reliably (semantic judgment required) — tracked here instead.

Sites (4, all still open as of this filing):
- \`src/a2a_server/adcp_a2a_server.py\` — several handlers bypass their tool's builder, worst instance reimplements only 5 of ~10 updatable fields from \`media_buy_update.py::_build_update_request\` (real behavior divergence, not just style)
- REST \`/creative-formats\` route bypasses \`build_list_creative_formats_request\` (the disease pattern's own canonical example)
- \`get_products_raw\` (A2A/REST path) calls \`create_get_products_request\` without \`adcp_validation_boundary\`
- \`create_media_buy\` A2A skill handler bypasses \`_build_create_media_buy_request\` (high-traffic, needs dedicated regression coverage)

Fix: route each site through its tool's existing builder inside \`adcp_validation_boundary\`, matching the pattern already correct elsewhere in the same files.

### LINKED ISSUE #1886: given_expired_token BDD Given doesn't actually drive a presented-but-invalid token through REST (no REST real-token harness seam)
Found while migrating the AUTH_MISSING/AUTH_INVALID split, surfaced again during PR #1721 review round 2 (F9, salesagent-2bpt.9) citation cleanup.

\`tests/bdd/steps/domain/uc011_accounts.py::given_expired_token\` is textually "expired token" but its implementation is identical to \`given_unauthenticated\` (no credential at all) -- it does not actually drive a PRESENTED-but-rejected token through the real resolution chain. That distinction was unobservable before the AUTH_MISSING/AUTH_INVALID split (both cases collapsed to the deprecated \`AUTH_REQUIRED\`); it is real now (absent -> AUTH_MISSING, presented-but-invalid -> AUTH_INVALID).

A real fix was attempted (an identity carrying a bogus \`auth_token\`, mirroring \`CapabilitiesEnv.invalid_token_identity()\`): A2A correctly exercises the real chain and rejects with AUTH_INVALID, but the REST harness's \`_configure_rest_auth\` dependency override always treats a non-None identity as a valid, already-resolved token (no real header/token-lookup path exists in-process for REST -- see \`tests/harness/capabilities.py\`'s "real header path is exercised on e2e_rest" note), so it collapses to AUTH_MISSING via \`require_principal_id\`.

The step is currently pinned to what it actually produces (AUTH_MISSING) rather than forcing a wrong assertion.

Fix: either build a REST real-token harness seam (so REST can exercise a presented-but-invalid token in-process, matching A2A), or rename this Given to match its actual behavior (e.g. "the Buyer Agent has no authentication credentials") and drop the "expired token" framing until the harness seam exists.

### LINKED ISSUE #2070: A2A failed-Task error envelopes are built, attached, then discarded
Found while triaging the 2026-08-24 inline review round on #1575. Verified at `c11f6846f`.

## The production defect

`src/a2a_server/adcp_a2a_server.py:975-982` builds the two-layer AdCP error envelope, clears the task artifacts, appends the envelope as a `processing_error` artifact, and sends the protocol webhook — and then, at `:986`, does `raise _internal_error_for("message processing", e)`, which throws the failed Task away. The envelope is constructed and never reaches the buyer.

## What it costs the test suite

Four sites cannot grade the real A2A wire and grade a harness-built envelope instead:

- `tests/integration/test_format_fetch_transient_errors.py:33` — the `a2a_has_wire=False` parameter exists to name this divergence. With it set, `RATE_LIMITED` / `SERVICE_UNAVAILABLE` are graded on `synthesized_error_envelope`, which the harness builds in the test process. Break the A2A boundary translator for either code and this stays green.
- `tests/integration/test_creative_sync_behavioral.py:351` — an `if transport is Transport.A2A:` branch inside a `_WIRE_TRANSPORTS` parametrization. One error contract, two oracles chosen by transport.
- `tests/integration/test_auth_suggestion_parity.py:60` — `AUTH_REQUIRED` was converted from `assert_wire_error` to `assert_synthesized_error`, and the docstring was edited to drop the wire claim rather than the claim being made true.
- `tests/unit/test_protocol_webhook_ssrf.py:471` — grades a buyer-visible gate (`VALIDATION_ERROR` / `recovery="correctable"` / `field="reporting_webhook.url"`) on the in-process reconstructed exception via `call_impl`; no transport is exercised.

## A second defect blocks the obvious fix

The natural repair for the SSRF test is to route it through `MediaBuyUpdateEnv.call_via(<transport>)`. That method accepts a `Transport` and **ignores it** (`tests/harness/media_buy_update.py:312-326`), dispatching everything through `ImplDispatcher`. It has zero callers today. Routing tests through it as-is would manufacture cross-transport coverage that does not exist — so either give the env real per-transport dispatch or delete the method.

## Scope

Return the failed Task with its envelope artifact (or capture it in `_run_a2a_handler`), then delete `a2a_has_wire` and put both call sites on `assert_wire_error`, remove the transport branch in the creative-sync test, and restore wire grading in the auth-parity test.

Related: #1574 is the downstream half (running the UC-002 layer-separation scenario across transports); this is the production root cause it waits on.

KonstantinMirin:
Closes #1210. Closes #1521. Closes #1598. Closes #2070. Closes #1880. Closes #1882. Closes #1884. Closes #1885. Closes #1886. **Advances** #1592. Also satisfies #1408 (capabilities advertise real tool surface) and #1329 (sync_accounts settings-update mode); closes the `BILLING_NOT_SUPPORTED` half of #1602.

## What this PR is

It started as #1210 — `get_adcp_capabilities` and the accounts surface. It finished as a
rebuild of how this seller handles a request at all.

**Requests, responses, error codes and the order in which a caller is authorized are now one
mechanism each, and every tool in the codebase goes through them.** A transport hands the raw
payload and the headers to a single boundary. That boundary validates the payload into the
tool's declared request shape, resolves the caller once, resolves the account the request
names, honours the idempotency key and stamps the buyer's context onto whatever leaves. A tool
receives a type that already carries its principal and tenant, so nothing below re-checks
authorization and nothing below can reach a header. Responses conform by inheriting the pinned
schema rather than being shaped per handler, and every buyer-facing error field is a function
of the error code through one table rather than text written at a raise site.

**That is what produces the compliance result.** The storyboard runner now passes 30 of the 51
checks that reach a verdict, on MCP and on A2A alike with zero disparity between them — the
highest this agent has scored. It is not the sum of thirty targeted fixes; it is what happens
when the request and response shapes are correct by construction, so conformance stops being
something each tool has to get right on its own.

**The behavioural suite grew with it, and then grew again.** Scenarios are transport-independent:
one Gherkin scenario runs over MCP, A2A, REST and the in-network REST stack and passes on all
of them. On top of that, this PR brings in the UC-026 package-media-buy suite, which had never
graded anything — its scenarios, its step module and its tag set were all written and
maintained while every node xfailed at fixture setup. Wiring it turns several hundred nodes that
reported nothing into roughly four hundred that pass, with the remainder stating precisely which
production gap or harness gap blocks them. That is a large single addition to real coverage.

The original #1210 scope is delivered too, and unchanged in intent: every capability claim is
derived from real configuration and backed by implemented code, so nothing on the wire is
hardcoded fiction.

## How it works now

**Capabilities are honest by construction.** The response is built from live sources: the tenant's adapter config (channels, targeting, creative formats), `tenants.capability_declarations` (a config-time-validated STRICT declaration store — a declaration whose `required_tools` aren't implemented is rejected at config load, not discovered on the wire), and `src/core/billing_policy.py`, the single source both the `account` capability block and `sync_accounts` enforcement read — the claim and the enforcement cannot drift. Unimplementable surfaces are declared honestly absent: signing emits `{supported: false}`, `reporting_delivery_methods` is omitted (blocked on RFC 9421 signing, #1291), and strict-excluded declarations are tracked in #1724.

**Accounts implement the 3.1.1 core contract.** `sync_accounts` supports provisioning and settings-update (`oneOf` modes per `sync-accounts-request.json`), with a `_FIELD_POLICY` table as the one mechanism both arms share, dry-run preview parity (a guard pins that live-create and preview go through the same row builder), account natural-key uniqueness + immutability (DB constraint, race-safe resolve-or-write, repair migrations for pre-existing mangled rows), and per-entry advisory errors routed through the shared `normalize_advisory_errors` lane (which preserves `field`/`suggestion`/`details` and knows the pinned supplement codes, so `BILLING_NOT_SUPPORTED` survives to the wire).

**The behavior is graded, not asserted.** BR-UC-010/BR-UC-011 scenarios run through one transport-independent harness across a2a/mcp/rest (+e2e_rest in network runs); error paths assert on the wire envelope via the sanctioned helpers only — a structural guard bans hand-rolled envelope parsing. The BDD strict-xfail tripwire now *classifies* failures: an xfail whose real cause is a missing step binding or Given-side setup error fails loud as a misclassification instead of masquerading as a production gap (this retroactively surfaced and fixed 9 pre-existing mislabels). The pinned error-code fixture is re-vendored verbatim from spec tag v3.1.1 and pinned by SHA-256, so drift is a red test.

**Architecture boundaries are enforced, not requested.** The repository-pattern guard discovers files by glob over `src/core/tools/` + `src/core/helpers/` (no hand-maintained list to fall out of), with a reverse rule that repository modules may not import SDK protocol-response types; policy logic lives in plain `src/core` modules (`billing_policy.py`, `idempotency_policy.py`); ORM row construction lives behind repository factory methods; `disallow_any_explicit` is enforced per-module on the files this PR owns; a shrink-only ratchet caps local-ticket-id citations in code (GH issues only for new code).

**Transport field parity (scope note).** Enforcing that `sync_accounts`/`list_accounts` request fields are built identically across MCP/A2A/REST (via a shared `select_request_fields()` seam and a new parity guard) meant the guard now scans all tools — and it surfaced five pre-existing wire-conformance bugs in *adjacent* tools where one transport silently dropped a spec field: A2A was dropping `list_creative_formats` disclosure filters, `list_authorized_properties` tags/domains, and `list_creatives` `media_buy_ids` + projection flags; REST was dropping `get_products` `property_list` and the `context` echo on two bodies. These are outside the accounts/capabilities surface this PR is titled for, but are fixed here (≈65 lines in the two transport handlers) rather than allowlisted, because allowlisting known bugs is strictly worse than fixing them. Reviewers of the adjacent tools: this is where those behavior changes come from.

**One interim carry:** `fix: outbound webhook SSRF gate silently rejected the e2e capture server` (originally PR #1860) rides along because #1697 without it fails all six outbound-webhook e2e tests; the whole validator it patches is deleted by the in-flight `feat/secure-outbound-fetch` rewrite, at which point the carry disappears with it.

## Also in this PR: every buyer-facing error derives from one code table

This branch grew a second, larger body of work. It belongs in this PR's history and is described here rather than left to be discovered in the diff.

**A graded wire field is a function of the code, never authored at a raise site.** `CODE_TABLE` (100 entries) is the sole authority for `message`, `suggestion`, and `recovery`. `AdCPError` has no `message` parameter, so no raise site can write buyer-facing text; provenance-bearing text goes to `internal_detail`, which is server-log only and deliberately absent from every serializer. `Error.of()` is the one advisory constructor and likewise exposes no `message`/`suggestion`/`recovery` parameter. Spec grounding: `error-handling.mdx` and the `enumMetadata` recovery classes in `enums/error-code.json` at the pinned 3.1.1; the deleted 150-line drift guard became unnecessary rather than improved, because the copy it guarded is gone.

**An error's `details` is a declared class, not a free-form dict.** `AdCPError` is now generic in its detail shape (`class AdCPError[DetailsT: ErrorDetails]`), so one `__init__` signature gives each error class its own exact type and mypy rejects both a dict and a foreign shape at every raise site. The measured problem this fixes: an AST survey of all 173 details-bearing sites found `AdCPValidationError` alone carrying **30 distinct key sets across 34 sites**, with `validation_errors`, `creative_errors`, `config_errors`, `adapter_errors`, and `violations` all naming "a list of problems" — each built by interpolating structured facts into an f-string and discarding the structure. 47 sites are converted here; the rest are tracked (see the table below).

**The exception owns its code; a shape owns only fields.** A detail shape names no code, which is what lets one shape serve several errors instead of being copied per class — `{package_id, media_buy_id}` serves both `AdCPPackageNotFoundError` and `AdCPGamUpdateError`. `rejected_value` and `accepted_values` follow the pin's canonical rejection-set keys (`core/error.json`, the `details` description, which names them explicitly "rather than seller-specific variants observed in the wild").

**`AdCPInventoryUnavailableError` is merged into `AdCPProductUnavailableError`.** An audit of the four codes claimed by more than one exception class found this the only genuinely redundant pair: same 422, same `PRODUCT_UNAVAILABLE`, same correctable recovery per the pinned `enumMetadata`, and nothing anywhere caught or `isinstance`-checked them apart. The other three groups are real and kept — `AdCPUrlNotAllowedError` has its own `except` clause, the `REFERENCE_NOT_FOUND` trio is a parent plus two children, and `AdCPAdapterError`/`AdCPServiceUnavailableError` split 502 from 503.

**Error verification has exactly one shape.** The `IMPL` pseudo-transport is deleted along with `synthesized_error_envelope` and the `ImplDispatcher` — all three computed an envelope from the same in-memory exception the assertion then read, so a regression in the production boundary translator could not have changed the result. `retry_after` became expressible, the last hand-indexed envelopes are gone, and a structural guard with **no allowlist** (`test_architecture_one_wire_error_assertion.py`) bans the older two-step spelling.

Three things this surfaced that were hiding behind the dict: three tests fed a `MagicMock`'s `.message` into what became the buyer's envelope (a dict took it silently; a typed field refuses it); `retry_after` was being written into `details` at the rate-limit site while already being a first-class field; and `src/adapters/base.py:410` authors four buyer-facing sentences into `details`, the last of them a **suggestion** — the exact field this epic made a function of the code.

### Still open, tracked

| Item | Where |
|---|---|
| BDD feature files bound by no test module, so their scenarios cannot run | #1947 |
| Upstream: SDK `RootModel` blocks an extra-policy override; the 15 error-details models aren't exported | `adcp-client-python#1077`, `#1080` |

Five items that stood here are gone rather than carried. Three are done: the
`details: dict` escape hatch and its migration scaffolding are gone,
`validate_media_buy_request` returns `list[ErrorProblem]` instead of authored text, and the
`issues[]` row described a constructor (`AdCPError.__init__`) that no longer exists. The other two were mis-tracked by me: the A2A discarded-envelope defect is FIXED by this branch and now sits on the Closes line above — `_internal_error_for` and the `processing_error` artifact are gone, and the failure path returns `to_wire(failure.response)` — and an `isinstance`-dispatch item I filed turned out to be invalid, because those sites classify an arbitrary caught exception at a boundary rather than deriving any graded field.

## The BDD suite: one scenario, every transport

The behavioural suite was rebuilt alongside the architecture, and it is what makes the
claims above checkable rather than asserted.

**A scenario is transport-independent by construction.** The same Gherkin runs over MCP,
A2A, REST and the in-network REST stack, and passes on all of them. Scenario text never
names a transport: the scenario says what the behaviour is, the environment says how to
reach it per transport, and guarded helpers decide whether the assertion touched real wire
bytes. Both suites now run clean — in-process at zero failures over thousands of
parametrized items, in-network the same. Full account of the layering:
[BDD harness architecture](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/design/bdd-harness-architecture.md) and
[tests/CLAUDE.md](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/tests/CLAUDE.md).

**Coverage expanded and the scenarios were reconciled against AdCP 3.1.1.** Roughly fifty
feature files carry a few thousand scenarios. A generated feature file is editable here and
several were wrong: they asserted error codes the pinned enum contradicts, and passed for
years because production emitted the same wrong code. Those are corrected in place with the
version cited at the change. Two scenarios demanded behaviour the specification forbids —
one wanted an idempotency response to echo cached state, which `L1/security.mdx` names as
turning key reuse into a read oracle — and grade the refusal instead. The authority order is
written down: pinned spec, then storyboard, then SDK, then production, and production is
never the authority for what a test should expect.

**The Gherkin is consolidated.** Many sentences that differed only in wording now bind to a
single step body, so one behaviour has one implementation rather than one per phrasing. Set-up
collapsed the same way: a single `Given` sentence seeds the tenant and principal a scenario
needs, idempotently, so a scenario that needs a caller says so in one line instead of
repeating a fixture block. Steps are split into reusable and use-case-specific layers, and
dispatch has exactly one writer of the result — nothing hand-stashes wire data.

**Response conformance is checked by one sentence, for any tool.** `the response is compliant
with the <tool> spec` validates the response against the pinned schema for whichever tool the
scenario exercised, with sibling sentences for the error envelope, the webhook payload and
the envelope's status. A scenario no longer hand-rolls a shape assertion, which is what let
the shapes drift before.

**Errors are asserted on the wire, never on a reconstruction.** The harness used to rebuild a
typed exception from a wire response; that covered under half the error classes and collapsed
two that share a code, so a test asserting on the rebuilt object graded the reconstruction
rather than the buyer's envelope. The reconstruction is deleted. Assertions go through helpers
that read the captured envelope and default the recovery semantics from the pinned enum.

**Webhook testing has an architecture, and now a document.** Tests stand up a real
programmable HTTP origin on an ephemeral loopback port, with a TLS sibling serving real
certificates from the repo's generated CA, and an MCP origin over TLS. Deliveries, retry
sequencing and the circuit breaker are exercised against it. None of this was written down
anywhere until this PR: [webhook testing architecture](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/design/webhook-testing-architecture.md).

**Liveness is measured, not inferred.** A tagged scenario that never executed a step used to
count as coverage — the classifier decided a structural question by matching substrings in a
free-text reason, so a reworded message could turn "never ran" into "covered". The verdict now
derives from whether a step body actually executed, a fact the run records. The expected-failure
ledgers only ever shrink, and the storyboard ledger is empty.

**Storyboard conformance: 30 of the 51 checks that grade, on MCP and on A2A alike.** The runner
attempts 300 checks per protocol; 249 skip for reasons that are not verdicts — no test controller
for the storyboard, an unimplemented tool, a failed prerequisite, or not applicable to this seller —
so 51 actually reach a pass-or-fail, and 30 pass. Before this work that number was in the low
teens. The two protocols score identically, with zero disparity between them, which is the
measurable form of "one boundary, many transports": the agent cannot drift on A2A while looking
correct on MCP.

## Documentation

The design documentation was rewritten against this branch rather than patched, because most of
it described mechanisms the rework deleted. Start here:

| To understand | Read |
|---|---|
| How a request reaches business logic, including credentials, identity, accounts and idempotency | [Request lifecycle](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/development/request-lifecycle.md) |
| How to add a tool, and what the registry row derives | [Building a tool](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/development/building-tools.md) |
| The error taxonomy and how a typed error becomes a wire envelope | [Error architecture](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/design/error-architecture.md) |
| The principles the layering follows | [Architecture principles](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/development/architecture-principles.md) |
| The behavioural suite and its transport independence | [BDD harness architecture](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/design/bdd-harness-architecture.md) |
| Webhook and outbound-delivery testing | [Webhook testing architecture](https://github.com/KonstantinMirin/prebid-salesagent/blob/feature/spec-gaps-1210/docs/design/webhook-testing-architecture.md) |

Two documents are new: the error architecture, and the webhook testing architecture for an area
that had none. Several were retired rather than updated — a document describing a deleted
mechanism is worse than no document, and a completed migration plan is not a design. The
registry design folded into the tool-building guide, and every citation to the retired files was
repointed first.

**Credentials are no longer stored in plaintext.** The seller mints a buyer token once, shows it
to the operator once, and stores `sha256(token)` plus a short display prefix; the resolver
hashes what a request presents and looks up the hash. The database holds nothing that
authenticates, and a test cannot read a credential back out of a row. A 256-bit random value is
not a password, so there is no slow hash and the lookup stays an index hit. The migration
converts an existing deployment in place, so credentials already issued to advertisers keep
working. Sandbox and demo credentials are not yet distinguished from production ones — giving
demo keys their own prefix belongs to the sandbox feature and is future work.

## The integration suite is ledgered

50 integration tests are xfailed, listed in `tests/integration/known_failures.txt`.

Their premise changed under them. This PR gives production one path to every implementation:
a transport hands `serve` the raw payload and the request headers, the resolver builds the
identity once, and a tool reads typed models — nothing below the boundary needs a dict, a
header or an identity of its own. The integration layer is the only thing still shaped the
old way, so what these tests fail on is mostly their own setup: injecting an identity
through the wire, mocking a collaborator we own, reading an envelope as a raw dict, seeding
a tenant partially, or naming a field that no longer exists.

Most of what they were protecting is already graded, on all four transports, by the BDD
suite this PR expands — which is green. Deciding per test whether the obligation belongs in
a scenario, genuinely needs integration, or is already covered is follow-up work, not a
blocker on this PR: **epic `salesagent-e30o7`**, one issue per failure cluster.

The ledger is strict and shrink-only: a ledgered test that starts passing fails until it is
graduated deliberately, and `tests/unit/test_integration_ledger_state.py` pins the contents
so entries cannot be added without a decision.

Two entries are not the testing layer's fault. `create_app` composes the test-credential
login blueprint only when `adcp_auth_test_mode` is set, so two template-scanning tests pass
wherever that flag is on and fail wherever it is off — production composition depending on
whether a suite is running. What sits behind it is a first-run bootstrap capability named,
credentialled and gated as a test fixture; **`salesagent-091d8`** asks how first-run admin
setup should work instead. And one entry may be a production defect rather than a stale
test — `list_accounts` answering an unauthenticated REST caller — which is the first thing
its cluster asks about.

## Verification

**Capabilities/accounts work:** full in-network suite (all 7 environments) at head `acdf61ab3` with no failures. The 2 e2e reds (`test_get_products_request_validation`, `test_get_media_buy_delivery_request_schema`) are pre-existing on that box and identical to baseline (a live schema fetch against a gitignored cache, tracked on #1868).

**Error-handling work:** `make quality` green at every commit. A full in-network baseline is running now and will be recorded before the `issues[]` lane proceeds; that number is NOT yet in hand, so treat the error-handling half as locally verified and not yet suite-verified.

**The conflict that kept CI from running is resolved.** The `CI`, `CodeQL` and `Security` workflows are `pull_request`-triggered, so none of them could fire while the PR reported `CONFLICTING` — GitHub builds no merge ref for a conflicted PR, and they had not run since August. `main` is now merged into the branch and pushed, and the PR reports `MERGEABLE`, so those workflows can run against this work for the first time. Treat any check older than that as meaningless: the only ones executing were `check-pr-title` and `ipr-check`, both `pull_request_target`.

## Deferred, and where each lands

Every row below was re-verified against the issue's own acceptance criteria on this
branch, not against whether the originally-cited lines changed. Those two come apart
whenever this rework touched a neighbourhood for an unrelated reason: #1877 read as
fixed by the line check while the defect it names is still there verbatim
(`proven` comes from the 2xx status alone), and two further rows failed the same way.

The pass moved five items off this table and onto the `Closes` line — #1880, #1882,
#1884, #1885 and #1886 — each verified fixed by THIS branch rather than by a merged
sibling (`_boundary.py` does not exist on `origin/main` at all). One row was dropped
because another PR had already closed it. Four rows are partly done and say so,
because a blocker that changes identity is not a blocker that cleared.

| Item | Where it lands |
|---|---|
| `reporting_delivery_methods`, request/webhook signing, and the `identity.key_origins` / `brand_json_url` signing anchors | #1291 (RFC 9421 adoption via `adcp.signing`, assigned) |
| Strict-excluded capability surfaces (e.g. `auto_approve` creative mode, compliance_testing) | #1724 |
| Presence-object capability families (creative extended, governance flags) — scenarios authored, dormant pending steps + production | #1855 (production), #1856 (account-config surface), #1857 (targeting dims) |
| Agency/account grants model | #1615 |
| UC-017 account financials | #1722 |
| `NotificationProofService` still proves nothing: `proven` is set from the 2xx status alone, and the POST body omits `challenge` plus three other required fields, so there is no value to echo and no echo check | #1877 — PR #1757 (open draft) owns the file |
| Raw sessions in a directory no guard scans: `src/services/order_approval_service.py` keeps seven `get_db_session()` sites, outside `_IMPL_DISCOVERY_DIRS` and uncounted by the admin ratchet. The admin-blueprint half of that issue IS fixed here — `accounts.py` routes through `AccountRepository`/`AccountUoW` | #1878 |
| Type erasure on the capabilities modules: both originally-cited sites are fixed and `mypy.ini` gained `[mypy-src.core.tools.capabilities]`. The acceptance also requires `capability_declarations.py` in strict scope, and `from_tenant(cls, declared: Any)` survives | #1879 — PR #1757 (open draft) |
| Make wrong wire-assertion paths fail by construction, and retire the TRANSPORT-BYPASS incentive. Partly done: the `list_accounts` bypass is gone and the allowlist is empty, but the wire-discipline guard's Check B is still name-scoped, so the call-graph blind spot behind five sibling guards remains | #1876 — no PR open; needs the #1740 conftest redesign first |
| `set_adapter_channels` e2e escape hatch — the hatch itself is gone and the scenario left the escape-hatch registry, but it is not in live grading either: `T-UC-010-channel-mapping` is now xfailed on all four transports as dormant, and unledgered | #1871 — the blocker changed identity rather than clearing |
| `CreativeSyncEnv` Given-step helpers (`setup_generative_build`, `set_run_async_result`) | #1887 |
| `_NO_E2E_REST_TAGS` drops a webhook-SSRF scenario from e2e_rest (consolidation task; the ported fix depends on `feat/secure-outbound-fetch`) | #1892 — three of its four bullets are met; the scenario still xfails on e2e_rest |
| Repository race-safe-write naming convention (`AccountRepository.create` routed through `resolve_or_write`; mixin premature until UoW moves callers into real repositories) | #1853 (admin UoW epic) — no PR open |
| `paused=true` on create accepted but not honored (pre-existing) | #1619 |
| Admin UoW migration (pre-existing debt, ratcheted shrink-only — this branch reduces admin raw sessions rather than adding any) | #1853 epic, #1854 sweep — no PR open |

## The fix list is almost certainly understated

The issues named in this description are the ones whose fixes were deliberate. The rework
replaced the request and response path wholesale — one boundary, one validated request shape
per tool, wire models that conform by inheritance, one error table, and delivery that goes
through a single gateway — so a whole class of reported defects is fixed as a side effect of
the shape being right rather than by a targeted change.

Areas where that is very likely true, and where a reported bug probably no longer reproduces:

- protocol version negotiation
- `get_products` responses and their non-conformant fields
- creative format discovery and the format identity contract
- non-conformant responses generally, on every tool, since the response is now the pinned
  model minus a declared omission rather than something each handler shaped
- webhooks that were never delivered, now that delivery and its failure semantics have one
  owner

Establishing which specific reported issues each of these closes means reading the open bug
list against the new behaviour, one by one, and that audit has not been done. The estimate is
a dozen or more beyond those listed. Nothing here should be read as a claim that a particular
issue is closed — it is a statement about where to look, and the reason a reviewer should not
treat the named list as the extent of what this changes.

## Scope corrections (added after review round 3)

**#1592 is advanced, not closed.** Two of its three named sections land here and grade
un-xfailed after the T-UC-010-main split: the account block, and
`supported_pricing_models`. The third, `media_buy.reporting_delivery_methods`, is
undeliverable without RFC 9421 webhook signing and is re-filed against #1291. Claiming
closure would have marked an unbuilt section done.

**Some adjacent wire fixes ship UNGRADED, deliberately.** The UC-001/UC-007 fixes — REST
`get_products` `property_list`/`context`, and A2A `list_authorized_properties` /
`list_creative_formats` — have no test module and no harness, and in UC-001's case no
scenario row either. Building two harnesses to grade fixes that were opportunistic
neighbours of the real work is inflation this PR refuses: the fixes stay because they are
correct, the PR says plainly that they are ungraded, and two tickets carry the harnesses.

**Declaration validator** is a *read-path backstop*, not a write gate; #1856 is the write
gate. The earlier description overstated it.

## A security fix found while reviewing this PR: audit-log forgery on token rotation

This branch introduced one new instance of #2110, on the endpoint that issues tenant
credentials, and fixing it uncovered a second defect underneath that the reorder alone would
have made worse. Both are fixed here.

**The reported bug.** `rotate_token` (`src/admin/blueprints/principals.py`) stacked its
decorators with auditing OUTSIDE authorization:

```python
@log_admin_action("rotate_principal_token")   # outer: runs first
@require_tenant_access()
```

`require_tenant_access` *returns* a redirect or a 401 rather than raising, and
`log_admin_action` sets `success = True` and only flips it when the wrapped function
*raises*. So an unauthenticated `POST /tenant/<t>/principal/<p>/rotate-token` wrote an
`audit_logs` row reporting a completed token rotation — `principal_name` "unknown", the
tenant taken from the URL, the caller's own form keys in `details.request_data` — while
nothing was rotated. Reproduced before the fix: exactly one `success=True` row. The order is
what `src/admin/utils/audit_decorator.py`'s own module docstring already asks for, and this
site contradicted it.

`record_admin_action_failure` is the sanctioned escape for a handler that catches its own
refusal and returns normally. It does not apply: the handler never runs, so no handler code
is in a position to declare the failure.

**The defect underneath, and why a one-line reorder was not the fix.** `log_admin_action`
resolved the tenant from `kwargs` only, and writes its row only `if tenant_id`. Flask hands
URL parameters to the *outermost* wrapper as kwargs, so reading kwargs worked precisely
while the audit decorator was outermost — and yielded `None` once it was not, because
`require_tenant_access` passes the tenant id positionally. Swapping the decorators therefore
turned auditing OFF on the route instead of fixing it:

| State | Unauthenticated POST | Authorized POST |
|---|---|---|
| original order | **1 row, `success=True`** (the bug) | — |
| decorators swapped only | no forged row | **0 rows** — auditing silently dark |
| swapped + tenant from `request.view_args` | no forged row | 1 row, `success=True` |

`request.view_args` holds the URL parameters however the wrapper chain passed them, so it is
the resolution that survives either order.

**This is the finding that matters beyond this PR.** The remaining 54 sites carrying the
inverted order are #2110's, and reordering them *without* the decorator fix would have
disabled admin auditing across the whole admin UI with every test still green. A
refusal-only test — the obvious test to write, and the one originally proposed — reports
state 2 above as a clean fix. Recorded on #2110 with the measurements, together with the
rule that every reordered site needs the authorized-path assertion as well, and a validated
ast-grep rule for the ordering that should land with that sweep rather than before it.

**What is graded.** `tests/integration/test_audit_decorator_auth_order.py` asserts both
directions against the real route: a refusal writes no `success=True` row, and an authorized
rotation both replaces the stored token hash and is audited. The endpoint had no test of any
kind before this. DB access sits in a new `AdminPrincipalEnv`
(`tests/harness/admin_principal.py`) rather than in the test bodies, which is what
`test_architecture_repository_pattern.py` requires.

**Ungraded on the BDD lane, deliberately, and tracked.** This repo has a hand-authored admin
BDD lane (`BR-ADMIN-ACCOUNTS.feature`, transports `integration` + `e2e`) and covering this
there would additionally grade it against a real deployment. That needs a feature file, a
steps module and conftest wiring on top of the env now in place, and is follow-up rather
than scope creep on a branch this size.

**Three other claims from the same review did not survive verification**, and are recorded
so nobody re-litigates them: a suspected new tenant-isolation leak in `append_step_comment`
turned out to have write scope identical to `main` (the real defect is guard blindness,
#1803); a "deleted redactor" referred to functions that exist on no branch in this
repository's history (they live only on an unmerged review branch of closed PR #1575); and
the `ValueError`/`PermissionError` branches that stopped carrying exception text carried no
buyer-facing text to lose — removing it is what the pinned 3.1.1 `transport-errors.mdx`
Security Considerations list requires, so it must not be re-added. A fourth,
`AccountReference1` reaching the buyer in `field` and `issues[].pointer`, does reproduce but
is pre-existing (`main` emits the byte-identical `field`); it is a real RFC 6901 conformance
defect and is recorded on #1996 with a fix shape.


numarasSigmaSoftware:
Flagging an overlap with #1547 (currently in review) on `tests/fixtures/adcp_schemas_pinned/enums/error-code.json` and `tests/harness/transport.py` — two independent, incompatible edits to the same files, verified against both PRs' current heads (`512c0ea11` here, `ce45f9a77` there).

**1. `error-code.json` — two different provenance models for the same file.**

This PR hand-edits the fixture to 92 codes. Independently fetching the real upstream `error-code.json` at the v3.1.1 tag (`467fd93d7`) and diffing: the `enumMetadata.suggestion` text for at least 4 codes here doesn't match upstream —

| Code | This PR's text | Upstream v3.1.1 |
|---|---|---|
| `CREDENTIAL_IN_ARGS` | "...Move the credential out of request args..." | "...Move authentication material or caller-supplied trust material out of request args..." |
| `MEDIA_BUY_NOT_FOUND` | "verify media_buy_id or buyer_ref" | "verify media_buy_id; for legacy correlation use get_media_buys plus context, such as context.internal_campaign_id" |
| `PACKAGE_NOT_FOUND` | "verify package_id or buyer_ref via get_media_buys" | "verify package_id; for legacy package correlation use get_media_buys plus package context, such as context.buyer_ref" |
| `REQUOTE_REQUIRED` | "re-negotiate via get_products in 'refine' mode..." | "adjust the update to stay within the current quote envelope, rediscover products/terms, add packages when available, or create a separate media buy..." |

#1547 takes the opposite approach for the same underlying gap: reverts the fixture's `PINNED_SHA` to the base commit it was actually vendored from, and declares only the two AUTH codes production requires as an explicit, digest-checked supplement — rather than a broader hand-edit that leaves the rest of the file's provenance ambiguous. Both PRs are patching the same "fixture doesn't match what it claims" problem with different, non-composable answers.

**2. `tests/harness/transport.py` — a direct hunk conflict, not just proximity.**

This PR keeps `_pinned_error_metadata()` in place and adds `is_pinned_error_code()` calling it locally. #1547 deletes `_pinned_error_metadata()` entirely, replacing it with an import from a new `tests/helpers/pinned_schema.py` module (`pinned_error_code_metadata`). Whichever merges second will have `is_pinned_error_code()` (if it survives the merge) calling a name that no longer exists in this file.

Separately, both PRs edit `TransportResult.assert_wire_error`'s `require_suggestion` block within a few lines of each other — this PR adds a `field:` kwarg threaded through to `assert_envelope_shape`; #1547 rewrites the suggestion check from a non-empty assertion to strict equality against the pinned enum text via the same `pinned_schema` module. Likely mergeable without a hard conflict marker there, but not obviously correct without checking the two visions actually compose.

Not asking either PR to change course — just flagging so whichever lands second isn't surprised, and so #1687/#1680 (the tracked full re-vendor) has both data points when it picks this file up.

KonstantinMirin:
Before continuing with implementation, I scanned for overlap against 5 in-flight sibling branches  — several findings were inside files already under active rewrite elsewhere, so fixing them here would have meant redoing work in flight and risking conflicts. Deferred as GitHub issues, referenced on the branch that now owns them:

- **F1** (NotificationProofService claims a webhook-control proof it never performs) → #1877, deferred to #1757
- **F3** (raw-session/raw-model sites in `admin/blueprints/accounts.py` + `order_approval_service.py`) → #1878, deferred to #1757
- **F4** (list_accounts BDD steps bypass wire dispatch via `TRANSPORT-BYPASS`) → #1880, deferred to #1728
- **F5** (type erasure on the tenant value) → **partially** deferred as #1879 (the `capabilities.py`/`capability_declarations.py` sites, to #1757); the `adapter_helpers.py`/`billing_policy.py`/`accounts.py:1309` sites (no overlap found) stay on this PR
- **F8** (`error-code.json` fixture pin diverges from `_refresh.py`) → #1881, deferred to #1868

Remaining scope for **this** PR — F2, F5 (rescoped to the 3 non-overlapping sites), F6, F7, F9 (rescoped to its non-deferred citation sites), F10 — tracked in beads (`salesagent-2bpt`).

This comment is here so a future re-review of this PR doesn't re-flag F1/F3/F4/F8 (or the deferred F5 sites) — they're real, they're just tracked as GitHub issues against the branches that already own the files, not against this one.

KonstantinMirin:
## Follow-up filed: #2093

Three commits on this branch fix a buyer-visible A2A defect and stop where the fix stops being a fix and starts being architecture. What was deliberately **not** done here is filed as **#2093**, so the surface of this PR doesn't grow for no reason.

### What landed

| | |
|---|---|
| `5035c0b36` | A2A generative creatives were silently written as **static**. The boundary pre-typed `format_id` into our `FormatId` subclass; pydantic doesn't re-validate a model instance that already fits its annotation, so A2A alone carried a different concrete class, and pydantic v2's class-sensitive equality made the registry match fail — `is_generative` never evaluated, no error. Boundary now emits a wire dict; format identity moved to one matcher. The harness bypass that hid it is gone, so the `a2a` leg frames a real Task. |
| `7ab4256b8` | `get_format` compared a declared `str` to a model (no-`agent_url` branch resolved nothing); `MockCreativeEngine` hashed an unhashable model. Both now compare the id. That matcher migrated onto the **pre-existing** `format_id_identity`, which canonicalizes `agent_url` as `core/format-id.json` requires; four more hand-rolled normalizations moved with it, including the product- and request-side key builders in `media_buy_create`, which are compared to each other. |
| `82117ece2` | Comment citations repointed from the closed #1388 (SDK upgrade) to #2093. |

Also corrected #2011: its recorded cause for the generative-as-static symptom was wrong — removing `CreativeAsset(**c)` did not fix it, and 11 `[a2a]` ids still failed afterwards. Its `idempotency_key` point is still open and wider than recorded (`sync_creatives_raw` has no such parameter at all, so no transport on that entry point honors the retry contract).

### Why #2093 is not in this PR

The work here kept uncovering that the tests were encoding outdated or absent behaviour rather than that the wiring needed a nudge. Following that to its root means: identity onto the model, our `FormatId` as the annotated type, repositories returning models instead of `JSONType` dicts, collapsing four `agent_url` normalizers, and wiring the dormant scenarios. That is a different change with a different risk profile, and none of it is blocked by what landed — every layer added here is additive and deletable in one pass once the model owns its identity.

The finding that makes #2093 worth doing properly, stated plainly: **no BDD scenario is violated by the defects that remain, because the scenarios specifying them do not run.** `BR-UC-020-build-creative.feature` mentions `format_id` on 89 lines and has no test module at all; the 8 `partition`/`boundary` scenarios for update format-compatibility xfail as *"harness not yet wired"*. The behaviour is specified, at the right level, across transports — and never asked.

### Verification

Full in-network run on the last code commit: `exit=0`, **0 failures** across unit / integration / bdd_inprocess / bdd_e2e / e2e / admin / ui. The 38 xpasses are byte-identical to the pre-change baseline, so nothing graduated by accident. `82117ece2` is comments only (73 tests over the touched modules re-run green).

florianerl:
Hi @KonstantinMirin , 

great work on this!  Big sorry in advance if I'm being blind or stupid (or both) here, i have some questions that i wasn' t able to answer from the docs. I have only read the markdowns, not the ~1000 changed code files. I assume the docs are the source of truth your agent is working from, so if two of them would disagree, I am wondering which instruction they are supposed to follow. 

1. Compatibility
[building-tools.md](https://github.com/KonstantinMirin/prebid-salesagent/blob/12bb1bcec7ef421aeab577a0718c5e5cb42231a8/docs/development/building-tools.md?plain=1#L347-L353) says compatibility lives in the DTO as a `mode="before"` validator. A bare domain becomes `brandReference`, old geos or format ids get upgraded here... 
[one-tool-registry-remaining.md](https://github.com/KonstantinMirin/prebid-salesagent/blob/12bb1bcec7ef421aeab577a0718c5e5cb42231a8/docs/design/one-tool-registry-remaining.md?plain=1#L30) however seems to say the opposite: the compat step is a layer before validation, it is currently [ABSENT], and it should come back as one designed seam or not at all.... 
The [same doc ](https://github.com/KonstantinMirin/prebid-salesagent/blob/12bb1bcec7ef421aeab577a0718c5e5cb42231a8/docs/design/one-tool-registry-remaining.md?plain=1#L111) also says it does not belong on the DTO either, because a BeforeValidator would make the model accept a shape AdCP does not define... 
On this commit, if a buyer sends "brand": "abc", which of MCP / A2A / REST accept it? And is that the intended final state?

2. idempotency
what is the payload hash protecting? On this branch the key is live for the four tools that declare it. Same key → lookup → replay or IDEMPOTENCY_CONFLICT. The work does not run again in either case.
Do i undestand this correctly as: 
-- The key stops a retry from running twice.
--  The hash prevents replaying the wrong work (same key, different buy).
--  The strip avoids a false conflict when a retry carries fields _impl never sees
Is that the rule? Then the line [building-tools.md](https://github.com/KonstantinMirin/prebid-salesagent/blob/12bb1bcec7ef421aeab577a0718c5e5cb42231a8/docs/development/building-tools.md?plain=1#L410)  about a buyer “defeating” idempotency with `cache=abcd` is about a false conflict, not a second execution, correct?

KonstantinMirin:
Hi @florianerl 
I really appreciate you spending time on digging through this. Too many changes, I know. Believe it or not, they all boil down to one thing: simplify and unify the flows around system boundary and what is relates is "collateral damage control" that prevents the SalesAgent from becoming broken because of the pre-existing layers of legacy logic.
To your questions
1. I will clean up the documents, but building-tools.md is the authoriatative. The final design _allows_ for any compatibility logic to live in beforeValidate and other extension points but intentionally ships *none* today to keep the flow simple. Similarly to other things like X-Adcp-Token which is declared "legacy" but no one actually uses it and I dropped it too. It is easlier to add properly designed compatibility features on top of the clean structure than to do both at the same time. So the intended state is "it passes 3.1.1 and doesn't necessarily pass anything before".
2. You are reading it absolutely correclty. Indeed, the difference is that by the spec `&cache=123` should be treated as a different request and lead to the new run. We check if the request _after_ stripping and validation is correct, so we will not react to the field we don't know about. For us, both requests are identical so re-run won't do anything differently, so we just replay. I think it is a feature but the real reason that this decision allows us not to thread through any data or logic through the layers. If this becomes problematic, then each transport will have to produce its request hash and pass it through to invoke_tool. I saw that as more complicated with very little added value and the behavior change in a very narrow case, so chose design simplicity instead.

Thanks to your question, I found the last bit of "compatibility" in there - format->FormatId which I'm killing now (and that uncovers a layer of buggy behavior, similarly to other "compat" logic I removed before).

florianerl:
Hi Constantine 

no worries at all!  I’m still trying to get a better picture of the project and reading through this cleared a bit more of it.

I’m generally in favor of anything that makes this project easier to live with. If this were a personal project I’d say push through 🙈: anything that helps is already a win. I just worry a bit about people who already run this in production, or who forked this and rely on current behaviour staying stable.

I’ll keep reading.

numarasSigmaSoftware:
@KonstantinMirin 

The shared boundary here is a good place to raise a layering question, since it is now the one path every transport takes:

Today SDK-derived types cross the boundary unchanged and reach three places that outlive the SDK pin: persisted payloads (media_buy.py stores req.model_dump as raw_request; idempotency_attempt.py stores response-model dumps and re-validates them on replay; workflow.py stores request payloads the approval path reads later), the adapter interface (adapters/base.py takes CreateMediaBuyRequest), and every _impl signature.

I measured what that costs by bumping the pin to 7.0.2 in a scratch worktree of this branch: 700 unit failures, about 85% of them the SDK's creative-identity rework (FormatId -> LegacyFormatId, agent_url becoming str, Product rejecting legacy format ids) landing in products, adapters and stored format ids at once. The spec itself moved almost nothing between 3.1.1 and 3.1.15; the churn is the SDK's API.

Proposal: keep the SDK-grounded DTO as the announced and validated shape, exactly as the registry enforces, and add a translation step inside invoke() after validation, so implementations, repositories and adapters receive application-owned models. Structural SDK changes then land in the SDK pin and the translation module; new protocol behavior still needs implementation. This does not conflict with CLAUDE.md pattern #1, which governs the wire schema, not the persistence or adapter contracts.

KonstantinMirin:
Thanks. 
0. 80% of those tests fail on this branch today, that's what I am busy with now.
1. Media buy has to be refactored. It is the scary 2500-lines piece that we disassemble piece by piece 
2. Creative-related tools have 1.3 shape. I have a separate analysis that I will post as RFC later. It has to be heavily refactored 
3. I think that you are right about persisting models. There are other problems with it- schema does not always match the request: tenant, webhooks, columns vs jsonb etc. so PersistenceModel.from(RequestModel) is a correct pattern that decouples these parts.

So: the idea is good, we have to follow it, it needs a separate RFC and I am happy if you want to take a first stab.

In support of that idea, done right, this .fromRequest() allows to keep only things we want and prevent unknown data to be ever persisted.

I am a bit concerned about duplicating the fields but I don't see another way around it.

KonstantinMirin:
@numarasSigmaSoftware here is the RFC for creative-related functionality. Currently we have it broken AF regardless of 1721's work. It's just written for AdCP 1.3 a year ago and then was patched many times. All the issues with products not being delivered or metrics or ad server not saving proerly - all from here.
https://github.com/prebid/salesagent/issues/2236

ChrisHuie (commented):
Four items, all should-fix. One is a behaviour regression against main; the rest are on the same root cause or are documentation that now contradicts the code. Each is anchored as a thread below.

## 1. A budget breach answers INTERNAL_ERROR / transient / 500

`src/core/tools/financial_validation.py:31` raises `exc_type(details={"reason": reason})`. `details` is a raw dict, and `_details_to_wire` (`src/core/exceptions.py:153`) now calls `details.to_wire()` with no Mapping branch, so the failure builder raises `AttributeError` before it can produce an envelope.

The typed error is correct when raised — `BUDGET_EXCEEDED` / `correctable` / 422 — and the audit record logs it correctly. The buyer then receives:

```
HTTP 500
{"code": "INTERNAL_ERROR", "recovery": "transient",
 "suggestion": "Retry with backoff; if it persists, report it to the seller's operator"}
```

`recovery: transient` tells the buyer to retry a request that can never succeed. The pinned `core/error.json` uses "budget too low" as its own example of `correctable`, and names `recovery` the normative carrier receivers classify on. The diagnostic naming which package and which limit is discarded with the details block.

Reproduction: a tenant with `CurrencyLimit.max_daily_package_spend` set, and a `create_media_buy` whose package budget exceeds it. Confirmed on REST, MCP and A2A, with and without `ADCP_TESTING`, and with `ENVIRONMENT=production`. The repo's own `TestMaxDailySpendExceeded` cases fail at `exceptions.py:153` for the same reason.

This is a regression: at the merge base `details` was declared `dict[str, Any] | None` and this helper passed the message positionally, so the path emitted `BUDGET_EXCEEDED` / 422.

Live call sites, all passing `exc_type` explicitly: `media_buy_create.py:2488`, `:2532`; `media_buy_update.py:656`, `:789`. The two legacy-mode sites (`:2504`, `:2548`) are unreachable — entering legacy mode requires every package budget to be falsy, and `get_total_budget()` sums only package budgets, so `validate_budget_positive` at `:2158` rejects first with a correctly-rendered `BUDGET_TOO_LOW`.

Fix: both classes already declare their shape — `AdCPSalesAgentError[BudgetDetails]` — and `BudgetDetails` carries `requested_budget` / `current_spend` / `budget_limit`, already used in the GAM adapter. Parameterizing the parameter closes the hole at the type level:

```python
exc_type: type[AdCPSalesAgentError[BudgetDetails]]
```

mypy currently reports `Success` on this file; with that annotation it reports `Argument "details" ... incompatible type "dict[str, str]"; expected "BudgetDetails | None"`. The default then has to go, which is fine — all six call sites pass `exc_type` explicitly. With the typed details in place the same request returns 422 / `BUDGET_EXCEEDED` / `correctable` and keeps the diagnostic.

## 2. The failure builder can raise, and a raise there has no envelope

`failure_response` (`src/core/tools/_boundary.py:185`) is invoked from inside the boundary's own `except Exception` at `:326`, so anything it raises propagates past `serve()`. Confirmed on all three transports: the `AttributeError` above escapes `serve()` rather than leaving as `AdcpFailure`.

Each transport's catch-all then converts it, so this degrades rather than 502s — but all three degrade to the same wrong answer, and two of the catch-alls describe a case this is not. `src/core/main.py:383` says "A tool's own failure never reaches here"; `src/a2a_server/adcp_a2a_server.py:302` says "Raised before any tool ran". Both are reached by a tool's own failure, and both then build the record with no identity and no echo.

One consequence is separately visible: the buyer's `context` is dropped. Same endpoint, same request shape, only the error differs —

```
budget breach    -> 500, INTERNAL_ERROR,    context: null
unknown product  -> 404, PRODUCT_NOT_FOUND, context: {"buyer_trace": "..."}
```

`_served` documents the opposite invariant: "a fresh run, a replayed one and a failure all pass through it".

Fix: make the renderer total — log and return `None` for a non-`ErrorDetails` value — and/or refuse one at construction, the posture `__new__` already takes for the code.

## 3. `raise_mapped_adcp_error` is the second dict site, with no callers

`src/core/helpers/adapter_helpers.py:102` raises `error_class(details={"agent": agent_label}, ...)` and fails the same way when rendered.

It has zero callers. Three statements disagree with that: the comment at `:40` says "the registries still catch the SDK's own ADCPError and delegate the mapping here" — neither registry catches `ADCPError` or imports this module; the docstring at `:47` names two callers that do not exist; and `outbound_error_mapping.py:31` says the function "was deleted". It is absent from main and present here.

Fix: delete it. If it returns, it needs a typed details class and a caller in the same change.

## 4. Two class docstrings state the inverse of what ships

`src/core/exceptions.py:1115` — `AdCPMediaBuyRejectedError` documents "(422 -> POLICY_VIOLATION) ... recovery=correctable so the buyer can adjust the request and resubmit". It emits `MEDIA_BUY_REJECTED` / terminal / "Do not retry this buy; contact the seller about the decision".

`src/core/exceptions.py:1086` — `AdCPBulkUpdateError` documents "Recovery=transient (inherited)". It emits `PARTIAL_FAILURE` / correctable, and "inherited" names a mechanism this change removes.

Both sentences are byte-identical to main, where both were true — `ERROR_CODE_MAPPING` mapped `MEDIA_BUY_REJECTED` to `POLICY_VIOLATION` there. With the class-level recovery knob gone, the docstring is the only per-class statement of a raise site's recovery contract in `src/`, and both classes are raised in production. A sweep of the file's 30 recovery-bearing docstrings found only these two.

## Not tested

The A2A JSON-RPC serialization round trip — the dispatch escape and the `InternalError.data` body were driven, the protobuf transport framing was not.

src/core/tools/financial_validation.py
```diff
@@ -6,30 +6,29 @@
...
     ceilings — so the wire code reflects the failure kind.
+
+    ``reason`` is the validator's first-party diagnostic, not wire text: the
+    buyer-facing sentence comes from the code's table entry, and the diagnostic
+    travels under ``details`` where it is machine-readable.
     """
-    if message:
-        raise exc_type(message, context=context)
+    if reason:
+        raise exc_type(details={"reason": reason})
```
ChrisHuie: `details` is a raw dict here, and `_details_to_wire` (`src/core/exceptions.py:153`) calls `details.to_wire()` with no Mapping branch — so the failure builder raises `AttributeError` and the buyer gets `INTERNAL_ERROR` / `transient` / 500 instead of `BUDGET_EXCEEDED` / `correctable` / 422.

Driven on REST, MCP and A2A, with and without `ADCP_TESTING`, and under `ENVIRONMENT=production`. Regression: at the merge base this helper passed the message positionally and `details` was declared `dict[str, Any] | None`, so the path emitted `BUDGET_EXCEEDED` / 422.

Both error classes already declare `AdCPSalesAgentError[BudgetDetails]`, and `BudgetDetails` carries `requested_budget` / `current_spend` / `budget_limit`. Annotating the parameter `exc_type: type[AdCPSalesAgentError[BudgetDetails]]` makes mypy flag the dict (it reports `Success` on this file today). The `AdCPValidationError` default then has to go — all six call sites pass `exc_type` explicitly, so nothing depends on it.

See item 1 of the review for the full trace.

src/core/tools/_boundary.py
```diff
@@ -0,0 +1,399 @@
...
+    request and a resolved caller -- so only ``_failed`` passes them.
+
+    The ORIGINAL exception goes to the recorder: the body carries no exception text (AdCP 3.1.1
+    transport-errors.mdx, Security Considerations), so the server-side record is the sole
+    answer to what broke. ``adcp_error_for`` types it -- an untyped ValueError is a
+    VALIDATION_ERROR, a PermissionError a PERMISSION_DENIED, anything else an INTERNAL_ERROR --
+    and that answer does not depend on which transport is asking.
+    """
+    record_boundary_error(protocol, operation, exc, identity=identity)
+    return _served(echo, AdcpErrorResponse.of(adcp_error_for(exc)))
```
ChrisHuie, line 198: This is reached from the boundary's own `except Exception` at `:326`, so anything it raises propagates past `serve()` rather than leaving as `AdcpFailure`. Confirmed on all three transports with a malformed `details` block.

Two consequences: the transports' catch-alls all degrade to the same wrong answer, and the buyer's `context` is dropped because `_served` is bypassed — which contradicts its own "a fresh run, a replayed one and a failure all pass through it".

The catch-alls at `src/core/main.py:383` and `src/a2a_server/adcp_a2a_server.py:302` both state that a tool's own failure cannot reach them; it can.

Making the renderer total (log and return `None` for a non-`ErrorDetails` value) would keep the boundary from ever raising. See item 2 of the review.

src/core/helpers/adapter_helpers.py
```diff
@@ -2,152 +2,348 @@
...
+            break
 
-        tenant = get_current_tenant()
+    logger.error("%s for %s: %s", failure_mode, agent_label, exc.message)
+    # The mode is an operator label, i.e. prose — it belongs in the log, not in
+    # details. Four lines above, this file's own comment says a mode needing its own
+    # buyer sentence needs its own AppErrorCode, "not a message argument"; parking
+    # the sentence in details would be the same thing by another route. The SDK
+    # exception itself is the cause, logged with its traceback by the boundary.
+    raise error_class(details={"agent": agent_label}, internal_detail=exc) from exc
```
ChrisHuie: Second dict-`details` site, same failure when rendered — driven: `AdCPServiceUnavailableError` with `details={'agent': ...}` raises `AttributeError` in `AdcpErrorResponse.of`.

This function has zero callers. The comment at `:40` says the registries still delegate here, but neither registry catches `ADCPError` or imports this module; the docstring at `:47` names two callers that do not exist; and `outbound_error_mapping.py:31` states the function was deleted. It is absent from main and present here.

Deleting it matches what main already did. See item 3 of the review.

src/core/exceptions.py
```diff
@@ -1113,191 +972,187 @@ class AdCPIdempotencyExpiredError(AdCPConflictError):
...
     detail (package_id, line_item_id) belongs in ``details`` as data.
     """
 
-    _default_error_code: ClassVar[str] = "GAM_UPDATE_FAILED"
+    _code: ClassVar[ErrorCodeT] = AppErrorCode.AD_SERVER_UPDATE_FAILED
 
 
-class AdCPMediaBuyRejectedError(AdCPError):
+class AdCPMediaBuyRejectedError(AdCPSalesAgentError[RejectionReasonDetails]):
     """The seller declined the media buy (422 → POLICY_VIOLATION).
```
ChrisHuie: This docstring documents `(422 -> POLICY_VIOLATION)` and `recovery=correctable so the buyer can adjust the request and resubmit`. Driven, the class emits `MEDIA_BUY_REJECTED` / **terminal** / "Do not retry this buy; contact the seller about the decision".

The sentence is byte-identical to main, where it was true — `ERROR_CODE_MAPPING` mapped `MEDIA_BUY_REJECTED` to `POLICY_VIOLATION` there. `AdCPBulkUpdateError` at `:1086` has the same problem in the other direction (documents transient, emits correctable, and "inherited" names a mechanism this change removes).

With the class-level recovery knob gone, these docstrings are the only per-class statement of a raise site's recovery contract in `src/`. A sweep of the file's 30 recovery-bearing docstrings found only these two. See item 4 of the review.

KonstantinMirin:
All four review items are fixed. Items 1 and 2 turned out to be one defect with one cause, so
the fix is smaller than the review proposed — and the review's own suggested remedies were
the wrong shape, for a reason worth recording.

Commits: f442b499e, 7525aefc8, 2bd048b68, ef9c081d5, 9cc20f16d.

## 1 and 2 are one bug: an annotation that erased a type parameter

Confirmed the reproduction exactly as filed. A budget breach raises
`BUDGET_EXCEEDED / correctable / 422`, the audit record logs that correctly, and the buyer
receives `500 INTERNAL_ERROR / transient`.

The cause is one annotation. `raise_if_validation_failed` declared
`exc_type: type[AdCPSalesAgentError]` with the parameter dropped, which erases `DetailsT` to
`Any`, and `Any` accepts the dict. `docs/design/error-architecture.md` already states the
enforcement — "mypy rejects both a dict and a foreign detail class at every raise site" — and
that holds at every site that spells its detail type. Two sites did not.

```python
exc_type: type[AdCPSalesAgentError[BudgetDetails]],   # no default: all six callers name their class
requested_budget: Decimal,
budget_limit: Decimal,
```

mypy then reports the dict at the raise site. The diagnostic sentence is logged rather than
carried: buyer text comes from `CODE_TABLE`, `internal_detail` takes an exception, and no
`ErrorDetails` class has a free-text field.

### Neither of the two remedies the review suggested belongs in the tree

The review proposed making the renderer total "and/or" refusing a details block at
construction, "the posture `__new__` already takes for the code". I implemented both, then
removed both. Recording why, because both are plausible and both are wrong:

* **A `__new__` refusal duplicates the type system and contradicts the document.**
  `error-architecture.md` enumerates exactly what `__new__` refuses — a bare base
  construction, and an `error_code=` on a class that already declares one. A third branch
  made that paragraph false. The dict got through because the type system was not consulted,
  not because it was insufficient.
* **A `try` around the failure builder institutionalises the damage.** The real harm was the
  server log saying `BUDGET_EXCEEDED` while the buyer received `INTERNAL_ERROR`. A catch there
  produces exactly that disagreement, by design, for every future renderer bug. It was also
  incoherent as written: `record_boundary_error` runs before the `try`, so the path most likely
  to raise stayed uncovered.

The structural observation in item 2 is correct and stands as a comment rather than a
mechanism: `_failed` runs inside the boundary's own `except Exception`, so a raise there is
past the catch. The answer is to make the raise impossible upstream, which the type parameter
does.

## 3: deleted, and it was contradicted three ways

`raise_mapped_adcp_error` had zero callers, is absent from `main`, and three statements
disagreed with that at once — a comment claiming both registries delegate to it (neither
imports it), a docstring naming two callers that do not exist, and
`outbound_error_mapping.py` already describing it as deleted while it was still in the tree.
`raise_mapped_outbound_error` is what the registries call. Deleting it also removed the second
of the two dict-details sites.

## 4: corrected, and one was wrong three times in one sentence

`AdCPBulkUpdateError` said "Recovery=transient (inherited)" where the table says
`correctable`, and no class-level recovery knob remains to inherit from.

`AdCPMediaBuyRejectedError` was worse than filed: it claimed the code was `POLICY_VIOLATION`,
the recovery `correctable`, and the wire code `POLICY_VIOLATION`. It emits
`MEDIA_BUY_REJECTED / terminal / 422`. All three were true on `main`, where `ERROR_CODE_MAPPING`
collapsed the code.

## Why the defect shipped: three markers, none with the real reason

This is the part the review did not reach, and it matters more than the annotation.

Scenarios demanding `BUDGET_EXCEEDED` on the wire already existed, and their `Then` asserts it
through `assert_wire_error` against the real envelope. They never ran: **48 items, 48 xfailed,
0 executed, 6.4 seconds.** Three markers covered the same scenarios, and each gave a different
wrong reason:

| Marker | Claimed | Actual |
|---|---|---|
| `uc002-not-wired` ENV_ROUTES catch-all | "harness not yet wired" | dispatched nothing |
| `_UC002_VALIDATION_XFAIL` | "production raises plain ValueError" | production raises the correct typed error; reason expired |
| `T-UC-002-ext-k` | "stale spec, pending upstream regen" | the pin agrees with production; the scenario was wrong |

Removing any one alone leaves them as dormant as before. All three are gone: the two outlines
moved onto the `uc002-ext` row whose seed they already needed, and the feature was corrected in
place per `tests/CLAUDE.md` (`3.1/enums/error-code.json` separates the two codes by direction in
`BUDGET_EXCEEDED`'s own text, and a daily ceiling is not a minimum).

**Graded, and the grading was verified by breaking it.** Reintroducing the defect:

```
9 failed, 18 passed
AssertionError: Expected error code 'BUDGET_EXCEEDED', got 'INTERNAL_ERROR'
```

The three breach rows on a2a, mcp and rest. The 18 that pass are the below-cap/at-cap/no-cap
rows, which raise nothing. With the fix restored: 27 passed.

## Carried in the same branch, from the same root cause

* **`AdCPNotCancellableError`.** A ledger entry diagnosed a code-specialization gap correctly
  and then set an unsatisfiable exit condition — "graduates when an issue lands" for what is
  three lines at the guard it cites. The pin reserves `NOT_CANCELLABLE` for a refused cancel;
  production emitted the generic `INVALID_STATE` for every mutation. The code also had no
  subclass, so a fixture named it on the base class, which is the one way to put a code on the
  wire with nothing bound to it. It also had no `_HTTP_STATUS` row and answered the 500 default,
  which that scenario's own "should NOT be a 500" line refuses; it is 410 now, in the band whose
  comment is this code's sentence. Measured as `XPASS(strict)` before the marker came out.
* **The setup-gate patch is gone.** `validate_setup_complete` sat in
  `MediaBuyCreateEnv.EXTERNAL_PATCHES` stubbed to return `None`, while the live e2e_rest server
  enforced it: one scenario, two productions, which BDD rule 1 forbids by construction.
  Removing it left 96 of 180 UC-002 scenarios failing `CONFIGURATION_ERROR` — not a production
  defect, a half-seeded tenant. Seeding the rows the gate grades, all 180 pass with the gate
  real on every transport.
* **One e2e_rest failure, found and fixed rather than ledgered.** Making the outline live
  exposed it on a fourth transport with a real adapter, where a row asked for 9,999,000
  impressions against the mock ad server's 1,000,000 goal ceiling. The row's subject is the
  daily figure, and `daily = budget / flight_days` is reachable two ways — only one of them
  asks anything of the total, which is what downstream limits see. `9999` over one day is the
  same daily figure at 999,900 impressions. Its sibling survives by exactly one impression:
  10,000 at $10 CPM is precisely 1,000,000, and the adapter's test is `> limit`.
* **The step that hid it.** `_assert_validation_pass` reported `ctx["error"]`, whose repr on a
  wire transport is the code and nothing else. Four candidate causes were eliminated by
  inspection and none was the real one; the envelope named it in one line. It now prints
  alongside.

## CLAUDE.md changes
Added instructions to point agents to the Error handling patterns we have now.

ChrisHuie (approved):

Issue #1726, opened 2026-07-27
UC-004: seller attribution default not implemented — production echoes the buyer's requested window (BR-092)


KonstantinMirin:
Split out of #1418. One of two `e2e_rest` ledger entries that survived the Wave 3 triage as genuine production gaps but were never filed anywhere.

## Gap

The seller does not implement an attribution default — production echoes the buyer's requested `attribution_window` back instead of applying and reporting the seller's own default.

BR-092 requires the seller to echo the **applied** window, so echoing the request verbatim is only correct when the seller has no default of its own to apply.

## Evidence

`tests/bdd/e2e_rest_known_failures.txt`, block "explicit inline-xfail production gaps (2)":

```
tests/bdd/test_uc004_deliver_media_buy_metrics.py::test_seller_ignores_attribution_request__returns_platform_default[e2e_rest]
```

The scenario is already inline-xfailed in the step body, so this is a known, deliberately-parked gap rather than a discovery.

## Relationship to #1462

Adjacent, not identical. #1462 is about `attribution_window.post_click` being **dropped** on the request path so the INV-5 validator never fires. This issue is about the seller not applying its own default on the **response** path. They meet at the same BR-092 echo requirement and are likely worth fixing together.

## Done when

- the seller applies and echoes its platform default when the buyer's request does not pin one
- the scenario passes on `e2e_rest` and the inline xfail is removed, not re-pinned
- the ledger entry is removed from `tests/bdd/e2e_rest_known_failures.txt`

KonstantinMirin:
Flagging something you'd hit on day one of this, found while closing #1462 next door.

## The scenario can't currently detect its own fix

Both Givens are inert. `supports_attribution_windows` has two writers and zero readers:

```
tests/bdd/steps/domain/uc004_delivery.py:654    ctx["supports_attribution_windows"] = True
tests/bdd/steps/domain/uc004_delivery.py:660    ctx["supports_attribution_windows"] = False
```

Nothing in `src/` reads that key. So `the seller supports configurable attribution windows` and `the seller does NOT support configurable attribution windows` establish byte-identical production state — the two legs are the same test. Three scenarios rest on it: BR-UC-004 feature `:588`, `:599`, `:625`.

[...]

PR #1728, opened 2026-07-27
fix(bdd): grade the UC-004 attribution echo on the wire and retire three false-green oracles

### LINKED ISSUE #1462: Delivery request path drops attribution_window.post_click in-process → BR-092 INV-5 campaign validation never fires
Found while building the clean BDD error-assertion reference for attribution_window (PR #1420 follow-up).

**Symptom:** `get_media_buy_delivery` with `attribution_window={"post_click": {"interval": 2, "unit": "campaign"}}` (no `model`) **succeeds** instead of being rejected. The response echoes `attribution_window=AttributionWindow(post_click=None, model=last_touch)` — i.e. `post_click` is gone.

**Why it matters:** the INV-5 validator (`_validate_attribution_window` in `src/core/tools/media_buy_delivery.py`) checks `req.attribution_window.post_click`; if it's `None` it can't enforce BR-RULE-092 INV-5 ("unit=campaign ⇒ interval must be 1"). So `campaign interval=2` is silently accepted on the in-process a2a/mcp/rest transports.

**Localization:** the request schema parses it fine in isolation — `GetMediaBuyDeliveryRequest(attribution_window={'post_click': {'interval': 2, 'unit': 'campaign'}})` preserves `post_click`. So the loss is in the request-building/echo path (likely the REST endpoint rebuilding the request from the body, or the a2a/mcp harness construction, or the seller-echo dropping post_click when `model` is absent). Worth checking whether production also drops a buyer-supplied post_click when model is omitted (BR-092 says the seller must echo the APPLIED window).

**Status in PR #1420:** `unit=campaign with interval=2` is parked as a tracked xfail in the attribution_window reference; the other three invalid rows (interval=0 / unit=weeks / model=last_click) pass the clean assertion. This issue tracks closing the campaign gap.

### LINKED ISSUE #1739: test(bdd): 10 parallel e2e_rest entries are mock-injection artifacts, not genuine gaps — realize the setup or declare E2EUnsupportedSetup
Split out of #1418. These are the last 10 entries in `tests/bdd/e2e_rest_known_failures.txt` with no GitHub issue. Tracked in beads as `salesagent-clyq`, but that id does not resolve for anyone outside the repo owner's machine, which is why this exists.

## What they are — not production gaps

All 10 appeared on 2026-07-09 when `perf/parallelize-test-suite` enabled parallel `e2e_rest` (`E2E_PER_WORKER`). They are mock-injection artifacts: the scenario's Given seeds state **in-process**, and the separate HTTP server the request actually reaches cannot see it. Nothing is wrong with production.

Three injection surfaces:

- **UC-004** — `set_adapter_response` (delivery numbers)
- **UC-005** — `set_registry_formats` (creative formats)
- **UC-018** — injected cross-principal creatives

## Why listing them as known failures is wrong

`e2e_rest_known_failures.txt` is scoped, per its own header, to "GENUINE production / harness gaps only" — entries that fail for a real reason and need a code change. These fail because the setup is structurally unavailable over HTTP, which is a different category with an existing mechanism: an env-level `E2EUnsupportedSetup` declaration raised during harness realization in `tests/harness/`, surfaced as xfail by the conftest report hook rather than listed here.

Leaving them in the ledger inflates the genuine-gap count and hides that the ledger is otherwise nearly clean.

## Entries

```
tests/bdd/test_uc004_deliver_media_buy_metrics.py::test_breakdown_complete_not_truncated__truncation_flag_set_false[e2e_rest]
tests/bdd/test_uc004_deliver_media_buy_metrics.py::test_breakdown_truncated_by_limit__truncation_flag_set_true[e2e_rest]
tests/bdd/test_uc004_deliver_media_buy_metrics.py::test_buyer_requests_supported_dimension__seller_returns_breakdown[e2e_rest]
tests/bdd/test_uc004_deliver_media_buy_metrics.py::test_multiple_dimensions_requested_simultaneously[e2e_rest]
tests/bdd/test_uc005_discover_creative_formats.py::test_baseline_list_creative_formats_response_carries_format_id_objects_with_agent_url_and_id[e2e_rest]
tests/bdd/test_uc005_discover_creative_formats.py::test_format_id_roundtrip__list_creative_formats_returns_the_same_format_object_that_get_products_advertised[e2e_rest]
tests/bdd/test_uc005_discover_creative_formats.py::test_format_id_with_agent_url_pointing_at_a_thirdparty_creative_agent_is_reported_as_observation_not_failure[e2e_rest]
tests/bdd/test_uc018_list_creatives.py::test_brrule034_inv1_counter__crossprincipal_creatives_never_visible[e2e_rest]
tests/bdd/test_uc018_list_creatives.py::test_brrule034_inv1_holds__query_always_scoped_by_principal[e2e_rest]
tests/bdd/test_uc018_list_creatives.py::test_list_creatives_filtered_by_concept_ids_returns_only_creatives_in_that_concept_carrying_concept_id_and_concept_name[e2e_rest]
```

## Preferred fix — realize the setup, don't just reclassify

Reclassification is the floor, not the ceiling. #1418's core invariant is that the harness realizes setup **intent** through the real source-of-truth surface per concept, and two of these three surfaces have one:

- UC-004 delivery numbers → write the `DeliverySimulationConfig` row the Mock adapter reads from the server DB
- UC-018 cross-principal creatives → seed the rows in the server DB
- UC-005 formats → the checked-in reference-format fixture both the `ADCP_TESTING` path and the harness read

Where the setup can be realized, realize it and the scenario graduates. Reclassify to `E2EUnsupportedSetup` only what genuinely cannot be reached over HTTP — and say which is which in the PR.

## Done when

- each of the 10 is either graduated (passes in-network) or carries an env-level `E2EUnsupportedSetup` declaration
- none of them remain listed in `e2e_rest_known_failures.txt`
- the ledger contains only genuine production gaps, tracked by #1582, #1318, #1319, #1726, #1727

### LINKED ISSUE #1751: BDD assertions nested under a presence-guard for the artifact they grade: absence passes silently (~10 sites need mutation triage)
> **Measured on:** `fix/uc004-delivery-breakdowns` @ `49a8abcbe` (PR #1728).
> Counts and line numbers below are from that tree. On `main`, or on another in-flight
> branch, they will differ — please re-measure before treating any figure as current.

---

Split out of the UC-004 BDD masking cleanup on `fix/uc004-delivery-breakdowns`.

## The defect

An assertion nested inside `if <expr>:` with no `else`, where the guard tests for the **presence of the very artifact the assertion grades**. When the artifact is absent the assertion never runs — so the regression the row exists to catch is exactly the one it cannot see.

## The instance that proved it is worth chasing

`_assert_valid_content` in `uc004_delivery.py` had:

```python
daily = getattr(pkg, "daily", None) or getattr(pkg, "by_day", None)
if daily is not None:
    assert isinstance(daily, list), ...
```

**Neither attribute exists on `adcp.types.ByPackageItem`** — the field is `daily_breakdown`. So both `getattr` calls returned their `None` default by construction, the guard was unconditionally false, and the assertion was unreachable: dead code no production change could ever reach. Mutation proof: `assert False` inside the guard body left all 18 daily-breakdown rows green.

Repairing it immediately surfaced the production gap it had been hiding (#1319 item C5: `include_package_daily_breakdown=true` accepted but never honoured).

## Census and triage

**37 guarded-assert sites** in `tests/bdd/` (AST: `ast.If` with no `orelse` whose body is entirely `ast.Assert`, inside `then_*`/`assert_*`/`_assert*`). They are NOT all defects — the classification that matters is **what the guard tests**:

- **guard on the EXPECTATION** (`if require_suggestion:`, `if seeded_count is not None:`, `if key in expected_data:`) — legitimate. The guard selects which obligation applies; it cannot hide a production artifact. **12 sites.**
- **VALUE guards** (`if len(...) > 1`, `if status == "rejected"`, `if isinstance(...)`) — legitimate scenario-shape branching. **11 + 1 mixed.**
- **guard on the ACTUAL** — the defect. **13 sites**, of which 2 are fixed on the branch.

## Remaining work: ~11 sites, each needing its own mutation verdict

```
uc004_delivery.py:2328   then_geo_system              if geo_system is not None
uc004_delivery.py:2079   then_error_no_reveal         if mb_id
uc004_delivery.py:2610   then_no_billing              if mock is not None
uc019_query_media_buys.py:1364  then_package_details  if pkg.bid_price is not None
uc026_package_media_buy.py:2453 then_pricing_defaults if list_price is not None
uc002_create_media_buy.py:1032  _assert_validation_pass       if hasattr(root, "account_id")
uc006_sync_creatives.py:4586    _assert_standard_processing   if hasattr(registry.build_creative, "called")
uc006_sync_creatives.py:2431    then_assignment_processing_should_abort  if bad_package
uc011_accounts.py:1236   then_no_accounts_in_response if error_payload is not None
uc011_accounts.py:2631   then_none_have_brand_domain  if hasattr(acct, "brand") and ...
generic/then_media_buy.py:325   then_media_buy_persisted      if principal is not None
```

**Do not sweep these blindly.** Removing a guard while the read still names a nonexistent attribute converts a silent pass into a permanent failure. Each needs the mutation treatment: remove the artifact in production, confirm the row goes red.

## A companion check worth running alongside

Scan for `getattr(<obj>, "<literal>")` in oracles naming a field absent from every `adcp` model. Build the field universe by walking **all** `adcp.*` modules (925 modules, 2658 field names) — **not** `dir(adcp.types)`, which misses submodule models and produces false positives. Verified absent: `daily`, `by_day`, `is_fixed`, `workflow_step_id`. Verified present (an earlier claim that it was absent was wrong): `list_price`, on `adcp.types.generated_poc.pricing_options.price_breakdown.PriceBreakdown`.

## Related

- #1381 (closed) — remediated 3 sites of the mock-conditional flavour (`if mock is not None: assert...`); the ~11 above are the residue, and the issue closed with no follow-up tracker
- #1319 item C5 — the production gap this class was hiding
- #1749 — the sibling vacuity shape (defaulted reads of never-written ctx keys); a complete anti-vacuity guard should cover both

KonstantinMirin:
UC-004 BDD wiring, plus the fallout from the review of this PR. The theme throughout is one defect class: **oracles that could not fail while reporting success**, and the guards that let them survive.

`src/` carries two changes, both small and both explained below — everything else is test, guard and doc.

## Spec grounding

Authoritative version: **AdCP 3.1.1** (repo pin, `adcp==6.6.0`, `docs/adcp-spec-version.md`).

| Obligation | Source |
|---|---|
| Seller echoes the applied window **when the buyer provided one** | `dist/schemas/3.1.1/media-buy/get-media-buy-delivery-request.json`, `attribution_window.description` |
| `attribution_window` present on every successful response | **repo-level rule, stronger than spec** — the response schema's `required` is `[reporting_period, currency, media_buy_deliveries]` |
| `interval` must be 1 when `unit` is `campaign` (INV-5) | `dist/schemas/3.1.1/core/duration.json` — stated in prose only; expressible in draft-07 `if/then`, but the published schema does not, so neither the SDK model nor request validation enforces it |
| `interval` minimum 1; `unit` enum; `model` enum | `core/duration.json`, `enums/attribution-model.json` |
| `VALIDATION_ERROR` / `recovery: correctable` | identical in the harness-pinned enum (64 codes) and v3.1.1 (92 codes) |

Conformance storyboard: **ungraded** — `attribution_window` appears in 0 of the 286 files under `dist/compliance/3.1.1/`. There is no non-BDD fallback either: `pytest tests/unit tests/integration -k attribution --co` collects nothing, and `_resolve_attribution_window` has no direct test. These scenarios are not *a* grader of the attribution contract; they are the only one.

## Oracles that could not observe what they graded

Each was proven false-green by mutation, not by inspection.

**The echo oracle compared against literals.** `then_attribution_echo` read `ctx["request_attribution"]`, a key with no writer anywhere in the repo, falling back to `.get(..., 7)` / `.get(..., "days")`. It stayed green only because its one scenario happened to request 7/days. Fixed at the choke point: `dispatch_request` now records `ctx["dispatched_kwargs"]`, covering all 99 call sites, read through one shared `dispatched_field` accessor that also handles the `req=<model>` dispatch shape used by 38 of them.

**`_assert_valid_content`'s attribution branch was three guards deep.** With `attribution_window` dropped from every response, all 48 valid rows stayed green. Under the same mutation, 63 params now go red.

**A guarded assertion read an attribute that does not exist.** `getattr(pkg, "daily") or getattr(pkg, "by_day")` — neither exists on `adcp.types.ByPackageItem` (the field is `daily_breakdown`), so the guard was permanently false. Repairing it made a pre-existing production gap observable, now filed as **#1776**.

**Breakdown and truncation oracles graded the typed payload.** `then_packages_limited`, `then_field_true` and `then_field_false` now read the wire through one shared `_wire_packages` reader. The threat model is worth stating precisely, because the obvious one is wrong: a *dropped* field is caught by the typed oracle too, since reconstruction turns a missing key into `None`. The blind spot is a value that **coerces back** — emitting the boolean truncation flags as the JSON strings `"true"`/`"false"` left all six rows green before and turns all six red now.

**The circuit-breaker probe oracle blamed a mock that was never missing.** `then_single_probe` looked its POST mock up as `env.mock["httpx_post"]`/`["webhook_post"]` — no env defines either — so it always missed and fell through to `pytest.xfail("HARNESS GAP: no webhook POST mock")`. `CircuitBreakerEnv` exposes it as `mock["post"]`. Its baseline was a second phantom (`ctx.get("pre_open_call_count", 0)`). Corrected, this exposes a real gap (**#1781**), and the xfail is narrowed to exactly that.

**Eleven assertions were nested under a guard testing for the artifact they graded** (#1751). They did not fall into one bucket, which is why each got its own verdict rather than a sweep:

- *Genuinely dead code.* `then_no_billing` looped over `env.mock.get(name)` for `"charge"`/`"create_billing_record"`/`"bill"`. No harness env has ever exposed any of those — `DeliveryPollEnv` patches exactly one external, `"adapter"` — so the body never executed. Deleted rather than repaired, and deliberately not replaced with an "adapter was not called" assertion: sandbox delivery *does* call the adapter. The mirror image is `hasattr(registry.build_creative, "called")` on a MagicMock, which was unconditionally *true*.
- *Dead ctx read, fixed at the write side.* `then_error_no_reveal` read two keys no step writes. The uc006 abort step's fallback key was written by one of its two Givens and not the other, so the check silently applied to one scenario and skipped the other.
- *Legitimate optionality still hiding vacuity.* `bid_price` is legitimately `None` for fixed-price options and `brand` may be absent — but absence across *every* item made the claim graded-by-nothing. A cross-agent-**leak** check passed while structurally unable to observe a leak.

Counterfactual, with production nulling every `by_geo` system: **old oracle 32 passed / exit 0; new oracle 6 failed.**

**`then_attribution_model` could not tell "seller applied the buyer's model" from "seller applied its own default."** The scenario requested `last_touch`, which *is* `PLATFORM_DEFAULT_ATTRIBUTION_MODEL`, so both produce byte-identical output. It now requests `data_driven` and cross-checks the Gherkin literal against the dispatched request; the ignore-the-buyer's-model mutation fails 3/3 transports.

**A campaign-unit window asserted `interval >= 1`,** which passes for any positive number — production could collapse the buyer's 1094-day flight lookback to 1 day invisibly. Now compared against the fixture-derived flight length; the collapse fails 9 rows.

**Nothing anywhere graded a seller fabricating an unrequested window.** The echo helper `continue`d when the buyer named no `post_click`/`post_view`. Two fabrication mutations left the entire UC-004 module byte-identical. Both now die (6 and 36 failures).

## Guards that let this survive

**The strict-xfail deselection has been deleted.** The redundant-transport optimization dropped mcp/rest items *because* they carried a strict xfail — so a row that started passing could never surface as `XPASS(strict)` in any configuration CI runs, and `strict=True` bought nothing on those transports. Removing it un-hid **19** strict xpasses, not the six that were known. Each was walked per `.claude/rules/workflows/xpass-graduation.md` — date-range ×8 (validation lives in `_impl`), MCP TypeAdapter ×6 (closed by `RequestCompatMiddleware`), REST `INVALID_REQUEST` ×3, plus ownership/sampling. The vacuous second-layer `strict=False` blocks that had been demoting graduated rows to silent XPASS went with it. Cost of running them: 321 of 7543 items, seconds. Answers **#1740**'s central question.

**The #1462 reason guard reached one of three mechanisms.** It scanned `pytest.mark.xfail(reason=<literal>)` only: 43 of 61 markers and **0 of ~144** runtime `pytest.xfail(...)` calls, which pass their reason positionally. It was also blind to the conftest tables, where `reason=` binds an `ast.Name` — writing `#1462` into `_UC004_PARTITION_SELECTIVE` left the file green. Now scans every non-docstring string literal, which reaches all three at once; comments and docstrings stay exempt so the history can still be explained. Its `_scan_source` duplicate walk — exercised only by the meta-tests, leaving the production path unprotected — is gone, and a new test pins that `_BDD_DIR` resolves to a non-empty tree, the one mutation `tmp_path` fixtures cannot catch. 5 → 13 tests; all four surviving mutations now redden.

**102 step definitions no feature step can reach.** `test_architecture_bdd_no_unbound_steps.py` implements pytest-bdd's real matching semantics (not literal comparison) and ratchets the backlog shrink-only under **#1801**. Two of the sites fixed above were among them — they carried the #1751 defect *and* were unreachable, so they read as coverage under two separate reviews.

**The pin guard checked the constant and not one word of the prose around it,** which is why `README.md` still advertised AdCP 3.1.0-beta.3 / `adcp==5.7.0` while `CLAUDE.md` said 3.1.1. Corrected, plus a presence-and-drift scan that excludes the deliberately historical shapes, mutation-verified both ways. The bump procedure in `docs/adcp-spec-version.md` now names `README.md` and `CLAUDE.md` — their omission was the root cause.

**Nothing type-checks `tests/`.** Stage 1 lands a hard mypy gate on `tests/bdd/steps/_outcome_helpers.py`, the accessor layer every step module reads through, proven by injecting a type error that passed `make typecheck` before and fails now. The ctx accessors return `Any` rather than `object` (each caller immediately indexes the result), which took `uc004_delivery.py` 4 → 0 and `_outcome_helpers.py` 5 → 0. Widening to `tests/harness/` (43 errors) and a `tests/`-wide ratchet (baseline 427) are tracked, not attempted here.

## Graduations

`T-UC-004-partition-attribution` is out of the xfail routing entirely — five rows, one per commit, each walked against the protocol: scenario checked for a named outcome, error code re-verified against the pinned enum, Givens confirmed to use only the shared cross-transport harness, and the request traced through `adcp_validation_boundary` to the wire envelope rather than inferred from a green mark.

`T-UC-004-boundary-attribution` graduates as a block (3 payloads × 2 transports) on the transitive argument that its payloads are byte-identical to partition twins already graduated under the stricter `with suggestion` assertion.

`T-UC-004-boundary-reporting-dims` **does not** graduate. Its rows assert a categorical "invalid" rather than a named code, and REST emits `INVALID_REQUEST` where MCP/A2A emit `VALIDATION_ERROR` for the same input (`src/app.py:248` special-cases only `attribution_window`) — the transports genuinely disagree, and that wants reconciling first.

Ten mock-injection ledger entries graduated with in-network evidence; the 20 whose cause is intact are kept, reason re-derived at the routing site. Four `*-pending_activation` substrings removed as dead.

## The two `src/` changes

**`adcp_a2a_server.py` — documentation only, no behaviour change.** Every A2A response DataPart routes through `_dict_to_value`, and `google.protobuf.Value`'s only numeric slot is `number_value`, a double — so **every integer widens to an integral float** on the A2A wire (`14` → `14.0`), while MCP and REST preserve ints. This is inherent to a2a-sdk 1.0's proto-first Part type, not repo-fixable, and **not** a conformance break: the AdCP schemas are draft-07, where `"type": "integer"` accepts any number with a zero fractional part. It does mean `assert echoed == req_window` passes on A2A regardless of numeric type. Documented at the choke point and pinned by `test_a2a_numeric_wire.py` so an SDK representation change surfaces here first.

**`webhook_validator.py` — the `ADCP_TESTING` allowance now covers private-range receivers.** Merging `origin/main` brought in #1697's send-time SSRF gate, which silently dropped **every** e2e webhook delivery: the capture receiver is reached by the compose alias (`ADCP_WEBHOOK_HOST=tests`) resolving to a private Docker IP, and the localhost rewrite targets `host.docker.internal`, but the allowance matched only localhost/loopback rejection shapes. Six webhook e2e tests ran with zero deliveries. This deliberately **reverses** an earlier pin asserting that testing mode blocks private networks — that pin asserted exactly the behaviour that broke e2e, and the reversal is recorded with its justification rather than quietly dropped. Cloud metadata endpoints fail the hostname check first with a message the needles do not match, so they stay blocked; verified directly:

```
BLOCK  http://169.254.169.254/latest/meta-data/
BLOCK  http://metadata.google.internal/x
ALLOW  http://172.20.0.5:8080/webhook
ALLOW  http://192.168.1.1/webhook
```

Production semantics are untouched and asserted. **This overlaps #1802 (`feat/secure-outbound-fetch`)**, which is reworking this area — coordinate on merge order; this change is the minimum needed to keep e2e webhook delivery observable on this branch.

## Ledger, guards, deps

`e2e_rest_known_failures.txt`: **17 → 9**, with `EXPECTED_LEDGER` kept in exact sync. Every remaining entry is a production code gap. Where an entry was *added* rather than graduated, it is because repairing a vacuous test unmasked a pre-existing production failure — the `daily_breakdown` rows cite **#1776**, and the metro-no-system row is the partition twin of a C10 gap the boundary row already recorded, unmasked when the vacuous `strict=False` blanket came out.

`test_architecture_bdd_no_orphan_ctx_reads.py` — Tier 1 (defaulted orphan reads, the silent-constant shape) is **empty and must stay empty**; Tier 2 seeded 11 pre-existing entries and is now at 8, shrink-only. Verified to flag the original defect when run against the pre-fix tree.

`verify_feature_error_codes.py` now classifies `LOWERCASE_VARIANT` separately from `NON_CANONICAL`, gated repo-wide via `make quality`; 13 casing-only sites corrected.

Two dependency changes: the unused `mocker` dev dependency is dropped (it was breaking the in-network Docker build), and `datamodel-code-generator` moves to `>=0.64.0`, clearing 9 advisories.

`.duplication-baseline` **lowered** 80 → 74 for `tests/` — the ratchet recording what the consolidations here earned, measured identically on macOS and in the CI-box image (pylint 4.0.5 / astroid 4.0.4). No guard allowlist grew.

## Verification

Full suite on the CI box (`innet_300726_1816`), all seven envs, **zero failures and zero errors** across 18,103 items:

| env | total | passed | failed | xfailed | xpassed |
|---|---|---|---|---|---|
| unit | 5718 | 5682 | 0 | 26 | 0 |
| integration | 2275 | 2226 | 0 | 41 | 0 |
| bdd_inprocess | 7543 | 1788 | 0 | 5743 | 9 |
| bdd_e2e | 2348 | 406 | 0 | 1923 | 17 |
| e2e | 125 | 93 | 0 | 4 | 0 |
| admin | 89 | 86 | 0 | 3 | 0 |
| ui | 5 | 5 | 0 | 0 | 0 |

`bdd_e2e` is the only job that grades the e2e_rest ledger, and it executes here. Branch is level with `origin/main` (merged, 0 behind).

Every fix in this branch is mutation-verified: the assertion is shown to fail against a deliberately broken production, not merely to pass against a working one. No assertion was deleted or weakened to reach green — the one reversal (testing-mode private networks) is argued explicitly above. No test skipped, deselected or newly xfailed.

The remaining xpasses are non-strict and pre-existing: the 9 in-process ones are UC-004 rows needing per-scenario walks, tracked separately.

Closes #1462 — not reproducible as reported; the root cause was BDD step binding, fixed in #1545.
Closes #1739 — all ten parallel e2e_rest mock-injection entries graduated, verified in-network.
Closes #1751 — all 11 presence-guard sites resolved, each with its own mutation verdict.

## Known follow-ups

- **#1801** — the 102-step dormant-step backlog the new guard ratchets
- **#1777** — UC-004 package oracles still on the typed payload, and the Check C allowlist tracking them
- **#1778** — the corpus-wide census: 243 success-path Then steps across 15 step modules
- **#1776** — the `include_package_daily_breakdown` production gap these ledger rows record
- **#1781** — the half-open probe never reaching httpx post
- **#1749** — orphan ctx reads; Tier 2 down 11 → 8, not empty
- **#1750** — the circuit-breaker probe mask
- **#1767** — beads ids in comments, extended from `src/` to `tests/`

Also open and unchanged: #1726 (INV-2 seller capability), #1740 (the rest of the collection hook), #1744 (success-wire guard, a prerequisite for #1778), #1752/#1753.

Not carried here: the REST/MCP `INVALID_REQUEST` vs `VALIDATION_ERROR` divergence at `src/app.py:248`, which blocks the reporting-dims graduation; widening the `tests/` mypy gate; and mirroring the two feature-file edits upstream.


ChrisHuie:
# Review — #1728 (verified at head `e3e13358a`)

The mechanism holds and it is the strongest part of the change. `wire_dict` (`tests/bdd/steps/_outcome_helpers.py:53-59`) is the guarded reader, not the inline-read-plus-`model_dump`-fallback shape: it degrades only for `transport in (None, IMPL)`, `_IMPL_ONLY` is empty (`tests/bdd/conftest.py:2773`), and `dispatch_request` raises when transport is `None` (`tests/bdd/steps/generic/_dispatch.py:56-60`), so the degrade path is unreachable from every UC-004 attribution param. Serializer-only mutations confirm it: popping `attribution_window` in the REST serializer (`src/routes/api_v1.py:400`), the A2A one (`src/a2a_server/adcp_a2a_server.py:1390`) and the MCP `structured_content` build (`src/core/tools/media_buy_delivery.py:808`) each redden +21 params, isolated to that transport, with no null-key hole. The counterfactual is the number worth quoting: those same three mutations redden **15** params at `origin/main` and **63** here, and the 48 newly caught are exactly the outline rows whose old `if aw is not None: assert "model" in aw` was fully skippable. Removal of the dead xfail block is outcome-neutral at collection across all four transports — 9561 items both sides, 0 gained or lost, 0 marker-kind or `strict=` changes — and the removed condition was provably unsatisfiable: 412 collected nodeids contain `error "INVALID_REQUEST"` and none contain "attribution".

One thing the PR body understates. `ctx["request_attribution"]` was not a key someone invented: `ctx["request_attribution"] = aw` existed, was added in `ebb527c67`, and was deleted by `61f6ad9a2` — the BDD-harness foundation commit — with the deleted line's own comment reading *"Capture the requested window so the echo/model Then-steps can cross-check the response against what the buyer actually asked for."* That history argues **for** the design chosen here: a per-step write is exactly what got orphaned, and a choke-point record cannot be orphaned the same way. Sweeping that commit for siblings comes back clean — it removed four `ctx` writes and the other three have no surviving readers.

Everything below is one theme. The strengthening is **asymmetric**: it fixes the axis it was looking at and leaves adjacent axes at the previous strength, in the same scenario and the same helper. Nothing blocks — no production code changes, suite green — and the two most severe items are a one-line and a four-line edit.

---

## 1. [SHOULD-FIX] The scenario the PR de-coincidenced is still coincident on the other half of the same line

**Where:** `tests/bdd/features/BR-UC-004-deliver-media-buy-metrics.feature:590-591`.

**Context:** commit 1 moves the requested window 7 → 14 days with the reasoning "14 equals no default anywhere, so the coincidence that hid this cannot recur." The same Gherkin line also carries `"model": "last_touch"`.

**Problem:** `PLATFORM_DEFAULT_ATTRIBUTION_MODEL = AttributionModel.last_touch` (`src/core/tools/media_buy_delivery.py:81`), and production resolves `model = requested.model or PLATFORM_DEFAULT_ATTRIBUTION_MODEL` (`:992`). So the buyer's model and the fallback produce the same byte:

```
buyer sends model=last_touch -> {'post_click': {'interval': 14, 'unit': 'days'}, 'model': 'last_touch'}
buyer sends NO model         -> {'post_click': {'interval': 14, 'unit': 'days'}, 'model': 'last_touch'}
byte-identical?  True
```

`then_attribution_model` (`uc004_delivery.py:2381-2389`) compares the wire against the Gherkin literal rather than against `dispatched_kwargs`, so the model half of INV-1 — *buyer provides, seller applies* — is untested in the one scenario named for INV-1. To be precise about what this is and is not: the step still reddens if production returns a **wrong** model; what it cannot see is production **ignoring** the buyer's model, which is the claim the scenario makes. Sweeping every attribution row that names a model: `:590`, `:706` and `:731` all request `last_touch` and cannot discriminate; only `:708` and `:732` (`data_driven`) do, so the obligation is carried by 2 rows × 3 transports.

**Fix:** one line, on both halves of the pair.

```diff
-    When the Buyer Agent requests delivery metrics for "mb-001" with attribution_window {"post_click": {"interval": 14, "unit": "days"}, "model": "last_touch"}
-    Then the response should include attribution_window with model "last_touch"
+    When the Buyer Agent requests delivery metrics for "mb-001" with attribution_window {"post_click": {"interval": 14, "unit": "days"}, "model": "data_driven"}
+    Then the response should include attribution_window with model "data_driven"
```

Better still, have `then_attribution_model` cross-check the parsed literal against `_expected_attribution_model(ctx)`, so the two model readers share one derivation instead of two policies.

## 2. [SHOULD-FIX] `interval >= 1` lets production collapse a 1094-day campaign window to 1 day

**Where:** `tests/bdd/steps/domain/uc004_delivery.py:2716` (new in this PR) and `:2542` (pre-existing check, in a function this PR rewrote onto the wire — base carried `pc.interval >= 1` at `:2556`).

**Context:** commit 2 deletes the campaign branch from `then_attribution_echo` because it "asserted only `interval >= 1`, which passes for any positive number — a near-vacuous check inside a fix whose purpose is removing vacuous checks." Its replacement docstring at `:2425-2432` goes further and prescribes the remedy: *"derive the expected day count from the flight dates the Given seeded."* Commit 5 then writes `assert echoed["interval"] >= 1` into the new helper.

**Problem:** mutating `_echo_duration` so the campaign branch yields `days = 1` (`src/core/tools/media_buy_delivery.py:986`) leaves the 87-param attribution slice **byte-identical to baseline** — 72 passed / 3 xfailed / 12 xpassed. Production can reduce the buyer's full-flight lookback from 1094 days to 1 and no oracle moves. The correct value is deterministic: `given_media_buy_with_status` (`:118`) passes no dates, so `MediaBuyFactory` supplies `2025-01-01` → `2027-12-31` (`tests/factories/media_buy.py:32-33`) and `campaign_length_days` is 1094.

Worth stating what this is *not*, because it is the obvious first hypothesis and it is wrong: the `campaign_length_days is None` fallback at `:988` is unreachable here. It is None only when `target_media_buys` is empty, and `media_buy_deliveries` is built by iterating that same list (`:288`, `:687`), so the oracle's own `assert wire.get("media_buy_deliveries")` fires first. The blindness is to the day count itself, not to the fallback.

**Fix:** assert the value. Deriving it from the response's own dates would mirror production's subtraction, so read it from the fixture:

```diff
+from tests.factories.media_buy import MediaBuyFactory
+
+# The seeded flight; a campaign-unit window must resolve to exactly this.
+_CAMPAIGN_FLIGHT_DAYS = (MediaBuyFactory.end_date - MediaBuyFactory.start_date).days
...
-            assert echoed["interval"] >= 1, (
-                f"Valid {field}: resolved campaign {window_name} must span >= 1 day, got {echoed['interval']}"
+            assert echoed["interval"] == _CAMPAIGN_FLIGHT_DAYS, (
+                f"Valid {field}: a campaign {window_name} must resolve to the seeded flight length "
+                f"({_CAMPAIGN_FLIGHT_DAYS} days), got {echoed['interval']}"
             )
```

and the same at `:2542`. That catches a wrong flight, a fallback, and a clamp; it shares production's subtraction so it would not catch an off-by-one in `.days` itself — pinning the literal `1094` would, at the cost of breaking when the fixture moves. Either is strictly stronger than `>= 1`.

## 3. [SHOULD-FIX] The echo oracle grades presence of what was asked for and never absence of what was not

**Where:** `tests/bdd/steps/domain/uc004_delivery.py:2703-2706`.

**Context:** the loop `continue`s when the buyer requested no `post_click`/`post_view`; the docstring at `:2684` claims the helper grades "BR-RULE-092 INV-1/2/3".

**Problem:** two production mutations that violate INV-4 — making the `requested is None` branch return `post_click=Duration(99, days)`, and making `_echo_duration(None)` return the same — each leave not only the attribution slice but the **entire UC-004 module** byte-identical (458 passed / 494 xfailed / 25 xpassed) and the full unit suite green. There is no non-BDD grader to fall back on: `pytest tests/unit tests/integration -k attribution --co` collects **nothing** (7811 deselected), and `_resolve_attribution_window` has zero direct tests. A seller fabricating lookback windows the buyer never requested is graded by nothing anywhere.

**Fix:** assert the converse for windows the buyer did not name. This was run rather than reasoned: on a clean tree it breaks **zero** currently-passing rows — including `seller_ignores`, `model_only` and `empty_object` on all three transports — and under the two mutations above it produces 6 and 20+ failures respectively.

```diff
         else:
             assert echoed == req_window, (
                 f"Valid {field}: {window_name} should be echoed verbatim as {req_window}, got {echoed}"
             )
+    for window_name in ("post_click", "post_view"):
+        if requested.get(window_name) is None:
+            assert aw.get(window_name) is None, (
+                f"Valid {field}: buyer requested no {window_name}, but the response echoed "
+                f"{aw.get(window_name)} — the seller must not fabricate a lookback window"
+            )
```

## 4. [SHOULD-FIX] The guard this PR re-points cannot see the reason string this PR edits

**Where:** `tests/unit/test_architecture_bdd_no_1462_xfail_reason.py:64-77`, self-tests at `:101-109`.

**Context:** commit 4 states "the rule, the AST scan and all four meta-tests are untouched, and the file still reports 5 passed." All true.

**Problem:** `_UC004_PARTITION_SELECTIVE` entries reach `pytest.mark.xfail(reason=reason, strict=False)` at `tests/bdd/conftest.py:1994` as an `ast.Name`, so `_string_parts` collects no `ast.Constant`. Writing `#1462` into `conftest.py:1947-1948` — the exact entry this PR rewrote — leaves the file at **5 passed**. So does a positional `pytest.xfail("…#1462")`. A positive control with a literal `reason=` kwarg at `conftest.py:1910` reddens with the expected message, so the guard does walk that file and the two blind results are real. Coverage, split by construct so the denominator is honest: **43 of 61 `pytest.mark.xfail` markers reached (70%), and 0 of 144 runtime `pytest.xfail(...)` calls** (88 f-strings, 55 literals, 1 concat — all human-authored reason text pytest reports as the xfail reason).

One thing this is not: the PR did not push a reason from visible to blind. Literal-reason markers in conftest went 44 → 43 (the PR removed one visible marker) and variable-reason markers stayed 18 → 18; the table entry was already the blind form at base.

Compounding it, the four meta-tests exercise `_scan_source`, a second copy of the walk, so the production path is unprotected: pointing `_BDD_DIR` at a non-existent directory leaves 5 passed, and so does making `_find_1462_reasons` return `[]` unconditionally. Only the two shared leaves are covered.

**Fix:** in `_find_1462_reasons`, take `node.args[0]` when the call is `pytest.xfail`/`xfail`, and resolve string constants bound to the name passed as `reason=`. Delete `_scan_source`, have the meta-tests write their fixture to a `tmp_path` `.py` and call `_find_1462_reasons(path)`, and add one case that points `_BDD_DIR` at a tmp dir holding a known-bad file and asserts the production test's violation list is non-empty. Give each newly-covered form a meta-positive, and state the residual out-of-scope forms in the docstring rather than leaving the current unqualified claim.

## 5. [SHOULD-FIX] The attribution xfail narrative is corrected in one of five places, and the new reason is contradicted by the run

**Where:** `tests/bdd/conftest.py:1269-1275`, `:1281-1284`, `:1319-1334`, `:1943-1949`.

**Problem:** `:1269-1275` still says the attribution reference asserts `error "INVALID_REQUEST" with suggestion` — the feature says `VALIDATION_ERROR` at `:713-717`; still blames "generic-step shadowing (#1417)", the cause the new guard docstring declares dead; still contrasts it against "the #1462 in-process drop", the framing the guard now forbids; says `campaign_interval_not_one is xfailed separately below` when no marker exists anywhere; and calls four rows "GRADUATED (removed)" while `:1944-1949` masks exactly those four. `:1281-1284` repeats "returns INVALID_REQUEST on all transports". The guard cannot help here — its docstring explicitly licenses comments mentioning #1462 to explain history, so it structurally cannot catch stale prose one screen from the prose the PR did rewrite.

`:1943-1949`, the entry the PR rewrites, now reads "retained on **observed status** pending per-row graduation". The observed status is **12/12 XPASS** on a2a/mcp/rest, every row emitting `VALIDATION_ERROR` / `correctable` with a suggestion. With `strict=False`, a regression of that wire contract lands as a silent xfail.

`:1319-1334` is the boundary twin: 6 mcp/rest rows carrying the same dead `INVALID_REQUEST` / "rest 422 detail" framing, contradicted by the measured wire (`{"adcp_error": {"code": "VALIDATION_ERROR", "recovery": "correctable", "field": "attribution_window.post_click.interval", …}}`). They are `XPASS(strict)` and the single-transport optimization deselects them in every configuration CI runs — `BDD_ALL_TRANSPORTS` appears nowhere in `.github/`, `tox.ini`, `run_all_tests.sh` or `Makefile`.

**Fix:** rewrite or delete `:1269-1275` and `:1281-1284`; drop the four-row set from `:1943-1949` so those rows grade unmasked; delete the six `:1319-1334` entries — removing the strict marker also removes the deselection trigger, so the rows run and pass in the default configuration. Confirm the first two with e2e_rest enabled before landing.

## 6. [SHOULD-FIX] The two INV-2 scenarios seed identical state and demand opposite behaviour

**Where:** `tests/bdd/steps/domain/uc004_delivery.py:654`, `:660`; `…feature:709`, `:739`.

**Context:** the PR carves INV-2 out as still unmet, tracked separately — honest about the obligation, but it understates how the suite currently represents it.

**Problem:** `ctx["supports_attribution_windows"]` is written by both Given steps and read by nobody — `git grep` returns exactly those two writes, and `git grep -E "supports_attribution|attribution_supported|configurable_attribution" -- src/` returns nothing, so production has no seller-capability concept at all. `ctx` is a per-scenario fixture (`conftest.py:2950-2951`), so `T-UC-004-attr-supported` and `T-UC-004-attr-unsupported` seed byte-identical database state and are separated only by that inert write — one requiring the echo, the other requiring the platform default. Whichever way a capability check is eventually implemented, **one of those two scenarios becomes unsatisfiable**. That also means the three baseline xfails at `then_attribution_default:2471`, labelled "PRODUCTION GAP", cannot graduate on a production change alone.

Separately, the two Example rows named for the same invariant — `seller_ignores` (`:709`) and `seller ignores field (no configurable window support)` (`:739`) — carry no such Given, route through `_assert_valid_content:2765`, and now assert `echoed == req_window` verbatim: 6 params asserting that the seller echoes the buyer's 30-day window, which is what the label says must not happen. Before this PR the branch asserted only `"model" in aw`, so they were inert; the strengthening makes them assert the negation of their own name.

**Fix:** make the Given wire a real capability so the pair is separable and the xfail is graduatable; and either rename the two rows so they stop claiming INV-2, or add the unsupporting-seller Given and let them xfail with the other three. In the carve-out, say INV-2 has zero graded coverage rather than three xfails.

## 7. [SHOULD-FIX] The pin correction stops one file short of the file outside contributors read first

**Where:** `README.md:8`, `README.md:25-26`, `docs/adcp-spec-version.md:46`, `:69`, and the checklist at `docs/adcp-spec-version.md:88-102`.

**Problem:** commit 6's rationale — that the stale line "sits directly above the Spec-Grounding Gate, which instructs contributors to cite the version the repo currently PINS" — applies verbatim, with wider reach, to `README.md:25` ("targets **AdCP spec version 3.1.0-beta.3** via the `adcp==5.7.0` Python SDK", same doc-link, same guard reference), and `:26` goes further with "That spec version is a **beta** — request/response shapes are not yet frozen", which is no longer true. Inside the canonical doc, `:46` says "The SDK **pin** (3.1.0-beta.3)" against `:3` and `:15`, and `:69` states "adcp 5.7 types the response…". The root cause is on the page: the bump procedure at `:88-96` and the Related-files list at `:100-103` name `pyproject.toml`, `EXPECTED_SPEC_VERSION` and "this document" — neither README nor CLAUDE.md, which is why this drifted. (`:28`'s mapping row and `:51`'s beta.3 storyboard path are historical and correctly stay.)

**Fix:** the same two-value edit at those four lines, drop the "beta" sentence, and add README.md and CLAUDE.md to both the numbered procedure and Related files. A follow-up worth filing: extend `tests/unit/test_adcp_spec_version.py` to derive the pin and assert the prose in those files matches — it is statically checkable from local files, so it belongs in pytest.

## 8. [SHOULD-FIX] The grounding table's echo citation is scoped wider than the sentence it quotes

**Where:** the PR body's obligation table; `tests/bdd/steps/domain/uc004_delivery.py:2351-2370`.

**Problem:** the cited sentence is conditional — "**When provided**, the seller returns conversion data using the requested lookback windows instead of their platform default. The seller echoes the applied window in the response." The response schema's `required` is `['reporting_period', 'currency', 'media_buy_deliveries']`, the SDK model types `attribution_window` as `required=False`, `| None`, and 3.1's `core/attribution-window.json` dropped 3.0's `required: ['model']` in favour of "SHOULD be populated … when absent, the seller's default applies." So making absence a hard failure on every valid row is a repo-level rule stronger than spec, not a spec obligation. The code docstring is honest about this — it says "BR-RULE-092 requires"; the table is where the over-read is.

**Fix:** split the table row into the spec-confirmed part (echo when the buyer provided one) and the repo-level part (present on every successful response, stronger than the schema). No code change.

---

## Nits

- `_dispatched_post_click:2405` and `_dispatched_attribution_window:2665` are handed the same unreadable input and behave oppositely — the first asserts and fails with a diagnostic, the second returns `{}`, which is also its deliberate return for the legitimate `omitted` row, so the two cases are indistinguishable by construction. 38 of the 99 `dispatch_request(` sites record the `req=<model>` form the `_dispatch.py:42-46` docstring warns about; none is in UC-004, `ctx` is per-scenario, and both readers are module-private, so nothing reaches it today. Worth making the second raise like the first, and worth the accessor pair the codebase already uses one file over (`wire_field` / `wire_dict`) rather than a raw ctx key with a three-paragraph prose contract.
- `then_attribution_has_model:2497` grades enum membership and `then_attribution_campaign_length:2532` grades `model is not None`, while the PR's own `_expected_attribution_model` derives the value. Mutating production to always return the wrong model reddens 57 params and leaves 6 green — these two, ×3 transports. Both are redundant with a stronger sibling on the identical dispatch (`attr-omitted` for the first, the `campaign_unit` partition and boundary rows for the second), so this is consistency rather than a coverage hole. `:2532` also bypasses `_wire_attribution_model`, giving a third distinct message for a missing model.
- `uc004_delivery.py:2404` and `:2665` add two mypy `attr-defined` errors (`"object" has no attribute "get"`) from `_require(...) -> object`, taking this file from 1 to 3. Nothing anywhere catches them: `Makefile:18,51` run `mypy src/`, the untyped-defs ratchet hard-codes `SRC_DIR = "src"`, and the pre-commit `files:` pattern excludes `tests/`. Making `_require` return `Any` also clears 5 pre-existing errors in `_outcome_helpers.py`.
- `_dispatch.py:38-39` says `ctx["request_kwargs"]` is popped "by several UCs". Exactly one UC pops it (`uc026_package_media_buy.py:293`, `:703`). The two adjacent claims about `request_params` in the same docstring are accurate.
- `_assert_valid_content` still has four branches with the same skippable shape the attribution one just left — `:2734` (`status_filter`), `:2746` (`resolution`), `:2775` (`daily_breakdown`), `:2791` (`date_range`) — plus a byte-identical duplicate pair at `:2781-2787` (`account`) and `:2797-2803` (`ownership`). Two of them grade against `ctx["request_params"]`, which `_dispatch.py:35-37` documents as an expectation channel deliberately holding values that were never dispatched. Worth carrying these four `path:line`s into the migration follow-up rather than a count.
- `uc004_delivery.py:2399-2404` and `:2660-2665` repeat a byte-identical `_require(..., hint=...)` block including the hint literal — the only duplicated hint string in all of `tests/bdd/`. `min-similarity-lines = 6` and the block is 5 lines, so the duplication ratchet cannot see it. Similarly `:2452/2458/2460`, `:2511/2516/2518`, `:2677/2680` triplicate the platform-default assertion with a byte-identical message, and `:2457` has already drifted from its twins.
- `_wire_attribution_window:2364` guards only `assert "error" not in ctx`; a 200 carrying `errors=[...]` (`media_buy_delivery.py:689`) grades as a clean echo. Latent — the measured success wire has `errors: null` — and the fix is verified safe: `assert not wire.get("errors")` gives `6 failed, 72 passed`, identical to baseline.
- Four `assert resp.reporting_period is not None` were dropped (`:2381`, `:2444`, `:2503`, `:2522`) and none of those four scenarios has another step grading it. `reporting_period` is `required=True` and not `| None`, so each was structurally vacuous and no coverage is lost, and `:1538` already documents "reporting_period check is left to its dedicated Then step" — but "no assertion deleted to get green" is inaccurate as worded. `assert "reporting_period" in wire` in `_wire_attribution_window` would make the removal a genuine upgrade.
- The diff adds `salesagent-1zy8` (new to the tree) at `_dispatch.py:30` and `uc004_delivery.py:2397`, and `salesagent-x18x` at `conftest.py:1932`, while commit 4 removes `salesagent-50hl` on the stated grounds that these ids don't resolve for outside contributors. The tree nets −5, so the issue is the internal inconsistency; `_dispatch.py:30` is the one that matters, being a shared docstring every UC reads, and the sentence works without the id.
- Commit 2's "24 further instances of that idiom across 7 step modules" does not match the tree: 28 at base → 20 at head, and this file lost 8 (9→1), not 5. One copy remains here at `:2956`, grading `status_filter` off the typed payload.
- `:2704`, `:2706`, `:2709`, `:2540`, `:2544` index Gherkin-authored data directly. Structurally safe today — `Duration` requires both `interval` and `unit`, so a partial window is a validation error that trips the error guard first — but `tests/unit/test_uc004_delivery_response_guard.py` exists for exactly this class and pins only two other steps.
- `_wire_attribution_window:2366` asserts `media_buy_deliveries` is non-empty for all six oracles; the spec permits an empty array and no current row hits it.
- `then_attribution_has_model` imports `AttributionModel` from `adcp.types.generated_poc…`, which the SDK's own `__init__` marks internal. There is no alternative in 6.6.0, and production carries the same TODO at `media_buy_delivery.py:69`; mirroring that marker on the test import would let a bump sweep find both.
- The docstring at `:2432` says the echo step "asserts exact equality"; on A2A the wire carries `interval` as a float, so `pc == requested` passes on `14.0 == 14`. Numeric type is not graded — worth saying so in the docstring. (The A2A int-to-float widening looks generic rather than attribution-specific — `sequence_number` and `aggregated_totals.media_buy_count` show it too — and is out of this diff; worth its own issue.)
- Claim (c) in the PR body — INV-5 is "cross-field, not expressible in JSON Schema" — is imprecise; draft-07 `if/then` can express it. The production docstring at `media_buy_delivery.py:33-36` already words it correctly: the published schema states it in prose only, so neither the SDK model nor request validation enforces it.
- `fix(bdd):` routes five test files and one doc line into release-please's visible "Bug Fixes" section; repo precedent for test-only merges is dominantly `test:` / `test(bdd):`.
- `dispatched_kwargs` is a new shared ctx channel on a 99-site choke point documented only in a helper docstring; the sibling channel `wire_response` is documented at `tests/CLAUDE.md:404-420`.

## Informational — why the existing guards could not have caught the original defect

`test_architecture_bdd_no_trivial_assertions.py` scores `assert pc.interval == req_interval` as meaningful purely because it is an `ast.Compare`; the comparison was to a constant, so the guard stayed green across the false oracle for as long as it existed. It also scans `@then`-decorated functions only, which puts `_assert_valid_content` and both new dispatch readers outside every guard's scan set. A fitness function for "an assertion whose expected side is `X.get(k, <literal>)` inside a `@then`" would close that class statically.

Two cheap hardenings, both better as their own issues than folded here. `tests/unit/test_bdd_partition_assertion_strength.py` already pins the "this branch is unconditional" property for `status_filter` and `resolution`; `attribution_window` has no case, and adding one is a unit-tier call needing no database. And `test_architecture_e2e_rest_escape_hatches.py:38-68` claims "adding OR removing a route fails here, forcing a reviewable pin update in the same change" — the block this PR removes gated on an intermediate variable, so both trees report 18 routes and an empty removed-set; three more indirect routes survive undetected at `conftest.py:1671`, `:1712`, `:1744`.

On the orphan-key follow-up the PR body names: the count is right. Seventeen `ctx` keys in `tests/bdd` are read but never written, and exactly four carry a non-None literal default — `webhook_url` (`uc004_delivery.py:1526`, which also disagrees with this module's own `_WEBHOOK_URL` at `:330`), `pre_open_call_count` (`:1806`), `last_max_results` (`uc011_accounts.py:417`) and `named_media_buy_ids` (`uc026_package_media_buy.py:2320`). The other thirteen degrade to `None` rather than to a wrong constant. The *write*-orphan direction is a different and milder class — an inert Given rather than a lying oracle — but it is what makes finding 6 permanent, so the proposed structural guard should cover both. Linking the filed follow-ups in the PR body would help; three of the four named there have no locatable issue.

---

Verified at `e3e13358a` (merge-base == `origin/main` `8cb8d0ed6`, 0 behind; `mergeable: MERGEABLE`, not `DIRTY`, so CI genuinely ran):

**Ran:** full BDD from the JSON report (`1738 passed / 5448 xfailed / 25 xpassed / 3 skipped`, 0 non-OK; reconciles to the 9561 collected figure as 7214 in-process + 2347 e2e_rest); `make quality` (exit 0; ruff, `mypy src/` clean on 282 files, 12 hooks, 5494 passed); 518 architecture guards, all BDD guards, the pin-drift guard; per-node outcome diff of the UC-004 module base vs head (977 nodes, 0 diffs) and a collection diff with `BDD_E2E_ENABLED=true` (9561 items, 0 gained/lost, 0 marker changes); **24 mutations** — three per-transport `attribution_window` pops, the MCP null-key variant, the inverse interval mutation, two INV-4 fabrication mutations, two wrong-model mutations, the campaign `days = 1` collapse, three mutations of the guard's own scanner plus a positive control, and a shallow-copy aliasing probe with its inversion; both proposed assertions run against a clean tree to confirm they break no passing row; duplication and complexity measured on both trees; five ratchet baselines diffed; the pinned 3.1.1 schemas read directly for the response `required` set and the `model` wording; the SDK response model introspected for field requiredness; `git log -S` provenance on the orphaned key; and all three GitHub comment endpoints plus GraphQL review threads.

**Not run:** e2e_rest at runtime — collected (2347 items, unchanged) and its wire path read at `tests/harness/dispatchers.py:315,330`, but no in-network stack was up locally, so every "grades the wire" statement is empirically proven for a2a/mcp/rest and inferred for e2e_rest; CI's own in-network job is green at this head. Also not run: `tox -e bdd` specifically (runs used pytest directly with `-p no:randomly`, matching the reported counts exactly), the integration/e2e/admin suites, a mutation of `PLATFORM_DEFAULT_ATTRIBUTION_MODEL` itself (the oracles import it from production, so they grade "production echoes its own default" — arguably correct, but unpinned), and an end-to-end re-derivation of the original #1462 report; its non-reproducibility was confirmed only in-process, by checking that the shared request builder preserves `post_click` and that `_validate_attribution_window` raises `VALIDATION_ERROR` / `correctable` with a suggestion.

KonstantinMirin:
Deferred from PR #1721's review (round 2) — filed as #1880 since this branch is already migrating the same file (`uc011_accounts.py`, close to your existing `oyiv.8`): list_accounts BDD steps bypass wire dispatch via an allowlisted `TRANSPORT-BYPASS`. Should be addressed as part of this PR.

Issue #1291, opened 2026-05-12
RFC 9421 request signing - Implementation


ChrisHuie:
## TL;DR

AdCP at v3.0.9 is explicit that request signing is **optional, capability-advertised**, and the `signed_requests` universal storyboard **self-gates on the seller's advertisement**. "AdCP Verified" is not a defined spec term.

The abstract question ("does AdCP require signing?") has a spec-level answer: no. What remains is a salesagent decision in three parts:

1. **Decision 1**: Implement signing at all? (Yes / No)
2. **Decision 2** (if implementing): Which layers of the signing stack do we use the SDK's implementation for, vs. roll our own?
3. **Decision 3** (if implementing): Stay on our current adcp 4.3 pin, or bump to 4.4+? (Decision 3 couples to other gaps — it's not just a signing-scoped choice.)

A diagnostic puzzle (our 2026-05-10 run reports 39 failures when the spec says non-advertising agents should be skipped) is captured as a footnote and should be resolved independent of the decisions above.

---

## Background: what the spec actually says

From the JSON Schema description on `request_signing` (verbatim):

> "Optional in 3.0 — capability-advertised so counterparties can opt into signing selectively."

From the `signed_requests` universal storyboard preamble:

> "Universal capability-gated storyboard — runs for any agent advertising `request_signing.supported: true`."
>
> "Absence of advertisement is not a failure, it is a declaration that the agent does not offer verified signed requests. Agents that do not advertise support are not tested against this storyboard."

From `security.mdx`: no `MUST/SHOULD/MAY` directive about whether sellers expose signed endpoints. The only normative RFC 9421 language in the doc applies to **webhook** signing (outbound) — separate track from inbound request signing.

So the spec's contract:

- Don't advertise → not tested, conformant.
- Advertise → must pass `signed_requests`.
- "AdCP Verified" tier: does not appear in the spec.

---

## Current salesagent state (verified 2026-05-10)

- No file imports anything from `adcp/signing/`.
- `src/core/tools/capabilities.py:67-247` does not emit a `request_signing` block.
- No JWKS endpoint exists.
- No signing-related issues, PRs, or commits in salesagent's history.
- By the spec's plain reading: conformant. Implicit Option-A posture (don't implement; don't advertise).

---

## A note on direction: inbound vs outbound signing

RFC 9421 signing in AdCP applies in both directions:

[...]

Issue #979, opened 2026-01-29
Update release workflow auth


dgirardi:
The [release workflow](https://github.com/prebid/salesagent/blob/main/.github/workflows/release-please.yml) is set up to use the default GITHUB_TOKEN. Because of this, release PRs to do not trigger any other workflows unless they are updated (for example, with an [empty commit](https://github.com/prebid/salesagent/pull/976/changes/8df8ae81ba1f53fa7c2a97c74074023814b28275))

This is documented here: https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials

However, the suggested solution (personal access tokens) is not ideal as they are tied to a specific individual and require frequent rotation. 

We should set up [github app authentication](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/about-authentication-with-a-github-app) to generate a token for release please.

olhaparamonova:
agreed with Chris to close it

Issue #977, opened 2026-01-29
Update docker hub repository


dgirardi:
The [release workflow](https://github.com/prebid/salesagent/blob/main/.github/workflows/release-please.yml) is set up to push images to Docker Hub:

```
      - name: Log in to Docker Hub
        uses: docker/login-action@v3
        with:
          username: ${{ secrets.DOCKER_HUB_USERNAME }}
          password: ${{ secrets.DOCKER_HUB_TOKEN }}
```

The credentials should be changed to point to [prebid](https://hub.docker.com/u/prebid), and any systems depending on [adcontextprotocol](https://hub.docker.com/u/adcontextprotocol) should be updated accordingly.

Issue #1589, opened 2026-07-13
Repo-wide secure outbound-fetch wrapper (SSRF/URL validation), sync + async


KonstantinMirin:
## Problem

SSRF / outbound-URL validation is applied **piecemeal** and keeps recurring. We fix it in one caller (a creative fetch here, a webhook there), then the next outbound call ships without it. The latest instance surfaced in the #1417 review:

- `src/services/webhook_delivery_service.py:483-485` posts buyer-supplied `reporting_webhook.url` / `push_notification_config.url` straight to `httpx.post` with **no** validation.
- `src/core/webhook_delivery.py:91` **does** go through `WebhookURLValidator`.

Same class of bug, fixed repeatedly in different files. Point-fixing each call site is unbounded and error-prone — this is the wrong way to address it.

## Solution

Build **one** repository-wide secure-fetch primitive — e.g. `salesagent_fetch_secure` — a single wrapper/class that embeds **all** the checks in one place (SSRF / private-IP / metadata-endpoint / scheme allowlist / redirect validation / DNS-rebind / timeouts). Every outbound HTTP call goes through it; call sites do zero validation of their own.

### Requirements
- Single module owning the policy; centralizes what `WebhookURLValidator` does today **plus** the HTTP call.
- **Both** sync and async variants (webhook delivery is sync; other callers async).
- Migrate every existing outbound caller onto it: `webhook_delivery_service.py`, `core/webhook_delivery.py`, creative fetch/upload, GAM upload, and any other raw `httpx`/`requests` egress.
- Add a structural guard (AST) that **bans raw `httpx`/`requests` outbound calls outside the wrapper**, so a new un-validated egress fails `make quality`. This is what stops the recurrence.
- Preserve existing `WebhookURLValidator` semantics (no weakening); fold it into the wrapper.
- Confirm the AdCP spec doesn't mandate accepting arbitrary buyer URLs before finalizing the allowlist policy.

## Context
Surfaced by PR #1417 review (round 8, item 5). SSRF is intentionally **not** fixed inside #1417 — filed here as a standalone architectural ticket. Verified still-valid @ `76f4304c6`.

KonstantinMirin:
Adding a concrete migration target that isn't named in the list above and is easy to miss under "creative fetch/upload": the **creative-agent connection path**, which is buyer-reachable and currently zero-validated.

[...]

PR #1757, opened 2026-07-28
feat: RFC 9421 request signing — key lifecycle, trust root and inbound verifier (#1291)

KonstantinMirin:
> ## ⚠️ Stacked on #1721 — REVIEW THE DELTA, NOT GITHUB'S DIFF
>
> This branch is **stacked on top of #1721** (`feature/spec-gaps-1210`, the capabilities/accounts contract work). #1721's branch lives on a fork, so GitHub **cannot** take it as a base — the base here is forced to `main`, and **the diff GitHub shows carries all of #1721's commits too**. GitHub's view is not the review scope. This section defines the review scope instead.
>
> ### What to review
>
> This branch forked from #1721's line at **`7f37e5252`** — that is `merge-base(#1721 tip, HEAD)`, and it is exactly the #1721 state merged in here at `9b955a123`. Everything this PR contributes sits on top of it:
>
> ```
> git diff  7f37e5252..HEAD      # 702 files — THE REVIEW SCOPE
> git log   7f37e5252..HEAD      # 134 commits
> ```
>
> **126 of those 134 commits are this branch's own work**; the other 8 arrived with the `main` merge at `12743116a`. GitHub's compare API confirms the span is clean — `status: ahead`, `total_commits: 134`.
>
> Cross-fork compare link: [`7f37e5252...KonstantinMirin:feat/rfc9421-request-signing`](https://github.com/prebid/salesagent/compare/7f37e5252...KonstantinMirin:prebid-salesagent:feat/rfc9421-request-signing) — **but it truncates the file list at 300** (GitHub's cap; the real count is 702). Use it for the commit list; use `git diff` for the file set. This is the third GitHub-side cap in this section that reads like a measurement — see the table below.
>
> ### What to ignore
>
> | | figure | why it is noise |
> |---|---|---|
> | GitHub's diff vs `main` | 320 commits · 680 files | **194 of those commits are #1721's**, not this PR's — GitHub has no way to exclude them |
> | The PR's "commits" tab | shows 100, then 250 | both are **GitHub API page caps**, not counts. The real number vs `main` is 320 |
| The compare page's file list | stops at 300 | also a cap. The real delta is **702 files** — `git diff 7f37e5252..HEAD` |
>
> ### Merge gate (not a review gate)
>
> **The stack is a STALE SNAPSHOT, and this section previously said otherwise.** The snapshot merged at `9b955a123` captured #1721's Aug 6 state (`7f37e5252`). #1721 has since advanced to `9585ace68`, which is **not an ancestor of this branch** — verified with `git merge-base --is-ancestor`. `git merge-tree` against that tip reports **64 conflicting files** (re-measured at `f5e286fdd`; the previous revision said *91*, which was the count of `CONFLICT` **lines** in merge-tree's informational block, not of files — 91 lines / 64 files). #1721 is itself OPEN and CONFLICTING against `main`.
>
> Reconciling with #1721's tip is a **merge** gate, not a review gate. **This cannot merge before #1721 does** — merging as-is would carry #1721's line into `main` through the back door. But a reviewer should not wait on that: the delta above is reviewable today and is what changes if the review changes anything.

Draft. Tracks #1291.

## What this delivers

RFC 9421 message signing, end to end: we verify inbound signatures, sign everything we emit, publish a trust root a counterparty can walk, and declare capability postures that match what we actually enforce.

Four outcomes, each checkable by a third party rather than asserted internally:

| | Outcome |
|---|---|
| **O1** | A third party fetches our JWKS and verifies a signature we produced |
| **O2** | A signed inbound request is accepted; an unsigned one to a `required_for` operation is refused, over real HTTP |
| **O3** | Everything we emit outbound is signed and verifiable — webhooks, the proof-of-control challenge, outbound AdCP requests |
| **O4** | What we advertise equals what we enforce |

## Spec grounding

Pinned version **AdCP 3.1.1** (`adcp==6.6.0`); guard `tests/unit/test_adcp_spec_version.py`.

- **Optional, capability-advertised.** `dist/schemas/3.1.1/bundled/protocol/get-adcp-capabilities-response.json` → `#/properties/request_signing`: *"Optional in 3.0 — capability-advertised so counterparties can opt into signing selectively."*
- **Graded by** the universal storyboard `signed_requests` (`dist/compliance/3.1.1/universal/signed-requests.yaml`) against **40** conformance vectors (12 positive + 28 negative) at `dist/compliance/3.1.1/test-vectors/request-signing/`, plus the coordination contract `test-kits/signed-requests-runner.yaml`. Negative vectors must return `401` with `WWW-Authenticate: Signature error="<code>"` matching byte-for-byte.
- **`identity.brand_json_url` is load-bearing** — schema-mandated the moment any signing posture is declared, which is why trust-root publication is in scope rather than optional.

Two upstream spec defects were found and fixed while grounding this: adcontextprotocol/adcp#6071 (vector count) and #6076 (seven documented error codes that do not exist), plus #6075 on the underlying docs/spec drift.

## Architecture

`src/core/signing/` owns the signing contract. Callers depend on it and on nothing beneath it — where the SDK is correct it delegates, where the SDK is wrong it carries the upstream fix verbatim with per-unit provenance pins. A structural guard forbids `adcp.signing` imports outside the layer, allowlist empty, mutation-verified. Consequence: a future SDK bump deletes copied functions and changes no caller.

Four SDK conformance bugs were found by grading against the real vectors and filed upstream — adcp-client-python #976 / #977 / #978 / #979, all fixed there — along with #975/#980 (the SDK's own vendored vector set was incomplete and its loader could not detect it).

## Verification

Full in-network suite: **12,652 passed / 0 failed** across unit, integration, bdd_inprocess, bdd_e2e, e2e, admin and ui — run `innet_210826_1308` (unit 6916 · integration 2668 · bdd_inprocess 2269 · bdd_e2e 543 · e2e 145 · admin 106 · ui 5). The earlier figure in this section (12,091) came from a superseded run and was not re-derived; it is replaced rather than relabelled.

**That run graded `4378e1e94` (2026-08-21 12:56), NOT this head.** Twenty-four commits have landed since, including D1 (`daabef6c3`) and the seven graduation commits (`0cebe90c9..f5e286fdd`). It is quoted for its suite breadth, which no CI job reproduces, and is **not** a statement about `f5e286fdd`. The current gate for this head is the **32/32 CI matrix on `f5e286fdd`**, which includes `BDD In-Network (e2e_rest)` — the job that grades the e2e_rest ledger — green.

Ratchets all tightened, none grew: duplication 74→73, mypy-untyped-defs 212→211, C901 183→182. **These are this PR's ORIGINAL span, and two of them are in a format the tree no longer uses** — the duplication baseline has since been split per-scope (`src`/`tests`/`scripts`), so "73" has no current counterpart. The truth-pass table below measures the LATER span (this epic's own before/after) at the final head. Both are true; they are different spans and, for duplication, different units. `mypy-untyped-defs 212→211` and `C901 183→182` are the same numbers seen from the two ends.

Signing behaviour is graded over **real HTTPS against the live stack** — a signature is produced by the server, the buyer's discovery chain is walked from served documents only (capabilities → `brand_json_url` → brand.json → JWKS), and the signature is verified from bytes fetched over the socket. Negative controls included: same `kid`, swapped key material must fail at the crypto step rather than at key lookup.

## Known open, tracked

| | |
|---|---|
| **#1890** (P1) | Outbound webhook delivery passes its own httpx client, so the SDK skips IP pinning — DNS-rebinding TOCTOU on buyer-supplied URLs. **Cannot be fixed in this PR**: the correct fix routes through the SSRF-guarded seam in #1802, which first needs a per-attempt signing hook (an RFC 9421 sender that signs once and retries is rejected by a conformant receiver on `negative/016-replayed-nonce`). This branch rebases onto #1802 and re-derives its destination-policy decision there. |
| `z6nr.15` | Sandbox grading endpoint: code-complete, pending an external grading run |
| `z6nr.21` | Six UC-010 xfail entries still parked on #1291; graduation is graded by e2e |
| **#2059** (P1) | **Buyer-facing.** `get_media_buy_delivery` serves **fabricated placement reporting data** to buyers with no testing gate, which also makes the AdCP `sort_by` fallback ungradable. Found by the graduation walk below. |
| **#2058** | Reporting webhook bodies violate AdCP 3.1.1 — a forbidden `aggregated_totals`, and an invalid flat body shape |
| **#2060** | Webhook circuit-breaker **recovery** path has zero behavioral coverage; a broken recovery silently reports serving campaigns as `reporting_delayed` |
| **#2061** | A `strict=False` duplicate xfail route silently converts a **future** production fix into a permanent XPASS. Measured live and reachable; `EXPECTED_XFAIL_ROUTES` records conditions only, never `strict`, so it cannot detect this. 62 tags routed more than once, one audited |
| **#2062** | Debt item C4's *"one change clears ~32 rows"* estimate over-counts — several xfails cited to it are coercion or silent-drop, not failed translation |

Deferred with rationale: KMS `SigningProvider`, the `canonical.py` seam deletion (fires on a future SDK bump, #1794), and the fabrication siblings found alongside `piyo` (#1845 and three beads).

---

## Truth pass (measured at `f5e286fdd`, not carried from notes)

**Measured at `f5e286fdd`. This section, the stack table at the top, and the Verification figures all go stale on every push, and must be RE-MEASURED — not patched row-by-row — before any further push.** It has already aged twice in exactly one way — an earlier revision cited `137b0e53d` while one row (the dark-primitives allowlist) was only true at `a7de571b3`, mixing two measurement points under one sha; a later one cited `a7de571b3` after D1 had changed the tree beneath it. **Any further push must re-measure the WHOLE table, not patch the row that moved.** A table that silently ages is worse than one that states its own expiry. It has now aged a third time — this revision re-derives every figure at `f5e286fdd`, and extends the expiry notice to cover the stack table and the Verification section, which were ageing silently with no marker at all.

Every claim below was re-derived from the tree at this head. Where an earlier note turned out already-true or stale, that is said rather than restated.

**Scope.** #1291 is explicitly inbound-only; this PR also reaches into the **outbound** webhook boundary. That was a deliberate keep, not drift: removing it would leave UC-010 rows *advertising* webhook signing with the enforcement gone — a direct O4 violation — and the extraction is semantic-merge-sized on a stacked branch for near-zero review benefit. A reviewer's frame should include outbound; #1291's text does not yet say so.

**xpass posture.** 23 xpasses as of `innet_210826_1308` (18 `bdd_e2e`, 5 `bdd_inprocess`) — not a graduation backlog to wave through. The mechanism matters: `Then` steps reading `env.mock["post"].call_count` are transport-blind, and `httpx.MockTransport` makes the logs *look* like real retries, so a vacuous pass and a real one are not distinguishable from the count alone. **This PR's scope CHANGED here, on an owner decision, and the previous revision's claim is now false.** It said graduation was *"tracked separately (`salesagent-n78j0.13`) and deliberately out of this PR's scope."* That was true when written. The owner then called for graduation to run **on this branch**; `salesagent-n78j0.13` is **CLOSED**, and seven of its rows are commits here (`0cebe90c9..f5e286fdd`). The section below is the record — it is not retroactive scope, it is a scope change, and it is stated as one so a reviewer can see why those commits exist.

**Deferrals that are owned, and one that is not.**
- `send_signed_challenge` opens a plain `httpx.AsyncClient` rather than the SDK's IP-pinned transport. **Owned**: the source states it at `webhook_sender_factory.py:580-586` with two issue citations (#1802, #1890), a rationale (adopting it changes behaviour for every existing receiver URL), and the named residual risk — the destination is validated once at fire time, not again at socket-open. That TOCTOU is stated, not hidden.
- **#1845 and #1867 are OPEN but UNASSIGNED.** An unassigned issue is not an established deferral. Both need an owner before they can be cited as one.

**Merge hazard to carry forward.** After #1802 lands, a *textually clean* merge can swap a pinned webhook dial for an unpinned one with every signing test still green — the tests grade signing, not which transport opened the socket.

**How to read the green.** CI ran on this branch for the first time on 2026-08-19; earlier heads carried only `check-pr-title` and `ipr-check`. The 32/32 on this head is a real full-matrix result, not a long-standing one.

### Xpass graduation — what it actually found

**THE RESIDUE WAS NOT A GRADUATION BACKLOG. It was a defect-discovery surface that occasionally retires a tag.** That is the finding a reviewer most needs here, because the tag counts alone read as underachievement. Two retirements out of seven rows looks thin until you see that all seven produced a defect or a correction.

`daabef6c3..f5e286fdd` is seven commits, seven rows — **7 inspected · 2 retired routing · 7 produced a finding or correction**:

| Commit | Row | Outcome |
|---|---|---|
| `0cebe90c9` | `T-UC-004-webhook-hmac` | **GRADUATED** |
| `ec66db971` | `T-UC-004-webhook-bearer` | **GRADUATED** |
| `dfae0e4a2` | `T-UC-004-webhook-no-aggregated` | PARKED → #2058 |
| `1aa29b7c9` | `T-UC-004-dim-sortby-fallback` | PARKED → #2059, **+ certificate corrected** |
| `e86d75daf` | `T-UC-004-webhook-circuit-recovery` | PARKED → #2060 |
| `f987106f4` | `T-UC-004-boundary-ownership` | PARKED (scenario under-specification) |
| `f5e286fdd` | ledger `:55` daily-breakdown | **STAY** + the `strict=False` shadow → #2061, #2062 |

Four defects filed (#2058, #2059, #2060, #2061 — plus #2062 on a stale debt estimate). **One prior graduation's certificate corrected** (`dim-sortby`, `1aa29b7c9`): it had cited an escape hatch as proof of the thing the escape hatch suppressed.

**Why so few retirements — the five failure modes.** Each row was graded on five questions, ordered cheapest-to-falsify:

| | question | what it asks |
|---|---|---|
| **Q1** | **seam** | Can the `Then` *see* the right thing at all? |
| **Q2** | **level** | Does it inspect the right *layer* of it? |
| **Q3** | **falsifiable** | Can the fixture make the assertion fail *at all*? |
| **Q4** | **fidelity** | Does the `When` do what the scenario *says*? |
| **Q5** | **certificate** | Does the comment claiming it was checked cite an *escape hatch* as proof? |

An xpass is **evidence, not a verdict** — it means either production now implements the behaviour, or the scenario is too weak to fail. Only the first graduates. A sixth mode has no question because none of the five asks it: *is it even an xpass?* At least one ledger row was never an xpass at all — `e2e_rest_known_failures.txt` is a known-**failures** list, and line `:55` genuinely XFAILs everywhere for a real, spec-grounded gap.

**Stopped by owner decision, with the remainder filed rather than dropped:** 10 rows unwalked (3 ledger lines + 7 tag-set entries, triage and per-row candidacy reads in **#2063**) and **61 of 62 doubly-routed tags unaudited** (#2061). Nothing was silently abandoned.

### Ratchets, stated because the charter requires it

| ratchet | before | at `f5e286fdd` |
|---|---|---|
| `check_untyped_defs` | 211 | 211 |
| duplication (`src` / `tests` / `scripts`) | 35 / 72 / 0 | 35 / **68** / 0 — hook prints `(unchanged)` for all three scopes |
| `check_fixme_citation_count` | 19 / 119 | 19 / 119 |
| `C901` / `PLR0912` / `PLR0915` / `F841` | — | 182 / 132 / 107 / 37 (unchanged) |
| `type-ignore` | — | 63 (unchanged) |
| `admin_get_db_session` / `admin_session_add` | — | 190 / 43 (unchanged) |
| `IMPL_SESSION_ALLOWLIST` | `set()` | `set()` |
| signing-layer boundary allowlists | — | `frozenset()` ×2 (`ADCP_SIGNING_ALLOWLIST`, `SUBMODULE_REACH_THROUGH_ALLOWLIST`) — still empty |
| dark-signing-primitives allowlist | — | `set()` — new, and **empty as of `a7de571b3`**; it shipped with one entry through `137b0e53d` |
| `EXPECTED_XFAIL_ROUTES` | — | 20 (unchanged) |
| `e2e_rest_known_failures.txt` | 17 | 17 — **unchanged through the whole graduation phase** |
| `_UC004_E2E_WEBHOOK_INTERNAL_TAGS` | 12 | **9** (12 → 11 pre-phase; 11 → 9 by the two graduations recorded above) |

Re-measurement is cheap and trustworthy here because it is **bounded**: `git diff --stat daabef6c3..f5e286fdd` touches exactly two files — `tests/bdd/conftest.py` and `tests/unit/test_e2e_rest_ledger_state.py`. No `src/` file, no baseline file, no allowlist file. The only ratchets that *could* have moved are the tag set (shrank) and the ledger (unchanged), and both were measured directly.

One breach occurred and was fixed at its cause rather than baselined: `e1f7588a2` took `check_untyped_defs` to 212 because `build_row` declared `operator: str` against a nullable column; the signature was widened and `--update-baseline` was never run.

**D1 landed in this PR rather than being deferred** (`daabef6c3`), on the owner's confirmation of his 2026-08-05 instruction. Two unvalidated `cast(BrandAgentType, …)` at the boundary that decides whether a request is *refused* became `narrow_agent_type()`, which raises `AdCPConfigurationError` on an unresolvable value — the `narrow_alg`/`narrow_purpose` shape this PR had already introduced two files away and then violated. A cast is a runtime no-op, so a typo in an env-overridable setting previously reached the resolver, matched no `agents[]` entry, and 401'd every signed counterparty with nothing naming the cause.

Worth stating because it is the argument for the change rather than a side note: **the two new types caught two real defects on contact.** `AnyUrl(jwks_origin)` exposed a `str | None` that mypy could not narrow through the `anchored` flag — an unvalidated URL had been reaching the *served* capabilities document via `cast(Any, …)`. `DeclaredBucket` exposed that the three `protocol_methods_*` buckets are generated `RootModel[str]` wrappers, not `str | Enum`. The `Any` was not hiding nothing.

One prescription in that scope was **unachievable and is recorded as such**: `posture_for_tenant(tenant: Tenant | None)` would require changing `_detect_tenant` in `resolved_identity.py`, which is identity resolution and out of scope under #1870. A `TypedDict` states the two keys actually read and the raw dict is projected at the signing layer's boundary; the ORM boundary stays where #1870 will find it.

---

## Self-review remediation: two epics, 33 commits (`f5e286fdd..21f1d8c53`)

The PR's own review artifact produced findings that did not share a symptom, but did share a root. Both epics were framed around that root rather than around the findings, so that closing a ticket removes a *class* of defect rather than an instance.

**Epic 1 — four architectural roots dissolved (13 lanes).** The headline defect was `SG-01`: a malformed `Signature` header was rejected only in the `required_for` bucket, so a present-but-malformed signature silently downgraded to bearer in every other posture. The spec places header well-formedness in a **pre-check phase, above the operation bucket**; the verifier had collapsed both phases into one. Fixed at `818af03f7` — the rejection now happens in every bucket, and the `none` bucket no longer records signed requests as `unsigned/"ignored"`.

**How the suite missed it, which is the more useful finding.** The test that should have caught it asserted a universal claim — *"a malformed signature blocks the bearer fallback regardless"* — over a fixture that declared exactly **one** bucket. Eight instances of that same *universal-oracle-over-point-fixture* shape were found and closed. It was systematic rather than accidental: the generated feature files carry a rule that each scenario differ by exactly one variable, and the 40 conformance vectors contain **zero** `warn_for` occurrences, because `warn_for` is a local extension the SDK's `VerifierCapability` drops on the floor.

**Epic 2 — "close the controls that only appear to be in force" (13 tickets, `salesagent-3ajkg`).** Scope test: a finding belongs here when something in the repository *claims* to enforce a rule and does not. Selected outcomes:

| ticket | the control that wasn't | commit |
|---|---|---|
| `.4` | vendored spec version restated as twelve literals instead of derived from the pin | `a77ca809a` |
| `.5` | signing-surface type discipline resting on a reviewer objecting, not on config — `mypy.ini` sections **5 → 9** | `c53f937ab` |
| `.6` | three hand-rolled declaration closures beside the factory that exists for them | `a4b9dd9c9` |
| `.7` | two byte-identical scrape guards whose *messages had already drifted* | `21f582391` |
| `.8` | `security.mdx@v3.1.1 :1464` "Sellers MUST log every request" implemented as a per-**registration** log — and the docstring recorded the narrowing as though it were the spec | `17f58bf6e` |
| `.9` / `.12` | five spec citations pointing at `dist/docs/3.1.1/`, a root that resolves at **no tag** | `5050df5be`, `19a837892` |
| `.13` | a bare generic reaching the SDK only via `dict[Any, Any]` compatibility | `5fe84449d` |
| — | the stale-citation guard carried markers for `rc.12` and missing-GA, but **none** for the dangling root it had just missed | `21f1d8c53` |

**`.8` is the one worth a reviewer's attention.** The spec obligation is per-*request*; the implementation logged per-*registration*, at the outbound delivery seam. Correcting it moved the seam: `_credentials_force_a_signature` reads `resolved.signature_forced **and** posture.supported`, so it is unreachable for a `supported: false` seller — and `:1465` routes exactly that population *into* the duty, naming the alternative reading ("3.0 migration note, not an exemption") in order to reject it. Logging at the escalation seam would have covered every credential-bearing request **except the ones the spec singles out**, with a green grader over it.

**Two limits are named rather than implied**, in the commit body and the ticket: requests on non-AdCP surfaces (`ADCP_SURFACE_PREFIXES`), and deployments with `verifier_enabled = false`, which return before the body is buffered.

**One departure from the BDD-first policy, recorded rather than quietly taken.** `.8`'s grader is an integration test, not a cross-transport scenario. That is forced, not chosen: a log record is process-local, and no carrier returns the running server's records to the test process — the only two `FileHandler`s in `src/` sit on `adcp.audit`, `/metrics` exposes counters rather than records, and the in-network runner holds no docker socket. A `caplog` `Then` on the `e2e_rest` leg would have seen nothing and **passed vacuously**. The e2e leg is left ungraded *on the record*, with the reasoning in the test's module docstring.

### Grading

Full six-suite in-network run at `17f58bf6e` — `test-results/innet_270826_0441`, `EXIT=0`:

| suite | passed | failed | xpass |
|---|---|---|---|
| admin | 106 | 0 | 0 |
| bdd | 2838 | 0 | 21 |
| e2e | 145 | 0 | 0 |
| integration | 2697 | 0 | 0 |
| ui | 5 | 0 | 0 |
| unit | 6934 | 0 | 0 |
| **total** | **12725** | **0** | |

Every suite hit a value **predicted before the run was read**. `integration 2695 → 2697` is the only movement and is the arithmetic of one test function with two parametrized cases; an unchanged count would have meant the grader never collected.

**Xpassed identity: 21 against 21, none appeared, none vanished**, compared on **bare node IDs with the prefix stripped on both sides** — the baseline file carries 16 `bdd_e2e::` and 5 `bdd_inprocess::`, and a local run emits a single `bdd.json`, so normalising one side alone reports 21-appeared/21-vanished when true drift is zero.

The four commits after `17f58bf6e` are graded by scoped runs with a stated basis rather than a second full suite: a `-U0` diff filter shows every changed line in `5050df5be` and `19a837892` is a comment, docstring, or markdown line — **zero executable lines** — and `5fe84449d`/`21f1d8c53` are graded by `mypy` clean over 322 files with the strict flags on, plus `unit 6936/0` (`6934` + the two new guard tests, so the arrival count pairs with the change).

### Ratchets

| ratchet | at `f5e286fdd` | at `21f1d8c53` |
|---|---|---|
| `check_fixme_citation_count` (src / tests) | 19 / 119 | **0** / 119 |
| duplication (`src` / `tests` / `scripts`) | 35 / 68 / 0 | 35 / 68 / 0 |
| `check_untyped_defs` | 211 | 211 |
| `type-ignore` | 63 | 63 |
| `C901` / `PLR0912` / `PLR0915` / `F841` | 182 / 132 / 107 / 37 | unchanged |
| `e2e_rest_known_failures.txt` | 17 | 17 |
| `mypy.ini` `[mypy-src…]` sections | 5 | **9** |

Every `src/` FIXME naming a beads id is gone (`350c8c562`, `427ac0be0`): a beads identifier resolves for nobody outside the session that wrote it, so it looked like tracking and reached no one. The 119 under `tests/` are deliberately retained — there a FIXME is frequently the *evidence itself* (an xfail reason, a known-gap marker) rather than a pointer to it.

### Filed, not swept

| issue | what it is |
|---|---|
| `xpfcm` | `BR-UC-010`'s citation fix reverts on the next regeneration — `compile_bdd.py` routes UC-010 through silent `TARGET-WINS`. **Measured**, not assumed. `21f1d8c53` turns that revert from silent into a red build. |
| `3kzh0` | 796 `ref=v3.1-04f59d2d5` citations. Ruled **drift** — every `ref=` must name 3.1.1. Not a swap: it means re-verifying each cited artifact still says what the citation claims *at* the pin. |

Both are out of scope for this PR. The merge-direction question behind `xpfcm` is recorded on the ticket: `compile_bdd.py` treats upstream `adcp-req` as authoritative, but for these files this repo holds the more recent, verified-at-3.1.1 text.

---

## Update — the #1721 architecture merge, and eleven regressions closed

**Measured at `a42865b6f`.** This section states its own scope: the figures below were
re-measured at that sha. The older tables above were measured at `f5e286fdd` and their
expiry notice still stands — they have **not** been re-measured here, and the ones that
need a full eight-suite run are called out at the end of this section rather than quietly
carried forward.

### What changed

This branch was re-based onto #1721's request boundary, which landed on `main` as a squash.
That is not a rebase in the ordinary sense: #1721 replaced the dispatch architecture
wholesale — one boundary `serve()` → `invoke_tool`, `CODE_TABLE` as the sole authority for
error text, and the deletion of `Transport.IMPL`, `ImplDispatcher` and
`synthesized_error_envelope`. The RFC 9421 work sits inside its identity resolver, so every
signing surface had to be re-expressed in those terms rather than merged textually.

The review surface was **725 files**, not the 180 a conflict count suggests; the auto-merged
files were the silent risk and were classified individually. Final PR diff: **479 files**.

Error raising was refactored to match the new architecture exactly: a raise site **names its
code by its class** and authors no message. 28 hand-written signature classes, no runtime
class generation — an SDK bump must produce the same classes, not regenerate different ones.

### The eleven

A full in-network run after the merge left eleven failures. Each was diagnosed to root cause
before any fix, and all eleven were **test-stale** — `production_change_needed: none` across
the board. They are closed:

| # | Group | What it actually was |
|---|---|---|
| 1–4 | deleted `GET /capabilities` | #1721 made `RestBinding.verb` `Literal["POST","PUT"]`; a GET is no longer expressible. Three e2e modules and one integration probe still addressed the deleted route and died at their first wire call, so nothing downstream had been exercised. |
| 5–6 | UC-004 e2e_rest webhooks | Four separate walls — see below. |
| 7 | stateless MCP mount | Asserted an `mcp-session-id` a deliberately stateless mount does not mint. The test's own recorded mutation no longer reddened it either, so it was **deleted** rather than repaired: under #1721 the failure mode it existed to catch is unreachable by construction. |
| 8 | malformed-signature probe | Carried **two** faults at once (no body *and* a malformed signature), so it graded neither. |
| 9 | provisioning audit sink | #1721 correctly restored auditing on the credential-minting route; its relative file sink then landed inside the test's sandbox. |
| 10–11 | template `url_for` scan | The scanner built its app from the **ambient environment**, so its verdict reported which environment the run happened in. |

Three of these deserve more than a row.

**#10–11 were not a ledger-bookkeeping job.** `create_app()` ran at module import, and
`create_app` registers the test-credential login blueprint only when `adcp_auth_test_mode
and not is_production`. The scanners are Jinja-blind, so they demanded that
`url_for('test_auth.test_auth')` — which sits inside `{% if test_mode %}` — resolve in an
app that deliberately did not compose that blueprint. A strict xfail on an environment fact
is red exactly where it is enforced and green everywhere else. The module now composes that
blueprint unconditionally, which is the **stricter** reading: every template branch is
scanned. Verified green under `ADCP_AUTH_TEST_MODE` both false and true. Integration ledger
`CEILING` 44 → 42. This does **not** discharge the open first-admin bootstrap question.

**#9 was relocated and then read.** Moving the sink alone would have spent the obligation
the snapshot assertion carried, so the relocated file is now graded: exactly one successful
`provision_signing_key` record naming the tenant, and **no key material in its bytes** —
checked against the encrypted PEM banner, the plaintext PKCS#8 banner, and the actual stored
ciphertext of both minted rows. The tempting alternative (absolutising `adcp_log_dir`) is
rejected in the docstring: it turns the test green while hiding every audit write from it
forever.

**#5–6 had four walls, and the fourth was an escape hatch.** The media buy was seeded
`pending_approval` while `get_media_buy_delivery` selects on `[active, completed]`, so the
server declined and sent zero POSTs. `assert_rejection_logged` filtered the *in-process*
sender's `task_type`, so on e2e_rest it could only ever find zero rows. The background sweep
(`DELIVERY_WEBHOOK_INTERVAL=5`) inflated the attempt count against a ~4s ladder. And the
backoff step was declared `e2e_unsupported`.

That declaration named what was missing — *"the capture service records no receipt times to
reconstruct them from"* — so the capability was **built rather than the declaration
re-justified**: every capture now carries a monotonic `received_at`, and the gaps between
consecutive attempts are the waits, measured on the wire. Attempts at one delivery are
separated from a concurrent sweep by **body bytes**, which retries share by signature
necessity (`content-digest` covers them) and separate deliveries never do — exact, not
heuristic. The hatch is gone and its pin shrank by one.

### Verification

| suite | result |
|---|---|
| `e2e` (in-network) | **138 passed, 0 failed** |
| `bdd_e2e` (in-network) | **1242 passed, 1598 xfailed, 14 xpassed, 0 failed** |
| `bdd` in-process (UC-004, all transports) | 570 passed, 492 xfailed, 0 failed |
| storyboard `signed_requests` | **mcp 59p/21f, a2a 59p/21f** — failure sets byte-identical, symmetric difference 0, **zero** signing failures |

| ratchet | before | now |
|---|---|---|
| integration ledger (`CEILING`) | 44 | **42** |
| `e2e_unsupported` declarations | 32 | **31** |
| `e2e_rest_known_failures.txt` | 11 | 11 |

Every repaired assertion was mutation-checked — the fix is not accepted until a deliberate
break in the thing it claims to grade turns it red, and the mutation is recorded in the
commit. For #10–11 that meant breaking the test-mode template branch **with the flag off**,
which was unreachable before the composition fix.

### Still outstanding, stated rather than carried silently

* The `f5e286fdd` tables above need a full eight-suite re-measure. This section re-measured
  only what it cites.
* One production question was raised by the UC-004 diagnosis and deliberately **not** folded
  into it: `delivery_webhook_scheduler.py:245` treats a degraded success (non-empty
  `errors[]`) as a hard failure and refuses to send, which reads against the advisory lane
  documented in `docs/design/error-architecture.md`. It predates this work. Filed separately;
  this PR must not be read as evidence about it.

---

### Footnote — the storyboard score depends on a locally patched conformance runner

The `signed_requests` figures above (**59 pass / 21 fail on both cards**) are measured with
four local fixes applied to `@adcp/sdk` through the existing `patch-package` precedent in
`tests/storyboard/runner/patches/`. Without them the same agent scores mcp 56p/25f and a2a
31p/52f — **those are runner defects, not verdicts about this agent**, and the four are kept
separable because each belongs to a different upstream conversation.

| fix | upstream | status |
|---|---|---|
| protocol-aware `resolveVectorTransport`; `jwks_override` synthesized 401 | [adcp-client#2958](https://github.com/adcontextprotocol/adcp-client/pull/2958) (closes #2954/#2955/#2956) | **merged** — our hunks are superseded and come out on the next SDK bump; upstream's `jwks_override` fix is the better one |
| `agentCapability` wiring + the `'either'` over-skip | [adcp-client#2963](https://github.com/adcontextprotocol/adcp-client/pull/2963) | open, **approved** |
| send an A2A request when the protocol under test is A2A | [adcp-client#2964](https://github.com/adcontextprotocol/adcp-client/pull/2964) | open |
| request-signing vectors need bodies a seller can parse | [adcp#7583](https://github.com/adcontextprotocol/adcp/pull/7583) | open, **approved** |

Related spec issue, filed and triaged, not patched here:
[adcp#7577](https://github.com/adcontextprotocol/adcp/issues/7577) — checklist step position
is graded in practice while `signed-requests.yaml` calls it informational.

**What this means for the number.** 59/21 is the score we expect to hold **once upstream
accepts these PRs or makes equivalent changes** — three of the four are already merged or
approved. It is not a score that depends on anything special about this agent's
configuration: the patches fix how vectors are dispatched and graded, not what the agent
answers. The 21 remaining failures are the same non-signing gaps both surfaces always had,
and the two cards' failure sets are identical, which is the property that says the grading
is now protocol-fair.


KonstantinMirin:
Deferred from PR #1721's review (round 2) — filed as issues since this branch now owns the files:

- #1877 — `NotificationProofService` claims webhook-control proof it never performs (`notification_proof_service.py:91-106`)
- #1878 — raw-session/raw-model construction in `admin/blueprints/accounts.py` + `order_approval_service.py`
- #1879 — type erasure (`Any`/`Mapping`) on `capabilities.py`/`capability_declarations.py`

All three should be addressed as part of this PR, not merged around. Full evidence in each issue.

KonstantinMirin:
Heads-up from #1802 (the outbound-egress seam), because this will land on you as a build failure rather than as a design question, and the design question is the more interesting one.

`src/core/signing/webhook_sender_factory.py:51` imports `httpx` directly. #1802 adds a src-scoped ruff TID251 ban on exactly that, run by `make quality-ci`. I checked this against this PR's head rather than guessing — dropping the file into a tree with #1802's config gives:

```
TID251 `httpx` is banned: Outbound HTTP goes through
src/core/security/outbound_http.py (send/asend) — the seam owns address,
TLS, redirect and retry policy (GH #1589).
  --> src/core/signing/webhook_sender_factory.py:51:8
```

Silencing it with a `# noqa: TID251` won't work either: `tests/unit/test_ruff_egress_bans.py::test_noqa_files_equal_recorded_constant` asserts set *equality* between the files carrying that noqa and a recorded `SEAM_FILES` constant, so a new exemption fails there instead. That's deliberate — the exemption list is meant to be a reviewed decision rather than something that accretes.

## The part worth actually deciding

The module docstring says destination policy is unchanged by C1:

> The sender is handed an `httpx.AsyncClient` we own, which puts it on the SDK's operator-supplied-client path — the operator's transport owns SSRF, exactly as today.

That was true against this branch's base. It stops being true once #1802 lands, because there the operator's transport *is* a pinned one: every webhook dial goes through `send`/`asend`, which builds `build_ip_pinned_transport` per call — DNS resolved once, connection pinned to the validated address. `adcp_webhook_sender` opening a plain `httpx.AsyncClient(timeout=...)` when the caller has none (and `send_signed_challenge` doing the same) means the sender skips that. So it isn't preserving behaviour after the merge; it's opting out of a protection that will exist by then. Same for the buyer-supplied webhook URLs it dials — that's where DNS rebinding actually matters.

## Two ways out, both already on #1802

Cheapest: pass `client=None` and let the SDK own its own client, which pins. Your own line 420-421 already documents that path.

Better, and the reason I'm writing rather than just filing: #1802 added a per-attempt signing hook to the seam for this exact case. `send`/`asend` take an optional `sign` callback invoked *inside* the retry loop, given `(method, url, body_bytes)` for that attempt and returning headers to merge. It exists because signing once above the loop replays the signature on retries — fine for legacy HMAC, wrong for RFC 9421, whose `nonce` a conformant receiver must reject on replay. Your own conformance suite ships `negative/016-replayed-nonce.json`, so a naive `send()` adoption would sign once, retry twice, and have both retries rejected — presenting as a flaky webhook rather than a bug.

With the hook you get a callback instead of a client: the dial stays on the seam's pinned transport, and the signature is fresh per attempt. That's one dial path in the codebase instead of two, which is the whole point of #1589.

## Practical suggestion

Rebase onto #1802 before you finish rather than after. It's green now (32/32) and only needs review. Rebasing first means you re-derive the destination-policy decision against the seam that will actually be there, instead of discovering it as a red build and reaching for a `noqa`.

Happy to pair on the `sign=` adoption if useful — the signature is in `src/core/security/outbound_http.py`, `SignAttempt`.

KonstantinMirin:
## Storyboard conformance: webhooks and signing both clean

Measured twice in-network (`innet_200926_0731`, `innet_200926_0742`), byte-identical:

| | mcp | a2a |
|---|---|---|
| passed | **101** | **101** |
| `signed_requests` failures | **0** | **0** |
| `webhook_emission` failures | **0** | **0** |

`webhook_emission` had never run at all before this branch — it reported `requirement_unmet: webhook_receiver` on both cards. Publishing the runner's receiver behind the TLS front made it red (13 checks), and four causes closed it:

1. **Runner matched brand.json on the card base.** `security.mdx` @ 3.1.1 :1142 step 1 defines `A` as the protocol endpoint; :1104 step 5 byte-equals `agents[].url` against it. Second call site of the defect in [adcp-client#2967](https://github.com/adcontextprotocol/adcp-client/pull/2967), patched locally in both runner bundles.
2. **The storyboard never discovers a product**, so its four `create_media_buy` steps ship the SDK's `"test-product"` placeholder and get `PRODUCT_NOT_FOUND`. `corrected_storyboards.py` inserts a `get_products` step into the discovery phase that already exists for it (adcp#7601).
3. **tox filtered `ADCP_SIGNING_DEV_KEK`** out of the seeder's env, so the tenant published an empty JWKS.
4. **We delivered every notification once per stored `PushNotificationConfig`** — the loop variable was never read, so the step's own registration went out once per row. Configs accumulate per operation: one create emitted the same payload to the same URL three times inside 17ms. This is the one agent defect of the four and the only production change; the loop is removed rather than guarded.

**Ledger: 46 → 16.** The fitness function named 30 entries as grading nothing once the fixtures were fixed. Removed. It only shrinks.

**Storyboard Conformance is still red**, on 8 checks per protocol, all `media_buy_seller/*` and none of them signing or webhook. They were masked before — the create path died on a fixture fault long before reaching them. Filed as `salesagent-basxl`; not ledgerable, since the allowlist only shrinks.

ChrisHuie:
## Review — RFC 9421 request signing

Reviewed `74a0d8a26`, plus `3ba4ec463` for the BDD surface. Spec references are `security.mdx` @ `v3.1.1`. Nits omitted.

**1 blocker · 24 should-fix.**

Checked by mutation and holding: verifier checklist order (9a → 10 → 13), per-attempt outbound signing, byte-identical refusal wire across REST/MCP/A2A, tenant scoping on every signing-key access, both migration downgrades against Postgres, and the duplication ratchet (base 25/45/0, head 25/43/0, both matching their committed baselines).

---

### Blocker

**Storyboard Conformance is red; main is green.** 16 failures on both heads, all `media_buy_seller/*`.

The defects are not yours — `src/core/tools/media_buy_update.py` is byte-identical to main. `req.canceled` is read only at `src/core/tools/media_buy_update.py:459`, and `_requested_actions` at `src/core/tools/media_buy_update.py:158-175` maps paused/budget/dates/packages only, so the first cancel is a no-op and the second can't return `NOT_CANCELLABLE`. This branch unmasks it by fixing the fixture faults that used to kill the create path first. Two others are runner gating defects.

So: not a fix request, a decision. Merging turns a green job on main red. Either ledger the eight with GitHub issue refs, or land the cancel fix separately first. A local task id won't resolve for anyone outside your session.

---

### Should-fix

**Recovery class is derived from how a code is spelled.** `src/core/errors/signature_codes.py:118` classifies all 28 codes by substring. The clear case is `request_signature_brand_json_malformed`: `_malformed` makes it `correctable`, but it is a counterparty document the buyer cannot edit, and L1122 says "do not retry; surface to operations". Because `message` and `suggestion` derive from the same class (`:155-171`), the buyer is also told to re-sign a request that was fine. The sibling splits are arguable rather than wrong — L1373 gives only `Retry? No` for every row, so `components_incomplete` vs `components_unexpected` landing in different classes is defensible; I'd still expect one condition family to share a remedy. Separately the docstring at `:126` says one code is genuinely correctable while the constant produces seven.

**Nothing grades any of it.** Dropping `"_required"` from the marker set flips a code's recovery and rewrites its wire text — 6196 unit, 68 integration, 114 BDD tests stay green. `tests/unit/test_architecture_error_recovery_enum_conformance.py:108` excludes the 28 on the grounds that the challenge string is pinned, but the challenge is a function of the code alone, so it can't catch a wrong recovery or a wrong sentence.

**The harness can't grade two of three transports.** `tests/harness/_base.py:531` refuses on `status_code >= 400` before parsing, so `wire_error_envelope` is `None` on MCP and A2A while the body carries a full envelope including `recovery`. The rationale is stated at `:507` — a bodyless 401 from the ASGI verifier — and that premise no longer holds now that verification moved into the resolver. `signing_enforcement.py` asserts only the header, so `recovery` is ungraded on every BDD transport too.

**A private key reaches the browser in a readable cookie.** `src/admin/blueprints/signing_keys.py:110-118` flashes the `env:` PEM, which serialises into the client-side session. Decoded from the running admin app with base64+zlib and no secret key — an encrypted PEM with a KEK configured, unencrypted PKCS#8 without one. The cookie flags at `src/admin/app.py:128-129` are pre-existing and not yours; routing key material through them is new. Render the hand-off from the POST instead of redirecting through a flash. The docstring's "never stored" is false as written. No test drives this path at all (`src/admin/blueprints/signing_keys.py:82-83`).

**Unauthenticated callers get internal issue numbers and operator prose.** `ConfigurationDetails.to_wire()` (`src/core/errors/details.py:509`) emits `tracked_by`, and a credential-less `POST /api/v1/capabilities` returns it carrying remediation text, a GitHub issue number, a module path and a DB constraint name. `details.py` is untouched here and one emitting arm predates you, so the systemic fix belongs in a follow-up. What is in scope: the three new payloads in `src/core/schemas/capability_declarations.py` are buyer-reachable. The other nine are admin-mint only — `provision_signing_key` has no caller outside the blueprint and the ops script. Worth noting `src/core/signing/provider.py:34` follows this rule and quotes the spec section forbidding it, so the two halves of the same change disagree.

**`expected_key_origins={}` opts into the stricter of two SDK behaviours — worth confirming that is what you want.** `src/core/signing/verifier.py:701` passes `{}` when the counterparty advertises no map, and your docstring at `:674-679` explains why: `None` makes the SDK skip the check with a warning. That is accurate, and the SDK's own warning text says to pass an empty dict "if the operator advertises no map and you want the missing-declaration rejection to fire" — so this is a documented opt-in, not a slip. The question is whether the spec supports it. L1108 and L1127 condition the rejection on a *declared* signing posture, and the schema says an absent map means shared-origin. The concrete case: a counterparty that signs outbound but declares no inbound buckets emits no `key_origins` — which is exactly what your own `posture.py:384` does, since it gates emission on `request_signing_buckets_declared` — and would be refused. Scope is bounded: the SDK returns early unless the JWKS came from brand.json, so publisher-pinned keys are unaffected. If you've already settled this against L1108, a line in the docstring saying so would stop the next reader re-opening it.

**Step 14 isn't implemented** (`src/core/signing/verifier.py:706`). L1246, on the request checklist: "Verifiers MUST reject bodies containing duplicate object keys (`request_body_malformed`)." Zero occurrences in `src/`; the SDK checklist returns after step 13. The primitive already exists at `src/core/security/webhook_strict_json.py:150`. No vector grades it, so it would ship unnoticed either way.

**`protocol_methods_required_for` is declarable but unenforced** (`src/core/signing/verifier.py:42-59`). The verifier's own docstring records that no protocol-method request reaches the bucket check, so a tenant can advertise a posture nothing applies. Against the stated goal that advertised equals enforced. A member gate refusing the declaration until the handlers reach the boundary would close it.

**An unreadable declaration fails open** (`src/core/signing/posture.py:277-284`). `posture_for_tenant` catches `AdCPSalesAgentError` and returns `UNSUPPORTED_POSTURE`, so a single typo in a declaration silently disables `required_for` enforcement for that tenant. The one log line interpolates `exc`, which after ADR-010 is just `Configuration error`; `exc.details` holds the block name. `record_boundary_error` logs `field` and `details` for precisely this reason elsewhere in this PR.

**Three new enforcement mechanisms don't enforce.**
- `tests/unit/test_architecture_signing_block_construction.py:70,161` uses relative `Path("src")`. I planted a violation and ran it twice: `1 failed` from the repo root, `13 passed` from `/tmp`. All 12 self-tests pass either way, so the self-tests can't catch it. The only one of 26 new guards not anchored on `repo_root()`.
- The three added exemptions (`ruff-boundary.toml:55`, `ruff-serialization.toml:51,52`) are inert today — I deleted each and both configs still pass. The ast-grep mirror of the serialization pair *is* load-bearing, so the "mirrored in" comment overstates the parity between the two mechanisms.
- The banned-api entry (`ruff-boundary.toml:189-201`) reaches the parent class only. A file importing a child and the parent gets one error, naming the parent. The comment says it reaches all 28, which would matter the day someone imports a child.

**Nine new branches no test evaluates**, each established by mutation rather than reading. The one that matters: `src/core/resolved_identity.py:215`, the operand stopping a verified signature from re-identifying an already-authenticated caller. Deleting it leaves 137 integration and 644 unit tests green — the test setup binds signer and bearer to one principal row, so nothing can distinguish them. Also `config.py:145`, `adcp_a2a_server.py:489`, `context_manager.py:850`, `errors/_entry.py:56`, `webhooks/delivery.py:163` (three operands), `http_utils.py:33-34`.

**Half of this morning's fix is ungraded** (`src/core/tools/media_buy_create.py:2814`, at `3ba4ec463`). The echo lands at two sites and the red-verification removed both. Removing only the manual-review arm leaves everything green; removing the auto-approve arm alone does redden. So the scenario is half-scoped rather than vacuous, and the commit's `3 failed / 21 passed` reads as covering both arms.

**Four claims about where coverage lives don't hold.**
- `tests/integration/test_mcp_tool_roundtrip_minimal.py:29` deleted as "belongs in BDD, parametrized" — nothing at head reads MCP text content or an A2A TextPart, so the obligation is gone rather than moved.
- `tests/harness/test_harness_circuit_breaker.py:36` cites three scenarios as the new home; this PR routes two of them to xfail and documents both as not grading production. The obligation does survive, stronger, at `tests/integration/test_delivery_service_behavioral.py:452` — so it's the citation to fix, not the coverage.
- `tests/bdd/conftest.py:1412` parks `identity-key-origins` with a reason saying the row "IS gradable today". The actual blocker is a missing step definition, and because the route fires imperatively it can never XPASS to tell you.
- `tests/bdd/steps/domain/uc002_create_media_buy.py:1831` describes a `PushNotificationConfig` read this PR deleted. Its six nodes XFAIL at the env gate, so it is wrong *and* dormant — not passing against broken production.

**Four smaller ones, same severity.** `src/core/http_utils.py:7` adds `headers_from_asgi_scope` with no production callers and last-wins repeat-header semantics, the inverse of the live `joined_headers` rule, and a guard constant pins it so that guard cannot currently fail. `mypy.ini:135-139` omits `disallow_any_explicit`, which the sibling `capabilities` section sets, leaving two explicit `Any`s at `capability_declarations.py:81` and `:747`. Three import cycles are held open only by lazy imports of leaf symbols (`src/core/signing/revocation.py:184`, `config.py:397`, `posture.py:201`), each of which a small relocation would break properly. And `templates/signing_keys_list.html:35-36` carries no CSRF token — repo-wide and pre-existing across 47 POST forms, but this is the first route whose cross-site invocation destroys cryptographic state.

**Also:** the deployment KEK is a published literal in `docker-compose.yml:96-97`, and the production-signal validator refuses three other test-kit relaxations but not this one. Nothing in `src/` writes `Principal.agent_url` (`src/core/database/models.py:698-710`), so bearer-less signed identity is unreachable for a real tenant and fails closed — which is also why the operand above has no oracle. The migration roundtrip (`scripts/ci/migration_helpers.py:162-166`) only downgrades one step, so `c41f8a6b2d90`'s destructive downgrade is never exercised in CI; generating that revision's SQL shows `ALTER TABLE principals ADD CONSTRAINT uq_principals_tenant_agent_url UNIQUE (tenant_id, agent_url)` (`alembic/versions/c41f8a6b2d90_add_inbound_request_signature_verification.py:48`), which takes ACCESS EXCLUSIVE on a populated table where `a164b85bab9e` uses `CONCURRENTLY`.

---

### Pre-existing — worth issues, not blocking this PR

Checked against `origin/main` rather than assumed. The `media_buy_seller` cancel defect (file byte-identical); `tracked_by` reaching the wire at all (`details.py` untouched); checklist-after-schema-validation; migration roundtrip depth (`migration_helpers.py` untouched); the admin cookie flags and the absent CSRF on 47 POST forms (`templates/signing_keys_list.html:35`); and `test_uc026_package_media_buy.py` in-suite instability — three whole-suite runs at one head gave three different verdicts, and a serial run on a fresh database put all 13 failures and 109 XPASSes in that one file, which passes 420/0 alone and which this PR does not touch.

---

### CodeQL log-injection, `src/core/signing/provider.py:57`

False positive. Every value reaching `diagnostic` is interpolated with `!r`, which escapes CR/LF; CodeQL doesn't model `repr()` as a sanitizer. All 13 `_refuse()` sites checked. The non-`!r` ones: `:190` a config list whose `__str__` reprs its elements, `:280` a `cryptography` `ValueError` whose two reachable messages echo no input, `:333` never request-derived, `:322`/`:333` `datetime.isoformat()`. Two hardening asks so the argument stops depending on a per-site audit: make `:280` `{exc!r}`, and add a test passing a CRLF-bearing diagnostic that asserts no raw newline lands in the record.

---

### Not checked

Full suite at either head; e2e and admin suites; the storyboard guards (pinned bundle absent locally, 34 skips); 16 of the 26 new guards were not mutation-tested; and the new head's storyboard failure count, so whether this morning's fix retired two of the sixteen is unknown.

KonstantinMirin:
## Response to review

Thank you — the mutation work is what made most of this actionable. Every point below is answered, and three of them found real defects I'd have shipped. Status against `<HEAD>`.

### Blocker — storyboard red, main green

**Taken your first option.** 16 entries ledgered in `tests/storyboard/known_failures.txt` (8 checks × 2 protocols), each citing a GitHub issue rather than a local id: gh#1041 (×2 paths), gh#1261, gh#1305, gh#1445, and three filed for this — gh#2267 (two packages naming one product refused; the pin has no `uniqueItems`), gh#2268 (the unknown-product refusal fails the response `oneOf`), gh#2269 (the mock adapter's 1M-impression cap refuses a conformance-sized buy).

`tests/helpers/storyboard_ledger_pin.py` moves in the same commit, as that file requires. The ledger's "only shrinks" rule is stated as an explicit exception in both files rather than quietly broken: these eight checks are pre-existing faults this branch made *reachable* by fixing the fixture faults that killed the create path first. The gap did not widen; the measurement did.

Your diagnosis was confirmed independently — `media_buy_update.py` is byte-identical to main.

### Fixed

**Recovery derived from spelling.** You were right, including about `brand_json_malformed`. Replaced with 28 explicit rows transcribed from L1373 and the discovery table at L1119-1127, each group citing its source. `_recovery_for` is now a lookup with **no default**, so an SDK taxonomy that grows raises at import rather than being classified by a fallback. The `:126` docstring you flagged went with the constant.

**Nothing grades any of it.** This one bit twice. The scenario now names the recovery class beside the code, and `assert_signature_challenge` takes `recovery` as a **required** keyword so no call site can omit it.

My first attempt read the expected class from `CODE_TABLE` — and your exact test destroyed it: flipping `request_signature_required` to `terminal` moved the wire and the expectation together, and every test stayed green. A classification checked against the thing that declares it grades nothing. The value now comes from `local-request-signing-enforcement.feature`, authored from the spec. Same mutation against the current version: **red on all three transports.**

**Harness can't grade two of three transports.** Correct, and you understated it: `recovery` was ungraded on **REST too**, because `assert_signature_challenge` read the header and nothing else. `_jsonrpc_body` now raises only on an empty or non-JSON body — parse first, classify second, the order REST already had and the E2E A2A leg was corrected to. `_a2a_task_outcome` threads the response into the failed-Task `WireError` so the challenge stays reachable. The stale "bodyless 401" rationale at `:507` is rewritten to say why it stopped being true.

**A private key reaches the browser.** Fixed more broadly than asked, because tracing it back the mint was the defect rather than the flash.

`env:` was the only mintable scheme that stored the private half nowhere, which is *why* it had to hand the PEM back. And it's the residue of a superseded design: `1b016fe55` (2026-07-28) stored private keys on disk with the row holding a scheme-prefixed reference; `6ae529688` (2026-07-29) replaced that with "encrypted in the row rather than on disk" and demoted `env:` to an option nobody removed. That commit already argued your case — "`env:` cannot be the default because keys are per tenant and minted at runtime".

So rather than render the hand-off from the POST: **`env:` minting is deleted**, and `provision_signing_key` returns a `str` kid. The return type has nowhere to put key material, so no provisioning path can hand any back. The `db:`-only collapse follows through — the scheme allowlist, the resolvers and `private_key_ref` itself are gone, since with the feature shipping whole in this PR there were never any legacy rows to accommodate.

The regression test is the one you'd want: POST the route with the form fields the page sent *before*, then read the session through Flask's own `session_transaction()` and assert no key material in body or cookie. Red on the old code, green now.

**Storyboard measured-status totals.** Not in your review — the ledger exposed it. `measured_status()` documents that a ledgered failure outranks the controller gate, but two of the four totals read the raw inputs (`measured_failing_protocols`, `requires_controller`) instead of the verdict, so a check that is both was counted twice. Latent until the ledger admitted the first controller-gated failure: 766 of 759. All four now read `r.measured`.

### Open, and owned

- **`tracked_by` to unauthenticated callers** — taking the in-scope half: the three buyer-reachable payloads in `capability_declarations.py`. The systemic `details.py` fix is a follow-up, as you scoped it. Your observation that `provider.py:34` quotes the spec section forbidding this is the sharpest part of the point.
- **`expected_key_origins={}`** — settled against L1108 as you guessed; owed a docstring line saying so.
- **Step 14 (duplicate object keys)** — unimplemented, primitive exists at `webhook_strict_json.py:150`. Needs a vector too, since nothing grades it either way.
- **`protocol_methods_required_for`** — refusing the declaration until the handlers reach the boundary.
- **Unreadable declaration fails open** — `posture.py:277-284`, logging `exc.details` and failing closed for that tenant.
- **Three mechanisms that don't enforce** — `repo_root()` in the guard, the inert exemptions deleted, the overstated banned-api comment corrected.
- **Nine unevaluated branches** — `resolved_identity.py:215` is the one that matters and has no oracle because nothing in `src/` writes `Principal.agent_url`. Needs a fixture binding signer and bearer to different principals.
- **Four coverage citations** — accepted, each is the citation to fix rather than the coverage.
- **Four smaller** — `headers_from_asgi_scope` deleted (no callers), `disallow_any_explicit` added with the two explicit `Any`s fixed, `{exc!r}` at `provider.py:280`. The import cycles and repo-wide CSRF are follow-ups.
- **Migration roundtrip / `CONCURRENTLY`** — `c41f8a6b2d90`'s downgrade, plus the roundtrip's one-step depth that hides it.

### Partially addressed

**Half the context-echo fix ungraded.** You're right and the gap is narrower but real. `1818ab8f9` adds element-level grading with a random per-package payload — the Then asserts positional per-element equality, so nothing about the payload's shape can be inferred by the production path. But the feature has **one** create scenario, so the manual-review and auto-approve arms still aren't both exercised. Recorded as open rather than claimed closed.

### Verification

Gate run: 21524 items, 15393 passed, **0 failed, 0 errors**, 0 regressions against baseline. Storyboard reproduces 102 passed / 15 failed on both protocols with 0 disparity across two independent runs. CI green.

ChrisHuie:
## Re-review at `7f2018abd`

Covers the six commits since `3ba4ec463` and the open items from the last review. Spec references are `security.mdx` / `core/error.json` @ v3.1.1 (`adcp==6.6.0`).

Fixed since last time:
- Provisioning returns a `str` kid, so key material can't reach the browser.
- The refusal body is graded on all three transports.
- The recovery table has no fallback and doesn't contradict v3.1.1.
- Storyboard Conformance is green.

### Corrections to my last review
- `config.py:145` is graded: `test_mcp_unknown_field_handling.py` runs with `ENVIRONMENT=production`.
- `webhooks/delivery.py:163` operands 2–4 predate this PR.
- Don't add `disallow_any_explicit` to the `capability_declarations` mypy section: it also fails on three pydantic class lines (`:320`, `:337`, `:375`). Typing `:81` as `object` and `:747` as `list[tuple[object, str, bool, str]]` is enough; `mypy src/` passes.
- The ACCESS EXCLUSIVE lock is in `c41f8a6b2d90`'s `upgrade()` (`op.create_unique_constraint` at `:48`), not its downgrade.

### Should-fix

**1. The KEK override described in `docker-compose.yml` doesn't work, and production accepts the dev KEK** (`docker-compose.yml:96`)
- `:96-97` are literal `environment:` entries, which Compose applies over `env_file: .env`. So the `.env` override the comment at `:93-94` tells real deployments to use is silently ignored. I checked this with a scratch compose file.
- This is the development file. But `docs/deployment/single-tenant.md:96-104` deploys with it, and `ENVIRONMENT=production` accepts the literal (`config.py:677` refuses the test-kit relaxations, not this).
- With `env:`/`file:` gone, every tenant's signing key is ciphertext under that passphrase, so anyone holding a DB dump of such a deployment can decrypt it. That does require database access, and no Postgres port is published.
- Fix:
  - Use `"${ADCP_SIGNING_KEY_PASSPHRASE_ENV:-ADCP_SIGNING_DEV_KEK}"` / `"${ADCP_SIGNING_DEV_KEK:-dev-only-signing-kek-not-a-secret}"`; with that, `.env` wins.
  - Refuse the literal under a production signal (it also appears at `docker-compose.e2e.yml:220` and `:584`).
  - Name the KEK in `.env.template`.

**2. A logged-in user who isn't a member of a tenant can make the agent sign with that tenant's key** (`src/core/creative_agent_registry.py:458`)
- `/api/formats/list` and `/search` are `@require_auth()` only and take `tenant_id` from the query string (`format_search.py:93`, `:34`). That read predates this PR; the signer on this path is new.
- Reproduced: such a session gets 403 on `/tenant/<victim>/signing-keys/` but 200 on `/api/formats/list?tenant_id=<victim>`, and the outbound `list_creative_formats` call carries the victim's `keyid`. The OAuth callback keeps `session["user"]` even when the account belongs to no tenant (`auth.py:527`).
- No key material leaves, and the call goes where that tenant's own format fetches go.
- Fix: check membership (super admin exempt), e.g. by moving the routes under `/api/tenant/<tenant_id>/formats/...` so `require_tenant_access` applies.

**3. After a KEK loss, the admin page and the setup checklist give the wrong answer** (`src/admin/blueprints/signing_keys.py:60`)
- With the KEK unset or wrong, the page still badges the active key "signing", because it doesn't check decryptability (`templates/signing_keys_list.html:71-72`).
- The checklist says "Published key is revoked or not yet active — provision a replacement" (`setup_checklist_service.py:1186`). Your runbook says the opposite: "restore the KEK. Do not re-mint" (`signing-key-runbook.md:242`).
- Fix: have the page use `signing_key_backed(...).signs` as the checklist and signer do, and give the checklist a distinct "active but can't be opened — check the KEK" message.

**4. The terminal suggestion gives the wrong remedy for most terminal codes** (`src/core/errors/signature_codes.py:212`)
- "…rotate or re-register the signing key" is sent for all 16 terminal codes. For these it points at the wrong fix:
  - `replayed`: use a fresh nonce.
  - `digest_mismatch`: recompute over the sent bytes and re-sign.
  - `revocation_stale`: the verifier's revocation list is stale.
  - The two key-origin codes and the five brand.json codes: document fixes, per L1119-1127.
- The correctable "missing or malformed" also goes to present, well-formed signatures (`alg_not_allowed`, `tag_invalid`, `window_invalid`, `components_unexpected`).
- Since the code is already on the wire twice, a per-code remedy discloses nothing new.
- Fix: text per code, from the spec's remediation column and #7647's `suggestion`.

**5. `2d325b348`'s `BREAKING CHANGE:` footer will cut a major release if squash-merged**
- The `7f2018abd` message notes the footer is moot. But this repo's squash message keeps sub-commit bodies (`COMMIT_MESSAGES`), and `b94ab979b` produced v1.7.0 → v2.0.0 the same way.
- Fix: add a `BEGIN_COMMIT_OVERRIDE … END_COMMIT_OVERRIDE` block to the description, or edit the squash message at merge.

### Spec questions

**6. `expected_key_origins={}` refuses signers that the schema doesn't let declare `key_origins`** (`src/core/signing/verifier.py:701`)
- With `{}`, adcp 6.6.0 refuses any brand.json-resolved signer that lacks `key_origins.request_signing`, without checking whether that signer declares a signing posture (`adcp/signing/key_origins.py:126-139`).
- v3.1.1's capabilities schema allows that entry only when the agent declares `request_signing` buckets ("MUST have a corresponding signing posture declared elsewhere"), and says an absent map means "assume shared-origin". L1127 also conditions the refusal on a declared posture.
- So a buyer that signs but doesn't verify can't comply either way. Your e2e counterparty passes only by declaring `key_origins.request_signing` with no `request_signing` block (`tests/e2e/counterparty_origin_service.py:149-155`).
- If you read L1108 as covering every signer, it seems worth raising upstream; otherwise pass `{}` only when the counterparty declares a posture.

**7. Six recovery rows now differ from adcp#7647** (`src/core/errors/signature_codes.py:137`)
- #7647 (closes your #7642; in the `v3.2.0-rc.6` prerelease) makes these correctable: `key_unknown`, `key_purpose_invalid`, `key_revoked`, `invalid`, `digest_mismatch` and `replayed`. They're terminal here, and the feature pins `digest_mismatch` as terminal.
- v3.1.1 doesn't decide these rows, so both are conformant at the pin.
- The `:156-157` rationale ("the bytes arrive identically on every retry") comes from the webhook paragraph (L1571); a request signer re-signs.
- Adopting them now avoids a flip at the bump.

### Still open from the last review
- Step 14 (duplicate keys) is unimplemented (`verifier.py:706`). Note that it must run before schema validation, which the boundary currently does first (`_boundary.py:230`, same order as main).
- `protocol_methods_required_for` can be declared but isn't enforced (`verifier.py:42-59`).
- An unreadable declaration still falls back to no enforcement (`posture.py:276-285`); your reply planned to fail closed.
- An anonymous `POST /api/v1/capabilities` against an unbacked declaration returns `tracked_by: "#1855 (no content-standards surface exists…)"` (`capability_declarations.py:163-167`).
- Nothing in `src/` writes `Principal.agent_url` (`models.py:698-710`), so signature-only identity can't work for a real tenant.

The test and tooling items from the last review still stand; I haven't repeated them here.

### Pre-existing (not blocking; worth issues)
- `/api/formats/search`, `/list` and `/agents` accept any `tenant_id` from any logged-in session (`format_search.py`).
- Create and update success responses can carry `errors`, which v3.1.1's success arms forbid (`_base.py:1125`, since `cfb7597da`). This is likely what fails `create_buy_with_lists`.
- Admin POST forms have no CSRF protection, and production sets `SameSite=None` (`app.py:129`).
- `ConfigurationDetails.to_wire()` sends `tracked_by` to unauthenticated callers (`details.py:509`).

---
Verified at `7f2018abd`: each item above was reproduced or read at this head. Not run: the full `./run_all_tests.sh ci`, e2e, admin, or the in-network storyboard.

Issue #1824, opened 2026-07-31
test(bdd): wire BR-UC-009 update_performance_index scenarios onto PerformanceEnv


KonstantinMirin:
Wire the BR-UC-009 `update_performance_index` scenarios onto `PerformanceEnv` so performance feedback is graded across transports against AdCP 3.1.1.

## Status: implemented, needs extracting

The implementation already exists and is green on `refactor/transport-boundary` (full suite `innet_300726_1114`, all 7 envs, exit 0). This ticket exists because that branch grew to 252 files and is being unbundled into reviewable PRs — one ticket per PR, per policy. **Do not re-implement**; extract the named commit onto a branch off latest `main`, run the suite, ship.

`main` currently wires **nothing** for this use case — every scenario xfails at the fixture via `_wire_simple_env`. Wiring does not force graduation: genuine spec-production gaps stay xfailed through `_SPEC_GAP_XFAILS`.

Umbrella: #1594 (covers UC-001/008/009/010 together — this ticket is its per-UC slice).
Precedent: #1555 (UC-019 wiring, same shape).

## Scope

Tags wired (5): `T-UC-009-main-mcp`, `T-UC-009-main-rest`, `T-UC-009-main-mcp-adapter`, `T-UC-009-main-rest-adapter`, `T-UC-009-main-mcp-audit`

Files: `tests/harness/performance.py` (new, +94), `tests/bdd/steps/domain/uc009_performance.py` (new), `tests/bdd/test_uc009_update_performance_index.py` (new), `wired_tags` entry.

Consumes `tests/bdd/steps/_datatable.py`, which lands with UC-001 (#1822).

Source commit: `0540ca47d`.

## Acceptance
- All 5 scenarios execute across the in-process transports, including the adapter and audit variants
- `./run_all_tests.sh` green; no ledger or allowlist growth
- Verified against **main**

PR #1837, opened 2026-08-03
test(bdd): wire BR-UC-001 get_products alt-empty/alt-filtered onto ProductEnv (#1822)

### LINKED ISSUE #1822: test(bdd): wire BR-UC-001 get_products scenarios onto ProductEnv (+ shared harness base)
Wire the BR-UC-001 `get_products` scenarios onto `ProductEnv` so discovery is graded across transports against AdCP 3.1.1.

## Status: implemented, needs extracting

The implementation already exists and is green on `refactor/transport-boundary` (full suite `innet_300726_1114`, all 7 envs, exit 0). This ticket exists because that branch grew to 252 files and is being unbundled into reviewable PRs — one ticket per PR, per policy. **Do not re-implement**; extract the named commit onto a branch off latest `main`, run the suite, ship.

`main` currently wires **nothing** for this use case — every scenario xfails at the fixture via `_wire_simple_env`. Wiring does not force graduation: genuine spec-production gaps stay xfailed through `_SPEC_GAP_XFAILS`.

Umbrella: #1594 (covers UC-001/008/009/010 together — this ticket is its per-UC slice).
Precedent: #1555 (UC-019 wiring, same shape).

## Scope

Tags wired (4): `T-UC-001-main`, `T-UC-001-alt-anonymous`, `T-UC-001-alt-empty`, `T-UC-001-alt-filtered`

Files: `tests/bdd/steps/domain/uc001_discover_inventory.py` (new), `tests/bdd/test_uc001_discover_available_inventory.py` (new), the `wired_tags` entry in `tests/bdd/conftest.py`.

**This slice also carries the shared harness/step base** the other three depend on — deliberately, so a reviewer sees each helper next to its first consumer rather than as infrastructure with no user:

- `tests/harness/_base.py` (+80/-22, `_json_safe` and friends)
- `tests/bdd/steps/_outcome_helpers.py` (+15, `wire_dict` / `wire_path`)
- `tests/bdd/steps/generic/_dispatch.py` (+6)
- `tests/bdd/steps/_datatable.py` (new — consumed by UC-009)

Ship this one first; #1823 (UC-008), #1824 (UC-009) and #1825 (UC-010) branch off `main` after it lands.

Source commit: `4e7b300d5`.

## Acceptance
- The 4 scenarios execute (not xfail-at-fixture) across the in-process transports
- `./run_all_tests.sh` green; no ledger or allowlist growth
- Verify against **main**, not the branch — a red result means the slice has a hidden dependency on branch production and must be resequenced

---

## Scope correction — this slice ships only what it actually grades

A wired scenario that is immediately strict-xfailed grades nothing; it is a dormant scenario with extra steps. So this ticket is de-scoped to the tags that genuinely assert, and each masked scenario moves to the production-gap ticket that unblocks it — **to be wired in the same change as the fix**, so the fix arrives verified and the wiring arrives grading.

**Ships (2 tags, genuinely grading):** `T-UC-001-alt-empty`, `T-UC-001-alt-filtered` — plus the shared harness/step base.

**Moves out:**
- `T-UC-001-main` → **#1595** — the wire never carries `brief_relevance`, a pinned v3.1.1 field production never serializes
- `T-UC-001-alt-anonymous` → **#1591** — token-less requests get `AUTH_REQUIRED` despite a public `brand_manifest_policy`; the gap is owned by the #1088 boundary work

This slice must be **fully green on `main`** — no new strict-xfail entries, no ledger growth. #1824 (UC-009) is the model: 5 of 5 tags grade, nothing masked.

KonstantinMirin:
Closes #1822.

Port-minus of `refactor/transport-boundary@4e7b300d5` onto main, per the issue's scope correction: only the two genuinely-grading BR-UC-001 `get_products` scenarios are wired — `T-UC-001-alt-empty` and `T-UC-001-alt-filtered`. The two masked tags move to their production-gap tickets, to be wired in the same change as the fix: `T-UC-001-main` → #1595 (relevance_score / brief_relevance never serialized), `T-UC-001-alt-anonymous` → #1591 (require_identity rejects token-less requests; owned by the #1088 boundary work). No `_UC001_XFAIL_TAGS` block ships — this slice adds zero strict-xfails, zero ledger entries, zero allowlist growth. The remaining 119 scenario blocks stay dormant at the UC-001 fixture catch-all; wiring forces no graduation.

## Spec grounding (AdCP 3.1.1, the pinned version via adcp==6.6.0)

- **alt-filtered**: `dist/schemas/3.1.1/core/product-filters.json` (via `get-products-request.json` `filters` $ref) — `delivery_type` (`enums/delivery-type.json`) and `countries` ("Filter by country coverage using ISO 3166-1 alpha-2 codes… Works for all inventory types.").
- **alt-empty**: `dist/schemas/3.1.1/media-buy/get-products-response.json` — `products` is an unconstrained "Array of matching products"; empty is a valid completed success (feature POST-S1).
- **format_id federation contract** (asserted on the real products wire body): `core/format-id.json` — every `format_id` is an object carrying `agent_url` + `id`, never a bare string.
- Conformance storyboards: `dist/compliance/3.1.1/domains/media-buy/scenarios/` grades only `get_products_async`, `refine_products`, `product_signal_targeting` — sync empty-result and structured-filter behavior are **ungraded**; both scenarios are schema-grounded.

## What's in the diff

- `tests/bdd/steps/domain/uc001_discover_inventory.py`, `tests/bdd/test_uc001_discover_available_inventory.py` — new, from the donor commit.
- `tests/bdd/conftest.py` — three hunks: plugin registration, `_detect_uc` UC-001 branch, `_harness_env` UC-001 block seeding a 3-product catalog (US/guaranteed, GB/non_guaranteed, one restricted to another principal) so visibility/filter asserts are non-vacuous.
- `tests/bdd/steps/_datatable.py` + `tests/bdd/steps/generic/given_config.py` — the shared `datatable_to_dicts` helper (donor `9988e37d5`), replacing given_config's private copy; whitespace-stripping folded in. Its second consumer (UC-009) arrives with #1824.
- `tests/bdd/steps/generic/then_success.py` — fix caught during verification: the ported status-completed step shadowed the generic one for every use case sharing that step text, breaking 3 UC-005 scenarios. Resolved by deleting the duplicate and extending the generic step's statusless-success list with `products`.
- `tests/integration/conftest.py` — cherry-pick of an existing fix for a real infra deadlock hit during the full run (mcp_server fixture wrote subprocess output to undrained pipes; now writes to files). May land independently from its home branch; harmless either way.
- `CLAUDE.md` — stale spec-pin sentence corrected to 3.1.1 / adcp==6.6.0 (matches `pyproject.toml` and `docs/adcp-spec-version.md`).

The issue's "shared harness base" (`_base.py`, `_outcome_helpers.py`, `_dispatch.py`) needed **no porting** — main already carries those changes (and is ahead of the donor branch in places), so copying donor files would have regressed main. Only the datatable helper was actually missing.

## Verification

Full suite (all 7 envs, `innet_310726_2139`): both wired scenarios are plain PASS on **all four transports** — a2a/mcp/rest in-process and e2e_rest in the bdd-in-network run — on main production unchanged (this branch contains zero `src/` changes). unit 5663 / integration 2226 / bdd_inprocess 1753 / bdd_e2e 393 / admin 86 / ui 5 passed, zero failures in every suite except `tests/e2e/`, where 6 pre-existing webhook failures (5× `test_a2a_webhook_payload_types.py`, 1× `test_delivery_webhooks_e2e.py`) trace to an already-tracked bug (SSRF validation from #1697 blocking the e2e webhook-capture server's Docker-internal hostname). This branch touches neither `src/` nor `tests/e2e/`, so it cannot have introduced them.

### DIFF
--- a/tests/bdd/conftest.py
+++ b/tests/bdd/conftest.py
@@ -58,6 +58,7 @@ pytest_plugins = [
     "tests.bdd.steps.domain.uc004_delivery",
     "tests.bdd.steps.domain.uc002_create_media_buy",
     "tests.bdd.steps.domain.uc002_nfr",
+    "tests.bdd.steps.domain.uc001_discover_inventory",
     "tests.bdd.steps.domain.uc003_update_media_buy",
     "tests.bdd.steps.domain.uc003_ext_error_scenarios",
     "tests.bdd.steps.domain.uc006_sync_creatives",
@@ -3075,6 +3076,8 @@ def _setup_existing_media_buy(ctx: dict, env: object, tenant: object, principal:
 def _detect_uc(request: pytest.FixtureRequest) -> str | None:
     """Detect which use case a BDD scenario belongs to via its tags."""
     marker_names = {m.name for m in request.node.iter_markers()}
+    if any(t.startswith("T-UC-001-") for t in marker_names):
+        return "UC-001"
     if any(t.startswith("T-UC-002") for t in marker_names):
         return "UC-002"
     if any(t.startswith("T-UC-003") for t in marker_names):
@@ -3517,6 +3520,39 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
                 yield
         else:
             pytest.xfail(f"UC-004 harness not yet wired for type: {harness_type}")
+    elif uc == "UC-001":
+        # get_products discovery (salesagent-cfrg port-minus of salesagent-8wf2/pli8).
+        # The wired set runs a real get_products through every transport on
+        # ProductEnv; the seeded catalog (open US/guaranteed, open GB/non_guaranteed,
+        # and one product restricted to another principal) makes the visibility
+        # and filter asserts non-vacuous. T-UC-001-main (#1595) and
+        # T-UC-001-alt-anonymous (#1591) are not wired on this slice. Everything
+        # else stays dormant here: wiring never forces graduation.
+        _UC001_WIRED = {
+            "T-UC-001-alt-empty",
+            "T-UC-001-alt-filtered",
+        }
+        marker_names = {m.name for m in request.node.iter_markers()}
+        if marker_names & _UC001_WIRED:
+            from tests.factories import PricingOptionFactory, ProductFactory
+            from tests.harness.product import ProductEnv
+
+            with _db_scope_for(request, e2e_config), ProductEnv(e2e_config=e2e_config) as env:
+                tenant, principal = env.setup_default_data()
+                ctx["env"] = env
+                ctx["tenant"] = tenant
+                ctx["principal"] = principal
+                open_us = ProductFactory(tenant=tenant, countries=["US"])
+                PricingOptionFactory(product=open_us)
+                open_gb = ProductFactory(tenant=tenant, delivery_type="non_guaranteed", countries=["GB"])
+                PricingOptionFactory(product=open_gb)
+                restricted = ProductFactory(tenant=tenant, countries=["US"], allowed_principal_ids=["someone-else"])
+                PricingOptionFactory(product=restricted)
+                ctx["seeded_products"] = [open_us, open_gb, restricted]
+                ctx["restricted_product_ids"] = {restricted.product_id}
+                yield
+        else:
+            pytest.xfail("UC-001 harness not yet wired for this scenario (salesagent-cfrg)")
     elif uc == "UC-GET-PRODUCTS":
         from tests.harness.product import ProductEnv
 
--- a/tests/bdd/steps/_datatable.py
+++ b/tests/bdd/steps/_datatable.py
@@ -1,0 +1,21 @@
+"""Shared pytest-bdd datatable parsing helpers.
+
+One canonical headered-table parser for all step modules — per-module copies
+drift (the uc009 copy stripped whitespace, the generic one didn't).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+
+def datatable_to_dicts(datatable: Sequence[Sequence[object]]) -> list[dict[str, str]]:
+    """Convert a pytest-bdd raw datatable (list of lists) to one dict per row.
+
+    The first row is treated as column headers. Remaining rows become dicts
+    keyed by those headers. Header and cell values are str()-coerced and
+    whitespace-stripped, so Gherkin table padding never leaks into keys or
+    values.
+    """
+    headers = [str(cell).strip() for cell in datatable[0]]
+    return [{h: str(v).strip() for h, v in zip(headers, row, strict=True)} for row in datatable[1:]]
--- a/tests/bdd/steps/domain/uc001_discover_inventory.py
+++ b/tests/bdd/steps/domain/uc001_discover_inventory.py
@@ -17,7 +17,7 @@ from typing import Any
 
 from pytest_bdd import given, parsers, then, when
 
-from tests.bdd.steps._outcome_helpers import _require_response, wire_field
+from tests.bdd.steps._outcome_helpers import wire_field
 from tests.bdd.steps.generic._dispatch import dispatch_request
 
 
@@ -117,13 +117,6 @@ def when_send_get_products(ctx: dict, datatable: list) -> None:
 def _wire_products(ctx: dict) -> list[dict[str, Any]]:
     """The products array as the buyer sees it on the serialized wire."""
     return wire_field(ctx, "products")
-
-
-@then(parsers.parse('the response status should be "completed"'))
-def then_status_completed(ctx: dict) -> None:
-    """get_products succeeded — a completed synchronous discovery response."""
-    assert ctx.get("error") is None, f"Request failed: {ctx.get('error')}"
-    assert _require_response(ctx).products is not None
 
 
 @then('the response should contain "products" array')
--- a/tests/bdd/steps/domain/uc001_discover_inventory.py
+++ b/tests/bdd/steps/domain/uc001_discover_inventory.py
@@ -1,0 +1,235 @@
+"""Steps for UC-001 discover available inventory (get_products) — salesagent-8wf2.
+
+Wired scenarios dispatch a real get_products through the parametrized wire
+transport (a2a/mcp/rest) via ProductEnv. The catalog is seeded by the UC-001
+conftest branch (three products: guaranteed/US, non_guaranteed/GB, and one
+restricted to another principal) so visibility and filter assertions are
+non-vacuous. Success-path asserts read the REAL serialized wire body via
+``wire_field`` where the transport stashes it (REST always; A2A/MCP via the
+client/handler dispatchers), so the v3.1.1 format_id object contract on the
+products wire is graded here (r50r follow-up landed with 8wf2).
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from pytest_bdd import given, parsers, then, when
+
+from tests.bdd.steps._outcome_helpers import _require_response, wire_field
+from tests.bdd.steps.generic._dispatch import dispatch_request
+
+
+def _datatable_to_kwargs(datatable: list) -> dict[str, Any]:
+    """Convert a two-column field/value datatable to request kwargs.
+
+    JSON-looking values (objects, arrays) are parsed; everything else stays a
+    string. ``buying_mode`` is dropped: production defaults it to brief
+    (BR-RULE-079 pre-v3 fallback) and GetProductsBody does not declare it yet
+    (spec-gap tracked in salesagent-ilbv) — sending it would trip the Pattern
+    #7 dev-forbid gate on REST rather than exercise discovery.
+    """
+    headers = [str(h).strip() for h in datatable[0]]
+    field_idx, value_idx = headers.index("field"), headers.index("value")
+    kwargs: dict[str, Any] = {}
+    for row in datatable[1:]:
+        field = str(row[field_idx]).strip()
+        raw = str(row[value_idx]).strip()
+        if field == "buying_mode":
+            continue
+        if raw.startswith(("{", "[")):
+            kwargs[field] = json.loads(raw)
+        else:
+            kwargs[field] = raw
+    return kwargs
+
+
+# ── Given steps ─────────────────────────────────────────────────────
+
+
+@given("a tenant exists with at least one product in the catalog")
+def given_tenant_with_catalog(ctx: dict) -> None:
+    """Background step: the UC-001 conftest branch seeded tenant + catalog."""
+    env = ctx["env"]
+    tenant, principal = env.setup_default_data()
+    ctx.setdefault("tenant", tenant)
+    ctx.setdefault("principal", principal)
+    assert ctx.get("seeded_products"), "UC-001 conftest branch did not seed the catalog"
+
+
+@given(parsers.parse('the tenant brand_manifest_policy is "{policy}"'))
+def given_brand_manifest_policy(ctx: dict, policy: str) -> None:
+    """Set the tenant's brand manifest policy column (products.py:194 reads it)."""
+    env = ctx["env"]
+    tenant, principal = env.setup_default_data()
+    ctx.setdefault("tenant", tenant)
+    ctx.setdefault("principal", principal)
+    tenant.brand_manifest_policy = policy
+    env.get_session().commit()
+
+
+@given("the tenant has an advertising_policy configured")
+def given_advertising_policy(ctx: dict) -> None:
+    """Give the tenant a non-empty advertising policy."""
+    env = ctx["env"]
+    tenant, _ = env.setup_default_data()
+    tenant.advertising_policy = {"prohibited_categories": ["weapons"]}
+    env.get_session().commit()
+
+
+@given(
+    "the product catalog contains products with valid schema "
+    "(format_ids, publisher_properties, pricing_options, reporting_capabilities)"
+)
+def given_valid_catalog(ctx: dict) -> None:
+    """The conftest UC-001 branch seeded the catalog; assert it is present."""
+    assert ctx.get("seeded_products"), "UC-001 conftest branch did not seed the catalog"
+
+
+@given("no products match the specified filters and brief")
+def given_no_matching_products(ctx: dict) -> None:
+    """Constrain the request so the seeded catalog matches nothing.
+
+    The catalog rows exist (seeded by the fixture); 'no products match the
+    SPECIFIED filters' is made literally true by specifying a country filter
+    no seeded product serves.
+    """
+    ctx.setdefault("request_preset", {})["filters"] = {"countries": ["ZZ"]}
+
+
+# ── When step ───────────────────────────────────────────────────────
+
+
+@when("the Buyer Agent sends a get_products request with:")
+def when_send_get_products(ctx: dict, datatable: list) -> None:
+    """Dispatch get_products with the table fields through the wire transport."""
+    kwargs = {**_datatable_to_kwargs(datatable), **ctx.get("request_preset", {})}
+    if "dispatch_identity" in ctx:
+        dispatch_request(ctx, identity=ctx["dispatch_identity"], **kwargs)
+    else:
+        dispatch_request(ctx, **kwargs)
+
+
+# ── Then steps ──────────────────────────────────────────────────────
+
+
+def _wire_products(ctx: dict) -> list[dict[str, Any]]:
+    """The products array as the buyer sees it on the serialized wire."""
+    return wire_field(ctx, "products")
+
+
+@then(parsers.parse('the response status should be "completed"'))
+def then_status_completed(ctx: dict) -> None:
+    """get_products succeeded — a completed synchronous discovery response."""
+    assert ctx.get("error") is None, f"Request failed: {ctx.get('error')}"
+    assert _require_response(ctx).products is not None
+
+
+@then('the response should contain "products" array')
+def then_contains_products_array(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    assert isinstance(products, list), f"products is not an array on the wire: {type(products).__name__}"
+    assert products, "products array is empty — seeded catalog should be visible"
+    # v3.1.1 core/format-id.json: every format_id on the products wire is an
+    # object carrying agent_url + id, never a bare string (federation contract).
+    from tests.helpers.format_assertions import assert_wire_format_id_is_object
+
+    for product in products:
+        for fid in product.get("format_ids", []):
+            assert_wire_format_id_is_object(fid)
+
+
+@then('the response "products" array should be empty')
+def then_products_array_empty(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    assert products == [], f"expected empty products array, got {len(products)} entries"
+
+
+@then("every product should have pricing_options as an empty array")
+def then_pricing_suppressed(ctx: dict) -> None:
+    """Anonymous requests get pricing suppressed — only meaningful non-vacuously."""
+    products = _wire_products(ctx)
+    assert products, "no products returned — pricing suppression is unobservable on an empty catalog"
+    for product in products:
+        assert product.get("pricing_options") == [], (
+            f"pricing_options not suppressed for anonymous buyer on {product.get('product_id')!r}: "
+            f"{product.get('pricing_options')!r}"
+        )
+
+
+@then("no products with allowed_principal_ids restrictions should be visible")
+def then_restricted_products_hidden(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    restricted_ids = ctx.get("restricted_product_ids", set())
+    assert restricted_ids, "fixture seeded no restricted product — assertion would be vacuous"
+    visible = {p.get("product_id") for p in products}
+    leaked = visible & restricted_ids
+    assert not leaked, f"restricted products visible to anonymous buyer: {leaked}"
+
+
+@then(parsers.parse('every product should match the delivery_type "{expected}"'))
+def then_products_match_delivery_type(ctx: dict, expected: str) -> None:
+    products = _wire_products(ctx)
+    assert products, "no products returned — delivery_type filter is unobservable on an empty result"
+    for product in products:
+        assert product.get("delivery_type") == expected, (
+            f"{product.get('product_id')!r} has delivery_type {product.get('delivery_type')!r}, expected {expected!r}"
+        )
+
+
+@then(parsers.parse("every product should have countries overlapping with {countries}"))
+def then_products_match_countries(ctx: dict, countries: str) -> None:
+    """Country filtering is observable by product identity, not a wire field.
+
+    ``Product.countries`` is an internal filter field (``exclude=True`` in the
+    schema — never serialized), so the filter's effect is asserted against the
+    seeded catalog: only products whose DB row serves an overlapping country
+    may come back, and at least one must (the catalog seeds a matching one).
+    """
+    requested = set(json.loads(countries))
+    products = _wire_products(ctx)
+    assert products, "no products returned — country filter is unobservable on an empty result"
+    matching_ids = {p.product_id for p in ctx["seeded_products"] if set(p.countries or []) & requested}
+    returned_ids = {p.get("product_id") for p in products}
+    off_target = returned_ids - matching_ids
+    assert not off_target, (
+        f"products not serving {sorted(requested)} leaked through the country filter: {sorted(off_target)}"
+    )
+
+
+@then(
+    "each product should have product_id, name, format_ids, publisher_properties, "
+    "pricing_options, and reporting_capabilities"
+)
+def then_products_have_required_fields(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    assert products, "no products returned"
+    required = ("product_id", "name", "format_ids", "publisher_properties", "pricing_options", "reporting_capabilities")
+    for product in products:
+        for field in required:
+            assert product.get(field) is not None, f"{product.get('product_id')!r} missing required field {field!r}"
+
+
+@then("the products should be ordered by relevance_score descending")
+def then_products_ordered_by_relevance(ctx: dict) -> None:
+    """Production gap: relevance_score is not emitted on the products wire.
+
+    The impl sorts internally when AI ranking is enabled but never serializes
+    a relevance_score field. Dormant until T-UC-001-main is wired (#1595).
+    """
+    products = _wire_products(ctx)
+    scores = [p.get("relevance_score") for p in products]
+    assert all(s is not None for s in scores), f"relevance_score missing on the wire: {scores}"
+    assert scores == sorted(scores, reverse=True), f"products not ordered by relevance_score desc: {scores}"
+
+
+@then("each product should include brief_relevance explanation")
+def then_products_have_brief_relevance(ctx: dict) -> None:
+    """Production gap twin of the relevance ordering assert. Dormant until T-UC-001-main is wired (#1595)."""
+    products = _wire_products(ctx)
+    assert products, "no products returned"
+    for product in products:
+        assert product.get("brief_relevance"), (
+            f"{product.get('product_id')!r} has no brief_relevance explanation on the wire"
+        )
--- a/tests/bdd/steps/generic/then_success.py
+++ b/tests/bdd/steps/generic/then_success.py
@@ -34,6 +34,7 @@ _STATUSLESS_SUCCESS_ATTRS: tuple[str, ...] = (
     "formats",  # ListCreativeFormatsResponse
     "media_buy_deliveries",  # GetMediaBuyDeliveryResponse
     "aggregated_totals",  # GetMediaBuyDeliveryResponse
+    "products",  # GetProductsResponse
 )
 
 
@@ -83,7 +84,7 @@ def then_response_status(ctx: dict, status: str) -> None:
         assert value is not None, (
             f"Status 'completed' claimed but response.{attr} is None — the schema-required success payload is missing"
         )
-        if attr in ("formats", "media_buy_deliveries"):
+        if attr in ("formats", "media_buy_deliveries", "products"):
             assert isinstance(value, list), (
                 f"Status 'completed' claimed but response.{attr} is {type(value).__name__}, expected a list"
             )
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -510,10 +510,13 @@ def sample_products(integration_db, sample_tenant):
 @pytest.fixture(scope="function")
 def mcp_server(integration_db):
     """Start a real MCP server for integration testing using the test database."""
+    import shutil
     import socket
     import subprocess
     import sys
+    import tempfile
     import time
+    from pathlib import Path
 
     # Find an available port
     def get_free_port():
@@ -564,13 +567,34 @@ from src.core.main import mcp
 mcp.run(transport='http', host='0.0.0.0', port={port})
 """
 
-    process = subprocess.Popen(
-        [sys.executable, "-c", server_script],
-        env=env,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        bufsize=1,  # Line buffered
-    )
+    # The server's output goes to FILES, never PIPEs. A PIPE nobody drains caps
+    # at 64KB; once server logging fills it, the server blocks on a log write
+    # INSIDE a request handler and the calling test awaits forever — this wedged
+    # every full CI run at integration's quiet tail until the >1h run reaper
+    # killed it (salesagent-cu12). Files keep the error-path diagnostics below
+    # without needing a drainer thread.
+    output_dir = Path(tempfile.mkdtemp(prefix=f"mcp-server-{port}-"))
+    stdout_path = output_dir / "stdout.log"
+    stderr_path = output_dir / "stderr.log"
+    stdout_f = stdout_path.open("wb")
+    stderr_f = stderr_path.open("wb")
+    try:
+        process = subprocess.Popen(
+            [sys.executable, "-c", server_script],
+            env=env,
+            stdout=stdout_f,
+            stderr=stderr_f,
+        )
+    finally:
+        # The child inherited the fds; the parent's handles are not needed.
+        stdout_f.close()
+        stderr_f.close()
+
+    def _server_output() -> str:
+        return (
+            f"STDOUT: {stdout_path.read_text(errors='replace') or 'N/A'}\n"
+            f"STDERR: {stderr_path.read_text(errors='replace') or 'N/A'}"
+        )
 
     # Wait for server to be ready.
     # Server startup is dominated by Python imports (fastmcp + adcp SDK + project)
@@ -593,29 +617,13 @@ mcp.run(transport='http', host='0.0.0.0', port={port})
         except (ConnectionRefusedError, OSError):
             # Check if process has died
             if process.poll() is not None:
-                stdout, stderr = process.communicate()
-                raise RuntimeError(
-                    f"MCP server process died unexpectedly.\n"
-                    f"STDOUT: {stdout.decode() if stdout else 'N/A'}\n"
-                    f"STDERR: {stderr.decode() if stderr else 'N/A'}"
-                )
+                raise RuntimeError(f"MCP server process died unexpectedly.\n{_server_output()}")
             time.sleep(0.3)
 
     if not server_ready:
-        # Capture output for debugging
-        try:
-            stdout, stderr = process.communicate(timeout=2)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            stdout, stderr = process.communicate()
-
-        process.terminate()
+        process.kill()
         process.wait(timeout=5)
-        raise RuntimeError(
-            f"MCP server failed to start on port {port} within {max_wait}s.\n"
-            f"STDOUT: {stdout.decode() if stdout else 'N/A'}\n"
-            f"STDERR: {stderr.decode() if stderr else 'N/A'}"
-        )
+        raise RuntimeError(f"MCP server failed to start on port {port} within {max_wait}s.\n{_server_output()}")
 
     # Return server info
     class ServerInfo:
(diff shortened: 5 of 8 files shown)


tests/bdd/conftest.py
```diff
@@ -3517,6 +3520,39 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
...
+            with _db_scope_for(request, e2e_config), ProductEnv(e2e_config=e2e_config) as env:
+                tenant, principal = env.setup_default_data()
+                ctx["env"] = env
+                ctx["tenant"] = tenant
+                ctx["principal"] = principal
+                open_us = ProductFactory(tenant=tenant, countries=["US"])
+                PricingOptionFactory(product=open_us)
+                open_gb = ProductFactory(tenant=tenant, delivery_type="non_guaranteed", countries=["GB"])
+                PricingOptionFactory(product=open_gb)
+                restricted = ProductFactory(tenant=tenant, countries=["US"], allowed_principal_ids=["someone-else"])
```
ChrisHuie, line 3549: **BLOCKER A1 — `alt-filtered` grades neither of the two filters it exists to grade.**

The 3-row catalog masks each filter with the other. `ProductFactory.delivery_type` defaults to `"guaranteed"` (`tests/factories/product.py:33`), so the seeded rows are `open_us(US, guaranteed)`, `open_gb(GB, non_guaranteed)`, `restricted(US, guaranteed, +allowed_principal_ids)`. **No seeded product is excluded by exactly one of the two filters** — `open_gb` fails both — so each filter is individually invisible.

Measured by disabling each filter in production independently:

| mutation | result |
|---|---|
| none (baseline) | 6 wired nodes passed |
| `src/core/tools/products.py:465` delivery_type guard disabled | **all 6 passed** |
| `src/core/tools/products.py:538` countries guard disabled | `alt-filtered` 3 passed (`alt-empty` reddened) |
| both disabled | 3 failed — `assert 'non_guaranteed' == 'guaranteed'` |

So `filters.delivery_type` is graded by nothing in the suite; `filters.countries` survives only via `alt-empty`'s `["ZZ"]` filter (`:100` of the new step module). This contradicts "so visibility/filter asserts are non-vacuous" and the spec-grounding section, which names both keys as what this scenario grades.

**Fix (prototyped, verified in both directions):** add two decoy rows here so each filter is the sole discriminator — `ProductFactory(tenant=tenant, countries=["GB"])` and `ProductFactory(tenant=tenant, delivery_type="non_guaranteed", countries=["US"])`, each with a `PricingOptionFactory`. On the 5-row catalog: unmutated → 6 passed; countries disabled → 3 failed (`products not serving ['US'] leaked through the country filter: ['prod_0015']`); delivery_type disabled → 3 failed (`'prod_0003' has delivery_type 'non_guaranteed', expected 'guaranteed'`).

Also tighten `:188` of the new step module from a subset check to equality against the expected visible set — as written, a filter that *over*-excludes stays green.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+
+    ``Product.countries`` is an internal filter field (``exclude=True`` in the
+    schema — never serialized), so the filter's effect is asserted against the
+    seeded catalog: only products whose DB row serves an overlapping country
+    may come back, and at least one must (the catalog seeds a matching one).
+    """
+    requested = set(json.loads(countries))
+    products = _wire_products(ctx)
+    assert products, "no products returned — country filter is unobservable on an empty result"
+    matching_ids = {p.product_id for p in ctx["seeded_products"] if set(p.countries or []) & requested}
```
ChrisHuie, line 186: **SHOULD-FIX A2 — the restricted seed row grades nothing here, and this line widens the country assert so a cross-principal leak stays green.**

`src/core/tools/products.py:373` (`elif principal_id in allowed_ids:` → `elif True:`) → **6 passed**. Two reasons:

1. The only step reading `ctx["restricted_product_ids"]` is `then_restricted_products_hidden` (`:155`), whose text appears solely in the dormant `T-UC-001-alt-anonymous`.
2. Because the restricted row carries `countries=["US"]`, it enters `matching_ids` here — so a leaked restricted product still satisfies `off_target == set()`.

Deferring the anonymous scenario is fine and clearly stated; the description sentence just overstates what the third seeded row buys.

**Fix:** correct that sentence, and exclude `ctx["restricted_product_ids"]` from `matching_ids` on this line.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+    (spec-gap tracked in salesagent-ilbv) — sending it would trip the Pattern
+    #7 dev-forbid gate on REST rather than exercise discovery.
+    """
+    headers = [str(h).strip() for h in datatable[0]]
+    field_idx, value_idx = headers.index("field"), headers.index("value")
+    kwargs: dict[str, Any] = {}
+    for row in datatable[1:]:
+        field = str(row[field_idx]).strip()
+        raw = str(row[value_idx]).strip()
+        if field == "buying_mode":
```
ChrisHuie, line 39: **SHOULD-FIX A3 — `buying_mode` is dropped from both wired scenarios, and the justification is wrong on three counts.**

Both wired scenarios declare `buying_mode` (feature `:107`, `:189`), and this `continue` discards it.

1. `dist/schemas/3.1.1/media-buy/get-products-request.json` has `"required": ["buying_mode"]` — it is the **only** required field — with *"v3 clients MUST include buying_mode. Sellers receiving requests from pre-v3 clients without buying_mode SHOULD default to 'brief'."* Both scenarios are tagged `@schema-v3.1` yet dispatch a request invalid against the pinned request schema, riding the pre-v3 fallback arm — which is what the feature's own dormant `@T-UC-001-inv-079` scenario (`:663`) exists to grade.
2. `docs/test-obligations/business-rules.md:1358` — BR-RULE-079 is "Enrichment Service Fail-Open with Exception Narrowing". The buying_mode fallback carries that tag only in the generated feature file (`:662`) — an id collision, not a registry entry.
3. The docstring names REST's `extra="forbid"` gate as the blocker. REST never sees the field: `tests/harness/product.py:115` whitelists the body to `("brief", "brand", "filters", "adcp_version")`. Removing the drop gives `a2a PASSED, rest PASSED, mcp FAILED` — `ValidationError: ... buying_mode Unexpected keyword argument`. The blocker is the MCP tool signature (`src/core/tools/products.py:776-791`), with the REST body model (`src/routes/api_v1.py:68-73`) one layer down.

Also: `git grep -n buying_mode -- src/` returns only `src/core/schemas/product.py:235-250`. Nothing reads or defaults it, so "production defaults it to brief" has no mechanism — its declared default is `None`, and the spec's seller-side SHOULD is unimplemented.

Two further divergences of the same kind, where the Gherkin is not the source of truth for the dispatched request: `:98` injects `filters={"countries": ["ZZ"]}` that `alt-empty`'s table does not declare (observed on the wire), and `:107` merges `request_preset` **after** the datatable, so a preset silently overrides any declared value (no live conflict today; latent for the next scenario reusing that Given).

**Fix:** correct the docstring to cite the request schema and name the MCP signature; drop the "defaults it to brief" and BR-RULE-079 clauses; make the suppression self-invalidating so it dies when the gap closes; give `request_preset` `setdefault` semantics. Adding the field to the MCP signature / REST body is a `src/` change and belongs in its own PR.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+
+# ── Then steps ──────────────────────────────────────────────────────
+
+
+def _wire_products(ctx: dict) -> list[dict[str, Any]]:
+    """The products array as the buyer sees it on the serialized wire."""
+    return wire_field(ctx, "products")
+
+
+@then('the response should contain "products" array')
```
ChrisHuie, line 122: **SHOULD-FIX B1 — this literal step and the generic parser step both match the same Gherkin line; which one runs is decided by `pytest_plugins` order.**

`tests/bdd/steps/generic/then_success.py:101` registers `@then(parsers.parse('the response should contain "{field}" array'))`. Both match the same concrete line — pytest-bdd 8.1.0: `parsers.string(...).is_matching(line)` → `True`, `parsers.parse(...).is_matching(line)` → `True`. `pytest_bdd/scenario.py:136-144` collects every match, sorts by `baseid`, and pytest activates the last.

Three mutations, both directions:
- raising in `then_contains_products_array` → 3 failed, so this literal wins **today**;
- breaking the `format_id` assert → 3 failed on real wire bytes, so the loop is reached and non-vacuous;
- then moving `"tests.bdd.steps.domain.uc001_discover_inventory"` from position 13 to 1 in `tests/bdd/conftest.py:49-73`, with that break still in place → **3 passed**. The generic step takes over, reads the model via `getattr`, and the v3.1.1 `format_id` federation contract stops being graded with nothing red.

`tests/unit/test_architecture_bdd_no_shadowed_steps.py` cannot see this: it keys on fixture name, and the two names differ (`..._the response should contain "products" array` vs `..."{field}" array`). It passes at HEAD along with all 19 BDD guards. This class already bit this PR once — commit `17f254240`, the `status should be "completed"` literal-vs-template pair, caught only by 3 UC-005 scenarios crashing.

Blast radius is bounded: that step text occurs at 14 sites, all in `BR-UC-001-discover-available-inventory.feature`.

**Fix — the repo already has a purpose-built order-independent mechanism.** `tests/unit/test_architecture_bdd_step_module_reachability.py:44-58` documents category (B) of `_ALLOWED_UNREGISTERED`: a module that is live but registered *locally* in its test module via `from ... import *`, "so its intentional generic-step overrides stay scoped to that UC instead of shadowing every other UC ... a PERMANENT exception, not pending work." The working precedent is `tests/bdd/test_uc019_query_media_buys.py:9-14`.

Drop this module from `pytest_plugins`, star-import it in `tests/bdd/test_uc001_discover_available_inventory.py`, and add one allowlist entry. Renaming the step text is not available — the feature file is generated.

tests/bdd/steps/generic/then_success.py
```diff
@@ -34,6 +34,7 @@ def then_no_media_buy_record_persisted(ctx: dict) -> None:
     "formats",  # ListCreativeFormatsResponse
     "media_buy_deliveries",  # GetMediaBuyDeliveryResponse
     "aggregated_totals",  # GetMediaBuyDeliveryResponse
+    "products",  # GetProductsResponse
```
ChrisHuie, line 37: **SHOULD-FIX B2 — this addition is dead code, and so is the block it extends.**

`ctx["response"]` is the typed model (`tests/harness/transport.py:103`, `tests/bdd/steps/generic/_dispatch.py:72`), and `GetProductsResponse` declares `status`, so `then_response_status` takes the branch at `:60` and returns at `:64`. A probe at `:59` prints `has_status_field=True status_value=<TaskStatus.completed: 'completed'>` x6. Reverting this whole file to its merge-base state → **6 passed, nothing reddens**.

Wider: a tripwire replacing the table with `("__SENTINEL_TRIPWIRE__",)` still passed, so `:66-95` is never entered at all. Of the 37 `*Response` classes in `src.core.schemas`, 19 lack `status`, and **none** of those carries `products`, `formats`, `media_buy_deliveries` or `aggregated_totals`. And `:47-48` states "UC-005 (ListCreativeFormatsResponse): no `status` field" — `'status' in ListCreativeFormatsResponse.model_fields` is `True` (it is declared on the library `ProtocolEnvelope` base).

Commit `17f254240`'s message gives the rationale as "(GetProductsResponse has no status field either)", which is false; the description repeats the framing. The load-bearing half of that commit — deleting the duplicate step from the UC-001 module — is correct and should stay.

**Fix:** revert both hunks in this file and correct the framing. Then decide the pre-existing block's fate rather than adding to it: delete `:66-95` and the table, or keep it for non-Pydantic doubles and drop the table (for a double, `resp_fields` is `{}`, so `:91` raises regardless of its contents). Worth confirming first whether any scenario depends on `:91` raising.

tests/bdd/conftest.py
```diff
@@ -3517,6 +3520,39 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
...
             pytest.xfail(f"UC-004 harness not yet wired for type: {harness_type}")
+    elif uc == "UC-001":
+        # get_products discovery (salesagent-cfrg port-minus of salesagent-8wf2/pli8).
+        # The wired set runs a real get_products through every transport on
+        # ProductEnv; the seeded catalog (open US/guaranteed, open GB/non_guaranteed,
+        # and one product restricted to another principal) makes the visibility
+        # and filter asserts non-vacuous. T-UC-001-main (#1595) and
+        # T-UC-001-alt-anonymous (#1591) are not wired on this slice. Everything
+        # else stays dormant here: wiring never forces graduation.
+        _UC001_WIRED = {
```
ChrisHuie, line 3531: **SHOULD-FIX B3 — a one-character drift in this set silently drops coverage with exit 0.**

Renaming `"T-UC-001-alt-empty"` → `"T-UC-001-alt-empty-RENAMED"` gives **`3 passed, 1107 xfailed`, exit 0**, against a baseline of `6 passed, 1104 xfailed`. Nothing reddens, and no membership pin exists anywhere (`git grep -rln "known_tags|feature_tags|_WIRED" tests/unit/ tests/bdd/test_*.py` → empty).

Three copies of the shape: this one plus the pre-existing `_UC002_IDEMPOTENCY_WIRED` (`:2807`) and `_UC002_MANUAL_APPROVAL_WIRED` (`:2817`) — both provenance-verified pre-existing, so the fix is the seam rather than this site.

**Fix:** one guard parsing every `BR-UC-*.feature` for `@T-*` tags and asserting that every member of every `_*_WIRED` set exists in some feature. That covers all three.

tests/bdd/conftest.py
```diff
@@ -3517,6 +3520,39 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
...
+                PricingOptionFactory(product=open_us)
+                open_gb = ProductFactory(tenant=tenant, delivery_type="non_guaranteed", countries=["GB"])
+                PricingOptionFactory(product=open_gb)
+                restricted = ProductFactory(tenant=tenant, countries=["US"], allowed_principal_ids=["someone-else"])
+                PricingOptionFactory(product=restricted)
+                ctx["seeded_products"] = [open_us, open_gb, restricted]
+                ctx["restricted_product_ids"] = {restricted.product_id}
+                yield
+        else:
+            pytest.xfail("UC-001 harness not yet wired for this scenario (salesagent-cfrg)")
```
ChrisHuie, line 3555: **SHOULD-FIX B4 — both de-scope tickets state a grading mechanism that this line cannot provide.**

The two issues this slice defers to say the gaps are held by a strict-xfail: one states *"Graded by the strict-xfail assert at the end of `T-UC-001-main` — emitting it forces graduation"*, the other *"Graded by the strict-xfail BDD scenario `T-UC-001-alt-anonymous`; the moment production allows anonymous discovery the scenario xpasses and must graduate."*

This PR ships no `_UC001_XFAIL_TAGS`, and this catch-all is an **imperative** `pytest.xfail(...)`, which raises immediately — the body never runs, so it can never xpass and can never force graduation. Verified by probe.

The scope correction itself is sound. The issue bodies should carry this PR's own accurate phrasing ("to be wired in the same change as the fix") instead, so whoever picks them up knows the wiring must be added.

tests/bdd/steps/_datatable.py
```diff
@@ -0,0 +1,21 @@
+"""Shared pytest-bdd datatable parsing helpers.
+
+One canonical headered-table parser for all step modules — per-module copies
```
ChrisHuie, line 3: **SHOULD-FIX C1 — the "one canonical parser" claim is not true at HEAD, and this PR adds a fifth copy of a second shape.**

Axis 1 (headered row→dict) is **4 adopted / 14 eligible**. Adopted: `given_config.py:114,193,211,232`. Holdouts:
- `tests/bdd/steps/generic/then_payload.py:271-272` — **byte-identical** to the body this PR deletes, in the same `generic/` package, and unstripped, which is the exact drift this docstring cites;
- `tests/bdd/steps/domain/uc011_accounts.py:835,1697,1716,1736,1753,2487,2611,2624` — eight copies of one two-liner in one module;
- `tests/bdd/steps/domain/admin_accounts.py:73-75`.

(`uc003_update_media_buy.py:489` lowercases headers — legitimate divergence.)

Axis 2 (two-column field/value) is **0 shared helper / 5 copies**: this PR's new `_datatable_to_kwargs` (`uc001_discover_inventory.py:24-45`) repeats the 3-line header-index preamble of `admin_accounts.py:32-34` almost verbatim, alongside `uc002_create_media_buy.py:1602-1607`, `uc003_update_media_buy.py:261` and `:357`.

The duplication gate clears none of it: R0801 measured directly in the tree is **74 at HEAD and 74 at the merge base**, matching the baseline, and zero of the 74 hits names any of the 14 sites — it cannot see same-module or sub-threshold clones. And `datatable_to_dicts` has no test of its own, so its `strict=True` contract is ungraded.

Two further notes on this docstring: there is no uc009 step module at HEAD or at the merge base, and the second consumer is scoped to an open issue rather than a merged PR; and the `.strip()` is a no-op, because `gherkin/gherkin_line.py:34-36` already strips every table cell.

**Fix:** add `datatable_to_field_map()` beside this helper, built on it, and have both field/value parsers call it; swap `then_payload.py:271-272` (one line, same package); add a ~10-line unit test pinning headers-stripped / ragged-row-`ValueError` / empty-body-`[]`. Migrating the `uc011_accounts` sites is a strict improvement — `_parse_sync_table` at `:800` does `value.lower() == "true"` on an unstripped cell. If the migration is out of scope, narrow the claim to what the code does.

tests/bdd/conftest.py
```diff
@@ -3075,6 +3076,8 @@ def _setup_existing_media_buy(ctx: dict, env: object, tenant: object, principal:
 def _detect_uc(request: pytest.FixtureRequest) -> str | None:
     """Detect which use case a BDD scenario belongs to via its tags."""
     marker_names = {m.name for m in request.node.iter_markers()}
+    if any(t.startswith("T-UC-001-") for t in marker_names):
```
ChrisHuie, line 3079: **SHOULD-FIX C2 — `_detect_uc` is a 13-arm ladder of one parameterized operation, and this arm crosses the threshold.**

This function holds nine structurally identical `if any(t.startswith("T-UC-0XX") for t in marker_names): return "UC-0XX"` blocks. Same file at both revisions, `ruff check --isolated --select C90,PLR091`:

- `_detect_uc`: C901 13→14, returns 13→14, and PLR0912 **newly crosses** 12→13
- `_harness_env`: C901 30→32, branches 36→39, statements 180→202
- file total: 8→9 errors

Nothing gates it — `.pre-commit-hooks/check_ruff_complexity_count.py` resolves its scan root to `src/`, so `tests/` is outside the ratchet's scan scope, not merely outside its trigger.

This is the shape the DRY invariant names as wrong (one logical operation repeated with only parameter substitution), and it matters now rather than later because the three sibling slices are queued to add three more arms plus three more wired-tag literals.

**Fix:** one prefix→UC mapping driving `_detect_uc`. And hoist `_UC001_WIRED` to a module-level annotated `set[str]` beside the UC-002 constants — it is currently a function-local literal re-allocated on all 1110 fixture invocations, with the same wired-set fact restated in four places and no sync mechanism.

tests/integration/conftest.py
```diff
@@ -564,13 +567,34 @@ def get_free_port():
...
-        stderr=subprocess.PIPE,
-        bufsize=1,  # Line buffered
-    )
+    # The server's output goes to FILES, never PIPEs. A PIPE nobody drains caps
+    # at 64KB; once server logging fills it, the server blocks on a log write
+    # INSIDE a request handler and the calling test awaits forever — this wedged
+    # every full CI run at integration's quiet tail until the >1h run reaper
+    # killed it (salesagent-cu12). Files keep the error-path diagnostics below
+    # without needing a drainer thread.
+    output_dir = Path(tempfile.mkdtemp(prefix=f"mcp-server-{port}-"))
```
ChrisHuie, line 576: **SHOULD-FIX D1 — this fix double-lands with another open PR, in a different implementation, on the same hunks.**

`tests/integration/conftest.py` is one of the 80 changed files in the open, mergeable PR titled "fix: media buy delivery correctness — scheduler serving statuses + webhook-only fields" (head `498f32b74`). It rewrites this same fixture pipe→file using `tempfile.TemporaryFile()` + `_server_logs()`/`_close_server_logs()`, keeping `terminate()` then `kill()`, at hunks `@@ -513,6 @@`, `@@ -564,12 @@`, `@@ -593,28 @@`, `@@ -635,10 @@`.

This PR uses `mkdtemp()` + on-disk files + `_server_output()` and replaces `terminate()` with `kill()`, at `@@ -510,10 @@`, `@@ -564,13 @@`, `@@ -593,29 @@`, `@@ -635,10 @@`. `git log -G'mcp-server-' origin/main -- tests/integration/conftest.py` is empty, so the change is on neither base.

Whichever merges second conflicts, and the two designs differ behaviourally — so "may land independently from its home branch; harmless either way" does not hold. One implementation should be chosen and dropped from the other PR before either merges.

tests/bdd/steps/generic/then_success.py
```diff
@@ -83,7 +84,7 @@ def then_response_status(ctx: dict, status: str) -> None:
         assert value is not None, (
             f"Status 'completed' claimed but response.{attr} is None — the schema-required success payload is missing"
         )
-        if attr in ("formats", "media_buy_deliveries"):
+        if attr in ("formats", "media_buy_deliveries", "products"):
```
ChrisHuie, line 87: **NIT C4 — this is a hand-maintained subset of the tuple above it.**

`("formats", "media_buy_deliveries", "products")` is a subset of `_STATUSLESS_SUCCESS_ATTRS` at `:33-38`. Adding one attribute required editing two literals, which this PR did; the next contributor edits one. Not drifted today — `aggregated_totals` is legitimately excluded, since it is a model rather than a list.

**Fix:** carry the discriminator at the source (`dict[str, bool]`, with the "object, not a list" reason inline) or derive it via `get_origin(resp_fields[attr].annotation) is list`, removing the second literal. Moot if the revert in the other finding on this file lands — sequence that decision first.

tests/integration/conftest.py
```diff
@@ -635,10 +643,7 @@ def __init__(self, port, process, db_name):
     except subprocess.TimeoutExpired:
         process.kill()
         process.wait()
-    if process.stdout:
-        process.stdout.close()
-    if process.stderr:
-        process.stderr.close()
+    shutil.rmtree(output_dir, ignore_errors=True)
```
ChrisHuie, line 646: **NIT D5 — temp dirs leak on both raise-before-yield paths.**

`:576` creates the dir; `:620` and `:624-626` raise; this `rmtree` is the only cleanup and sits after the `yield` at `:637`, so neither raise path reaches it. Probed: 1 `mcp-server-*` dir leaked per failure, holding `stdout.log` + `stderr.log` — the full server log, i.e. the output the new comment says can exceed 64 KB — and 10 tests use this fixture. A `Popen` failure at `:582` leaks it too.

**Fix:** `request.addfinalizer(lambda: shutil.rmtree(output_dir, ignore_errors=True))` immediately after `mkdtemp`.

Two adjacent items in the same hunk: `:625` `process.wait(timeout=5)` can raise `TimeoutExpired` and discard the `RuntimeError` at `:626` along with its diagnostic (same shape at the merge base, so not a regression) — `contextlib.suppress` it; and if `stderr_path.open("wb")` raises at `:580`, `stdout_f` is never closed, so move both opens inside the `try` or use `ExitStack`.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+@then("the products should be ordered by relevance_score descending")
+def then_products_ordered_by_relevance(ctx: dict) -> None:
+    """Production gap: relevance_score is not emitted on the products wire.
+
+    The impl sorts internally when AI ranking is enabled but never serializes
+    a relevance_score field. Dormant until T-UC-001-main is wired (#1595).
+    """
+    products = _wire_products(ctx)
+    scores = [p.get("relevance_score") for p in products]
+    assert all(s is not None for s in scores), f"relevance_score missing on the wire: {scores}"
```
ChrisHuie, line 216: **NIT D6 — `relevance_score` is not an AdCP 3.1.1 field, and this step also passes vacuously on an empty array.**

`dist/schemas/3.1.1/core/product.json` has 49 properties including `brief_relevance` and **no `relevance_score`** (`countries` is likewise absent, which independently confirms that `Product.countries`' `exclude=True` is spec-correct). The tracking issue's own title scopes to `brief_relevance`; the PR description widened it to "relevance_score / brief_relevance". So this assertion grades an invariant 3.1.1 does not define.

Separately: with `products == []`, `all(s is not None for s in [])` is `True` and `[] == sorted([], reverse=True)` is `True`, so the step passes — see the vacuity finding on `:158`.

**Fix:** re-express the ordering assert against `brief_relevance` plus response order, and correct the description sentence.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+            f"pricing_options not suppressed for anonymous buyer on {product.get('product_id')!r}: "
+            f"{product.get('pricing_options')!r}"
+        )
+
+
+@then("no products with allowed_principal_ids restrictions should be visible")
+def then_restricted_products_hidden(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    restricted_ids = ctx.get("restricted_product_ids", set())
+    assert restricted_ids, "fixture seeded no restricted product — assertion would be vacuous"
```
ChrisHuie, line 158: **NIT E1 — 2 of the 9 new Then steps pass on an empty-but-successful products wire, drifting from the 7 that anchor.**

Direct invocation with `ctx={'wire_response': {'products': []}, 'transport': Transport.REST}`: this step and `then_products_ordered_by_relevance` (`:208-217`) both **pass vacuously**. Positive control — `then_pricing_suppressed`, `then_products_have_required_fields`, `then_products_have_brief_relevance` and `then_contains_products_array` each raise their non-empty anchor.

This line guards the *fixture* side (`assert restricted_ids`) but not the *response* side; `:216-217` relies on `all([])` and `[] == sorted([])`.

Both are dormant today, so neither passes on broken production right now. But the day `T-UC-001-main` is wired, an empty array makes the step that exists to document a gap report the gap as closed. This is drift inside a same-PR assertion family, which is what the DRY invariant exists to prevent.

**Fix:** add the sibling anchor — `assert products, "..."` — at both sites, matching `:126, 146, 167, 185, 200, 224`.

ChrisHuie (commented):
## Review — `d7d13ec15`

Diff range `origin/main...HEAD`; merge base is current `main` (`d912a7ee9`), 0 commits behind. Spec pin derived at HEAD: **3.1.1 / `adcp==6.6.0`**. Tests-only — `git diff --stat origin/main..HEAD -- src/` is empty.

**1 BLOCKER / 12 SHOULD-FIX / 11 NIT.** Anchored findings are inline; the items below have no line inside a diff hunk, so they live here.

### What holds up

- Wire reads go through the guarded `wire_field`/`wire_dict` (`tests/bdd/steps/_outcome_helpers.py:18-59`), which raise when a real-wire transport did not stash `wire_response` — not an inline read with a `model_dump` fallback. No serializer tautology.
- The `format_id` assertion reuses `assert_wire_format_id_is_object` (`tests/helpers/format_assertions.py:20-37`) rather than hand-rolling it, and it is confirmed reached on real wire bytes on all three in-process transports: `fid={'agent_url': 'https://creative.adcontextprotocol.org/', 'id': 'display_300x250_image', ...}`.
- 1104 previously-uncollected scenario nodes now collect, 0 failures, zero allowlist growth, no `_XFAIL_TAGS` entries. Whole-feature `scenarios()` binding matches the convention the CI shard-splitter requires; the "119 dormant" figure is exact against 121 scenario blocks.
- The `CLAUDE.md` pin correction is right — `pyproject.toml` already pinned `adcp==6.6.0` at the merge base.
- 6 of 9 new wire-consuming Then steps carry an explicit non-vacuity anchor before asserting.

### CI state

`E2E Tests` and `Summary` fail on `d7d13ec15`; `mergeStateStatus: BLOCKED`. The failures are `test_a2a_protocol_compliance.py::test_get_media_buy_delivery_request_schema` and `test_schema_validation_standalone.py::test_get_products_request_validation`.

**They are not attributable to this diff**, proven two ways: `git diff --stat origin/main..HEAD -- src/ tests/e2e/` is empty (byte-identical), and both reproduce locally at HEAD with the cause visible — `tests/e2e/adcp_schema_validator.py:63` downloads schemas live from `adcontextprotocol.org`, and stdout reports `Could not resolve schema reference /schemas/latest/core/version-envelope.json - using strict fallback`. Upstream registry drift since `main`'s last green E2E run on the identical SHA (2026-07-30). `tests/e2e/test_schema_validation_standalone.py:27-31` already documents this class for a sibling test that was deleted for it.

---

### SHOULD-FIX D2 — the verification section does not describe this head's CI

It accounts for six webhook failures traced to an SSRF issue; CI on `d7d13ec15` fails on two schema-validation tests (`== 2 failed, 91 passed, 28 skipped, 4 xfailed ==`). The cited local run is stamped 07-31 against CI that ran 08-03. The real cause — a moving-target `/schemas/latest/` dependency — is recorded nowhere in the PR.

### SHOULD-FIX D3 — the storyboard grounding sentence is false on its premise

"Conformance storyboards ... grade only `get_products_async`, `refine_products`, `product_signal_targeting`" does not hold: of the 44 files in `dist/compliance/3.1.1/domains/media-buy/scenarios/`, **39 carry a `task: get_products` step**.

The *conclusion* survives. Scoped precisely: `filters.countries` appears in **none** of them, `filters.delivery_type` in exactly one (`canonical_formats.yaml`) with no validation on its effect, and no storyboard grades an empty synchronous `products` array. `pricing_currency_filter.yaml` does grade structured-filter effect on a sync `get_products`, but for `pricing_currencies`.

Suggested replacement: state the measured numbers and conclude that both scenarios are schema-grounded rather than storyboard-graded.

### SHOULD-FIX D4 — the pin correction has three unswept twins, two in the public README

`README.md:8` ("(AdCP 3.1.0-beta.3), under active development") and `README.md:25` ("targets **AdCP spec version 3.1.0-beta.3** via the `adcp==5.7.0`", plus a paragraph calling the spec unfrozen beta — 3.1.1 is stable), and `docs/adcp-spec-version.md:46` ("The SDK **pin** (3.1.0-beta.3) fixes the request/response *type shapes*"), which contradicts its own line 3.

All pre-existing — `git log origin/main...HEAD -- README.md docs/adcp-spec-version.md` is empty — but the README is the first thing an outside contributor reads, and that `adcp-spec-version.md` line makes "matches `docs/adcp-spec-version.md`" only half true. Three-line fold-in. (`:28` and `:51` in that file are legitimately historical.)

### NIT D8 — the shadow guard's matcher gap has 18 pre-existing instances

A parser-vs-literal probe over all 1181 registered step definitions finds 19 live shadow pairs; this PR adds the 19th (see the inline finding on the new literal step). `tests/unit/test_architecture_bdd_no_shadowed_steps.py:55-72` groups by fixture *name*, so all 19 pass it, while `_ALLOWED_SHADOWS` is documented as "MUST stay empty". Widening it — compare each parser's `is_matching()` against the others' text — is ~20 lines, but it would surface all 18, so it belongs in its own change rather than here.

### NIT E2 — 8 of 15 new step definitions ship unreached

`uc001_discover_inventory.py:62, 73, 85, 143, 155, 198, 208, 221`, plus the dead `if "dispatch_identity" in ctx` branch at `:108-109` (`dispatch_identity` is set only for the dormant anonymous scenario). Two of them assert fields the PR states production never serializes, so they are guaranteed-red the moment `T-UC-001-main` is wired, with no reason string tying them to the tracking issue.

---

### Not verified

The fourth transport (`e2e_rest` — only the three in-process ones were measured; neither wired nodeid appears in `e2e_rest_known_failures.txt`). The full-suite numbers cited in the description were not re-derived. All DB runs used a persistent Postgres, not a fresh one. Whether `tests/e2e/` is red on `main` today — `main` has not run CI since 07-30, so the attribution above rests on byte-identity plus local reproduction. No trial merge for the conflicting-PR finding. Per-scenario xfail reasons for the dormant 1104 were not enumerated, and whether any dormant block would pass if wired was not measured. Whether either pre-existing `_UC002_*_WIRED` set has already drifted. The reverse shadow sweep across the whole suite. Adjacent and unfiled: `src/core/tools/products.py:794` says "aligned with adcp v3.6.0 spec" against a 6.6.0 / 3.1.1 pin — pre-existing, same family as D4.

tests/bdd/conftest.py
```diff
@@ -3517,6 +3520,39 @@ def _harness_env(request: pytest.FixtureRequest, ctx: dict) -> Generator[None, N
                 yield
         else:
             pytest.xfail(f"UC-004 harness not yet wired for type: {harness_type}")
+    elif uc == "UC-001":
+        # get_products discovery (salesagent-cfrg port-minus of salesagent-8wf2/pli8).
```
ChrisHuie, line 3524: **NIT D7 — this comment's local tracking ids do not resolve for outside contributors, and the sibling xfail reasons do not use them.**

Among the ten `pytest.xfail("... harness not yet wired ...")` siblings in this dispatch chain, `:3555` is the **only** one carrying a local id: `:3288` and `:3431` use `tracked in #NNNN`, and `:3296`, `:3393`, `:3455`, `:3522`, `:3580` carry none. Same for the added ids here and at `uc001_discover_inventory.py:1`/`:30`, `test_uc001_discover_available_inventory.py:6`, and `tests/integration/conftest.py:574` — 9 occurrences of 5 distinct ids across 6 sites.

The convention is written down at `tests/unit/test_guards_no_beads_ids.py:3-5` (local ids "don't resolve for outside contributors"), with enforcement deliberately scoped to structural-guard test files, so these paths are unguarded but not unconvention'd.

This block also mixes both id systems (`:3524` local, `:3528-3529` GitHub refs), so one grep does not find the whole family.

**Fix:** GitHub refs, or no id, matching the nine siblings.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+@then('the response should contain "products" array')
+def then_contains_products_array(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    assert isinstance(products, list), f"products is not an array on the wire: {type(products).__name__}"
+    assert products, "products array is empty — seeded catalog should be visible"
+    # v3.1.1 core/format-id.json: every format_id on the products wire is an
+    # object carrying agent_url + id, never a bare string (federation contract).
+    from tests.helpers.format_assertions import assert_wire_format_id_is_object
```
ChrisHuie, line 129: **NIT D9 — this in-body import is not load-bearing.**

All three reasons that would justify it are absent, each checked: `tests/helpers/format_assertions.py` imports only `__future__` and `typing.Any`, so no cycle is possible; both import orders succeed in a live probe; nothing in `tests/` patches this target; and the two sibling step modules in the same package import it at top level (`uc005_format_id_roundtrip.py:17`, `uc005_format_id_shape.py:26`).

**Fix:** hoist it to the module's import block.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+import json
+from typing import Any
+
+from pytest_bdd import given, parsers, then, when
+
+from tests.bdd.steps._outcome_helpers import wire_field
+from tests.bdd.steps.generic._dispatch import dispatch_request
+
+
+def _datatable_to_kwargs(datatable: list) -> dict[str, Any]:
```
ChrisHuie, line 24: **NIT D10 — this annotation contradicts the helper the module calls.**

`:24` and `:114` annotate `datatable: list` (bare), while this PR's own new `tests/bdd/steps/_datatable.py:12` uses `Sequence[Sequence[object]]`. Repo-wide there are five variants across `tests/bdd/steps`: `list[list[str]]` (10), `Any` (9), bare `list` (6), `Sequence[Sequence[object]]` (6).

**Fix:** use `Sequence[Sequence[object]]` here, matching the helper this function feeds.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+
+@then(
+    "each product should have product_id, name, format_ids, publisher_properties, "
+    "pricing_options, and reporting_capabilities"
+)
+def then_products_have_required_fields(ctx: dict) -> None:
+    products = _wire_products(ctx)
+    assert products, "no products returned"
```
ChrisHuie, line 200: **NIT D11 — the wire-read-plus-non-vacuity-guard pattern wants a seam.**

`_wire_products(ctx)` is called 9 times in this module, 6 of them immediately followed by an emptiness guard (`:126, 146, 167, 185, 200, 224`), two of which are byte-identical (`"no products returned"` here and at `:224`).

The per-message specificity is a deliberate anti-vacuity design, so the seam is `_nonempty_wire_products(ctx, why)` rather than collapsing the messages. R0801 cannot see this (same-module, sub-threshold), so it needs the hand sweep.

tests/bdd/steps/domain/uc001_discover_inventory.py
```diff
@@ -0,0 +1,228 @@
...
+            kwargs[field] = raw
+    return kwargs
+
+
+# ── Given steps ─────────────────────────────────────────────────────
+
+
+@given("a tenant exists with at least one product in the catalog")
+def given_tenant_with_catalog(ctx: dict) -> None:
+    """Background step: the UC-001 conftest branch seeded tenant + catalog."""
```
ChrisHuie, line 53: **NIT E3 — these `ctx.setdefault` calls are always no-ops.**

`tests/bdd/conftest.py:3542-3544` sets `ctx["tenant"]` and `ctx["principal"]` unconditionally before any step runs, so the `setdefault` calls here and at `:62-63` can never take effect. The repeated `env.setup_default_data()` calls are safe — it is documented and implemented as get-or-create (`tests/harness/_base.py:1238-1274`) — but this derives one concept in four places.

**Fix:** read `ctx["tenant"]` directly, or make the step assert the fixture set them.

PR #1842, opened 2026-08-04
fix: emit brief relevance from product ranking

### LINKED ISSUE #1595: get_products: emit brief_relevance per v3.1.1 product schema

gagliathebuilder:
## Summary

- emit the existing AI ranking explanation as the AdCP 3.1.1 `brief_relevance` product field
- keep relevance scores internal while preserving the existing sort and threshold behavior
- wire `T-UC-001-main` across A2A, MCP, and REST

## Why

The product ranker already returned a relevance score and a human-readable reason for each product. `get_products` used the score for ordering and filtering, but discarded the reason before constructing the response. As a result, curated products never carried the spec-defined `brief_relevance` explanation on the wire.

This change copies the ranker's reason onto each surviving ranked product. Products returned without AI ranking remain unchanged and do not gain a synthetic explanation.

## Spec grounding

- AdCP 3.1.1 `core/product.json` defines `brief_relevance` as the explanation of why a product matches the supplied brief.
- AdCP 3.1.1 Brief Expectations directs publishers to explain relevance in curated responses.
- `T-UC-001-main` grades relevance ordering and the presence of product explanations.

## Validation

- `T-UC-001-main`: 3 passed across A2A, MCP, and REST
- product and `get_products` suites: 138 passed
- AdCP contract and BDD architecture checks: 151 passed
- Ruff formatting and lint checks passed
- code-duplication guard passed with no baseline growth

## Coordination note

PR #1837 is adding shared UC-001 BDD wiring that overlaps with the test scaffolding currently included in this draft. I plan to rebase after #1837 lands and remove the duplicate harness changes, while retaining the focused `brief_relevance` implementation and coverage for issue #1595.

I also identified a stale integration expectation and an overly broad BDD step in this draft. I will address both as part of that cleanup before marking the PR ready for review. Thank you.

Fixes #1595

### DIFF
--- a/src/core/tools/products.py
+++ b/src/core/tools/products.py
@@ -687,6 +687,11 @@ async def _get_products_impl(
                 # Filter out products with very low relevance (score < 0.1)
                 eligible_products = [p for p in eligible_products if ranking_map.get(p.product_id, (0.0, ""))[0] >= 0.1]
 
+                # AdCP 3.1.1 core/product.json: explain why each curated
+                # product matches the supplied brief.
+                for product in eligible_products:
+                    product.brief_relevance = ranking_map[product.product_id][1]
+
                 # Log the ranking results
                 for r in ranking_result.rankings:
                     logger.info(f"[AI_RANKING] {r.product_id}: score={r.relevance_score:.2f}, reason={r.reason}")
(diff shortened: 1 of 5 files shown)


gagliathebuilder:
Hi, thank you for the opportunity to contribute here. Since this draft has been parked for some time and now overlaps with other work in progress, I’m going to close it rather than leave a stale contribution open.

I appreciate the guidance and context shared along the way. Issue #1595 can remain open for any future implementation. Thank you.

Issue #1532, opened 2026-07-03
fix: channel_selector.html emits invalid MediaChannel values — product channel tags silently dropped


ChrisHuie:
## Summary

The shared admin channel picker `templates/components/channel_selector.html` offers four `<option>` values that are not members of the AdCP `MediaChannel` enum. A product tagged with one has that channel silently discarded on read, then falls back to adapter-default channel matching — so the publisher's explicit intent is lost, while the admin edit form keeps showing the channel as selected, masking it.

## Severity — P2

A publisher who tags a product **Video / Audio / Native / Retail Media** has that channel silently discarded on read (`src/core/product_conversion.py:441`, raw `MediaChannel(ch)` → `ValueError` → log-and-skip). The product then reaches `get_products` with `channels = None` and falls into the adapter-default matching branch of the channel filter (`src/core/tools/products.py:561-590`), so it is matched on `get_adapter_default_channels(adapter_type)` rather than on the publisher's tag — over- or under-matching — while the edit form re-selects the stale value and hides the loss. The failure is silent and drops explicit intent; the adapter-default fallback softens it into wrong-matching rather than guaranteed invisibility.

## Verified facts

- Invalid option values (`channel_selector.html:16-24`) vs the 20-member enum (`adcp.types.generated_poc.enums.channels.MediaChannel`): **`video`, `audio`, `native`, `retail`**. Valid: `display, dooh, ctv, podcast, social`.
- Silent-drop site: `product_conversion.py:437-445` — raw `MediaChannel(ch)` in `try/except ValueError: logger.warning(...skipping)`. `CHANNEL_MAPPING` is not on this path (it maps adapter names in `capabilities.py`), so all four values drop, not two. Conversion runs before the filter (`products.py:346`, then the channel filter at `:561`).
- Write path stores raw strings, unvalidated: create `products.py:973-975`, edit `products.py:1483-1488` (`request.form.getlist("channels")` → JSONType column `models.py:257`, no CHECK constraint, no Pydantic gate). Invalid values can therefore be persisted. Whether any production row is affected is not established here — the repair migration is defensive.
- Masking: the edit form pre-selects from the raw column (`products.py:1977` → `selected_channels`); `channel_selector.html:17` re-checks `{% if 'video' in selected %}`, so the stale string re-selects and appears valid.
[...]

PR #1848, opened 2026-08-04
feat: Property channel disclosure on PublisherPartner (Phase 1)

### LINKED ISSUE #1533: feat: Property channel disclosure — supported_channels on PublisherPartner (Property half of #1032)
> Spec grounding: AdCP **3.1.0-beta.3** (`adcp==5.7.0`). File:line refs verified against `main` on 2026-07-03.

#1032 delivered the Product half (`channels` field, `get_products` filter, product-form UI, tests). This issue covers the Property half. It is **not** a mirror of `get_products` — the sound design is smaller and lives on a different table.

## What this is / isn't (verified)

- `list_authorized_properties` returns `publisher_domains: list[str]`, not `Property[]` (`src/core/tools/properties.py:82,145`). Its read source of truth is the `PublisherPartner` table via `TenantConfigUoW.list_publisher_partners()` (`properties.py:79`; `src/core/database/repositories/tenant_config.py:44`). It does not read `AuthorizedProperty`.
- `Property.supported_channels` already exists (inherited library field, `src/core/schemas/_base.py:2258`) and is always `None` — nothing populates it.
- Per spec, per-property `supported_channels` is published in the publisher's out-of-band `adagents.json`, not in this API response (`dist/docs/3.1.0-beta.3/governance/property/authorized-properties.mdx`). Agent-level channel disclosure already exists via `get_adcp_capabilities → portfolio.primary_channels` (`src/core/tools/capabilities.py:155`). There is no spec request/response schema for this endpoint in 3.x — the local types (`_base.py:2268`) are a salesagent extension, so any channel field or filter added here is a local extension and is **ungraded**.

## Design decision (settle before coding)

Channels must live on `PublisherPartner`, not `AuthorizedProperty`. There is no foreign key between `PublisherPartner.publisher_domain` and `AuthorizedProperty.publisher_domain`, and a `PublisherPartner` commonly has zero `AuthorizedProperty` rows (property IDs are fetched fresh from `adagents.json`, not cached). Aggregating channels over `AuthorizedProperty` rows would yield an empty set for those domains, which under `[] = match-all` can never be excluded — a silent no-op. `PublisherPartner` has `UniqueConstraint(tenant_id, publisher_domain)` (one row per domain) and is already returned by `list_publisher_partners()`, so it needs no join and no new repository.

Recommended scope, in two phases:

- **Phase 1 — disclosure (recommended, self-contained).** Add `supported_channels` (`JSONType`, nullable `list[str]`) to `PublisherPartner` (`src/core/database/models.py:2024`). Admin UI edits it; the endpoint populates the already-dormant `primary_channels` response field (`_base.py:2301`) from it. No join, no new repository, no request filter. This is spec-aligned disclosure and matches the tested read path (`tests/integration/test_list_authorized_properties_integration.py::...reads_from_publisher_partner`).
- **Phase 2 — request-side `channels` filter (optional, defer).** Only after deciding the fate of the two sibling filters that are accepted but never applied today (`req.publisher_domains` / `req.property_tags`, `properties.py:196-208`, never read in the impl). A working third filter beside two dead ones is inconsistent — fix all three or none. If built, mirror `get_products` intersection semantics exactly.

Alternative considered (not recommended): return `properties: list[Property]` on the response and filter those. This rewrites the response contract and diverges from the pinned spec's domains-only model; pursue only if a graded scenario or explicit buyer requirement appears.

## Acceptance criteria — Phase 1

1. `PublisherPartner.supported_channels` — `JSONType`, nullable, `list[str]`, no backfill (mirrors the Product `channels` rollout). Alembic migration from the single head `a164b85bab9e` (verified via `uv run alembic heads`; re-check before writing), with non-empty `upgrade()` and `downgrade()`.
2. Admin publisher-partner create/edit persists `supported_channels`. Blocked on the channel-picker vocabulary fix (#1532) — `channel_selector.html` currently emits invalid `MediaChannel` values — or use a corrected picker.
3. The impl populates `primary_channels` on the response from the partners' `supported_channels`, kept typed `list[str]` (do not retype to `list[MediaChannel]` — separate breaking change, see Pitfalls). Values validated against `MediaChannel` defensively at the write boundary (unknown → logged and skipped, never fatal; an unknown enum value is a hard `ValidationError` even in production — `extra=ignore` does not rescue enum values).
4. Round-trip: unset → `primary_channels` omitted (`exclude_none`); populated → present. Extend `tests/integration/test_list_authorized_properties_integration.py` and `tests/unit/test_authorized_properties.py`.

## Acceptance criteria — Phase 2 (only if greenlit)

5. `channels` field on `ListAuthorizedPropertiesRequest`, threaded through the MCP wrapper (`properties.py:195`); the A2A raw wrapper forwards `req` already.
6. Filter keeps a domain iff `declared ∩ requested ≠ ∅`, comparing `MediaChannel.value.lower()` on both sides — matching `get_products` (`src/core/tools/products.py:561-590`). `None` / `[]` declared = match-all. Do not port the adapter-`default_channels` fallback (property channels are publisher-declared, not adapter-derived).
7. Resolve the two dead sibling filters in the same change.

## Pitfalls & guards

- The `primary_channels` `list[str] → list[MediaChannel]` retype is a breaking change (it crashes 7 sites that hardcode `["display","video"]`: `test_authorized_properties.py:179,188,220`; `test_adcp_contract.py:2314`; `test_response_shapes.py:570`; `test_adcp_json_serialization.py:46,55` — `list[MediaChannel]` rejects `"video"`) and is not required for disclosure. Land it separately if desired.
- No `get_db_session()` in `_impl` — it is already UoW-compliant (`properties.py:72`); channels on `PublisherPartner` need zero new data access. (`AuthorizedProperty` access is raw-select allowlisted debt, `tests/unit/test_architecture_no_raw_select.py:120-125` — another reason to prefer `PublisherPartner`.)
- JSONType columns only (`test_architecture_jsontype_columns.py`); migration completeness and single-head guards; tenant-scoped queries; factory-based tests (no `session.add()` in bodies).
- The response uses `NestedModelSerializerMixin` (`@model_serializer`, honored by FastMCP); the MCP wrapper passes the model, not `model_dump()` (`properties.py:223`) — do not add a custom `model_dump()`.
- `sponsored_intelligence` (the 20th enum value, new in 3.1) is missing from `CHANNEL_MAPPING` (`capabilities.py:47`) — do not hardcode enum lists.

## Test plan

- `tests/integration/test_list_authorized_properties_integration.py` already exercises the real read path (calls `_impl`, including `reads_from_publisher_partner` and `tenant_isolation`). Extend it with disclosure — and, if built, filter — cases.
- Optional: `tests/bdd/features/BR-UC-007-list-authorized-properties.feature` exists but is unwired (no `scenarios()` binding, no steps). Wiring it is a worthwhile adjacent cleanup; it is `e2e_rest`-capable (all DB-backed → no xfail-ledger entries).
- Obligations: `docs/test-obligations/BR-UC-007-list-authorized-properties.md`; UC-007 behavioral obligations sit in `obligation_coverage_allowlist.json` (which may only shrink). A behavioral test that calls the impl and carries `Covers: UC-007-*` lets you shrink the allowlist.

## PR must cite (Spec-Grounding Gate)

Version `3.1.0-beta.3`; field `dist/schemas/3.1.0-beta.3/core/property.json#supported_channels` and `enums/channels.json`; prose `governance/property/authorized-properties.mdx` and `reference/media-channel-taxonomy.mdx`; grading: **ungraded**; note that the channel field/filter is a salesagent local extension.

gagliathebuilder:
## Summary

- Adds `PublisherPartner.supported_channels` (`JSONType`, nullable) via migration `15238294201b` (revises `823974a5553e`).
- Persists publisher-partner channels at the admin write boundary (POST create + PATCH update via `TenantConfigUoW`). Unknown/malformed values return 422 with no overwrite; empty list stores `None`.
- Populates `list_authorized_properties` response `primary_channels` from the union of already-canonical partner channels.
- Extends `CHANNEL_MAPPING` with `sponsored_intelligence` and adds disclosure + write-boundary tests.

**Base:** merged `origin/main` in `7ec5f7b2` (includes #1868).

Closes #1533 (Phase 1 only).

## Spec-Grounding Gate

- **Version:** AdCP 3.1.1 (`adcp==6.6.0`)
- **Schema:** `dist/schemas/3.1.1/core/property.json#supported_channels`, `enums/channels.json`
- **Prose:** `governance/property/authorized-properties.mdx`, `reference/media-channel-taxonomy.mdx`
- **Grading:** ungraded (local salesagent extension; no spec request/response schema for this endpoint in 3.x)

`primary_channels` is the union of explicitly declared, write-canonicalized partner channels — not a completeness guarantee for partners that have not set the field.

## Out of scope

- **Phase 2:** Request-side `channels` filter and dead sibling filters (`publisher_domains`, `property_tags`) — not started.
- Global `channel_selector.html` vocabulary (#1532) — untouched; publisher-partner editing uses enum checkboxes in tenant settings only.
- Adapter/`product_conversion` mapper consolidation — deferred (Konstantin thread 4).
- Wire-level BR-UC-007 across transports — deferred (threads 7/8).

## Test plan

- [x] `make quality-ci`
- [x] `uv run pytest tests/unit/ tests/harness/`
- [x] `tests/integration/test_publisher_partner_channel_write.py`
- [x] `tests/integration/test_list_authorized_properties_integration.py` (with agent-db)
- [x] `tests/unit/test_authorized_properties_behavioral.py::TestPrimaryChannelsDisclosure`
- [x] `tests/unit/test_channel_normalization.py`

Made with [Cursor](https://cursor.com)


gagliathebuilder:
CI status update: Unit Tests and Integration (infra) are now green as of 94f4adb (ruff import ordering + dropped a flaky caplog assertion — the behavior it checked is still covered by the passing assert on the same test plus test_channel_normalization.py).

Remaining red checks are unrelated to this change:

E2E Tests / Summary (both Required) — failing on an external schema-fetch redirect (adcontextprotocol.org), confirmed also failing right now on #1837, an unrelated open PR. Repo-wide/environmental, not caused by this diff.
Security Audit / pip-audit (not Required) — pre-existing aiohttp/cryptography CVEs unrelated to this code path; a fix is in progress on a separate branch.

This PR is ready for review on the merits of the change itself — the remaining CI red isn't a defect here.

src/core/tools/properties.py
```diff
@@ -24,11 +24,22 @@
 from src.core.schemas import ListAuthorizedPropertiesRequest, ListAuthorizedPropertiesResponse
 from src.core.testing_hooks import AdCPTestContext
 from src.core.tool_context import ToolContext
+from src.core.tools.capabilities import normalize_channel_strings
 from src.core.validation_helpers import safe_parse_json_field
 
 logger = logging.getLogger(__name__)
 
 
+def _aggregate_primary_channels(publishers) -> list[str] | None:
```
KonstantinMirin, line 33: `publishers` is the one unannotated signature in these two modules. The concrete type is in-repo: this is always `TenantConfigUoW.list_publisher_partners() -> list[PublisherPartner]`, and the body reads `partner.supported_channels`. Untyped, a future caller passing domain strings or dicts fails only at runtime inside the loop. Annotate `publishers: Iterable[PublisherPartner]` (not `dict[str, Any]`/`object`).

src/core/tools/properties.py
```diff
@@ -148,6 +159,10 @@ def _list_authorized_properties_impl(
             if advertising_policies_text:
                 response_data["advertising_policies"] = advertising_policies_text
 
+            primary_channels = _aggregate_primary_channels(all_publishers)
```
KonstantinMirin, line 162: `normalize_channel_strings()` runs here on every `list_authorized_properties` request — lowercasing, alias-resolving, deduping and re-validating each partner's channels per read, warning per read instead of once at ingestion. AC#3 put this validation at the write boundary; it landed on the read path because the write boundary doesn't exist (`grep supported_channels src/admin src/core/database/repositories` is empty). Build the admin/repository write of `supported_channels`, normalize at the point of write, and this aggregation unions already-canonical values.

src/core/tools/capabilities.py
```diff
@@ -68,9 +68,38 @@
     "influencer": MediaChannel.influencer,
     "affiliate": MediaChannel.affiliate,
     "product_placement": MediaChannel.product_placement,
+    "sponsored_intelligence": MediaChannel.sponsored_intelligence,
```
KonstantinMirin, line 71: AC#5's `sponsored_intelligence` mapping has no test. Delete this line and every test in the PR still passes — the union/omit/skip cases use `display/ctv/olv/video` and `not_a_real_channel`, none touch this value. Add `normalize_channel_strings(["sponsored_intelligence"]) == ["sponsored_intelligence"]` in `test_channel_normalization.py` or as a value in the wired union scenario.

src/core/tools/capabilities.py
```diff
@@ -68,9 +68,38 @@
...
 }
 
 
+def normalize_channel_strings(raw: list[str]) -> list[str]:
```
KonstantinMirin, line 75: This is the third hand-rolled "raw channel strings -> canonical MediaChannel vocabulary" mapper. The other two diverge from it: the adapter loop at `capabilities.py:151-153` (silent-skip, no dedupe, `list[MediaChannel]`) and `product_conversion.py:463` (`MediaChannel(ch)`, no alias resolution — a DB `"video"` is dropped there but resolves to `olv` here). Make this the single vocabulary seam, add an enum-returning sibling (`normalize_channels(raw) -> list[MediaChannel]`, this one wrapping it), and route both sibling loops through it so unknown-handling and alias resolution have one definition.

src/core/tools/capabilities.py
```diff
@@ -68,9 +68,38 @@
...
+    """
+    canonical_values = {member.value for member in MediaChannel}
+    seen: set[str] = set()
+    normalized: list[str] = []
+
+    for channel in raw:
+        key = channel.lower().strip()
+        if key in CHANNEL_MAPPING:
+            value = CHANNEL_MAPPING[key].value
+        elif key in canonical_values:
```
KonstantinMirin, line 89: This `elif` is dead: `{m.value for m in MediaChannel} - set(CHANNEL_MAPPING) == set()` (verified), so any `key` that would satisfy it already matched the preceding `if key in CHANNEL_MAPPING`. Coverage can never green it. Remove the branch — the mapping lookup alone is the seam.

src/core/tools/capabilities.py
```diff
@@ -68,9 +68,38 @@
...
+            value = key
+        else:
+            logger.warning("Unknown channel value %r; skipping", channel)
```
KonstantinMirin, line 92: `%`-style logging against the file's f-string convention — the two neighbouring warnings (`:155`, `:171`) use f-strings, as do 128 sites across `src/core/tools/`. Align: `logger.warning(f"Unknown channel value {channel!r}; skipping")`.

tests/integration/test_list_authorized_properties_integration.py
```diff
@@ -279,15 +266,69 @@ def test_list_authorized_properties_tenant_isolation(integration_db):
...
+        tenant = TenantFactory(tenant_id="test_channels_unset", subdomain="chunset")
+        PublisherPartnerFactory(tenant=tenant, publisher_domain="alpha.com")
+        PublisherPartnerFactory(tenant=tenant, publisher_domain="beta.com", supported_channels=None)
+        env._commit_factory_data()
+
+    identity = PrincipalFactory.make_identity(tenant_id="test_channels_unset")
+    response = _list_authorized_properties_impl(req=None, identity=identity)
+
+    assert response.primary_channels is None
+    assert "primary_channels" not in response.model_dump()
```
KonstantinMirin, line 292: `model_dump()` is not the serialization any transport uses — this proves the model omits the field, not that the wire envelope does via `exclude_none`. `primary_channels` is a four-transport response field; grade it on the wire by wiring the dormant `BR-UC-007` feature and asserting the key is absent from `wire_dict(ctx)`. Same applies to the union/present assertions at `:316`/`:334`.

KonstantinMirin (commented):
## Review — PR #1848

**Overview** — #1533 Phase 1 asks for narrow property-channel disclosure: a `supported_channels` column on `PublisherPartner`, an admin path that writes it, and `list_authorized_properties` surfacing the union as `primary_channels`. The read half is built as the issue prescribed — right table, single-head migration, model returned unchanged, `exclude_none` omission, `list[str]` kept. Phase 1 is **incomplete**: AC#2 (admin create/edit persists `supported_channels`) is absent, so no shipped path can write the column and `primary_channels` can only be populated by a direct DB write. The disclosure is also graded entirely in-process against `_impl`/`model_dump()` while its four-transport BDD vehicle sits dormant, so nothing proves the field crosses a wire.

### Should fix

**No write boundary — the feature is unreachable, and validation landed on the read path as a result.**
`grep -rn supported_channels src/admin src/core/database/repositories` is empty. No transport and no admin UI writes the column; `primary_channels` is fillable only by a direct DB write. AC#3 asked for `MediaChannel` validation "at the write boundary"; with that boundary dropped, `normalize_channel_strings()` runs inside `_aggregate_primary_channels` on every `list_authorized_properties` request (`properties.py:162`) — lowercasing, alias-resolving, deduping and re-validating each partner's channels per read, warning per read instead of once at ingestion. The misplaced validation is a direct consequence of the missing write path: a disclosure that surfaces a column no user-facing path can fill is plumbing, not a feature. Build the admin/repository write of `supported_channels`, normalize at the point of write, and the read aggregation then unions already-canonical values. The deferral is parked on #1532, which is OPEN and UNASSIGNED (verified), so it is not an established, author-owned park. Either land the write path here (missing infra is a reason to build it) or file and self-assign an owned follow-up naming this exact scope, and treat Phase 1 as incomplete rather than done.

**The disclosure never crosses a wire — graded in-process while BR-UC-007 sits dormant.**
`list_authorized_properties` is a four-transport tool (MCP `properties.py:210`, A2A, REST, ADK); `primary_channels` must appear identically on all four and omit identically via `exclude_none`. That holds only by construction (the impl returns the model) and is verified on zero transports. Every new test calls `_list_authorized_properties_impl(...)` in-process and asserts `response.primary_channels` / `"primary_channels" not in response.model_dump()` (integration `:292,:316,:334`; unit `:934,:951`). `model_dump()` proves the Pydantic model is self-consistent, not what a buyer receives. The vehicle already exists: `tests/bdd/features/BR-UC-007-list-authorized-properties.feature` carries `primary_channels` scenarios but is dormant — no `scenarios()` binding, no step defs, no UC-007 harness, so it hits `pytest.xfail("No harness wired for UC-007")`. Its scenarios are forked per transport (`@main-rest`/`@main-mcp`, and the `@rest` body sends "via A2A"), and its Given seeds a tenant-level `primary_channels` config that does not match the partner `supported_channels` seam this PR built.

Wire `BR-UC-007` as transport-blind scenarios (Given/When/Then never name a transport; Given seeds partner `supported_channels` via env methods; `dispatch_request` is the sole writer of `ctx["result"]`/`wire_response`), register a UC-007 harness, and grade `primary_channels` through `wire_field`/`wire_dict` across a2a/mcp/rest/e2e — present-when-declared, key-absent-from-`wire_dict` for omission, sorted-deduped union, unknown-skipped. Fix the mismatched Given, collapse the per-transport forks into one parametrized scenario, then delete the `_impl`/`model_dump()` field grading and the redundant mock disclosure copy (`test_authorized_properties_behavioral.py:915`). Keep `test_channel_normalization.py` as the pure-function unit.

**`sponsored_intelligence` — the one enum AC#5 chartered — has no test.**
The PR adds `"sponsored_intelligence": MediaChannel.sponsored_intelligence` to `CHANNEL_MAPPING` (`capabilities.py:71`), but nothing asserts it survives normalization. Delete the line and every test still passes; the union/omit/skip cases use `display/ctv/olv/video` and `not_a_real_channel`. Add `normalize_channel_strings(["sponsored_intelligence"]) == ["sponsored_intelligence"]` in `test_channel_normalization.py`, or fold the value into the wired union scenario.

**`normalize_channel_strings()` is the third hand-rolled channel-vocabulary mapper.**
Three implementations of "raw channel strings → validated canonical MediaChannel vocabulary" now diverge on the two axes any bug fix must touch together:
- `normalize_channel_strings()` (new, `capabilities.py:75`): alias-resolve, log+skip unknown, dedupe+sort, `list[str]`.
- the adapter loop in `_get_adcp_capabilities_impl` (`capabilities.py:151-153`, same file, ~60 lines below): silent-skip unknown, no dedupe, `list[MediaChannel]`.
- `product_conversion.py:463`: `MediaChannel(ch)` + `except ValueError`, no alias resolution — a DB-stored `"video"` is dropped as unknown there but resolves to `olv` in the new helper, a live inconsistency the moment a partner and a product both store `"video"`.

This is the extract-a-helper-then-skip-the-siblings shape, with the un-converted sibling in the same file the PR edits. Make `normalize_channel_strings()` the single vocabulary seam, add an enum-returning sibling beside it (`normalize_channels(raw) -> list[MediaChannel]`, the string version wrapping it), and route both sibling loops through it so unknown-handling and alias resolution have one definition. Two things fall out of the same seam: the `elif key in canonical_values` branch (`capabilities.py:89`) is dead — `{m.value for m in MediaChannel} - set(CHANNEL_MAPPING) == set()`, so every canonical value already matched the preceding `if` — remove it; and the `%`-style `logger.warning` at `:92` is the minority style in a file whose neighbouring warnings use f-strings — align it.

**`_aggregate_primary_channels(publishers)` ships without its type contract.**
`def _aggregate_primary_channels(publishers) -> list[str] | None:` (`properties.py:33`) leaves `publishers` bare — the only unannotated signature in two modules whose every other helper is fully typed. The concrete type is in-repo: the argument is always `TenantConfigUoW.list_publisher_partners() -> list[PublisherPartner]`, and the body reads `partner.supported_channels`. Untyped, a future caller passing domain strings or dicts fails only at runtime inside the loop. Annotate `publishers: Iterable[PublisherPartner]` (`Sequence`/`list[PublisherPartner]` equally fine) — not `dict[str, Any]`/`object`.

**Spec-Grounding Gate cites a superseded version.**
The PR description cites AdCP `3.1.0-beta.3` / `adcp==5.7.0` with schema paths under `dist/schemas/3.1.0-beta.3/...`. The repo pins **3.1.1 / `adcp==6.6.0`** (`pyproject.toml:10`, `EXPECTED_SPEC_VERSION="3.1.1"`, `adcp.get_adcp_spec_version() → 3.1.1`). The substance holds — `sponsored_intelligence` and every tested value are present in the pinned 3.1.1 `channels.json`, and `native` is correctly unknown — but the provenance points at a version the repo no longer pins. Re-cite to 3.1.1 / `adcp==6.6.0`, repath the refs to `dist/schemas/3.1/core/property.json#supported_channels` and `enums/channels.json`, and keep the "ungraded — local extension" note.

- The `ResolvedIdentity` inline cap for `test_list_authorized_properties_integration.py` drops 6 → 0, with all six literals replaced by `PrincipalFactory.make_identity(...)` — a genuine shrink of a shrink-only ratchet.
- The two accepted-but-never-applied sibling filters on `list_authorized_properties` (`req.publisher_domains`, `req.property_tags`) and the Phase-2 `channels` request filter are out of scope here — the issue scoped them to Phase 2. If Phase 2 is greenlit, resolve all three filters in one change; a working third filter beside two dead ones is the inconsistency to avoid.

tests/unit/test_authorized_properties_behavioral.py
```diff
@@ -910,3 +910,42 @@ def test_policy_none_value(self):
             result = _list_authorized_properties_impl(req=None, identity=identity)
 
         assert result.advertising_policies is None
+
+
+class TestPrimaryChannelsDisclosure:
```
KonstantinMirin, line 915: This mocks `TenantConfigUoW` and grades `_impl` in-process — the union assertion is byte-identical to the integration test, and both would pass even if the transport serialization crashed. Once `BR-UC-007` is wired across the four transports, delete this class; keep the pure-function coverage in `test_channel_normalization.py`.

src/core/tools/properties.py
gagliathebuilder: Fixed in `83fcc22e` (follow-ups `aadb4dcb`, `86be0983`). POST create and PATCH update go through `TenantConfigUoW`; `canonicalize_supported_channels()` runs at write. Malformed or unknown values return 422 with no overwrite; empty list stores `None`; missing PATCH key leaves existing channels. `_aggregate_primary_channels()` now unions already-canonical stored values. Scoped to publisher-partner editing in tenant settings; `channel_selector.html` / #1532 untouched.

src/core/tools/capabilities.py
gagliathebuilder: Fixed in `464584f7` — `normalize_channel_strings(["sponsored_intelligence"]) == ["sponsored_intelligence"]` is now asserted in `test_channel_normalization.py`.

src/core/tools/properties.py
gagliathebuilder: Fixed in `464584f7` — `_aggregate_primary_channels(publishers: Iterable[PublisherPartner])`.

src/core/tools/capabilities.py
gagliathebuilder: Deferred for this pass. The adapter loop in `capabilities.py` and the `product_conversion.py` mapper still have their own seams. Follow-up unless this needs to land before ready-for-review.

src/core/tools/capabilities.py
gagliathebuilder: Fixed in `464584f7`. `{m.value for m in MediaChannel} - set(CHANNEL_MAPPING)` was empty, so the `elif` was deleted. Suite still green.

src/core/tools/capabilities.py
gagliathebuilder: Fixed in `464584f7` — warning now uses `logger.warning(f"Unknown channel value {channel!r}; skipping")`.

tests/unit/test_authorized_properties_behavioral.py
gagliathebuilder: Deferred until the wire-level BR-UC-007 coverage in the sibling thread lands, so we do not drop in-process coverage ahead of a replacement.

tests/integration/test_list_authorized_properties_integration.py
gagliathebuilder: Deferred. Model/integration coverage remains; BR-UC-007 is not yet wired through the post-#1868 harness, so omit/present is not graded on the wire in this pass.

src/admin/blueprints/publisher_partners.py
```diff
@@ -89,80 +137,87 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
...
+        return jsonify(payload), 201
 
+    except InvalidChannelInput as e:
+        return jsonify({"error": str(e)}), 422
     except Exception as e:
         logger.error(f"Error adding publisher partner: {e}", exc_info=True)
         return jsonify({"error": str(e)}), 500
 
 
+@publisher_partners_bp.route("/<tenant_id>/publisher-partners/<int:partner_id>", methods=["PATCH"])
```
KonstantinMirin, line 183: This new mutating route carries no `@require_tenant_access()`. The blueprint is 0/6 decorated while `authorized_properties.py` is 11/11, `principals.py` 13/13 and `products.py` 7/7, and `src/admin/app.py` has no global auth `before_request` — so an anonymous caller who puts a tenant id in the path can rewrite that tenant's disclosed channels. Add `@require_tenant_access(api_mode=True)` here and on the POST this diff rewrote, and authenticate the tests so the guard is exercised.

src/admin/blueprints/publisher_partners.py
```diff
@@ -89,80 +137,87 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
...
+        with TenantConfigUoW(tenant_id) as uow:
+            assert uow.tenant_config is not None
+            partner = uow.tenant_config.update_publisher_partner(partner_id, **fields)
+            if partner is None:
+                return jsonify({"error": "Publisher not found"}), 404
+            payload = _partner_json(partner)
+
+        return jsonify(payload)
+
+    except InvalidChannelInput as e:
```
KonstantinMirin, line 214: This `except InvalidChannelInput` is unreachable: `_parse_supported_channels` is the only raiser and the inner `try` at `:198-201` already catches it. Below it, `except Exception: jsonify({"error": str(e)}), 500` returns raw exception text (SQLAlchemy messages, `AttributeError` text) to the HTTP client. Root: exception-to-status translation is written per route. Make the channel error an `AdCPValidationError` subclass in `src/core/exceptions.py`, register one blueprint `errorhandler` for it, delete all three inline copies, and log with `exc_info=True` behind a fixed 500 message.

src/admin/blueprints/publisher_partners.py
```diff
@@ -89,80 +137,87 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
...
 
+@publisher_partners_bp.route("/<tenant_id>/publisher-partners/<int:partner_id>", methods=["PATCH"])
+def update_publisher_partner(tenant_id: str, partner_id: int) -> Response | tuple[Response, int]:
+    """Update display_name and/or supported_channels on a publisher partner."""
+    try:
+        data = request.get_json()
+        if not isinstance(data, dict):
+            return jsonify({"error": "JSON object required"}), 422
+
+        fields: dict[str, object] = {}
```
KonstantinMirin, line 191: `fields: dict[str, object]` exists only because the repository takes `**fields: object` and "key not provided" is encoded as absence from a dict — which is also why `_parse_supported_channels` returns a `(present, value)` tuple. Give `update_publisher_partner` explicit keyword params with an `UNSET` sentinel (see the comment on `tenant_config.py:87`) and pass real keywords here; the dict, the presence flag and the repository's runtime `allowed`-set check all disappear together.

src/admin/blueprints/publisher_partners.py
```diff
@@ -89,80 +137,87 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
     """Add a new publisher partner."""
     try:
         data = request.get_json()
-        publisher_domain = data.get("publisher_domain", "").strip().lower()
+        if not isinstance(data, dict):
+            return jsonify({"error": "JSON object required"}), 422
+
+        publisher_domain = _normalize_publisher_domain(data.get("publisher_domain", ""))
         display_name = data.get("display_name", "").strip()
```
KonstantinMirin, line 144: `data.get("display_name", "").strip()` is unguarded, so a non-string `display_name` raises `AttributeError` and surfaces as a 500 — while PATCH rejects the same input with 422 at `:193-196`. The two verbs hand-roll the same body validation and have already drifted after one push. Hoist it into one typed parser both verbs call (`_parse_partner_write(data, *, require_domain: bool) -> PublisherPartnerWrite`) so this field has one definition and one status code.

src/admin/blueprints/publisher_partners.py
```diff
@@ -89,80 +137,87 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
     """Add a new publisher partner."""
     try:
         data = request.get_json()
-        publisher_domain = data.get("publisher_domain", "").strip().lower()
+        if not isinstance(data, dict):
+            return jsonify({"error": "JSON object required"}), 422
```
KonstantinMirin, line 141: This is the first `422` in `src/admin/` — the layer's entire client-error vocabulary is `400` (66 sites) — and `:147` in this same handler answers a missing required field with `400`, so one endpoint uses two codes for the same class of failure. Return `400` like every sibling admin endpoint, or decide 422 for the admin layer as a whole through `src/core/exceptions.py`'s status mapping rather than in one of six routes.

src/admin/blueprints/publisher_partners.py
```diff
@@ -17,13 +17,71 @@
...
+from src.core.database.repositories.uow import TenantConfigUoW
 from src.core.domain_config import get_tenant_url
+from src.core.tools.capabilities import InvalidChannelInput, canonicalize_supported_channels
 
 logger = logging.getLogger(__name__)
 
 publisher_partners_bp = Blueprint("publisher_partners", __name__)
 
 
+def _partner_json(partner: PublisherPartner, *, extra: dict | None = None) -> dict:
```
KonstantinMirin, line 29: `-> dict` with an open `extra: dict | None` merge leaves the payload three routes emit and `static/js/tenant_settings.js:2184-2210` consumes declared nowhere: nothing catches a renamed or dropped key. This is an internal admin shape with no AdCP counterpart, so declare it — a `PublisherPartnerPayload` `TypedDict` with `property_count` and `message` as `NotRequired`, and named keyword params replacing `extra`. Drop the `partner.created_at if ... else None` guard too; the column is NOT NULL with `server_default=func.now()`, so it only widens the declared type.

src/admin/blueprints/publisher_partners.py
```diff
@@ -17,13 +17,71 @@
 from src.core.config import get_config
 from src.core.database.database_session import get_db_session
 from src.core.database.models import PublisherPartner, Tenant
+from src.core.database.repositories.uow import TenantConfigUoW
 from src.core.domain_config import get_tenant_url
+from src.core.tools.capabilities import InvalidChannelInput, canonicalize_supported_channels
```
KonstantinMirin, line 22: This imports an AdCP tool implementation (`capabilities.py` is the `get_adcp_capabilities` handler) into an admin blueprint for a validator; `tenants.py:25` does the same for UI label strings, so rendering the settings page now transitively pulls `fastmcp`, adapter construction and `src.services.targeting_capabilities`. Move the channel vocabulary to a core-owned module (e.g. `src/core/media_channels.py`, on the zero-project-import pattern of `src/core/enum_helpers.py`) and import that from both places.

src/admin/blueprints/publisher_partners.py
```diff
@@ -57,17 +115,7 @@ def list_publisher_partners(tenant_id: str) -> Response | tuple[Response, int]:
...
-                        "publisher_domain": partner.publisher_domain,
-                        "display_name": partner.display_name,
-                        "is_verified": partner.is_verified,
-                        "last_synced_at": partner.last_synced_at.isoformat() if partner.last_synced_at else None,
-                        "sync_status": partner.sync_status,
-                        "sync_error": partner.sync_error,
-                        "created_at": partner.created_at.isoformat(),
-                        "property_count": property_counts.get(partner.publisher_domain, 0),
-                    }
+                    _partner_json(partner, extra={"property_count": property_counts.get(partner.publisher_domain, 0)})
```
KonstantinMirin, line 118: This diff edits `list_publisher_partners` to call the new `_partner_json()` but leaves its `get_db_session()` + `select(Tenant)` + `select(PublisherPartner)` body and its raw-select allowlist entry, ten lines from the repository seam the same PR just built and used in `add_publisher_partner`. The blueprint's query also carries `.order_by(PublisherPartner.publisher_domain)` while `TenantConfigRepository.list_publisher_partners()` does not — two definitions of "the tenant's partners". Convert this handler to `TenantConfigUoW`, add `count_properties_by_domain() -> dict[str, int]` to the repository for the aggregate, give the repository the deterministic `order_by`, and drop the allowlist entry so the ratchet shrinks by two.

src/core/tools/capabilities.py
```diff
@@ -69,9 +69,59 @@
     "influencer": MediaChannel.influencer,
     "affiliate": MediaChannel.affiliate,
     "product_placement": MediaChannel.product_placement,
+    "sponsored_intelligence": MediaChannel.sponsored_intelligence,
```
KonstantinMirin, line 72: Hand-adding `sponsored_intelligence` keeps a hand-maintained copy of the enum in step by hand — the "do not hardcode enum lists" guardrail #1533 names. `CHANNEL_MAPPING` restates all 20 canonical `MediaChannel` values, this line being the last of them. Verified pure duplication: `{m.value for m in MediaChannel} - set(CHANNEL_MAPPING)` is empty and the only information the dict adds is `{"video", "audio"}`. Derive it instead: `CHANNEL_MAPPING = {m.value: m for m in MediaChannel} | {"video": MediaChannel.olv, "audio": MediaChannel.streaming_audio}`, in a core-owned vocabulary module the adapter loop at `:174`, `product_conversion.py:463`, the admin write and the picker all route through.

src/core/tools/capabilities.py
```diff
@@ -69,9 +69,59 @@
...
+    """
+    if not isinstance(raw, list) or not all(isinstance(item, str) for item in raw):
+        raise InvalidChannelInput("supported_channels must be a list of strings")
+    unknown = [item for item in raw if item.lower().strip() not in CHANNEL_MAPPING]
+    if unknown:
+        raise InvalidChannelInput(f"Unknown channel values: {unknown}")
+    return normalize_channel_strings(raw)
+
+
+def normalize_channel_strings(raw: list[str]) -> list[str]:
```
KonstantinMirin, line 100: After the write boundary landed this function's log-and-skip branch has no reachable production caller: `grep -rn normalize_channel_strings src/` returns one call site, `canonicalize_supported_channels` at `:97`, which has already rejected every unknown. So the module ships two public functions with contradictory unknown-value policies and only one is live, with the same key-normalize-and-membership test written twice per call. Fold both into one private resolver walked once, exposing a strict entry point and an enum-returning `resolve_channels(raw) -> list[MediaChannel]` the adapter and product paths wrap, and delete the standalone public name.

src/core/database/repositories/tenant_config.py
```diff
@@ -46,6 +50,54 @@ def list_publisher_partners(self) -> list[PublisherPartner]:
...
+            supported_channels=supported_channels,
+            sync_status=sync_status,
+            is_verified=is_verified,
+            last_synced_at=last_synced_at,
+        )
+        self._session.add(partner)
+        self._session.flush()
+        return partner
+
+    def update_publisher_partner(self, partner_id: int, **fields: object) -> PublisherPartner | None:
```
KonstantinMirin, line 87: `**fields: object` plus a runtime `allowed` set spends a compile-time contract on a string check: `supported_channels="display,ctv"` type-checks and fails at the DB, and `setattr(partner, key, value)` with `value: object` lets any type reach the column. It also forks the sibling convention (`product.py:140`, `account.py:232`, `media_buy.py:456` use `_IMMUTABLE_FIELDS` denylists over wide models; this is an allowlist over two fields). Use explicit keywords with a sentinel — `def update_publisher_partner(self, partner_id: int, *, display_name: str | Unset = UNSET, supported_channels: list[str] | None | Unset = UNSET)` — and delete the `allowed` set and its `ValueError`; the signature is the check. Better still, take `list[MediaChannel] | None` so "stored implies on-wire canonical" is a type fact rather than one handler's discipline.

templates/tenant_settings.html
```diff
@@ -2008,6 +2009,19 @@ <h3 style="margin: 0 0 1rem 0;">Add Publisher Partner</h3>
                             <small style="color: #6b7280;">Friendly name for this publisher</small>
                         </div>
 
+                        <div class="form-group">
+                            <label>Supported Channels</label>
+                            <div id="publisher-channels" style="display: grid; grid-template-columns: 1fr 1fr; gap: 0.35rem 1rem; max-height: 220px; overflow-y: auto; padding: 0.5rem; border: 1px solid #e5e7eb; border-radius: 6px;">
+                                {% for channel in media_channels|default([]) %}
```
KonstantinMirin, line 2015: This grid renders the picker from the enum, which is the right mechanism — applied in a sixth place. `templates/components/channel_selector.html` still hand-lists a stale vocabulary including `native`, which this PR's own test asserts is not a valid channel, and five product templates include it. Convert the shared component to render from the vocabulary seam's `media_channel_choices()` (that is #1532, and it is a template loop), parameterize it for multi-select against checkbox grid, and include it here instead of the inline copy. The `media_channels|default([])` fallback also means a view that forgets the context key renders zero checkboxes silently.

KonstantinMirin (commented):
## Review — PR #1848

**Overview** — #1533 Phase 1 asks for narrow property-channel disclosure: a `supported_channels`
column on `PublisherPartner`, an admin path that writes it, and `list_authorized_properties`
surfacing the union as `primary_channels`. The problem is framed the way the issue framed it, and
the storage/read half has the right shape: one column on the right table, single-head migration
`15238294201b`, canonicalization moved to the write boundary, `_impl` returning the declared model
so no transport can fork the field. Phase 1 is still partial. The two design obligations raised on
2026-08-05 are unaddressed — the channel vocabulary still has no owning module (this push added two
more restatements of it), and the disclosure is verified on zero wires while the harness that would
verify it already exists. The new write surface also sits beside the admin-blueprint standard
rather than on it: the mutating routes carry no tenant-access guard, and the tests that would have
caught that never authenticate.

### Should fix

**1. ↩ raised 2026-08-05, still open — the disclosure is graded on zero wires, because the tests
were built on an env that has no transport by construction.**

The cause is the test setup, not the assertions. This PR wrote a new
`_PropertiesEnv(IntegrationEnv)` with `EXTERNAL_PATCHES = {}` and nothing else
(`tests/integration/test_list_authorized_properties_integration.py:16`) for the one tool that
already owns `tests/harness/authorized_properties.py::AuthorizedPropertiesEnv`
(`call_impl`/`call_a2a`/`call_mcp`/`REST_ENDPOINT`, driven by `call_via(transport) ->
TransportResult`). The stripped copy exposes no dispatch surface, so in-process grading was the
only thing available, `model_dump()` became the stand-in for serialization, and the tests then
reach past the harness API (`env._commit_factory_data()`, `:286`, `:308`).

What that costs, concretely:

- `tests/integration/test_wire_omission_matrix.py:175` already grades `primary_channels` absent on
  a2a/mcp/rest for this exact tool, through `AuthorizedPropertiesEnv.call_via`. The new omission
  pair (`:292`, and `test_authorized_properties_behavioral.py:934`) is a weaker duplicate of
  coverage that already runs on three wires. The one case the matrix cannot cover, the populated
  union, is the actual new behavior and the one graded only in memory, twice.
- The mock copy does not exercise the branch it names. `_make_publisher()` returns a bare
  `MagicMock`, so `partner.supported_channels` is truthy and iterates empty
  (`bool(MagicMock().supported_channels)` is `True`). Delete the `if partner.supported_channels:`
  guard at `src/core/tools/properties.py:37` and the test stays green while production raises on
  the first row with `supported_channels IS NULL`.
- `Covers: UC-007-SCHEMA-02` (`test_authorized_properties_behavioral.py:918`) claims a
  serialization obligation ("when the channels are serialized, each value is one of the
  standardized enum values") that the test neither serializes nor checks membership for, and no
  guard scans that file, so nothing catches the mismatch.
- `tests/bdd/features/BR-UC-007-list-authorized-properties.feature` carries the two scenarios that
  assert exactly this field and is fully dormant: no `scenarios()` binding, no steps, no `UC-007`
  branch in `tests/bdd/conftest.py`. `grep -rn "UC-007" tests/bdd --include=*.py` is empty, so the
  scenarios are not even collected.

Delete `_PropertiesEnv` and use `AuthorizedPropertiesEnv`; express the populated-union case
through `call_via` on a2a/mcp/rest, asserting `result.wire_response["primary_channels"]`; and wire
BR-UC-007 with a `scenarios()` binding, a `UC-007` branch in `tests/bdd/conftest.py`, and
transport-blind steps whose Given seeds partner `supported_channels` (replacing the feature's
tenant-level `primary_channels` Given at `:29`, `:48`) and whose Then reads `wire_field` /
`"primary_channels" not in wire_dict(ctx)`. Collapse the `@main-rest`/`@main-mcp` forks; the
`@rest` scenario's When already says "via A2A", so there is no transport mechanic to fork on.
Assert the union values against `{m.value for m in MediaChannel}` there, which is what
UC-007-SCHEMA-02 actually says, and carry the `Covers:` tag on a file the obligation guard scans.
Then delete the `model_dump()` grading and `TestPrimaryChannelsDisclosure`. Because the value is a
response field on the wire, this needs no `e2e_rest_known_failures`, `_NO_E2E_REST_TAGS` or
`*_E2E_*_INTERNAL_TAGS` entry; a follow-up that proposes one would be growing a shrink-only
ratchet, not fixing this. The deferral reason given on 2026-08-05 ("BR-UC-007 is not yet wired
through the post-#1868 harness") does not hold at this head: the env, the three-transport dispatch,
the REST endpoint, the wire helpers and a working precedent are all already in the tree.

**2. ↩ raised 2026-08-05, still open — the channel vocabulary has no owning module, so every
consumer restates it; this push added two more restatements plus a hand-written enum entry.**

The root is not "N mappers exist". "The set of valid channels and how raw strings map into it" is
data with no owner, so each consumer either re-implements it or reaches into
`src/core/tools/capabilities.py`, which is the `get_adcp_capabilities` handler, not a vocabulary
home. Six restatements at this head:

- `capabilities.py:50-73` hand-lists all 20 canonical `MediaChannel` values plus 2 aliases, and
  this PR hand-adds the 22nd line (`sponsored_intelligence`) — the exact thing #1533's guardrails
  forbid. Verified pure duplication: `{m.value for m in MediaChannel} - set(CHANNEL_MAPPING)` is
  empty and `set(CHANNEL_MAPPING) - {m.value for m in MediaChannel}` is `{"video", "audio"}`, with
  nothing guarding the equality.
- `capabilities.py:174-175`, the adapter loop ~60 lines below the helpers this PR adds, in the file
  this PR edits: `channel_name.lower() in CHANNEL_MAPPING`, silent skip, no `.strip()`, no dedupe.
- `src/core/product_conversion.py:463`: `MediaChannel(ch)` with no alias resolution, so a stored
  `"video"` becomes `olv` on the partner path and is dropped on the product path.
- `templates/components/channel_selector.html:15-25`, a hand-listed `<select>` including `native`,
  included by five product templates, while this PR's own test asserts `native` is not a valid
  channel.
- `templates/tenant_settings.html:2015` plus `src/admin/blueprints/tenants.py:424`: a new bespoke
  checkbox grid fed by `canonical_media_channel_values()`. The right mechanism, applied in a sixth
  place instead of at the seam.
- `capabilities.py:85-121`: two public functions with contradictory unknown-value policies
  (`canonicalize_supported_channels` raises, `normalize_channel_strings` logs and skips), the same
  key-normalize-and-membership test written twice per call, and the log-and-skip branch left with
  no reachable production caller — `grep -rn normalize_channel_strings src/` returns one call site,
  inside `canonicalize_supported_channels`, which has already rejected every unknown.

Fallout from the current placement: `src/admin/blueprints/publisher_partners.py:22` and
`tenants.py:25` now import an AdCP tool implementation for a validator and for UI label strings, so
rendering the settings page transitively pulls `fastmcp`, adapter construction and
`src.services.targeting_capabilities`. `InvalidChannelInput` is the only exception class in `src/`
defined outside `src/core/exceptions.py`, and its message renders a raw Python list repr straight
into an HTTP body (`Unknown channel values: ['native']`).

Give the vocabulary one owner: a core module (`src/core/media_channels.py`, on the
zero-project-import pattern of `src/core/enum_helpers.py` so tools, admin blueprints and template
context can all import it without a cycle) that derives the canonical half from the SDK enum and
hand-lists only the true input aliases — `CHANNEL_MAPPING = {m.value: m for m in MediaChannel} |
{"video": MediaChannel.olv, "audio": MediaChannel.streaming_audio}`. One private resolver walked
once, three entry points over it: strict `canonicalize_channels(raw) -> list[str]` that raises for
write boundaries, `resolve_channels(raw) -> list[MediaChannel]` that the lenient consumers wrap,
and `media_channel_choices()` for the UI. Route the admin write, the adapter loop,
`product_conversion.py:463` and the picker through it; `channel_selector.html` renders from the
same list and is reused (multi-select vs checkbox grid) for both forms. That is #1532, and it is a
template loop. Make the strict failure an `AdCPValidationError` subclass in
`src/core/exceptions.py` so it carries a code and a status instead of being hand-mapped per route.
Do not vendor a schema copy or re-list values; the SDK enum is the single upstream pin. On the
deferral: `gh issue view 1532` returns OPEN with `assignees: []`, and the mapper consolidation was
deferred in the 2026-08-05 thread with no issue number, so neither slice is parked on an owned
ticket today.

**3. The new mutating admin routes ship without the tenant-access guard, and the hand-rolled test
client is what let that pass.**

`src/admin/blueprints/publisher_partners.py:183` (the new PATCH) and the POST this diff rewrote
carry no `@require_tenant_access()`. The blueprint is 0/6 decorated while `authorized_properties.py`
is 11/11, `principals.py` 13/13 and `products.py` 7/7, and `src/admin/app.py` has no global auth
`before_request`. `tests/integration/test_publisher_partner_channel_write.py` builds a module-level
`create_app()` and a local `client` fixture that establishes no session, and still gets 201/200
with real DB mutations. `test_patch_is_tenant_isolated:152-172` asserts 404 for a cross-tenant
PATCH from an anonymous client, so it grades repository path-scoping, not authorization, and stays
green after a regression that lets any user edit any tenant's partners.

The root is that a new state-changing admin API was authored without adopting the admin-blueprint
standard — route guard, error vocabulary, exception-to-status mapping, test harness, BDD vehicle —
because the one blueprint it landed in is the local outlier and the tests hand-rolled their own app
and client instead of going through `tests/admin/conftest.py` / `AdminAccountEnv`. The rest of this
cluster follows from the same choice: six new `422`s in a layer whose entire client-error
vocabulary is `400` (66 sites), with `400` and `422` for the same class of failure inside one
handler (`:141` vs `:147`); `except InvalidChannelInput -> 422` written three times, with `:214`
unreachable because `:199` catches it first; and `except Exception: jsonify({"error": str(e)}), 500`
echoing raw exception text to the client.

Put `@require_tenant_access(api_mode=True)` on the two routes this PR authors. Move the tests to
`tests/admin/test_publisher_partners_blueprint.py` under `pytest.mark.admin` with the shared
authenticated fixtures, which makes the isolation case grade 403 against 404 instead of
path-scoping. Map the channel error once, as the `AdCPValidationError` subclass from finding 2 plus
a blueprint `errorhandler`, deleting all three inline copies and the `str(e)` 500 body. Use the
layer's `400` for client input, or decide 422 for the admin layer as a whole through
`src/core/exceptions.py`'s status mapping rather than in one of six routes.

The same boundary's contract is graded partially. POST with unknown channels returning 422 and
creating no row is untested, and it depends on the parse sitting outside the UoW block at `:149`.
`[]` storing `None` is untested on either verb, absent-PATCH-key-preserves is untested, and a
malformed body is untested. The repository write test
(`tests/integration/test_tenant_config_repository.py:120-135`) asserts after the `with _RepoEnv()`
block on a detached, never-committed instance: the repository flushes without committing and
`IntegrationEnv.__exit__` discards the transaction, so `assert updated.supported_channels ==
["ctv", "olv"]` reads back the value `setattr` put in `__dict__`. Drive it through
`TenantConfigUoW`, the boundary that owns the commit, and re-read through a fresh session the way
the sibling `_stored_channels()` (`test_publisher_partner_channel_write.py:44-49`) already does.
The admin BDD vehicle exists — `BR-ADMIN-ACCOUNTS.feature` bound by `tests/bdd/test_admin_accounts.py`
with `tests/harness/admin_accounts.py` providing an integration (Flask client) and an e2e (Docker)
transport behind one step vocabulary — so "rejected write leaves the stored value untouched"
belongs there as a transport-blind scenario instead of four more hand-written client calls.

**4. One agent now gives two answers to "what are your primary channels", and the spec-graded one
still answers from an adapter default.**

`src/core/tools/capabilities.py:164-181` builds the spec-defined `portfolio.primary_channels` from
adapter `default_channels` and hardcodes `primary_channels = [MediaChannel.display]` when that
yields nothing, inside `_get_adcp_capabilities_impl`, which twenty lines later already calls
`uow.tenant_config.list_publisher_partners()` (`:188`). This PR gives those same partner rows
explicit publisher-declared channels and surfaces their union on
`list_authorized_properties.primary_channels`, the local ungraded twin. A buyer calling both tools
on one agent gets contradictory answers, and the spec-graded one is the wrong one:
`dist/schemas/3.1.1/protocol/get-adcp-capabilities-response.json` → `portfolio.primary_channels` is
the only place AdCP 3.1.1 names this field. The divergence did not exist before this diff, because
nothing populated declared channels.

Move the partner load above the channel block and union the declared `supported_channels` first,
using the adapter loop only as the fallback, and drop the `[display]` invention. Routed through the
seam in finding 2, both tools then read the same vocabulary and the same source. If the two fields
are meant to differ, that is a legitimate answer, but it needs to be stated in the Spec-Grounding
note with the spec sentence that supports it.

**5. The write API was designed as a field bag, so "stored implies canonical" is a convention of
one Flask handler rather than a property of the API.**

`src/core/database/repositories/tenant_config.py:87` is
`update_publisher_partner(self, partner_id: int, **fields: object)` with a runtime
`allowed = {"display_name", "supported_channels"}` set and a `ValueError`. A two-field update API
has a compile-time contract; spending it on a string set means
`supported_channels="display,ctv"` type-checks and fails at the DB, and
`setattr(partner, key, value)` with `value: object` lets any type reach the column. It also forks
the repository convention: `product.py:140`, `account.py:232` and `media_buy.py:456` use an
`_IMMUTABLE_FIELDS` denylist over wide models, while this is an allowlist over two fields.

The weakness propagates outward by construction. The blueprint assembles `fields: dict[str, object]`
(`:191`) and encodes "key not provided" as absence from a dict, which is why
`_parse_supported_channels` has to return a `tuple[bool, list[str] | None]` presence flag. Two
hand-written body validators have already drifted after one push: PATCH rejects a non-string
`display_name` with 422 (`:193-196`), while POST calls `data.get("display_name", "").strip()`
unguarded (`:144`), so the same input is an `AttributeError` surfaced as a 500. And
`_partner_json(partner, *, extra: dict | None = None) -> dict` (`:29-44`) plucks nine ORM
attributes and merges an arbitrary caller mapping, so the shape
`static/js/tenant_settings.js:2184-2210` consumes is declared nowhere and open by construction.

Replace `**fields: object` with explicit keyword parameters and a module-level sentinel —
`def update_publisher_partner(self, partner_id: int, *, display_name: str | Unset = UNSET,
supported_channels: list[str] | None | Unset = UNSET) -> PublisherPartner | None` — which deletes
the `allowed` set, its `ValueError` and the blueprint's `dict[str, object]` in one move, because
the signature becomes the check. Hoist body parsing into one typed parser both verbs call
(`_parse_partner_write(data, *, require_domain: bool) -> PublisherPartnerWrite`, a dataclass
carrying `display_name`, `supported_channels` and the sentinel) so `display_name` validation has
one definition and one status code. Give `_partner_json` a `PublisherPartnerPayload` `TypedDict`
with `property_count` and `message` as `NotRequired` keys and named parameters, replacing the open
`extra` bag; these are internal admin shapes with no AdCP counterpart, so a named record is the
target rather than `dict[str, Any]`. Then put the vocabulary on the write API's type — accept
`list[MediaChannel] | None`, or a `SupportedChannels` newtype only the strict canonicalizer can
produce — so "stored values are on-wire canonical" is a fact every writer inherits rather than one
handler's discipline. This is a missing type contract, not a request to defend against the DB: the
guarantee moves into the signature and no validation layer is added over the model. Drop the
`partner.created_at if ... else None` guard while you are there; the column is NOT NULL with
`server_default=func.now()`, so the guard only widens the declared type.

**6. The publisher-partner blueprint has no data or policy seam; this PR built both and wired one
call site of each.**

`TenantConfigRepository.get_publisher_partner()` / `get_publisher_partner_by_domain()` are new, and
`add_publisher_partner` was converted, correctly dropping its raw-select allowlist entry. Three
byte-equivalent hand-rolled queries remain in the same file: `publisher_partners.py:226` and `:542`
(`select(PublisherPartner).filter_by(id=..., tenant_id=...)`), `:98` and `:254`
(`select(PublisherPartner).filter_by(tenant_id=...)`). `list_publisher_partners` is edited by this
diff — its body now calls the new `_partner_json()` at `:118` — while keeping `get_db_session()`,
`select(Tenant)`, `select(PublisherPartner)`, a function-local `from sqlalchemy import func`, and
its allowlist entry. So one file expresses one kind of data access two ways, one of them still
pinned as debt. The two definitions of "the tenant's partners" also differ: the blueprint's query
carries `.order_by(PublisherPartner.publisher_domain)`, the repository's does not.

Same shape on the policy side: `_auto_verify_reasons(uow)` and `_publisher_added_message(is_dev,
is_mock)` were extracted by this PR (`:63-82`) and applied to `add_publisher_partner` only, while
`sync_publisher_partners` (`:264-274`) re-derives `is_dev`/`is_mock` and rebuilds the same
`" and ".join(reasons)` string inline, already sourcing the adapter differently
(`tenant.adapter_config` lazy relationship against `uow.tenant_config.get_adapter_config()`).

Route all publisher-partner data access through `TenantConfigUoW`/`TenantConfigRepository`. Add
`count_properties_by_domain() -> dict[str, int]` (the one genuinely missing piece; build it rather
than keep the raw group-by in a route) and a deterministic `order_by` on
`list_publisher_partners()` so the repository is the single definition. Convert
`list_publisher_partners`, the handler this diff already edits, and drop its allowlist entry so the
ratchet shrinks by two instead of one. The auto-verify rule is publisher-partner domain policy, not
Flask code: put it in a core/service function (`should_auto_verify_partner(...) -> AutoVerifyDecision`
carrying the flags and the reason list) that both the create route and the sync path call, with the
blueprint keeping only message formatting. `_normalize_publisher_domain()` (`:58`) is the same
category, write-boundary canonicalization that belongs beside the channel canonicalizer.

**7. The publisher modal was made dual-mode without a mode state, so mode-dependent values are set
per entry point and one was missed.**

`static/js/tenant_settings.js:2244-2255` (`showEditPublisherModal`) sets title, id, domain and
checkboxes but never the submit label, and `:2326` / `:2342` force
`submitBtn.textContent = 'Add Publisher'` unconditionally in both the `.then` and the `.catch`. The
button the user clicks reads "Add Publisher" throughout an edit and is reset to the create-mode
literal after any save or error, while every other mode-dependent value in `savePublisher` is
chosen from `isEdit`. The row cache is stashed on `window._publisherPartnersById` (`:2154-2157`,
`:2245`), the only `window._`-prefixed global in `static/js/`, so `showEditPublisherModal` silently
returns if `loadPublishers()` has not run. Introduce the mode state the feature needs: one
`setPublisherModalMode(mode, partner)` owning every mode-dependent value (title, submit label,
readOnly, target URL and method) and one restore path that restores the mode's label rather than a
literal, and carry the partner id on the row via `data-partner-id`, the pattern
`static/js/targeting_widget.js` uses, instead of a global.

- The 2026-08-05 items other than the two flagged above all landed: the write boundary through
  `TenantConfigUoW`, the `Iterable[PublisherPartner]` annotation, the dead `elif`, the f-string
  log, the `sponsored_intelligence` test, and the corrected 3.1.1 / `adcp==6.6.0` citation (the
  `dist/schemas/3.1.1/...` paths are correct as written). Both ratchets moved down: raw-select
  allowlist −1, `ResolvedIdentity` per-file cap 6 → 0.
- `delete_publisher_partner`, `get_publisher_properties` and `sync_publisher_partners` keep their
  raw-select allowlist entries and `get_db_session()` bodies. Out of scope: this diff does not
  touch those function bodies, unlike `list_publisher_partners`. They are the rest of the seam
  finding 6 opens, and `sync_publisher_partners` additionally fetches `adagents.json` inside the
  route, which is the natural home for the publisher-partner service.
- The rest of BR-UC-007 (auth, domain-filter, context-echo and policy scenarios, the three
  `@T-UC-007-enum-*` outlines) is equally dormant, and the feature also asserts `primary_countries`,
  `portfolio_description` and `last_updated`, which this PR does not populate. Out of scope:
  separate response fields and a larger wiring job. Finding 1 asks only for the `primary_channels`
  scenarios plus the binding; the rest going visibly red once the binding exists is the correct
  outcome and must not be silenced with an xfail or e2e-internal entry.
- `_normalize_publisher_domain` (`publisher_partners.py:58`, lines the diff only relocated) and
  `src/services/property_discovery_service.py:33-44` disagree on what a normalized domain is (the
  latter strips `www.`/`m.` and does not strip the scheme). Out of scope: the diff does not touch
  the discovery service, and reconciling them is a separate boundary. Worth an owned follow-up if
  `UniqueConstraint(tenant_id, publisher_domain)` is ever expected to match discovery results.
- Deferral homes checked: `gh issue view 1532` returns OPEN with no assignee, and the mapper
  consolidation, the BR-UC-007 wiring and the unguarded GET/DELETE/sync routes were deferred in
  threads with no issue number at all. None of those is an established, author-owned deferral, so
  they stay above. If a slice is genuinely split off, file and self-assign an issue naming the
  exact scope.
- Phase 2's request-side `channels` filter and the two accepted-but-never-applied sibling filters
  (`req.publisher_domains`, `req.property_tags`) are out of scope: #1533 scopes them to Phase 2 and
  this diff does not touch them.

tests/integration/test_list_authorized_properties_integration.py
```diff
@@ -8,8 +8,15 @@
 
 from src.core.database.database_session import get_db_session
 from src.core.database.models import PublisherPartner, Tenant
-from src.core.resolved_identity import ResolvedIdentity
 from src.core.tools.properties import _list_authorized_properties_impl
+from tests.factories import PrincipalFactory, PublisherPartnerFactory, TenantFactory
+from tests.harness._base import IntegrationEnv
+
+
+class _PropertiesEnv(IntegrationEnv):
```
KonstantinMirin, line 16: This tool already owns `tests/harness/authorized_properties.py::AuthorizedPropertiesEnv` (`call_impl`/`call_a2a`/`call_mcp`/`REST_ENDPOINT`, driven by `call_via(transport) -> TransportResult`), used by `test_wire_omission_matrix.py`. The stripped local copy has no dispatch surface, which is why the new assertions can only grade in process and why they reach past the harness API with `env._commit_factory_data()` at `:286`/`:308`. Delete `_PropertiesEnv`, use `AuthorizedPropertiesEnv`, and dispatch through `call_via`.

tests/integration/test_list_authorized_properties_integration.py
```diff
@@ -279,15 +266,51 @@ def test_list_authorized_properties_tenant_isolation(integration_db):
...
+        tenant = TenantFactory(tenant_id="test_channels_unset", subdomain="chunset")
+        PublisherPartnerFactory(tenant=tenant, publisher_domain="alpha.com")
+        PublisherPartnerFactory(tenant=tenant, publisher_domain="beta.com", supported_channels=None)
+        env._commit_factory_data()
+
+    identity = PrincipalFactory.make_identity(tenant_id="test_channels_unset")
+    response = _list_authorized_properties_impl(req=None, identity=identity)
+
+    assert response.primary_channels is None
+    assert "primary_channels" not in response.model_dump()
```
KonstantinMirin, line 292: `model_dump()` is not the serialization any transport performs — it proves the Pydantic model is self-consistent, not what a buyer receives. `test_wire_omission_matrix.py:175` already grades `primary_channels` absent on a2a/mcp/rest for this tool, so this assertion is a weaker duplicate of coverage that already runs on three wires, while the populated union at `:316` (the actual new behavior) is graded on none. Move the union case onto `AuthorizedPropertiesEnv.call_via(...)` with `wire_field`/`wire_dict`, wire BR-UC-007, then delete these.

tests/unit/test_authorized_properties_behavioral.py
```diff
@@ -910,3 +910,42 @@ def test_policy_none_value(self):
...
+    """Phase 1 channel disclosure via PublisherPartner.supported_channels.
+
+    Covers: UC-007-SCHEMA-02
```
KonstantinMirin, line 918: UC-007-SCHEMA-02 is a serialization obligation ("when the channels are serialized, each value is one of the standardized media-channel enum values"). This class never serializes and never checks membership against `MediaChannel`; it asserts a hardcoded list produced from mock inputs it supplied. The tag is also inert — this file is in neither `_UNIT_ENTITY_FILES` nor the obligation guard's scan surface — so nothing catches the mismatch. Attach the claim to the wired BR-UC-007 scenario asserting the serialized values against `{m.value for m in MediaChannel}`, on a file the guard scans, and drop the tag here.

tests/unit/test_authorized_properties_behavioral.py
```diff
@@ -910,3 +910,42 @@ def test_policy_none_value(self):
...
+    """
+
+    def test_primary_channels_omitted_when_no_partner_declares_channels(self):
+        """Partners without supported_channels do not contribute to primary_channels."""
+        from src.core.tools.properties import _list_authorized_properties_impl
+
+        tenant = _make_mock_tenant("tenant-ch")
+        identity = _make_identity(tenant)
+        publishers = [_make_publisher("example.com")]
```
KonstantinMirin, line 927: `_make_publisher()` returns a bare `MagicMock`, so `partner.supported_channels` is truthy and iterates empty (`bool(MagicMock().supported_channels)` is `True`). This test takes the True branch of `if partner.supported_channels:` (`properties.py:37`), the opposite of a real row with an unset column: delete that guard and this test stays green while production raises on the first `NULL` row. Delete this class once the union and omission cases are graded on the wire against real rows; `test_channel_normalization.py` keeps the pure-function coverage.

tests/integration/test_publisher_partner_channel_write.py
```diff
@@ -0,0 +1,172 @@
...
+from src.core.database.repositories.tenant_config import TenantConfigRepository
+from tests.factories import PublisherPartnerFactory, TenantFactory
+from tests.harness._base import IntegrationEnv
+
+app = create_app()
+
+pytestmark = [pytest.mark.integration, pytest.mark.requires_db]
+
+
+class _WriteEnv(IntegrationEnv):
```
KonstantinMirin, line 24: `_WriteEnv` is a line-for-line copy of `tests/harness/_base.py::BareIntegrationEnv` (`EXTERNAL_PATCHES = {}` plus a `get_session()` that commits factory data), which four integration modules already import. Import it and delete this class — and since the diff touches `test_tenant_config_repository.py`, collapse `_RepoEnv` onto it in the same pass so the copies stop multiplying.

tests/integration/test_publisher_partner_channel_write.py
```diff
@@ -0,0 +1,172 @@
...
+    EXTERNAL_PATCHES: dict[str, str] = {}
+
+    def get_session(self):
+        self._commit_factory_data()
+        return self._session
+
+
+@pytest.fixture
+def client():
```
KonstantinMirin, line 33: This fixture establishes no session, and the tests still get 201/200 with real DB mutations — that is the evidence the new routes have no `@require_tenant_access()`, not that they do not need one. It is also a strict subset of the shared `admin_client` (`tests/integration/conftest.py:76-88`), missing `PROPAGATE_EXCEPTIONS`, which the shared fixture sets explicitly so template/handler errors surface as tracebacks rather than 500 bodies. Move this file to `tests/admin/test_publisher_partners_blueprint.py` with the shared authenticated fixtures; the isolation test then grades 403 against 404 instead of repository path-scoping.

tests/integration/test_tenant_config_repository.py
```diff
@@ -112,3 +112,39 @@ def test_sorted_domains(self, integration_db):
...
+            partner = repo.create_publisher_partner(
+                publisher_domain="news.example",
+                display_name="News",
+                supported_channels=["display"],
+            )
+            session.flush()
+            updated = repo.update_publisher_partner(partner.id, supported_channels=["ctv", "olv"])
+
+        assert updated is not None
+        assert updated.supported_channels == ["ctv", "olv"]
```
KonstantinMirin, line 134: These assertions run after the `with _RepoEnv()` block on a detached instance. The repository flushes without committing (`TenantConfigUoW` owns the commit) and `IntegrationEnv.__exit__` discards the transaction, so this reads back the value `setattr` put in the instance's `__dict__` — it holds whether or not the row reaches the database. Drive the write through `TenantConfigUoW` and re-read through a fresh session, the way `test_publisher_partner_channel_write.py::_stored_channels` already does.

tests/unit/test_channel_normalization.py
```diff
@@ -0,0 +1,46 @@
...
+        assert normalize_channel_strings(["video", "audio"]) == ["olv", "streaming_audio"]
+
+    def test_deduplicates_and_sorts(self):
+        assert normalize_channel_strings(["ctv", "display", "ctv"]) == ["ctv", "display"]
+
+    def test_sponsored_intelligence_is_canonical(self):
+        assert normalize_channel_strings(["sponsored_intelligence"]) == ["sponsored_intelligence"]
+
+    def test_skips_unknown_values(self, caplog):
+        import logging
```
KonstantinMirin, line 23: This pins a log-and-skip policy no shipped path can execute: `normalize_channel_strings`'s only production caller pre-rejects unknowns, so the test reads as "unknown channels are tolerated somewhere" while every live path rejects them. Delete it along with the branch when the two functions fold into one seam. Separately, `canonical_media_channel_values()` — the sole feed for the new picker — has no test at all; deleting `media_channels=canonical_media_channel_values()` from `tenants.py:424` leaves the suite green and the picker empty. Add a rendering assertion (see `tests/integration/test_admin_ui_pages.py::test_settings_page_renders`) that a checkbox exists for every `MediaChannel` value, derived from the enum.

static/js/tenant_settings.js
```diff
@@ -2241,47 +2286,53 @@ document.addEventListener('click', function(e) {
...
-            display_name: displayName || domain
-        })
+        body: JSON.stringify(body)
     })
-    .then(response => response.json())
-    .then(data => {
+    .then(response => response.json().then(data => ({ ok: response.ok, status: response.status, data })))
+    .then(({ ok, data }) => {
         submitBtn.disabled = false;
         submitBtn.textContent = 'Add Publisher';
```
KonstantinMirin, line 2326: This forces the create-mode literal after every save, and `:2342` does the same in the `.catch`, while `showEditPublisherModal` never sets the label at all — so the button reads "Add Publisher" through an entire edit. Every other mode-dependent value in this handler is chosen from `isEdit`. Add the modal mode state the feature needs: one `setPublisherModalMode(mode, partner)` owning title, submit label, readOnly, URL and method, with a restore path that restores the mode's label. While there, replace the `window._publisherPartnersById` lookup (the only `window._` global in `static/js/`) with `data-partner-id` on the row, the pattern `static/js/targeting_widget.js` uses.

PR #1860, opened 2026-08-04
fix: outbound webhook SSRF gate silently rejected the e2e capture server

KonstantinMirin:
## Summary

`main`'s e2e suite has 6 tests failing (`test_a2a_webhook_payload_types.py`, `test_delivery_webhooks_e2e.py`) -- confirmed pre-existing, reproduces on a clean checkout with no other changes. Found while establishing a full-suite baseline for an unrelated PR.

Root cause: #1697's `ADCP_TESTING` SSRF carve-out only recognized literal `localhost`/`127.0.0.1`/`loopback` in the *error message text*, never the Docker Compose service-name address (`ADCP_WEBHOOK_HOST=tests`) this project's own e2e webhook harness actually uses -- and that env var was only ever set on the `tests` runner container, never on `adcp-server`, the process that performs the send-time SSRF check.

Every outbound webhook delivery attempt in e2e was silently SSRF-rejected, always, even with `ADCP_TESTING=true`, because `tests` DNS-resolves to a private compose-network IP (`172.x`) and nothing in the bypass logic recognized it as trusted.

## Fix (two gaps, both required)

1. **`src/core/webhook_validator.py`** -- replaced the fragile error-text string-matching (`"localhost" in error.lower()`) with `WebhookURLValidator._is_trusted_test_host()`, a hostname-based allowlist: literal `localhost`/`127.0.0.1`, or the exact value of `ADCP_WEBHOOK_HOST`. This is intentionally narrow -- arbitrary private IPs (e.g. `192.168.1.1`) remain blocked even under `ADCP_TESTING`, matching the existing `test_validate_for_testing_blocks_private_networks` assertion.
2. **`docker-compose.e2e.yml`** -- added `ADCP_WEBHOOK_HOST: tests` to `adcp-server`'s environment block (it was already set on the `tests` runner service, just never propagated to the server that actually needs it).

3 new regression tests in `tests/unit/test_webhook_security.py` cover the trusted-host allowlist directly.

## Verification

Root-caused via live `py-spy`/`docker logs -f` tracing against the real e2e stack, not just unit-level logic -- confirmed the exact SSRF rejection log line before the fix (`Protocol webhook URL failed SSRF validation ... blocked IP range 172.16.0.0/12`), and its disappearance after.

- `tests/unit/test_webhook_security.py`: 64/64 passing (61 existing + 3 new)
- `ruff format` / `ruff check` / `mypy`: clean on all touched files
- Full e2e suite (`tests/e2e/`): 91 passed, 28 skipped, 4 xfailed, 2 failed (unrelated -- pre-existing schema-validation gap, tracked separately, not touched by this PR)
- Targeted: `test_a2a_webhook_payload_types.py` + `test_delivery_webhooks_e2e.py` -- all 6 previously-failing tests now pass

Marked as draft pending final review pass.

### DIFF
--- a/docker-compose.e2e.yml
+++ b/docker-compose.e2e.yml
@@ -58,6 +58,16 @@ services:
       SUPER_ADMIN_EMAILS: ${SUPER_ADMIN_EMAILS:-}
       ADCP_TESTING: ${ADCP_TESTING:-true}
       ADCP_AUTH_TEST_MODE: ${ADCP_AUTH_TEST_MODE:-true}
+      # Must match the `tests` service's ADCP_WEBHOOK_HOST below: this is the
+      # SERVER's own trusted-test-host allowlist entry for outbound webhook SSRF
+      # validation (WebhookURLValidator._is_trusted_test_host), not just the
+      # runner's callback-URL construction. Without it here, every webhook the
+      # server tries to send to the e2e capture server (http://tests/webhook,
+      # which DNS-resolves to a private compose-network IP) is rejected by the
+      # SSRF gate even under ADCP_TESTING=true, because the gate only trusts
+      # literal localhost/127.0.0.1 plus whatever ADCP_WEBHOOK_HOST names --
+      # and this service never saw that name.
+      ADCP_WEBHOOK_HOST: tests
       # The SERVER must know about the pinned in-network creative agent: sync/
       # create flows resolve format specs server-side, and references carry the
       # PUBLIC canonical agent_url — without this, the registry's connection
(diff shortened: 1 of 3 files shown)


KonstantinMirin:
Closing — this is a narrow patch on code that's already scheduled for deletion. `feat/secure-outbound-fetch` (in progress) is a comprehensive rewrite of the entire outbound-webhook security path: it deletes the hand-rolled SSRF validator this PR touches (`ab2664fb2`), drops the localhost-rewrite logic this PR reads around (`70717da14`), and migrates `ProtocolWebhookService` onto a single unified egress seam that handles SSRF, redirects, retries, and (via the companion `feat/rfc9421-request-signing` branch) request signing in one place instead of scattered validator calls.

The underlying bug this PR reported (e2e webhook delivery silently SSRF-rejected because the ADCP_TESTING carve-out never recognized the Docker Compose service-name webhook host) is real and still worth fixing on `main` if it isn't already covered by the in-progress branch -- but the fix belongs there, not as a standalone patch to code being replaced.

PR #1862, opened 2026-08-04
fix(a2a): derive A2A skill roster from production instead of hand-copied test literals

### LINKED ISSUE #1940: A2A transport-contract tests hand-copy a stale skill roster: 3 shipping skills have no boundary/auth coverage
## Problem

`tests/unit/test_a2a_transport_contract.py` — the file its own docstring calls *"the regression gate between every Phase 2 step"* — hand-copies the A2A skill roster as a literal instead of importing it from production. The copy went stale: it lists **13** skills against production's **18**.

The five it misses:

| skill | status |
|---|---|
| `list_accounts`, `sync_accounts`, `get_media_buys` | **implemented, shipping** |
| `create_creative`, `assign_creative` | unimplemented stubs (`raise UnsupportedOperationError`) |

Three implemented, shipping skills therefore have **no boundary or auth-contract coverage at all** in the designated regression gate.

The same file also hand-copies `DISCOVERY_SKILLS` (4 entries vs production's 5, missing `list_accounts`). Because `AUTH_REQUIRED_SKILLS = ALL_SKILLS - DISCOVERY_SKILLS` and `list_accounts` appears in neither literal, **whether that endpoint requires authentication is currently ungraded**.

A third instance of the same defect lives in `tests/integration/test_a2a_skill_invocation.py::test_skill_handler_mapping`, whose hand-typed allowlist carried a phantom `get_creative_delivery` that has never been a real skill.

## Root cause

The roster has no single source of truth. Skills were declared in three independently hand-edited places — the dispatch map (a local dict inside `_handle_explicit_skill`, not importable at module scope), the `AgentCard`, and each test's own literal — with nothing cross-checking them. A local dict is precisely why tests copied a snapshot rather than importing.

## Fix

Extract the dispatch map to a module-level `SKILL_HANDLER_NAMES` (skill name -> handler method name, resolved via `getattr` at dispatch time, preserving identical behavior) and derive the test rosters from it.

## Related defect surfaced by the fix

`test_agent_card_skills_match_dispatch` asserted its containment backwards — roster ⊆ advertised — which structurally requires every dispatchable skill to be advertised on the public agent card, including unimplemented stubs. The correct invariant is the reverse: advertised ⊆ roster, since a card entry with no handler is the real defect while the roster legitimately contains skills deliberately kept off the card.

## Out of scope

The agent card separately advertises 5 skills that are unimplemented stubs and have no AdCP schema at any pinned version. That is tracked separately and is not fixed here.

KonstantinMirin:
Closes #1940

## Problem

`tests/unit/test_a2a_transport_contract.py` — the file its own docstring calls *"the regression gate between every Phase 2 step"* — hand-copied the A2A skill roster instead of importing it. The copy had gone stale: **13 skills listed against production's 18.**

| missed skill | status |
|---|---|
| `list_accounts`, `sync_accounts`, `get_media_buys` | **implemented, shipping** |
| `create_creative`, `assign_creative` | unimplemented stubs |

So three shipping skills had **no boundary or auth-contract coverage at all** in the designated regression gate.

The same file hand-copied `DISCOVERY_SKILLS` (4 vs production's 5, missing `list_accounts`). Since `AUTH_REQUIRED_SKILLS = ALL_SKILLS - DISCOVERY_SKILLS` and `list_accounts` was in neither literal, **whether that endpoint requires authentication was ungraded entirely.**

A third instance sat in `tests/integration/test_a2a_skill_invocation.py::test_skill_handler_mapping`, whose hand-typed allowlist carried a phantom `get_creative_delivery` — never a real skill.

## Root cause

A skill was declared in three independently hand-edited places — the dispatch map, the `AgentCard`, and each test's own literal — with nothing cross-checking them. The dispatch map was a local dict inside `_handle_explicit_skill`, not importable at module scope, which is exactly why tests copied a snapshot instead of importing one.

## Change

Extract the dispatch map to a module-level `SKILL_HANDLER_NAMES` (skill name -> handler method name, resolved via `getattr` at dispatch time — identical dynamic-dispatch behavior), and derive the test rosters from it. Adds 15 test executions across the 5 previously-missing skills, 9 of them on real shipping skills.

## Two defects this surfaced

**1. The card advertised vaporware.** An earlier revision of this branch added `AgentSkill` entries for `create_creative`/`assign_creative`, described as "(AdCP spec)". That was wrong: both are `raise UnsupportedOperationError("not yet implemented")` stubs, and neither is an AdCP task — there is no `create-creative`/`assign-creative` schema at any pinned version. Advertising a skill promises a buyer can call it; these error on every call. Caught by #1868's `test_all_adcp_skills_have_schemas` when this branch merged with main — neither side failed alone. Reverted here: both stay in `SKILL_HANDLER_NAMES` so dispatch returns a typed *"not implemented"* rather than *"unknown skill"*, but they are off the card. Advertised 18 -> 16.

This is why the original framing of this PR (and issue #1863, now closed) was wrong: the missing card entries were not a discoverability bug to fix, they were correct.

**2. The card assertion ran backwards.** `test_agent_card_skills_match_dispatch` asserted roster ⊆ advertised, which structurally requires every dispatchable skill to be advertised — including stubs. That is the mechanism that put those two on the card. Flipped to advertised ⊆ roster: a card entry with no handler is the real defect, while the roster legitimately contains skills deliberately kept off the card.

No allowlist grew; `_KNOWN_MISSING_SCHEMA_SKILLS` is untouched.

## Testing

Full suite on the CI box: unit 5768p, integration 2250p, bdd_inprocess 1747p, bdd_e2e 391p, admin 86p, e2e 107p, ui 5p.

The 6 remaining e2e failures are all webhook-delivery tests that fail identically on `origin/main` — introduced by d912a7ee9 (#1697, this branch's merge base) and root-caused separately in #1860. Cross-run comparison shows no suite lost a passing test and nothing regressed to xfail or skip.

## Not fixed here

- The agent card still advertises 3 other unimplemented stubs (`approve_creative`, `get_media_buy_status`, `optimize_media_buy`), all currently allowlisted. Same defect class, tracked separately.
- `SKILL_HANDLER_NAMES` is adopted at 3 of 5 eligible sites; two e2e subset smoke tests still hand-copy 4-of-18 literals.


tests/unit/test_a2a_transport_contract.py
```diff
@@ -505,13 +495,27 @@ class TestAgentCardContract:
...
+        create_creative/assign_creative got advertised in the first place.
+        """
         response = client.get("/.well-known/agent-card.json")
         card = response.json()
         advertised_skills = {s["name"] for s in card.get("skills", [])}
 
-        for skill in ALL_SKILLS:
-            assert skill in advertised_skills, f"Skill '{skill}' in dispatch map but not advertised in agent card"
+        for skill in advertised_skills:
+            assert skill in ALL_SKILLS, f"Skill '{skill}' advertised in agent card but not in the dispatch map"
```
ChrisHuie, line 518: **BLOCKER** — the flip from `roster ⊆ advertised` to `advertised ⊆ roster` leaves the other direction ungraded, and the direction it drops is the one that catches a shipping skill falling off the card.

Measured on `b19dd1a8e`: deleting the `get_products` `AgentSkill` entry from `create_agent_card()` leaves this entire file at **76 passed**. The pre-change assertion reddens on that same mutation. 12 of the 16 advertised skills are now pinned by nothing; the other 4 only by `tests/e2e/test_a2a_endpoints_working.py:241`, itself a 4-literal hand copy.

The docstring is right that `roster ⊆ advertised` would force the stubs onto the card. Encoding the exception in production gives you both directions instead of trading one for the other, and it moves the "deliberately not advertised" policy out of prose (stated three times — `src/a2a_server/adcp_a2a_server.py:2316`, and `:39` / `:497` in this file) into a data structure that a test can grade.

```diff
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -223,6 +223,18 @@ SKILL_HANDLER_NAMES: dict[str, str] = {
     # Note: legacy get_pricing/get_targeting removed - use get_products and get_adcp_capabilities instead
 }
 
+# Dispatchable but deliberately NOT advertised on the agent card. Advertising a
+# skill promises a buyer can call it; these raise UnsupportedOperationError on
+# every call and have no AdCP schema at any pinned version. They stay in
+# SKILL_HANDLER_NAMES so dispatch returns a typed "not implemented" rather than
+# "unknown skill".
+UNADVERTISED_SKILLS: frozenset[str] = frozenset(
+    {
+        "create_creative",
+        "assign_creative",
+    }
+)
+
```

```diff
--- a/tests/unit/test_a2a_transport_contract.py
+++ b/tests/unit/test_a2a_transport_contract.py
@@ -20,7 +20,7 @@
-from src.a2a_server.adcp_a2a_server import SKILL_HANDLER_NAMES
+from src.a2a_server.adcp_a2a_server import SKILL_HANDLER_NAMES, UNADVERTISED_SKILLS
@@ -498,15 +498,13 @@ class TestAgentCardContract:
     def test_agent_card_skills_match_dispatch(self, client):
-        """Every ADVERTISED skill is dispatchable — advertised is a subset of the roster.
-        ...
-        """
+        """The advertised set is EXACTLY the roster minus the deliberately-unadvertised skills.
+
+        Equality, not containment. Both directions are defects: a card entry with no
+        handler points buyers at an endpoint that cannot dispatch, and a shipping skill
+        missing from the card is undiscoverable. Asserting either containment alone
+        leaves the other direction ungraded -- which is how a stale roster survives.
+
+        The exception set lives in production beside SKILL_HANDLER_NAMES rather than
+        being hand-copied here, so adding a skill forces a deliberate choice: advertise
+        it, or name it in UNADVERTISED_SKILLS.
+        """
         response = client.get("/.well-known/agent-card.json")
         card = response.json()
         advertised_skills = {s["name"] for s in card.get("skills", [])}
 
-        for skill in advertised_skills:
-            assert skill in ALL_SKILLS, f"Skill '{skill}' advertised in agent card but not in the dispatch map"
+        assert advertised_skills == set(ALL_SKILLS) - set(UNADVERTISED_SKILLS)
```

What I ran against this patch on `b19dd1a8e`:

| | result |
|---|---|
| unmutated | 76 passed; full unit+harness `5768 passed, 10 skipped, 26 xfailed` — identical counts to the unpatched control |
| `get_products` dropped from the card | **fails** — the case that is silent today |
| card entry with no handler | **fails** — the direction this PR kept |
| new skill in `SKILL_HANDLER_NAMES`, never added to the card | **fails** — the drift class this PR exists to close |
| a name quietly removed from `UNADVERTISED_SKILLS` | **fails** — so the exception set is graded too, not a rubber stamp |

`ruff format --check`, `ruff check` and `mypy` are clean on both files.

Not run with this patch: integration, e2e, bdd, admin. The card's contents are byte-identical under it, so the card-reading tests in those suites should be unaffected — but I did not measure that.

ChrisHuie (commented):
Reviewed at `b19dd1a8e` (CI 30/30 on this head, pin `adcp 6.6.0` → spec 3.1.1).

Extracting the roster to `SKILL_HANDLER_NAMES` is the right seam, and the central spec claim checks out — `create-creative` and `assign-creative` exist at no pinned version, confirmed against the SDK-bundled index at 3.1.1 with `sync-creatives` / `get-products` as positive controls, and re-adding them to the card does trip `test_all_adcp_skills_have_schemas`. The auth contract now covers 18 skills instead of 13.

One blocker below, anchored inline, with a patch I mutation-tested.

Separately, and verified the same way but not written up here: the auth-boundary assertion has the same shape as the card one. `DISCOVERY_SKILLS` is now read from production on both sides of `AUTH_REQUIRED_SKILLS`, so widening production's public set does not fail a test — it deletes the test and substitutes its negation. Adding `approve_creative` to `DISCOVERY_SKILLS` leaves the full unit+harness suite at `5768 passed`, byte-identical to the control; 10 of the 13 auth-required skills behave that way, while 3 retain catchers elsewhere. The same "pin the exception, derive the inventory" shape fixes it. I can write that one up with a patch as well.

PR #1838, opened 2026-08-03
fix(tests): pin e2e AdCP schema validation to the installed SDK's 3.1.1 schemas

KonstantinMirin:
Fixes the E2E CI breakage of Aug 1–3, then follows the thread all the way down: with schema validation actually grading something, it surfaced (and this PR fixes) a set of real production bugs the old validator had been silently passing.

## Part 1 — the outage (original scope)

`tests/e2e/adcp_schema_validator.py` validated against **live** schemas fetched from `adcontextprotocol.org` — a registry that serves "latest" and drifts ahead of the repo's pin. On 2026-08-01 upstream adcp PR #6133 ("canonicalize schema URIs") changed the live schemas' `$ref` URIs from relative to absolute. No contract change — but the validator's cache-keyed sync resolver missed on the new URI shape and substituted a `{"additionalProperties": false}` fallback that rejects **everything**, failing two e2e tests on every branch overnight.

Fix: schemas load from the installed `adcp` SDK's schema tree — pinned by the same `pyproject.toml` pin the CI guard (`tests/unit/test_adcp_spec_version.py`) protects, offline and deterministic (graded by a socket assertion, not claimed). An unresolvable `$ref` raises naming the ref instead of degrading to a reject-everything schema.

## Part 2 — review round 1 (ChrisHuie)

Applied in full: `SchemaError` passthrough so resolution failures don't read as contract violations; parametrized `_normalize_ref` contract tests including the outage's absolute-URL form; resolve-and-contain traversal check; pin test asserts the validator's own `schema_root` + the guard's `EXPECTED_SPEC_VERSION`; pin literals replaced with pointers to `docs/adcp-spec-version.md`; dead decorators, `.gitignore` cache entries, and `/schemas/v1/` literals removed.

## Part 3 — the review's follow-up findings, executed here (not deferred)

Every remaining review finding was triaged and fixed on this branch:

[...]

Issue #1706, opened 2026-07-23
fix: REST has no catch-all envelope for untyped top-level failures (cross-transport parity)


numarasSigmaSoftware:
Follow-up from #1547 re-review. The "top-level / pre-handler failure surfaces the AdCP two-layer error envelope on the wire (not a transport-native error)" contract is currently graded — and on REST, implemented — only partially across transports.

## State per transport
- **A2A** — covered. `on_message_send`'s outer handler returns a failed Task carrying the two-layer envelope; graded by `@T-UC-002-ext-nl-unsupported` (A2A-only, correctly, since NL dispatch is an A2A entry point).
- **MCP** — envelope is carried. `_translate_to_tool_error` (`src/core/tool_error_logging.py`) runs `normalize_to_adcp_error()` on any exception and builds the two-layer envelope, so an untyped top-level failure still surfaces `AdCPToolError` with the envelope. (Message-scrub for untyped exceptions is the separate concern in #1587.)
- **REST** — **gap.** `src/app.py` registers exception handlers for `AdCPError`, `ValueError`, `RequestValidationError`, `PermissionError`, and `ToolError`, but there is **no** `@app.exception_handler(Exception)` catch-all. An untyped top-level exception (e.g. `RuntimeError`, a DB/OS error) escaping a REST endpoint gets FastAPI's default bare `500` with no AdCP envelope.

## Proposed fix
1. Add a catch-all `@app.exception_handler(Exception)` in `src/app.py` that routes through `safe_adcp_error(exc)` (leak-safe: untyped → generic `SERVICE_UNAVAILABLE` envelope, no `str(exc)`) and returns `_envelope_response(...)`, mirroring the MCP boundary. Framework exceptions with their own handlers (`HTTPException`, `RequestValidationError`) are unaffected — a generic `Exception` handler only catches what no more-specific handler claims.
2. Add per-transport integration tests grading "untyped top-level failure → two-layer envelope" on REST and MCP (A2A already covered), each asserting the same envelope values in that transport's shape.

## Scope note
Split out of #1547 deliberately: #1547's scope is A2A error routing, and a REST catch-all changes every REST endpoint's untyped-`500` shape — a cross-transport parity change that deserves its own review. Adjacent to #1587 (which scrubs the untyped-error *message*); the catch-all here should use the same `safe_adcp_error` sink, so the two are natural to land together.

KonstantinMirin:
[...]

Issue #1623, opened 2026-07-14
Sweep MCP structured_content serialization to omit nulls (model_dump) across all tool wrappers


numarasSigmaSoftware:
Follow-up spun off from #1575 / #1570.

MCP tool wrappers that pass the raw Pydantic model to `structured_content` leak explicit `null`s for unset optionals, because fastmcp serializes the model via `pydantic_core.to_jsonable_python`, bypassing `AdCPBaseModel`'s `exclude_none` default. A2A/REST omit those keys, so MCP diverges (confirmed on `get_media_buys`: `context: null`, `errors: null`).

`get_products` and (as of #1575) `get_media_buy_delivery` are fixed via `structured_content=response.model_dump(mode="json")`. The remaining wrappers still pass the raw model, e.g.:

- `src/core/tools/capabilities.py:319`
- `src/core/tools/accounts.py:206,725`
- `src/core/tools/creative_formats.py:580`
- `src/core/tools/media_buy_list.py:330`
- `src/core/tools/media_buy_update.py:1422`
- `src/core/tools/media_buy_create.py:4375`
- `src/core/tools/properties.py:223`
- `src/core/tools/signals.py:210,306`

**Fix:** sweep all MCP wrappers to `model_dump(mode="json")`, or (preferred, DRY) add a shared MCP-return helper so the serialization choice lives in one place and can't split again. Add a cross-transport wire-omission test per wrapper carrying optional/None fields (see `tests/integration/test_delivery_poll_behavioral.py::TestPollOmitsWebhookOnlyFieldsOnEveryTransport`).

PR #1868, opened 2026-08-05
fix: pin AdCP schema validation to the installed SDK and fix the wire-serialization bugs it surfaced

### LINKED ISSUE #1710: MCP structured_content emits schema-invalid null for absent optional response fields
On MCP, `structured_content` serialization emits `"supported_pricing_models": null` when the field degrades to `None`, but the 3.1.1 response schema types it `array` (non-nullable, `minItems: 1`), so a strict validator rejects it. REST/A2A correctly OMIT the field via `model_dump(mode="json")` exclude-none.

Pre-existing and repo-wide: `main` already emits `null` for this field and ~12 sibling optional fields. #1677 REDUCES it (the common path now emits a real array), so it is not a regression there — filed as the separate follow-up. Direction: make the MCP structured_content path drop None-valued optional fields (match the REST/A2A omission) so absent optionals are absent, not null.

### LINKED ISSUE #1843: E2E AdCP schema validation validates against the live registry, breaking CI when upstream drifts ahead of the pin
## Summary

E2E AdCP schema validation resolves against the **live** schema registry
(`adcontextprotocol.org/schemas/v1`), which serves "latest" and drifts ahead of the
repo's pinned spec version (`adcp==6.6.0` → AdCP 3.1.1). A benign upstream URI change
therefore breaks E2E CI on **every branch**, including PRs with no schema-adjacent
changes, and does so with an opaque failure mode.

This surfaced over the weekend of Aug 1–3, 2026 as the blocker that stopped work on
#1822 (BR-UC-001 `get_products` BDD wiring). It was meant to be a quick unblock but is
a focused fix of its own.

## Root cause

`tests/e2e/adcp_schema_validator.py` validates payloads against schemas fetched live
from `adcontextprotocol.org/schemas/v1`. On 2026-08-01 upstream adcp PR #6133
("canonicalize schema URIs") rewrote the live schemas' `$ref` URIs from relative
(`/schemas/3.1.1/core/version-envelope.json`) to absolute
(`https://adcontextprotocol.org/schemas/latest/core/version-envelope.json`). This is
**not** a contract change — the property sets are identical to pinned 3.1.1 — but the
validator's sync ref resolver derives its cache key from the ref string, missed on the
new URI shape, and substituted its "strict fallback"
(`{"additionalProperties": false, "properties": {}}`) — a schema that rejects
everything. Two E2E tests went red overnight on branches that changed nothing:

- `tests/e2e/test_schema_validation_standalone.py::test_get_products_request_validation`
- `tests/e2e/test_a2a_protocol_compliance.py::test_get_media_buy_delivery_request_schema`

The failure was confusing precisely because the validator degraded silently instead of
naming the unresolved ref.

## Why it matters

E2E schema validation grades production payloads against a moving target the repo does
not control. Any upstream URI/shape change — even one with no contract impact — reds
every branch, and the reject-everything fallback makes the outage read like a contract
violation rather than a resolver miss.

## Required outcome

- E2E schema validation resolves against the **installed `adcp` SDK's pinned schema
  tree** (the spec version production is built against), not the live registry —
  deterministic and offline, with no network in CI.
- The validated version moves **only** with the `pyproject.toml` `adcp` pin (guarded by
  `tests/unit/test_adcp_spec_version.py`), never with the upstream site.
- An unresolvable `$ref` raises a **named** `SchemaError`, never a silent degrade to a
  reject-everything schema.
- All historical ref forms callers use (`/schemas/v1/…`, absolute URLs, index-relative
  paths) normalize to the pinned tree.
- Offlineness is **graded**, not just claimed (a socket assertion in the pin test).

## Out of scope (follow-ups)

- The repo has a second, independent pinned-schema source
  (`tests/fixtures/adcp_schemas_pinned/`, own `PINNED_SHA`) used by the unit alignment
  suite. Unifying the two is a separate follow-up; the coexistence is documented in
  `docs/adcp-spec-version.md`.

Fixed by #1838.

### LINKED ISSUE #1881: error-code.json fixture $id pin diverges from _refresh.py's handling
## Deferred from PR #1721 review round 2

Must be addressed as part of **PR #1868** (`fix/adcp-schema-pin`) before merge — that branch owns the schema-fixture-pin territory and already touches `_refresh.py` plus ~200 sibling fixture files (including `enums/billing-party.json`, `error-code.json`'s unversioned sibling), but does not yet touch `enums/error-code.json` itself.

## Finding

The `tests/fixtures/adcp_schemas_pinned/enums/error-code.json` fixture's `$id` is `/schemas/3.1.1/...` (versioned), while its sibling `billing-party.json`'s `$id` is unversioned. `_refresh.py:31,61` still lists `error-code.json` under `PINNED_SHA=04f59d2d5...` without accounting for the versioned `$id` — running the documented refresh procedure regresses the fixture, caught only after the fact by the SHA-pin test, not by the refresh tooling itself.

## Architecturally correct fix

Requires a decision, not just a patch: determine which `$id` convention is canonical across the fixture set (match the unversioned sibling pattern, or make versioned `$id` the standard and update the other fixtures to match) — then make `_refresh.py` handle that convention explicitly so running the documented refresh no longer silently regresses the file. Document the decision in `_refresh.py` itself.

## Acceptance

`error-code.json`'s `$id` convention matches the decided standard (documented in `_refresh.py`); running the documented refresh procedure no longer changes the `$id` unexpectedly; the SHA-pin test continues to catch any future divergence.

Full triage evidence: `.claude/notes/pr1721-review-triage-and-bdd-rootcause.md` on `feature/spec-gaps-1210`.

### LINKED ISSUE #1888: get_products property-list resolution relabels implementation bugs as AdCPValidationError
`src/core/tools/products.py:409-413`:

```python
except AdCPError:
    raise
except Exception as e:
    logger.error(f"Property list resolution failed: {e}")
    raise AdCPValidationError(f"Failed to resolve property list: {e}", recovery="transient") from e
```

After correctly re-raising `AdCPError`, the blanket arm wraps **everything else** as `AdCPValidationError`. An `AttributeError` or `TypeError` inside `resolve_property_list()` / `filter_products_by_property_list()` therefore reaches the buyer as `VALIDATION_ERROR` — "your request is malformed" — when it is a server-side bug. The buyer edits a request that was never wrong.

Same shape as the `tests/helpers/adcp_schema_validator.py` boundary narrowed in #1868, where a blanket `except Exception` relabelled validator bugs as `SchemaValidationError`. That fix's rule: name the failure types that genuinely mean "the operation could not be performed", and let everything else propagate to the transport boundary's internal-error handler.

Allowlisted (shrink-only) in `tests/unit/test_architecture_no_broad_except_validation_relabel.py`, the ratcheting guard added by #1868.

### Acceptance
- Injecting a non-`AdCPError` implementation exception (e.g. `AttributeError`) into the property-list resolution path does **not** surface as `AdCPValidationError` on the wire.
- Assertion is on the wire envelope per `tests/CLAUDE.md` § Error Verification Policy (`assert_envelope_shape(result.wire_error_envelope, ..., recovery=...)`), not a reconstructed exception.
- The entry is removed from the guard's allowlist.

### LINKED ISSUE #1889: PolicyService currency-code validation reports environment failures as an invalid currency_code
`src/services/policy_service.py:135-146`:

```python
name = babel_numbers.get_currency_name(currency_code, locale="en")
if name == currency_code:
    raise ValidationError({"currency_code": f"Invalid currency code: {currency_code}. ..."})
except Exception as e:
    raise ValidationError({"currency_code": f"Error validating currency code: {e}"}) from e
```

The blanket arm wraps the `babel` call, so a missing/corrupt babel locale database — an environment problem — is reported to the admin as "your currency code is invalid". They will keep retrying valid ISO 4217 codes.

Same shape as the `tests/helpers/adcp_schema_validator.py` boundary narrowed in #1868, and the sibling production instance in #1888. Fix: catch the named babel failure types; let anything else propagate.

Allowlisted (shrink-only) in `tests/unit/test_architecture_no_broad_except_validation_relabel.py`, the ratcheting guard added by #1868.

### Acceptance
- A non-babel implementation exception raised inside the `try` propagates unwrapped.
- An actual unknown currency code still raises `ValidationError` with the existing message.
- The entry is removed from the guard's allowlist.

KonstantinMirin:
Supersedes #1838. Closes #1843. Closes #1710. Closes #1881. Closes #1888.
Closes #1889. Advances #1623 (its sweep-all-wrappers clause is done; its
per-wrapper live wire-test clause isn't — see below).

## The outage this fixes (#1843)

`tests/e2e/adcp_schema_validator.py` validated AdCP responses against schemas
fetched **live** from `adcontextprotocol.org`, a registry that serves
"latest" and drifts ahead of the repo's pinned SDK version. On 2026-08-01
upstream adcp PR #6133 ("canonicalize schema URIs") changed the live
schemas' `$ref` URIs from relative to absolute — no contract change, but the
validator's cache-keyed resolver missed on the new URI shape and substituted
a `{"additionalProperties": false}` fallback that rejects everything,
failing e2e on every branch overnight.

**Fix:** schemas load from the installed `adcp` SDK's own schema tree
(`tests/helpers/pinned_schema.py`), pinned by the same `pyproject.toml` pin
`tests/unit/test_adcp_spec_version.py` already guards — offline,
deterministic, and it can't drift out from under the SDK version production
actually ships against. An unresolvable `$ref` raises naming the ref instead
of degrading into a schema that lies about the result. `pinned_schema.py` is
the single resolver every schema-shape consumer in the repo reads through —
one ref normalizer, one containment check, one error type — replacing a
vendored 244-file fixture that had already drifted a spec-minor behind
(shrunk to 1 file, the error-code `suggestion`-text exception — see
Deferred) and two other ref-resolution mechanisms that disagreed with it and
each other.

Tightening validation also made several previously-decorative checks able to
fail for the first time: 48 of the schema index's task names were
unreachable and silently recorded as "validated"; two independent, divergent
skill→task maps existed (now one, derived from `create_agent_card()` so a
skill added there is graded the same day); `test_a2a_adcp_compliance.py` was
100%-skipped (its fixture probed the wrong endpoint).

## Wire-serialization bugs the tightened validation surfaced

**`Product`/`Creative`/`SyncCreatives` null-leaks.** Optional fields were
serialized as `null` instead of omitted (`Creative.model_dump`,
`strip_none_deep`'s list branch), and `SyncCreativesResponse` was missing
the spec-required `status` field outright.
`Product.reporting_capabilities` had the inverse problem: it's
schema-*required*, so omitting it when unset just traded one violation for
another. It's now a properly typed field with a validated `default_factory`
— the model can't be constructed with a missing or invalid value, so no
serializer-side backfill exists to forget, and each product gets its own
instance rather than sharing the constant's nested lists.

**MCP `structured_content` bypassed serialization on most tool wrappers.**
FastMCP serializes whatever a wrapper hands `structured_content` directly —
wrappers passing the raw pydantic model bypassed `model_dump()`'s
`exclude_none` and any custom nested serialization. 13 of 16 registered MCP
tools now route through one shared `mcp_result()` helper, guarded so a new
wrapper can't bypass it again. **Not yet swept:** `list_tasks`/`get_task`/
`complete_task` (`src/core/tools/task_management.py`) build hand-written
response dicts outside both the helper and the guard — `get_task` hand-
writes `"updated_at": None` by the same null-leak mechanism. Left for a
follow-up (see Deferred); this is the remaining half of #1623's
sweep-all-wrappers clause, which is why this PR advances rather than closes
it. #1623's *other* clause — a live cross-transport wire test per wrapper —
is in place for the wrappers whose null-omission fix landed in this PR (via
a new `assert_wire_omits_unset()` harness helper); most of the remaining 12
registered wrappers (`get_media_buys` among them) aren't wire-graded yet.

**A2A `list_creatives` serialized `format_id` as a bare string** instead of
the spec's `{agent_url, id}` object. Root cause: `_reconstruct_response_object`
fed an already-serialized A2A artifact back through `Model(**data)` to
regenerate a text string that was already sitting in
`artifact_data["message"]` — the only outbound dict-to-model reconstruction
anywhere in `src/` — and a `mode="before"` validator's in-place dict
mutation corrupted the shared payload before that reconstruction re-read it.
Deleted the reconstruction; the A2A `TextPart` is now built directly from
the stamped `message`. Separately (unrelated bug, same file): 5 pydantic
before-validators in `src/core/schemas/` mutate their input dict in place,
a live hazard for any caller holding a reference to the original — each now
copies first via a shared `copy_before_mutating()` helper.

**A2A success/message-stamping** had 3 near-duplicated call sites, one of
which silently diverged on the `success`-from-`errors` derivation
(`errors[]` populated but `success` still forced `true`). Consolidated into
one helper, graded by a live wire-level assertion rather than a mocked unit
test.

## Two production bugs unrelated to serialization, also surfaced and fixed

`get_products`'s property-list resolution and `PolicyService`'s
currency-code validation each caught their own implementation bugs internally
and relabeled them as buyer-facing validation errors. Both `except` blocks
deleted outright.

What that buys is not the same on both paths. `get_products` is reachable
over AdCP, so a crash there now reaches the buyer as an honest
`INTERNAL_ERROR`, remapped at the MCP and A2A wire boundaries to the pinned
enum's `SERVICE_UNAVAILABLE` (`transient`, HTTP 500). REST is not covered:
the FastAPI app registers handlers for `AdCPError`, `ValueError`,
`RequestValidationError`, `PermissionError` and `ToolError`, and no bare
`Exception` — so an untyped crash out of a REST route is Starlette's default
500 with no AdCP envelope. That is a real, already-filed gap — #1706.
`validate_currency_code` has no AdCP transport caller at all: the only chain
is `validate_currency_limits` → `update_policies` → a Flask admin route,
which catches the module-local `ValidationError` and returns a plain
`jsonify(...)`, never reaching error normalization. What its deletion buys is
the precise message reaching the caller instead of a double-wrapped generic
one, which is what `tests/unit/test_policy_service.py` grades.

A new shrink-only structural
guard (`test_architecture_no_broad_except_validation_relabel.py`, `set()` at
zero known violations) stops a third instance from landing silently
anywhere else in the tree. Closes #1888, #1889.

## Alignment-suite accuracy

The pydantic/pinned-schema alignment suite's protocol-envelope field
exclusion (`_PROTOCOL_ENVELOPE_FIELDS`) is now derived directly from the
pinned schema's `required` array instead of a hand-copied list of all 11
envelope property names. Only `status` is actually spec-required; the other
10 were inert (verified — there's exactly one place the set is consumed, and
none of the other 10 ever appear in a graded `required[]`). The prior
rationale comment was wrong on why `status` itself is excluded: it claimed
the field is stamped by the transport layer, but 10 of 11 registered
response models already carry a spec-correct `status` default straight from
the `adcp` library base — the real reason is two suite limitations (the
requiredness walk doesn't merge `allOf` properties; the sample generator
can't synthesize enum defaults), not a layering decision. A tripwire now
fails loudly if a future pin bump ever requires a second envelope field.
Actually grading `status` (fixing those two suite limitations, closing the
one genuine model gap at `SyncAccountsResponse`) is separate follow-on work.

`Product.model_dump()`'s `core_fields` safety-net set — and its dead
`pricing_options == []` sibling branch — is removed. Both were no-ops:
`reporting_capabilities` becoming a typed field (above) left every remaining
member already guaranteed non-null by the library base, confirmed directly
against its Pydantic field definitions rather than assumed. A new test pins
that invariant against the SDK type itself, so a future loosening is caught
rather than needing a safety net re-added.

## Test-infra fix

The integration suite's `mcp_server` fixture spawned the real server with
undrained `stdout/stderr=PIPE`; past ~64KB of logging the server blocked
forever inside a log write mid-request and the whole suite wedged at ~98%
(root-caused with py-spy: server in `anon_pipe_write`, worker parked in
asyncio select). Output now goes to temp files — no pipe exists, so the
deadlock is impossible by construction.

## Deferred (not in this PR)

- **`tests/fixtures/adcp_schemas_pinned/enums/error-code.json`** stays as
  the one remaining non-SDK-sourced schema fixture, for its `suggestion`-text
  field only: 4 codes' wording diverges between it and the SDK's own enum
  (`recovery` is byte-identical across all 64 shared codes and migrated
  everywhere else). Reconciliation is a separate, precisely-scoped track —
  #1883.
- **`task_management.py`'s 3 unswept MCP wrappers** (above) and
  **`update_performance_index`'s cited AdCP task schema not structurally
  matching what production sends** — both out of scope for a schema-pin PR.
- **`get_signals`/`activate_signal`** are not registered on any transport in
  production at all — tracked at #1353, unrelated to this PR.
- **`get_products`'s `property_list` filter is unreachable over REST** — the
  REST request body doesn't declare a field both the MCP and A2A wrappers
  accept. Same class of gap already tracked at #1596 (REST body models
  missing declared fields); folded in there rather than filed again.

## Unrelated riders

`aiohttp 3.14.3` / `cryptography 50.0.0`: 4 CVE advisories
(CVE-2026-59881/69243/69244, CVE-2026-69247) published upstream 2026-08-04
fail the audit job on every branch. Carried here as an unrelated,
already-verified one-line floor bump so this PR's security-audit check
isn't red for an unrelated reason.

`docker-compose.e2e.yml` / `run_all_tests.sh`: the e2e test container's
non-root execution now derives its uid:gid from the invoking host
(`TEST_UID`/`TEST_GID`, exported by `run_all_tests.sh`) instead of a
hardcoded value, defaulting to `0:0` for any caller that doesn't export
them — so a bare `docker compose run tests` keeps working the same as
before on hosts that never set those variables. Unrelated to schema
validation; carried here because it landed on this branch.

## Verification

Full suite (`unit`, `integration`, `bdd_inprocess`, `bdd_e2e`, `admin`,
`ui`): all green. Local full-stack `e2e`: 107 passed, 6 failed,
byte-identical to the pre-change baseline in both directions (no failure
added, none accidentally fixed) — pre-existing webhook-SSRF gap on `main`
(PR #1697's testing carve-out doesn't cover the in-network Docker Compose
topology the local full-stack harness uses), confirmed against a clean
`origin/main` baseline and out of scope here; it's being fixed independently
elsewhere. This PR's own `E2E Tests` CI job doesn't exercise that topology,
so it's unaffected and green. Every fix above is mutation-verified: revert
it, confirm the regression is observable (a test goes red or the wire
response is wrong), restore, confirm green.


KonstantinMirin:
While working on #1721, filed as #1881 since this branch owns the schema-fixture-pin territory: `error-code.json`'s `$id` diverges from `_refresh.py`'s handling. Should be addressed as part of this PR.

KonstantinMirin:
@ChrisHuie would appreciate a look on this one when you get a chance.

Heads up on scope: this started as a narrow schema-pin fix, but tightening the validator kept surfacing real production non-conformance one layer down — fix one thing, the next thing it touches turns out to need the same treatment. Despite trying to keep this descoped, it pulled in more than intended. Net result is things are in noticeably better shape than where we started, but it's a bigger diff than the title suggests, so flagging that up front rather than let it be a surprise in review.

tests/unit/test_pydantic_schema_alignment.py
```diff
@@ -85,10 +89,32 @@
 # These have defaults or are managed by the library base class — exclude from all comparisons.
 _VERSION_FIELDS: frozenset[str] = frozenset({"adcp_version", "adcp_major_version"})
 
-# Fields the pinned AdCP schema (04f59d2d5) defines but the adcp 5.7.0 Python library
-# / our local model does not yet model. These are spec-vs-library mismatches, not bugs
-# in our code. Re-derived against the pinned schemas — entries describing fields the
-# pin no longer defines were dropped. #1388 tracks the adcp 5.7 alignment.
+# Every AdCP response schema's Protocol Envelope arm (core/protocol-envelope.json,
+# shared via a top-level allOf) types these fields, and spec-requires "status" —
+# but they are populated by the PROTOCOL LAYER (_serialize_for_a2a et al) at the
```
ChrisHuie, line 94: **SHOULD-FIX** — this rationale is not true for `status`, and the precedent it cites argues the other way.

Nothing populates `status` at any transport boundary: a search for a `status` assignment across `src/routes`, `src/a2a_server`, `src/core/tools/_mcp.py` and `src/core/version_compat.py` is empty, and the `_serialize_for_a2a` named here stamps only `message` and `success` — its own docstring says so. `apply_version_compat` just returns `model_dump(mode="json")`.

The cited precedent points the opposite way too. `tests/helpers/adcp_schema_validator.py`'s docstring says `message`/`context_id` "are populated by the protocol layer, not the Pydantic response model, **but they ARE modeled and typed on the wire envelope the buyer actually receives, so they must be graded, not exempted**" — a decision in this same PR that went the other way.

This exclusion is also doing two jobs at once. Removing `"status"` from the set moves 44 passed → 12 failed: one is a genuine `SyncAccountsResponse` requiredness gap, four are synthetic-generator artifacts (a `Literal` vs a sample value), and six are a real suite limitation — `_allof_required_fields` merges allOf `required` but the property walk never merges allOf `properties`.

Smallest useful change now is the comment: say it's excluded because no boundary stamps it and the models are being migrated one at a time. Un-excluding it belongs with the response-model work, which I've kept out of this PR.

tests/unit/test_architecture_no_broad_except_validation_relabel.py
```diff
@@ -0,0 +1,211 @@
...
+    # FIXME(#1889): a babel locale-data failure is reported to the admin as
+    # "your currency_code is invalid".
+    ("src/services/policy_service.py", "validate_currency_code", "ValidationError"),
+}
+
+
+def _is_broad_handler(handler: ast.ExceptHandler) -> bool:
+    """True for ``except:``, ``except Exception:``, ``except BaseException:``.
+
+    A tuple of named types is never broad, even if one member is ``Exception``
```
ChrisHuie, line 68: **NIT** — "A tuple of named types is never broad, even if one member is `Exception`" is the one part of this that isn't right. `except (KeyError, Exception)` catches exactly what `except Exception` catches, so that form *is* broad.

The rest of the sentence is accurate and the limitation is deliberate and documented ("caught by the plain `ast.Name` arm below only when written alone"), so this is narrower than it first looks. The consequence is still real: feeding `_is_broad_handler` a tuple containing `Exception` returns `False`, so `except (Exception, KeyError): raise MyValidationError(...)` evades a shrink-only allowlist. There are 13 live tuple-with-`Exception` handlers in-tree (`src/core/tools/accounts.py:85`, `src/core/main.py:64`, `src/core/format_resolver.py:161`, `src/admin/blueprints/products.py:778,1401`, …) and **zero** currently relabel as a `*ValidationError`, so nothing is being missed today.

Widening the matcher to model tuples adds 0 violations at the current scope, so the allowlist wouldn't need to grow. If you'd rather keep the arm as-is, the fix is just the first clause of the docstring — as written it tells the next contributor the form is fine, and `:137-139`'s hint ("name the failure types … and delete the broad arm") is satisfiable with `except (Exception, ValueError)` while changing nothing.

tests/unit/test_guards_before_validator_no_mutation.py
```diff
@@ -0,0 +1,227 @@
...
+
+def test_no_unsafe_before_validator_mutation():
+    src_files = {str(p.relative_to(REPO_ROOT)): t for p in sorted(SCHEMAS_DIR.rglob("*.py")) if (t := _parse(p))}
+    violations = find_unsafe_before_validators(src_files)
+    assert not violations, (
+        '@model_validator(mode="before") mutates its input dict without a '
+        "defensive copy first -- pydantic-core hands list-item dicts to "
+        "before-validators BY REFERENCE, so this can corrupt a dict the caller "
+        "still holds (traced production bug: A2A list_creatives format_id). Add "
+        "`values = values.copy()` before any mutation. Violations:\n  " + "\n  ".join(violations)
```
ChrisHuie, line 140: **SHOULD-FIX** — this failure message instructs the exact shape the guard rejects.

It says ``Add `values = values.copy()` before any mutation``, but `test_positive_raw_copy_then_mutate` (`:174-185`) asserts a raw `.copy()` **is** a violation, and the module docstring at `:23-27` says "the abstraction must be used, not just a raw `.copy()`/`dict()` call". A contributor who trips this guard and follows its instruction gets a second failure.

Should name `copy_before_mutating(values)`.

Two related things while you're in here, both optional and neither a live defect: the guard doesn't see `del values[k]`, `values |= {...}`, nested writes (`values['a']['b'] = 1`), or aliasing (`v = values; v['x'] = 1`) — and since `copy_before_mutating` is a shallow `values.copy()`, the nested form would still corrupt the caller's dict after a compliant copy. All five current adoption sites write top-level keys only, so the shipped fix is sufficient; it's the unqualified docstring that promises more than `.copy()` delivers.

src/core/schemas/product.py
```diff
@@ -158,18 +174,37 @@ def model_dump(self, **kwargs):
...
+        # (R3-8, salesagent-1zq3.8). When actually unset, it must be OMITTED
+        # like any other optional field — it is only required via anyOf with
+        # format_options, not unconditionally.
+        #
+        # reporting_capabilities is different: the pinned schema's top-level
+        # required array lists it unconditionally, so omitting it is ALSO
+        # schema-invalid. It is not listed in core_fields (that would force
+        # in a raw None); instead it gets a real default backfilled below,
+        # after this loop, when still unset.
         core_fields = {
```
ChrisHuie, line 193: **NIT** — `core_fields` is now provably dead and can go with it, since you're already rewriting these lines.

All four remaining members are required, non-`Optional` fields on the library base (`product_id`, `name`, `description` are `str`; `delivery_type` is `DeliveryType`), so `value is not None` is always true for them and `key in core_fields` never changes the outcome of `if key in core_fields or value is not None:`. Your own added comment at `:179-181` says as much ("listing them is a no-op safety net").

Deleting the set and the `key in core_fields or` disjunct leaves behaviour identical and removes the thing a future reader has to re-derive. Worth keeping the comment's *explanation* of why `format_ids` must not be listed — that part is load-bearing.

src/services/policy_service.py
```diff
@@ -142,6 +142,9 @@ def validate_currency_code(currency_code: str) -> None:
                         "currency_code": f"Invalid currency code: {currency_code}. Please use a valid ISO 4217 currency code."
                     }
                 )
+        # FIXME(#1889): broad catch reports a babel locale-data failure (an
+        # environment problem) as "your currency_code is invalid". Allowlisted in
+        # tests/unit/test_architecture_no_broad_except_validation_relabel.py.
         except Exception as e:
```
ChrisHuie, line 148: **SHOULD-FIX** — this handler mangles the common path, not the rare one the FIXME describes.

The `raise ValidationError(...)` for a genuine unknown currency is *inside* this `try` (`:140-145`), so this arm re-wraps it. Every real invalid-currency response is already double-prefixed:

```
PolicyService.validate_currency_code("XYZ")
-> ValidationError currency_code: Error validating currency code: currency_code: Invalid currency code: XYZ. Please use a valid ISO 4217 currency code.
```

`tests/unit/test_policy_service.py:31` asserts `"Invalid currency code" in str(...)`, a substring that survives the corruption, which is why it's green. `ValidationError` isn't an `AdCPError` subclass, so this path also carries no wire envelope and no `recovery`.

One line fixes it — `except ValidationError:` / `raise` before this arm, or move the `:140` raise below the `try`. The sibling site already does exactly this (`src/core/tools/products.py:410` `except AdCPError: raise`). Worth tightening the test to an exact-message assertion at the same time so the wrapper can't come back.

The babel-locale case the FIXME names is real and separate — that half is fine to keep deferred; it's the issue's scope that should say "and the invalid-currency path is re-wrapped today".

ChrisHuie (commented):
Reviewed at `850a53540`. **0 blocker.** This is the only open branch with a green `E2E Tests`
job — every other open PR is red on it, and most are also red on `Security Audit`/`pip-audit`, which
the rider bump here clears. So the framing for this review is: land it, and don't let my list grow it.

Everything I found is either pre-existing and *surfaced* by this PR's own tightened validation, or
latent behind an unreachable writer. Nothing the PR introduces is broken on a live wire — I checked
that specifically rather than assuming it.

Six small things I'd fix before merge, five of them anchored as threads below. All are prose or
one-liners, none changes a wire shape or needs a new test fixture:

1. **SHOULD-FIX — the description makes five claims that don't survive tracing.** Worth correcting
   because the body is the grounding artifact and a wrong sentence in it tends to get copied into test
   docstrings later:
   - "**Every** MCP wrapper now passes `model_dump(mode="json")`" — 13 of 16 registered tools.
     `list_tasks`/`get_task`/`complete_task` (`src/core/tools/task_management.py`) return hand-built
     dicts, outside both the helper and the new guard; `get_task:159` hand-writes `"updated_at": None`,
     the same wire-null class by another mechanism. `list_tasks` is an SDK-defined tool.
   - "20 call sites (15 tool wrappers + 5 e2e/integration validators) … one `mcp_result()` and one
     `copy_before_mutating()` helper (validators)" — the other five are production pydantic
     before-validators in `src/core/schemas/`, not test validators; `copy_before_mutating` has zero
     callers under `tests/`, and it addresses in-place mutation, not null omission.
   - "`strip_none_deep`'s list branch didn't actually drop `None` elements" is under **"Production bugs
     this surfaced, fixed here"** — it's absent from `origin/main`; introduced by `6954486b6` and fixed
     by `ab74a635a`, both inside this PR. It never shipped.
   - "`verify_feature_error_codes.py`'s dead `RuntimeError` arm removed" — the base has no `except`
     tuple at all; added by `fa4d65396`, removed by `ec808bbd6`, both inside this PR.
   - "4 of the 20 swept sites (`capabilities`, `properties`, `performance`, `signals`) … each now has
     one, via `assert_wire_omits_unset()`" — it's 3 of 4.
     `tests/integration/test_signals_wire_shape.py` doesn't import that helper, dispatches
     `Transport.IMPL` only, grades `result.payload.model_dump(mode="json")`, and its own docstring says
     it "is NOT wire-transport coverage."

2. **SHOULD-FIX — `fix(tests):` on a diff that changes buyer-visible wire shape on three transports.**
   `release-please-config.json` maps `fix` → "Bug Fixes" with no scope filtering, so the changelog will
   read "**tests:** pin e2e AdCP schema validation…" and describe none of: `Product.model_dump()`'s
   omission changes and fabricated default, `Creative.model_dump()`'s asset null-stripping,
   `SyncCreativesResponse.status`, 14 MCP wrappers' `structured_content`, and A2A `list_creatives`
   `format_id` becoming the spec object. Reverting the change the changelog *does* describe now reverts
   all of that with it.

Plus the four threads below, and one you may want to batch with them: the diff adds **36** local
tracker ids across 23 tracked files (`git diff origin/main...HEAD | grep -E '^\+' | grep -cE
'salesagent-[a-z0-9]{3,}'`), e.g. `src/core/schemas/product.py:537`, `adcp_a2a_server.py`,
`tests/helpers/skill_to_adcp_task.py`, `.beads/last-touched`. The guard rules ask for a GitHub
issue/PR number and say explicitly never a local id, since those don't resolve for outside
contributors. The `FIXME(#1888)`/`FIXME(#1889)` markers are correctly formed — it's narrowly the local
ids and the `R…-N` labels. Pure cosmetics, so your call whether it rides along.

---

## Not for this PR — filed separately, not merge conditions

I found more, and I'm deliberately not asking for any of it here. Recording it so it isn't lost:

- **Envelope-required `status` is declared on 1 of 4 eligible models.** `GetMediaBuysResponse`,
  `SyncAccountsResponse` and `ActivateSignalResponse` have no `status` field, and their
  `model_dump(mode="json")` fails the pinned response schema with `at <root>: 'status' is a required
  property`. `get_media_buys`/`sync_accounts` are live; `activate_signal` isn't registered on any
  transport. This is my most substantive finding and it still doesn't belong here — fixing it means
  three more wire-shape changes plus tests, and those wires were invalid before this PR too.
  (`CreateMediaBuyResponse`/`UpdateMediaBuyResponse` are union aliases whose arms already declare it;
  `ListAuthorizedPropertiesResponse` has no pinned schema. And `GetMediaBuysMediaBuy`'s own required
  `status` is the per-media-buy lifecycle enum — a different field.)
- **`list_accounts` emits three schema-invalid nulls on MCP and A2A.**
  `src/core/schemas/account.py:50-57`'s `_ALWAYS_INCLUDE` re-inserts `advertiser`/`rate_card`/
  `payment_terms` as `None` after `super().model_dump()`; `core/account.json` types all three as
  non-nullable and not required. Last unconditional force-include-null override under
  `src/core/schemas/`. Outside this diff.
- **`test_all_adcp_skills_have_schemas` iterates the map it exists to police.** The agent card declares
  16 skill ids; `SKILL_TO_ADCP_TASK` has 12; the four unmapped are exactly the ones carrying the two
  items above. Deriving the iteration set from the card would close it.
- **The narrowed exception boundary's primary consumer can't see an implementation bug.**
  `tests/e2e/test_adcp_schema_compliance.py` catches `except Exception` → `"warning"` and has zero
  `assert` statements. Injecting an `AttributeError` into `_validate_against_schema` leaves it at 8
  passed while `test_schema_validation_standalone.py` goes 4 failed / 15 passed. Because the validator
  no longer wraps implementation bugs, they moved *out* of the `"fail"` arm and *into* `"warning"`.
- **`_INSTRUMENT_FAILURES` is incomplete one way and dead the other.** `OSError`/`PermissionError` and
  `UnicodeDecodeError` on the tree read, and `JSONDecodeError`/`FileNotFoundError` on `index.json`
  (read outside the try), all escape a caller written to the documented `except SchemaError:` contract;
  `jsonschema.exceptions.SchemaError` can't be raised by `Draft7Validator.iter_errors` at all. And a
  structurally-invalid schema (`{"required": "notalist"}`) becomes 7 payload-level ValidationErrors —
  the inversion the module exists to prevent.
- **`_normalize_ref` discards the URI authority.**
  `https://evil.example/schemas/latest/core/product.json` resolves to the local `core/product.json`.
  Latent — I censused all 669 plain-tree schema files at this pin: 0 absolute `$ref`s, 0 hosts. Worth
  closing eventually because this PR exists precisely because upstream moved `$ref`s to absolute.
  Note the parametrized rejected-forms case labelled `# foreign host` passes on its *path*, not its
  host.
- **`Product.model_dump()`'s backfill is presence-only.** A present-but-incomplete
  `reporting_capabilities` skips it and stays schema-invalid; `{"metrics": ["impressions"]}` → 6 errors,
  `{"available_metrics": [...], "expected_delay_minutes": 30}` → 4. Both are fixtures in
  `tests/unit/test_product_schema_model_dump.py`, and `:174` asserts the second is preserved onto the
  wire unchanged. The default itself validates clean, and the only in-tree partial writer
  (`broadstreet/managers/inventory.py:281`) has no call sites, so it's latent.
  `{**DEFAULT_REPORTING_CAPABILITIES, **(value or {})}` would close it.
- **The fabricated default is a binding commitment.** Per 3.1.1, a product's declared
  `available_metrics` becomes the binding reporting contract, so `["impressions"]` understates what the
  seller actually reports — and no operator can set the field (zero `reporting_capabilities` hits
  across `templates/`). Its two sibling required fields both fail loudly instead
  (`product_conversion.py:395`, `:440`).
- **Internal `Targeting` fields still ship on the `content`/`message` channel.** `mcp_result`'s
  `content=str(response)` and the A2A `message` use the default repr, which ignores both the
  `model_dump()` override and `Field(exclude=True)`. Verified end-to-end: `structured_content` gives
  `{"geo_countries": ["US"]}` while the content text carries `key_value_pairs` (incl. its value),
  `tenant_id` and `had_city_targeting`. Reachable at `media_buys[].packages[].targeting_overlay`.
  Exactly one rerouted response lacks a curated `__str__` — `GetMediaBuysResponse`. Both `str(response)`
  sites are pre-existing.
- **That invariant's only oracle is a serializer tautology.**
  `tests/unit/test_get_media_buys.py:566` asserts on `model_dump(exclude_none=True)` — the fix itself.
  Reverting `media_buy_list.py` to the pre-PR `ToolResult(...)` leaves the full unit suite at 1 failed
  / 5614, and the single failure is the AST shape guard.
- **Guard-infrastructure gaps the new guards inherit.** `iter_architecture_guard_trees` globs only
  `test_architecture_*.py`, so all 12 `test_guards_*.py` files sit outside the two meta-guards (6 real
  violations there, one added here). The new broad-except allowlist is a `set` keyed
  `(path, func, type)`, so a second violation anywhere in `_get_products_impl` — a 600-line function
  with three `except Exception` handlers — is silently absorbed; a `Counter` fixes it. And
  `test_architecture_test_marker_coverage.py` can never fail: it parses `--collect-only -q` for `::`,
  but `pytest.ini:11` sets `-v`, netting verbosity 0, so pytest emits the tree format with zero `::` —
  648 unmarked tests across 57 files, guard green. `-q -q` fixes the parser and exposes all 648, which
  is emphatically its own track.
- **`copy_before_mutating` is a shallow copy** and 4 of its 5 sites have no behavioural oracle
  (removing all four `_base.py` calls reddens only the AST guard). All five write top-level keys only,
  so the shipped fix is sufficient today; the unqualified docstring isn't. The guard also doesn't see
  `del values[k]`, `values |= {...}`, nested writes, or aliasing.
- **Two of the three new guard files hand-roll a byte-identical `_parse`/`REPO_ROOT`** while the third
  imports `safe_parse` from `tests.unit._architecture_helpers` in the same commit — and the canonical
  one is stronger (checks `exists()`, reads UTF-8). The five-line body is below the duplication gate's
  threshold, so it can't see it.
- **`scripts/verify_feature_error_codes.py:77`** catches `(AssertionError, KeyError,
  ModuleNotFoundError)`; `json.JSONDecodeError` and `AttributeError` escape `load_enum` uncaught, and an
  uncaught traceback exits 1 — the code the script uses for "findings exist". Its own comment at
  `:78-85` names that exact hazard.
- Smaller: the `success`-from-`errors` derivation now emits `status: "completed"` with `success: false`
  for a spec-defined partial success (3.1.1 says non-fatal errors populate only the payload);
  `products.py:417` emits `recovery="transient"` on `VALIDATION_ERROR`, whose canonical recovery is
  `correctable`; `Provenance.verification` is the unswept sibling of the `Creative.assets`
  null-stripping fix (and the pinned schema types it as an array, not a dict);
  `validate_response([])` dies with an unwrapped `AttributeError` where `validate_request([])` gives a
  proper validation error; the `ToolResult` guard misses the 2nd-positional form `ToolResult(str(r), r)`;
  and `test_collection_list_targeting.py:158-159`'s two spec-drift tests are dormant on a path that no
  longer exists, but `core/collection-list-ref.json` is in the pinned tree now.

## What the PR gets right, verified rather than taken from the description

`mcp_result` adoption is complete on its `ToolResult` axis (15/15, content byte-identical, and the
extraction kept the stronger absorbed behaviour). The reroute adds **no** new exposure on
`structured_content` — all 15 models sentinel-filled, old-vs-new serialization diffed key-path by
key-path, `added=0` for all 15. `SyncCreativesResponse.status` is spec-mandated, not invented.
`DEFAULT_REPORTING_CAPABILITIES` is a real single source of truth and validates clean.
`list_authorized_properties` is handled honestly with `schema=None` because no pinned schema exists.
No allowlist grew, both new broad-except entries are pre-existing with correctly-formed FIXMEs and a
count-pinning meta-test, and the duplication baseline moved `tests: 74 → 73`. The relocation to
`tests/helpers/` *widens* guard coverage. The 243-file fixture deletion can't regrow — `_refresh.py`'s
`ROOTS` is one leaf enum with zero `$ref`. And the CVE rider is minimal-and-sufficient rather than
scanner-silencing: `CVE-2026-69244` affects aiohttp ≤3.14.2 so 3.14.3 is required, and
`CVE-2026-69247` affects cryptography ≥44 <50 so the major bump is the only remediation.

## Not checked

No local full-suite run — CI is green on all six suites here. The `status`/nulls findings come from
constructing the model and validating `model_dump(mode="json")`, which is the same bytes
`mcp_result()` puts in `structured_content`, but not a live transport. 5 of ~60 conformance storyboards
read, so "the default's content is ungraded" is scoped to those. REST not confirmed for the two
schema-invalid wires, though `api_v1.py` uses the same dump at 8 sites. `Provenance.verification`
reachability is inferred at the last link.

KonstantinMirin:
@ChrisHuie thanks for the thorough pass. Went through everything:

The 5 SHOULD-FIX description-accuracy items are all corrected in the current PR description (2 of them — `strip_none_deep` and the `verify_feature_error_codes.py` `RuntimeError` arm — turned out to have never existed on `main` at all; both were introduced and removed entirely within this branch's own commits, so I dropped them from the "surfaced and fixed" framing rather than leave a description that credits this PR with fixing a bug that was never real). Title's fixed too — dropped the `(tests)` scope, since it undersold the wire-shape changes.

For the "not for this PR" list — confirmed none of it is code this branch touched (checked file-by-file against `git log main..HEAD`), so none of it belongs here. Filed each as its own GitHub issue with the reproduction evidence carried over, rather than let it get lost in a review comment:

#1900 (envelope `status` missing on 2 live models), #1901 (`list_accounts` null-leak), #1902 (schema-compliance test's broad except), #1903 (`_INSTRUMENT_FAILURES` gaps), #1904 (`_normalize_ref` foreign-host), #1905 (fabricated-default conformance question), #1906 (`Targeting` leak on the content/message channel), #1907 (serializer-tautology test), #1908 (guard-infrastructure gaps, all three bundled), #1909 (`copy_before_mutating` shallow-copy docs/oracle gap), #1910 (`Provenance.verification` unswept sibling), #1911 (`validate_response([])` crash), #1912 (`ToolResult` guard blind spot), #1913 (dormant `collection_list_targeting` tests, now resolvable).

Two more from the same pass turned out to already be fixed by this branch's later commits, so I closed the internal tickets rather than file anything: the `#1623`-adjacent MCP status-null bug (its specific reproduction is gone, mechanism now in place — a residual 3-of-16-wrapper gap is called out in the description) and the factory `reporting_capabilities` default shape mismatch (the factory doesn't set a custom default anymore — it just inherits the model's own compliant one now).

KonstantinMirin:
One more from the same review pass, worth closing the loop on explicitly: the "backfill is presence-only" concern (\`Product.model_dump()\` accepting a partially-populated \`reporting_capabilities\` and silently shipping a schema-incomplete result).

Verified directly against the library type: \`LibraryReportingCapabilities\` has 6 required fields (\`available_reporting_frequencies\`, \`expected_delay_minutes\`, \`timezone\`, \`supports_webhooks\`, \`available_metrics\`, \`date_range_support\`), pydantic-enforced at construction time. Since the field is now typed (not a dict-merge backfill), a partial value missing any of those raises \`ValidationError\` immediately — there's no way to construct a \`Product\` with an incomplete \`reporting_capabilities\` that passes validation. The mechanism change (typed field + \`default_factory\`, replacing the old dict-merge default) closes this structurally, not just relocates it.

tests/unit/test_pydantic_schema_alignment.py
KonstantinMirin: Fixed in 9fa436344. You were right the rationale was false, but not quite why — status isn't unstamped-and-therefore-excluded, it's already carried: 10 of 11 registered models inherit a spec-correct default straight from the adcp library base (verified directly against the library's Pydantic field defaults, not assumed). `_serialize_for_a2a` really does stamp only message/success, as you found — the honest reason for the exclusion is two suite limitations (the property walk doesn't merge allOf properties; the generator can't synthesize enum/Literal values) plus the one genuine gap you measured (SyncAccountsResponse, now #1900). The comment states that now.

Also derived the hand-copied 11-field set from the pin's own `required` array instead of hand-listing it — proven equivalent to the old set (the other 10 were genuinely inert), plus a tripwire so a future pin bump can't silently widen the exclusion. Un-excluding `status` for real grading is filed as separate follow-on work, since it needs both suite limitations fixed.

tests/unit/test_architecture_no_broad_except_validation_relabel.py
KonstantinMirin: Already fixed — landed incidentally in 6a70e3ddb when that commit reused this same guard file for the products.py/policy_service.py violations. `_is_broad_handler` now does `return any(_is_broad_name(e) for e in node.elts)` for the tuple case, and the docstring says a tuple containing `Exception`/`BaseException` is exactly as broad as the bare name. Verified against current HEAD directly, not assumed.

tests/unit/test_guards_before_validator_no_mutation.py
KonstantinMirin: Already fixed — this guard predates `copy_before_mutating()` (created in round 2, before that helper existed in round 3), and the message was updated to match when the helper was introduced. Current text: "Add `values = copy_before_mutating(values)` before any mutation -- the shared helper, not a raw `.copy()`, so the rationale lives in one place." Verified against current HEAD directly.

src/core/schemas/product.py
KonstantinMirin: Fixed in 6f9910fe5. Agreed and confirmed rather than assumed: checked the library base's actual Pydantic field definitions directly (`adcp.types.Product.model_fields`), not just schema requiredness — all four are genuinely required with no default. `core_fields` and the disjunct are gone. Also found and removed its twin: the `elif key == "pricing_options" and value == []` branch two lines down was equally dead (`[] is not None` is already True), same statement, same fix. Kept the `format_ids` rationale — that one's real. Added a guard test pinning the library-requiredness invariant so a future SDK loosening fails loudly instead of quietly needing a safety net back.

src/services/policy_service.py
KonstantinMirin: Already fixed — the whole try/except this bug lived in is gone (deleted in 6a70e3ddb, along with the sibling in products.py, as the actual fix for the broad-except-relabel class rather than a narrower except clause). The comment now there states directly why: "No try/except here on purpose: the previous one caught the very ValidationError raised below and re-wrapped its precise, actionable message inside a generic wrapper..." Verified against current HEAD directly.

tests/unit/test_architecture_no_broad_except_validation_relabel.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    node = handler.type
+    if node is None:
+        return True
+    if isinstance(node, ast.Tuple):
+        # A tuple CONTAINING Exception is exactly as broad as Exception alone —
+        # the named siblings are already subsumed by it. Treating tuples as
+        # narrow left "except (ValueError, Exception)" as a one-token way past
+        # this guard. Measured when tightened: zero new violations tree-wide.
+        return any(_is_broad_name(element) for element in node.elts)
```
ChrisHuie, line 91: **SHOULD-FIX** — this tightening works but nothing pins it.

I ran `except (KeyError, Exception)` through the guard and it is caught, so the fix is real. But all
seven `_POSITIVE_SAMPLES` are `None`/`Name`/`Attribute` handlers, and the only tuple sample is a
*negative* that stays satisfied either way — so reverting this line to `return False` leaves all 14
meta-tests in this file green.

The comment just above says "Measured when tightened: zero new violations tree-wide", which makes
this a claimed invariant with no failing oracle. One line in `_POSITIVE_SAMPLES` closes it:

```python
"tuple-with-Exception": "try:\n    f()\nexcept (KeyError, Exception) as e:\n    raise MyValidationError('x') from e\n",
```

tests/unit/test_architecture_no_broad_except_validation_relabel.py
```diff
@@ -0,0 +1,197 @@
...
+
+# Known violations — shrink-only, never grow. Keyed by (path, enclosing function,
+# raised type) rather than line number: a fix elsewhere in the file shifts lines,
+# and a line-keyed allowlist then reads as "fixed" when nothing changed.
+# EMPTY, and it must stay empty. Both original entries (#1888 get_products
+# property-list resolution, #1889 currency-code validation) were fixed by
+# deleting the handlers, not by allowlisting them. An AST sweep of src/, tests/
+# and scripts/ finds no other instance of this shape anywhere in the tree, so a
+# new entry here would be a NEW defect, never inherited debt.
+_KNOWN_VIOLATIONS: set[tuple[str, str, str]] = set()
```
ChrisHuie, line 65: **SHOULD-FIX** — the comment above says this must stay empty, and nothing makes that true.

`test_allowlist_entries_cite_a_github_issue` was removed in the same commit rather than rewritten. It
string-split on `"_KNOWN_VIOLATIONS: set[...] = {"`, which the empty-set form no longer contains, so
it had to change — but nothing replaced it. `test_known_violations_not_stale` only requires that an
allowlisted violation still exists; it passes for a re-added entry with no tracker and no FIXME, and
there's no repo-wide FIXME-presence guard to catch it either.

Shipping at zero is the strong position, so assert it directly:

```python
assert _KNOWN_VIOLATIONS == set(), "this guard ships empty -- delete the handler instead of allowlisting it"
```

That's stronger than the test that was removed.

ChrisHuie (commented):
Re-reviewed at `f7dba0b49`. **0 blocker**, and nothing below changes a wire shape or needs a new
fixture. All five threads from the last round are addressed — I diff-verified each at this head
rather than trusting the resolved state, and two were swept further than I asked. The 14 deferred
items are all filed and accurately titled.

Two things I tried to break and could not: the six-file consolidation into
`test_wire_omission_matrix.py` is complete rather than weaker (17 collected before and after, and
removing `strip_none_deep` from `Product.model_dump` reddens it on all three wire transports plus
IMPL), and the `_PROTOCOL_ENVELOPE_FIELDS` tripwire holds in both directions — simulating a pin that
also requires `timestamp` reddens `test_protocol_envelope_requires_only_status` and nothing else.

I've scoped this round to things that are one line to a few lines each, so this can land. Eight are
anchored as threads below. Three more that aren't anchorable to a changed line:

**Description — two sentences about the wire contract don't survive tracing.**

- "a crash now reaches the buyer as an honest `INTERNAL_ERROR`" is true for the `get_products`
  deletion and not for the currency one. `validate_currency_code` has no AdCP transport caller —
  the chain is `validate_currency_limits` -> `update_policies` -> a Flask admin route — and its
  `ValidationError` is a module-local `Exception` subclass that never reaches
  `normalize_to_adcp_error`. What that deletion actually buys is the precise message reaching the
  caller instead of a double-wrapped generic one, which is what the unit test grades.
- "remapped at the wire boundary" reads as unconditional; it holds on MCP and A2A. The FastAPI app
  registers handlers for `HTTPException`, `RequestValidationError`,
  `WebSocketRequestValidationError`, `AdCPError`, `ValueError`, `PermissionError` and `ToolError` —
  no bare `Exception` — so an untyped exception out of a REST route is Starlette's default 500 with
  no AdCP envelope.

**Description — the aliasing sentence describes something that never shipped.** "the default was one
shared dict instance" — the assignment took exactly one form across every commit on this branch,
`dict(DEFAULT_REPORTING_CAPABILITIES)`, which copies per call. What was shared is the nested lists,
which is what the new `_default_reporting_capabilities` docstring says correctly. And
`DEFAULT_REPORTING_CAPABILITIES` doesn't exist on `main` at all, so it sits under "bugs the tightened
validation surfaced" describing something that never reached production — the same shape as the two
claims you already pulled for that reason.

**One registry row.** `_RESPONSE_MODEL_REGISTRY`'s preamble excludes `GetMediaBuysResponse` as "not
spec-grounded", but `media-buy/get-media-buys-response.json` is present in the pinned tree this PR
ships — I loaded it. The model genuinely has no library base, so the stated criterion holds; the
inference doesn't. The registry keys membership on how a model was built rather than on whether the
pin has a schema for it, and `GetMediaBuysResponse` is where those diverge — which matters because
#1900's acceptance requires that model to declare `status`, and this exclusion puts it out of reach
of the suite meant to grade it. Not anchorable here since the line is unchanged.

**Deliberately not asking for these now**, recording so they aren't lost. The before-validator guard
is the only oracle for four of its five call sites — remove the four `_base.py`
`copy_before_mutating()` calls and deselect that guard, and the whole unit suite is green (5607
passed). It's a real gap but it's about the safety net, not the code: the calls themselves are all
correct today. Two parts, both worth a follow-up rather than a push here: it inspects only each
validator's own body, so the three that delegate to `_upgrade_legacy_format_ids` are ungraded; and 12
of its 15 non-first compound operands are dead to the suite, the notable one being `arg.id == param`,
whose loss would make `copy_before_mutating(some_other_dict)` count as a valid defensive copy.
Separately, the untyped-exception branch classifying a permanent defect as `transient` lives at one
repo-wide seam, not at your call site, and the pair you removed was strictly worse — also follow-up.

## Not checked

No local full-suite run; CI is green on all suites at this head, and E2E moved 107 passed / 31
skipped to 111 / 26. Integration coverage came from targeted per-file runs, not a fresh CI database.
Nothing was executed inside the e2e container. REST's behaviour above is from the handler registry
plus a TestClient probe, not a running server.

tests/integration/test_wire_omission_matrix.py
```diff
@@ -0,0 +1,332 @@
...
+    absent_paths: Sequence[str] = ()
+    transports: Sequence[Transport] = _ALL_WIRE
+    # Required when transports is narrower than every wire transport.
+    narrowed_because: str | None = None
+    # A response key that must be non-empty, so a schema-valid EMPTY body cannot
+    # pass as coverage.
+    must_be_non_empty: str | None = None
+
+    def __post_init__(self) -> None:
+        if self.schema is None and not self.absent_paths:
```
ChrisHuie, line 79: **SHOULD-FIX** — this is the check that makes the consolidation safe, and it has no failing oracle.

Deleting both `raise ValueError` branches leaves this file at 17 passed and the wider guards slice at
655 passed. Nothing outside this module references `WireOmissionCase`, `narrowed_because` or
`must_be_non_empty`, so no other caller exercises them either.

That matters more here than it normally would: this is what stops a future row being silently
narrowed to one transport, or added with no schema and no `absent_paths` so it grades nothing — the
exact regression the six-file merge would otherwise make invisible. Both operands are evaluated at
import (the two `schema=None` rows, and the narrowed row), but neither `raise` is ever reached under
test.

Two small tests cover it:

```python
with pytest.raises(ValueError, match="grades nothing"):
    WireOmissionCase(tool="x", setup=..., schema=None)

with pytest.raises(ValueError, match="narrowed_because"):
    WireOmissionCase(tool="x", setup=..., schema="...", transports=(Transport.MCP,))
```

tests/helpers/pinned_schema.py
```diff
@@ -1,73 +1,250 @@
...
+    through the walk.
+    """
+    path, schema = _resolve_and_load(ref)
+    return _canonicalize_refs(schema, file_dir=path.parent, root=schema_root())
+
+
 def _retrieve(uri: str) -> referencing.Resource:
-    """referencing retrieve callback: resolve a ``/schemas/...`` ref from the pinned tree."""
-    return DRAFT7.create_resource(_load_by_ref(uri))
+    path = _contained_path(Path(url2pathname(urlparse(uri).path)), what=f"Schema URI {uri!r}")
```
ChrisHuie, line 228: **SHOULD-FIX** — the foreign-host `$ref` gap moved here rather than closing, and the filed issue now
points at the wrong function.

`normalize_ref` correctly rejects absolute URLs now — I probed it and got `PinnedSchemaError`. But
this line resolves `urlparse(uri).path`, dropping scheme *and* netloc, so a `$ref` naming any host
still lands on the same-path file in the pinned tree. `_contained_path` is the only gate and it gates
on path, never authority. Probed at this head: an `https://evil.example/...` URI pointing at the
pinned `core/product.json` path loads it and returns `title=Product`.

Still latent, and I re-derived that rather than carrying it: 669 plain-tree schema files at this pin,
0 absolute `$ref`s, 0 hosts. So the deferral itself is defensible.

The reason I'd fix or repoint it now is that #1904's title, body and acceptance all name
`_normalize_ref` — which is clean. Anyone verifying at the named site sees it fixed and closes a live
issue. Either add a scheme/netloc guard before the containment check, or repoint the issue text here.

tests/e2e/test_schema_validation_standalone.py
```diff
@@ -87,29 +103,115 @@ async def test_get_products_request_validation():
...
+def test_normalize_ref_accepted_forms(ref, expected):
+    """The one accepted ref form resolves against the pinned version root."""
+    validator = AdCPSchemaValidator()
+    assert validator._normalize_ref(ref) == expected
+
+
+@pytest.mark.parametrize(
+    "ref",
+    [
+        # The absolute-URL form upstream #6133 introduced — the outage's trigger.
```
ChrisHuie: **NIT** — this renders as a link to `prebid/salesagent#6133`, which 404s.

The number is upstream `adcontextprotocol/adcp`. The word "upstream" right before it means a reader
isn't misled, but the autolink is broken. `adcp#6133` renders correctly and resolves cross-repo.

Same in the module docstring of `tests/helpers/adcp_schema_validator.py`.

tests/helpers/adcp_schema_validator.py
```diff
@@ -0,0 +1,312 @@
+"""AdCP JSON Schema Validator — pinned to the installed SDK's schemas.
+
+Validates requests and responses against the AdCP schemas shipped with the
+pinned ``adcp`` SDK (see docs/adcp-spec-version.md for the current pin and
+bump procedure) — never against the live adcontextprotocol.org registry. The
+live registry serves "latest", which
+drifts ahead of the repo's pin: on 2026-08-01 upstream PR #6133 canonicalized
```
ChrisHuie: **NIT** — same broken autolink as the other site: bare `#6133` resolves to `prebid/salesagent#6133`,
which doesn't exist. `adcp#6133` is the cross-repo form.

tests/unit/test_pydantic_schema_alignment.py
```diff
@@ -85,49 +90,79 @@
...
-# in our code. Re-derived against the pinned schemas — entries describing fields the
-# pin no longer defines were dropped. #1388 tracks the adcp 5.7 alignment.
+# Every AdCP response schema composes the same shared Protocol Envelope arm
+# (core/protocol-envelope.json) via a top-level allOf, and that arm spec-requires
+# exactly one field: "status". Excluding it from requiredness grading is NOT a
+# layering principle — it is a temporary concession to two mechanical limits of
+# this suite, and one tracked model gap:
+#
+#   1. the property walk in _resolve_response_item_schema merges allOf `required`
+#      but not allOf `properties`, so 6 of the registered models cannot be graded
```
ChrisHuie: **NIT** — measured 7, not 6.

Emptying `_PROTOCOL_ENVELOPE_FIELDS` reddens `test_declared_fields_present_in_schema_and_model` for
`GetProductsResponse`, `GetMediaBuyDeliveryResponse`, `GetCreativeDeliveryResponse`,
`ListAccountsResponse`, `ListCreativesResponse`, `ListCreativeFormatsResponse` and
`GetSignalsResponse` — all with the same `'status' not defined by pinned schema` cause.

Claims 2 and 3 in this comment both verify exactly as written (4 enum-synthesis false failures;
`SyncAccountsResponse` genuinely lacking the field), and I confirmed 10 of 11 models carry a
non-None `status` default. Worth noting the two buckets overlap — `GetProductsResponse`,
`ListAccountsResponse`, `ListCreativeFormatsResponse` and `GetSignalsResponse` appear in both — since
the next reader will use these numbers to size the work.

tests/integration/test_get_products_property_list_error_wire.py
```diff
@@ -0,0 +1,67 @@
...
+Two things are wrong with that on the wire, and only the wire can show them:
+
+1. The buyer is told ``VALIDATION_ERROR`` — "your request is malformed" — for a
+   fault entirely on our side (an ``AttributeError``, a ``TypeError``, a DNS
+   failure inside the resolver).
+2. The arm forced ``recovery="transient"`` while the pinned AdCP enum classifies
+   ``VALIDATION_ERROR`` as ``correctable``. The buyer is simultaneously told the
+   request is invalid AND that retrying the identical request may work.
+
+Graded on ``wire_error_envelope`` across every transport that has a wire.
```
ChrisHuie: **NIT** — "every transport that has a wire" vs `_WIRE_TRANSPORTS = [Transport.MCP, Transport.A2A]`.

REST has a wire. The exclusion comment further down gives the correct reason and I verified it —
`GetProductsBody` in `src/routes/api_v1.py` declares no `property_list`, so the request can't be
expressed over REST at all. It's only the absolute sentence up here that overstates.

Something like "across every transport that can carry this request (see the REST note below)" would
match what the table does.

KonstantinMirin:
One more for the deferred pile, since it changes whether #1904 is actionable as written.

#1904 ("`_normalize_ref` discards the URI authority — a foreign-host `$ref` resolves to the local
schema tree") no longer describes current code. `normalize_ref` in `tests/helpers/pinned_schema.py`
rejects any ref containing `://` outright with `PinnedSchemaError`, and
`AdCPSchemaValidator._normalize_ref` is a thin delegate to it — so the named site raises instead of
resolving.

There is a residual gap, but it's narrower than the one described and I don't think it's reachable:
`_retrieve` in the same module does `url2pathname(urlparse(uri).path)`, dropping scheme and netloc
before `_contained_path` runs. Containment still confines the result to the pinned tree, so a
foreign-host URI can only ever re-load a file already inside that tree — one a legitimate caller
could load anyway. Hardening, not an exploit path.

I'm leaving the issue's text alone rather than retargeting or closing it unilaterally. Flagging it
because as written it reads as live, and the failure mode is someone verifying at `_normalize_ref`,
finding it already fixed, and closing #1904 with the `_retrieve` behaviour never looked at.

tests/unit/test_architecture_no_broad_except_validation_relabel.py
KonstantinMirin: Fixed in abbac21d8. Added a tuple-with-`Exception` case to `_POSITIVE_SAMPLES`. Verified the mutation you described: patching the tuple arm back to `return False` now fails exactly `test_guard_detects_disease[tuple-with-Exception]`, 17 passed otherwise — before this fix it left the whole file green, as you found.

tests/unit/test_architecture_no_broad_except_validation_relabel.py
KonstantinMirin: Fixed in abbac21d8, though not exactly as you sketched it. Went with `assert _KNOWN_VIOLATIONS == set()` folded into the existing `test_known_violations_not_stale` rather than a standalone test — a standalone test would just pin a constant and only catch a rule-break a diff already shows. Verified with a real oracle: planted an actual violation in the tree and allowlisted it (not a stale fake), confirmed it fails, reverted.

tests/integration/test_wire_omission_matrix.py
KonstantinMirin: Fixed in abbac21d8 — both construction tests added as you sketched them. Verified each independently: deleting either `raise` arm in `__post_init__` now kills exactly its own new test, nothing else.

tests/helpers/pinned_schema.py
KonstantinMirin: Ended up disputing the reproduction rather than fixing it as filed — replied in a separate comment (https://github.com/prebid/salesagent/pull/1868#issuecomment-5256026154) since it changes what #1904 should say rather than confirming a fix. Short version: the foreign-host `$ref` load doesn't reproduce (checked two ways — direct `_retrieve()` call and the full resolution pipeline a real nested $ref would use — both correctly reject it). #1904 itself is stale against current `normalize_ref` though, which is the part worth fixing.

tests/e2e/test_schema_validation_standalone.py
KonstantinMirin: Fixed in 0b7b18c4a. `upstream #6133` -> `upstream adcp#6133`.

tests/helpers/adcp_schema_validator.py
KonstantinMirin: Fixed in 0b7b18c4a, same commit as the sibling site. Also checked the repo's own convention while in there — `adcp#NNN` is used 21 times across src/tests; these two were the only bare ones.

tests/unit/test_pydantic_schema_alignment.py
KonstantinMirin: Fixed in 0b7b18c4a — 6 -> 7, measured the same way you did (emptied `_PROTOCOL_ENVELOPE_FIELDS`, reran the alignment check). Also wrote the decomposition and the bucket-overlap note into the comment itself so the next reader gets the real numbers, not a total that reads like a share of 11.

tests/integration/test_get_products_property_list_error_wire.py
KonstantinMirin: Fixed in 0b7b18c4a. Reworded to point at the existing `_WIRE_TRANSPORTS`/REST-exclusion note below instead of restating it, so there's one place this is explained, not two.

KonstantinMirin:
The two unanchored description findings are fixed in 0b7b18c4a:

- The "production bugs" paragraph is split now — `get_products` keeps the `INTERNAL_ERROR`/wire-remap claim (accurate, MCP+A2A), the currency-code path gets its own accurate description (Flask-admin-route-scoped, no AdCP reachability at all, the real benefit is just an unwrapped message). While splitting it I also verified your REST observation directly (`src/app.py`'s 5 explicit handlers, no bare `Exception`) and stated it as a real, unfiled gap in the description itself rather than letting the "remapped at the wire boundary" line imply REST is covered when it isn't.
- The aliasing sentence is removed rather than rewritten — `DEFAULT_REPORTING_CAPABILITIES` never existed on `main` or this branch's HEAD, and `_default_reporting_capabilities`'s docstring already states the real (nested-list) sharing correctly, so restating it would just duplicate.

On the REST gap specifically: leaving it as a named-but-unfiled note in the description for now rather than opening a GitHub issue immediately — it surfaced as a side effect of verifying your finding, not something this PR's own commits touch either way, and I'd rather get your read on whether it's worth its own issue (a REST catch-all `Exception` handler mirroring the existing `ValueError`/`PermissionError` ones) before filing something that might turn out narrower or broader than it looks from here.

ChrisHuie (approved):

Issue #1849, opened 2026-08-04
create_media_buy does not enforce Product.allowed_principal_ids -- any authenticated principal can purchase ACL-restricted products by ID


KonstantinMirin:
## Summary

`Product.allowed_principal_ids` is a per-product access-control field. It IS enforced when listing products (`get_products` filters the returned catalog to only unrestricted products, or restricted products the calling principal is explicitly listed on). It is **not enforced anywhere else** in the codebase. The restriction only hides a restricted product from the catalog listing -- it does not prevent any authenticated principal from acquiring it.

## Evidence

`create_media_buy`'s product lookup (`src/core/tools/media_buy_create.py`, around line 2307) filters only by `tenant_id` and `product_id`:

```python
products_stmt = (
    select(ProductModel)
    .where(ProductModel.tenant_id == tenant["tenant_id"], ProductModel.product_id.in_(product_ids))
    .options(selectinload(ProductModel.pricing_options))
)
```

The calling principal's identity is never compared against the fetched product's `allowed_principal_ids` anywhere in this function. The only downstream check is product existence, not authorization.

The same tenant-only lookup pattern repeats in `execute_approved_media_buy` (same file, around line 842) when a pending manual-approval buy is executed.

`ProductRepository` (`src/core/database/repositories/product.py`) has no `allowed_principal_ids` awareness in any of its methods -- the ACL filter exists only as an inline loop inside the `get_products` tool handler (`src/core/tools/products.py`, around line 361-390). `create_media_buy` does not go through `ProductRepository` for this lookup at all; it is a raw `select(ProductModel)`.

## Impact / exploit path

Principal A creates a product restricted to `allowed_principal_ids=["B"]`. Principal C, who is not on that list, obtains the product's ID through any channel other than `get_products` (it is a plain string identifier, not a scoped or unguessable token -- it could be shared between buyers, logged, enumerated, or simply retained from before C's access was revoked). C calls `create_media_buy` with that product_id using C's own valid credentials. The lookup above returns the product regardless of C's identity (it only filters by tenant), passes the existence check, and the media buy is created -- a real spend-committing action against a product C was never authorized to see or buy.

## Why this is structural

Because the ACL check is duplicated logic living in one read-path tool handler rather than a shared repository accessor every product-consuming call site is forced through, every other call site that resolves products by ID (create, execute-approved-buy, and by inheritance anything built on top of them) has no mechanism reminding it to apply the same filter, and none currently do.

[...]

Issue #1861, opened 2026-08-04
Auth code masks DB/infra errors as auth decisions — and google_callback fails open


KonstantinMirin:
Authentication and authorization code across `src/core` and `src/admin` catches broad
exceptions from the database layer and converts them into a **deny decision**. A transient
DB error is therefore indistinguishable from a genuinely bad credential, and in one case
an infrastructure failure results in access being **granted**.

Found while measuring the AdCP 3.1.1 storyboard baseline (#1247 / #1858): `sync_accounts`
rejected `ci-test-token` as invalid while the *identical* token authenticated every other
call in the same run. There is no `sync_accounts`-specific auth code — the identity path
is provably identical across tools — so the rejection was a swallowed DB error surfacing
as "Authentication token is invalid".

### Root site

`get_principal_from_token()` (`src/core/auth_utils.py`) wraps `execute_with_retry` in a
bare `except Exception` and returns `(None, None)`. `resolve_identity()` renders that as
`Authentication token is invalid`. Callers cannot distinguish "no such principal" from
"the database was briefly unavailable", and a buyer agent is told to fix its credentials
when nothing is wrong with them.

### The same shape elsewhere

A disease scan found 13 further sites:

- `is_super_admin`, `is_tenant_admin`, `require_tenant_access`, `get_tenant_config_from_db`
  (`src/admin`) — a genuine outage silently denies instead of surfacing a 500.
- `config_loader`'s four tenant-lookup functions — match the exception by **string
  sniffing** rather than by type.
- `extract_user_info` — catches everything where only `jwt.PyJWTError` is meant.
- `google_callback` — **fail-open**: a failed User-record write is swallowed and the
  session is granted anyway. The sibling path at `select_tenant()` denies correctly, so
  this is an inconsistency as well as a security defect.

Two further sites were investigated and are *not* instances: the login page's deliberately
uniform auth-failure response, and `LazyTenantContext`'s documented DB-unavailable fallback.

### Expected

- Auth code catches the exception types that mean "no matching credential", nothing wider.
- Infrastructure failures surface as transient errors — `AdCPServiceUnavailableError` /
  503 on the protocol surface, a real 500 in admin — never as an auth decision.
- `google_callback` denies when the User record cannot be written.
- A structural guard pins the fixed sites so the shape cannot return.

### Notes

A complete fix was written and validated on #1858 and then **reverted** from it, because
that PR's scope (#1247) is gap analysis and roadmap, not production changes. The work is
[...]

Issue #1859, opened 2026-08-04
security: unauthenticated MCP/A2A calls to protected tools return HTTP 200 instead of 401


KonstantinMirin:
## Summary

An **unauthenticated** call to a protected tool over MCP or A2A returns **HTTP 200** instead of 401/403. Confirmed real, not a harness artifact.

Surfaced by the first measured AdCP conformance run (`universal/security.yaml`, `probe_unauth` check) in #1858. That storyboard grades this explicitly:

> *"an agent that 200s on an unauth call and 200s on a bad-credential call is ignoring credentials entirely."*

REST is unaffected — it already maps `AdCPError.status_code` to a real HTTP status via FastAPI exception handlers.

## Root cause — transport boundary, not tool logic

The per-tool `_impl` enforcement is **correct and is not the defect**. `require_principal_id` raises `AdCPAuthenticationError` as it should. The gap is that no transport turns that into an HTTP status:

- **MCP** — `src/app.py` mounts FastMCP's `mcp.http_app()` directly, with no ASGI-level auth gate in front of `/mcp/`. FastMCP's Streamable-HTTP JSON-RPC transport always returns 200 for a completed tool call, even when the body reports `AUTH_REQUIRED`.
- **A2A** — identical defect one layer shallower: `a2a-sdk`'s `jsonrpc_dispatcher.py` hardcodes `status_code=200` for every JSON-RPC error response, so auth failures raised via `InvalidRequestError` in `src/a2a_server/adcp_a2a_server.py` never surface as a real status either.

Compounding it: `resolve_identity()` (`src/core/resolved_identity.py:129-203`) only raises when a token is **present but invalid**. With **no token at all** it silently returns `ResolvedIdentity(principal_id=None)`, and `MCPAuthMiddleware.on_call_tool` (`src/core/mcp_auth_middleware.py:39-72`) stores that on context state regardless.

The probe targets `list_creatives` — a genuinely protected tool, not a discovery endpoint.

## Proposed fix — extend existing middleware, don't add a new one

`UnifiedAuthMiddleware` (`src/core/auth_middleware.py`) is already pure ASGI, already wraps the whole app including `/mcp` and `/a2a` regardless of mount order, and already parses auth headers into request state. **It just never rejects** — it only populates.

Extend it to short-circuit with a real 401 `JSONResponse` + `WWW-Authenticate` header when all of:

1. the target is `/mcp` or `/a2a`,
[...]

PR #1812, opened 2026-07-30
fix: principal-scope durable get_task / get_by_step_id_or_raise (#1741)

mkostromin-sigma:
## Summary

- Principal-scope durable AdCP `get_task` / `WorkflowRepository.get_by_step_id_or_raise` via resolve-then-authorize (select by `step_id`+`tenant_id`, then compare `Context.principal_id` in Python with equal-shape work on miss) — AdCP 3.1.1 L3 error-handling MUST.
- Sibling principal in the same tenant who learns a durable `step_id` receives the same wire `REFERENCE_NOT_FOUND` (internal `TASK_NOT_FOUND`) as an unknown id on `get_task` / `complete_task` (no ownership oracle). `list_tasks` / `list_by_tenant` remain tenant-wide — tracked as #1808; `docs/security.md` Principal Isolation names that boundary.
- Buyer input contract: empty **and** non-string `task_id` validated in shared L2 (`_require_task_id`); MCP signatures restore `task_id: str` (required) and middleware translates FastMCP missing-required `ToolError` into AdCP envelopes; A2A rejects unknown keys (typo `statas` → `VALIDATION_ERROR`); `@validates("error_message")` + L2 type check + `normalize_to_adcp_error` SQL-leak gate close the non-string `error_message` wire disclosure.
- Skills `get_task` / `complete_task` remain outside pinned manifest vocabulary — allowlisted with `FIXME(#1942)`; durable manifest pin tracked as #1943.

Closes #1741

Related: #1702, #1720, #1808, #1942, #1943, #1944, #1945, #1946, #1947, #1957

## Spec-Grounding (AdCP 3.1.1)

- **Authority:** pinned AdCP **3.1.1** (`docs/adcp-spec-version.md`).
- **Prose:** `by-layer/L1/security.mdx` § Agent and Account Isolation — bind state on create, verify on access, fail closed without distinguishing unauthorized from not-found. Plus `by-layer/L3/error-handling.mdx` (Not-found precedence; Uniform response; resolve-then-authorize two-step) and `protocol/get-task-status-request.json` (sellers MUST return `REFERENCE_NOT_FOUND` for a task_id that exists only under a different account or principal).
- **Storyboard:** sibling-principal durable `get_task` — **not_applicable** for this repo's tool surface (`get_task_status` MAY). This seller advertises durable `get_task` / `complete_task` (non-manifest; #1942/#1943) rather than the optional alias.
[...]

PR #1870, opened 2026-08-05
docs: add ACL/authorization architecture investigation notes

KonstantinMirin:
Pure docs addition, no code changes. Records findings from a direct architecture investigation (2026-08-04) into per-buyer tool ACL, AdCP `specialisms` enforcement, and admin RBAC — separate lineage from the PR #1838/#1868 review rounds. Tracked by the `salesagent-0krw` epic, which cites this file as its full writeup; it was previously uncommitted, sitting only in a local worktree with no durable home.

Related: #1849 (the P0 finding this investigation surfaced).

### DIFF
--- a/.claude/notes/acl-authorization-architecture.md
+++ b/.claude/notes/acl-authorization-architecture.md
@@ -1,0 +1,96 @@
+# Access control / authorization architecture — current state and target design
+
+Investigation done 2026-08-04 while triaging PR #1838 (salesagent-1zq3). Two
+questions drove it: (1) can this codebase check whether a calling party has
+access to a specific tool, and vary tool access by caller — the AdCP-ACL
+question; (2) is `get_adcp_capabilities`' `specialisms` field enforced
+anywhere, or purely descriptive.
+
+## 1. What exists today
+
+| Mechanism | Gates | Varies by | Enforced consistently? |
+|---|---|---|---|
+| `AUTH_OPTIONAL_SKILLS` | tool call (binary) | valid token, not identity | scattered across transports — see salesagent-3c4m |
+| Tenant isolation (`tenant_id`) | data rows | tenant | yes, pervasive |
+| `Product.allowed_principal_ids` | product visibility in `get_products` | buyer principal | **no — read-path only.** `create_media_buy` never re-checks it. **Filed as salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849), P0.** |
+| Admin UI `User.role` (`admin`/`manager`/`viewer`) | admin routes | human user + tenant | **no** — two correct decorators exist (`require_auth(admin_only=...)`, `require_tenant_access()` in `src/admin/utils/helpers.py:259,299`), but several blueprints hand-roll their own inline `session.get("role")` check instead: `src/admin/blueprints/policy.py:29,146,149,226`, `inventory.py:562`, `core.py:273,277`. Same "one correct pattern, several hand-rolled duplicates" disease as the transport auth-policy scatter. **Not yet filed as a ticket — see §3.** |
+| `SUPER_ADMIN_DOMAINS`/`_EMAILS` | cross-tenant admin access | human email domain | yes, routed through the same admin_only checks |
+| `WorkflowStep.assigned_to` | — | — | **not a permission at all** — a queue label; nothing restricts who can act on an assigned step |
+| Webhook HMAC signing | outbound message integrity | — | different category (authenticity, not access control) |
+| GAM adapter | — | — | **no native concept** — flattened to "tenant's service account can do whatever GAM allows" |
+| Creative approval | — | status transition, not identity | any principal owning the buy can act; not actor-gated |
+| `Account.account_scope` | — | — | pass-through AdCP spec field, not internally enforced |
+| `specialisms` (`get_adcp_capabilities`) | — | seller/tenant (declared, not buyer) | **no — purely descriptive.** See §2. |
+
+## 2. `specialisms` is a self-report, not a gate
+
+`specialisms` is a seller-wide (tenant-axis, not buyer-axis) array declared
+once — e.g. `src/core/tools/capabilities.py:100,272` hardcodes
+`specialisms=[AdcpSpecialism.sales_non_guaranteed]`. Grepped all of `src/`
+for every specialism value outside those two declaration sites — zero
+results. It's written once, serialized into the capabilities response, and
+never read anywhere else in the codebase.
+
+Confirmed concretely: whether a buy requires manual approval is decided by
+`tenant_approval_required or adapter_approval_required`
+(`media_buy_create.py:2707-2709`) — completely disconnected from
+`specialisms`. A tenant can declare only `sales-non-guaranteed` while its
+config forces every buy through guaranteed-style approval anyway, and
+nothing catches the divergence. `sync_governance`/`check_governance` don't
+even exist as tools in this codebase, so the `governance-*` specialisms
+aren't just unenforced — there's nothing to enforce.
+
+**Risk**: a tenant's declared specialisms and its actual runtime behavior
+can silently diverge. Low urgency (no financial exposure, unlike
+salesagent-7kwq) but a real compliance-integrity gap — the AAO compliance
+badge is only as honest as this self-report.
+
+## 3. Per-tool ACL by principal: doesn't exist, and isn't spec-mandated
+
+Confirmed: `ResolvedIdentity` (`src/core/resolved_identity.py:24-44`) carries
+only `principal_id`, `tenant_id`, `tenant`, `auth_token`, `protocol` — no
+roles or scopes. Once a token is valid, that principal can call every tool
+any other valid principal can call. There is no mechanism to vary
+tool/skill *access* by caller.
+
+But this is **not spec-mandated**. Read the pinned 3.1 schema
+(`get-adcp-capabilities-response.json`) directly: `capabilities` is a single
+global declaration, identical to every caller, no per-principal field
+anywhere. AdCP's actual per-buyer variance mechanism is **data-catalog
+filtering** (`get_products` returning a different set per buyer via
+`allowed_principal_ids`), not tool-access gating — and the codebase already
+implements that correctly in principle (product.py's filter), just doesn't
+enforce it past the read path (§1, salesagent-7kwq).
+
+### Choke-point mapping, per transport
+
+- **MCP: one already exists.** `mcp.add_middleware(MCPAuthMiddleware())`
+  (`src/core/main.py:174`) runs `on_call_tool` before every tool call — a
+  per-principal-per-tool check could slot in here with no restructuring.
+- **A2A: partial.** One dispatcher (`_handle_explicit_skill`), but auth is
+  checked at ~10 independent raise sites feeding it, only one of which
+  builds the correct two-layer envelope (see salesagent-3c4m, R3-4/R3-32).
+- **REST: none.** 13 routes each independently pick `require_auth` vs
+  `resolve_auth` as a per-route `Depends` default
+  (`src/routes/api_v1.py:224-486`) — no shared dependency all routes pass
+  through.
+
+### Target shape (if this is ever built)
+
+One typed `AuthPolicy` declared once per skill, consumed through a single
+accessor by all three transports — REST derives its route dependency from
+it instead of a hand-typed subset map, A2A's wire-translation seam is fixed
+so no raise site can bypass the envelope. This is the same fix already
+scoped for salesagent-3c4m (the AuthPolicy centralization epic) — a real
+per-buyer *tool* ACL (as opposed to policy-by-skill) would be an **additive
+axis on top of that**, not a replacement, and would need REST's missing
+choke point built regardless of whether the ACL layer itself gets built.
+
+**Not currently scheduled.** Legitimate future improvement, not required by
+spec, not blocking anything currently in flight.
+
+## 4. Tracking
+
+- salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849) — P0, the confirmed exploitable gap (§1, `allowed_principal_ids`).
+- salesagent-3c4m — AuthPolicy centralization epic; the choke-point/target-shape work in §3 is additive on top of it, not part of it yet.
+- salesagent-0krw — epic for this note. Children: salesagent-0krw.1 (specialisms non-enforcement, §2), salesagent-0krw.2 (admin RBAC duplicated checks, §1). Separate from the PR #1838 lineage (salesagent-1zq3 and its 4 successors) since neither finding came from that review.
--- a/.claude/notes/acl-authorization-architecture.md
+++ b/.claude/notes/acl-authorization-architecture.md
@@ -12,8 +12,8 @@ anywhere, or purely descriptive.
 |---|---|---|---|
 | `AUTH_OPTIONAL_SKILLS` | tool call (binary) | valid token, not identity | scattered across transports — see salesagent-3c4m |
 | Tenant isolation (`tenant_id`) | data rows | tenant | yes, pervasive |
-| `Product.allowed_principal_ids` | product visibility in `get_products` | buyer principal | **no — read-path only.** `create_media_buy` never re-checks it. **Filed as salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849), P0.** |
-| Admin UI `User.role` (`admin`/`manager`/`viewer`) | admin routes | human user + tenant | **no** — two correct decorators exist (`require_auth(admin_only=...)`, `require_tenant_access()` in `src/admin/utils/helpers.py:259,299`), but several blueprints hand-roll their own inline `session.get("role")` check instead: `src/admin/blueprints/policy.py:29,146,149,226`, `inventory.py:562`, `core.py:273,277`. Same "one correct pattern, several hand-rolled duplicates" disease as the transport auth-policy scatter. **Not yet filed as a ticket — see §3.** |
+| `Product.allowed_principal_ids` | product visibility in `get_products` | buyer principal | **no — read-path only.** `create_media_buy` never re-checks it. **Filed as salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849), P0.** Not an isolated instance — see §1a: this is a recurring disease, not a single bug. |
+| Admin UI `User.role` (`admin`/`manager`/`viewer`) | admin routes | human user + tenant | **no** — two correct decorators exist (`require_auth(admin_only=...)`, `require_tenant_access()` in `src/admin/utils/helpers.py:259,299`), but several blueprints hand-roll their own inline `session.get("role")` check instead: `src/admin/blueprints/policy.py:29,146,149,226`, `inventory.py:562`, `core.py:273,277`. Same "one correct pattern, several hand-rolled duplicates" disease as the transport auth-policy scatter. **Not yet filed as a ticket — see §3.** Adjacent: **[GH #1805](https://github.com/prebid/salesagent/issues/1805)** — `require_tenant_access()` sits outside `log_admin_action`'s decorator, so a correct denial from the decorator itself goes unaudited. Different bug (audit gap, not a missing check), same decorator-stack neighborhood. |
 | `SUPER_ADMIN_DOMAINS`/`_EMAILS` | cross-tenant admin access | human email domain | yes, routed through the same admin_only checks |
 | `WorkflowStep.assigned_to` | — | — | **not a permission at all** — a queue label; nothing restricts who can act on an assigned step |
 | Webhook HMAC signing | outbound message integrity | — | different category (authenticity, not access control) |
@@ -21,6 +21,35 @@ anywhere, or purely descriptive.
 | Creative approval | — | status transition, not identity | any principal owning the buy can act; not actor-gated |
 | `Account.account_scope` | — | — | pass-through AdCP spec field, not internally enforced |
 | `specialisms` (`get_adcp_capabilities`) | — | seller/tenant (declared, not buyer) | **no — purely descriptive.** See §2. |
+
+## 1a. The `allowed_principal_ids` gap is one instance of a recurring disease
+
+`#1849` is not isolated. The same shape — a scoping check that exists on the
+read path but is never re-checked on a write/action path, surfacing as a 200
+(sometimes with an empty body) instead of a 403 — recurs across resources:
+
+- **[GH #1318](https://github.com/prebid/salesagent/issues/1318)** (2026-05-18,
+  predates this investigation) — `MediaBuyRepository.get_by_principal` filters
+  silently by `principal_id`; a request for a media buy the caller doesn't own
+  returns an empty deliveries list, not a denial. The earliest known instance
+  of this exact pattern.
+- **[GH #1808](https://github.com/prebid/salesagent/issues/1808)** — MCP task
+  tools are tenant-scoped only; a sibling principal can read and terminalize
+  another principal's task. Partially closed by #1812 (`get_task`/
+  `complete_task`); `list_tasks`/`list_by_tenant` still open.
+- **[GH #1702](https://github.com/prebid/salesagent/issues/1702)** — A2A
+  `tasks/get`'s in-memory task-map fast path has no identity check at all
+  (the durable path added by #1544 does check). Same subsystem as #1808, same
+  underlying gap.
+
+None of these three are filed under `salesagent-0krw` — they were each found
+independently, in their own review threads, before this note existed. Listed
+here because §3's root cause (`ResolvedIdentity` carries no scoping beyond
+`tenant_id`; nothing enforces per-principal ownership on non-read paths) is
+exactly what all four bugs share. A structural fix — one ownership-check
+helper every mutating/task-reading path calls, instead of each repository
+method re-deriving its own scoping — would close all four at once instead of
+four independent patches. Not scoped as its own epic yet.
 
 ## 2. `specialisms` is a self-report, not a gate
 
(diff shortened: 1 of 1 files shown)


.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+questions drove it: (1) can this codebase check whether a calling party has
+access to a specific tool, and vary tool access by caller — the AdCP-ACL
+question; (2) is `get_adcp_capabilities`' `specialisms` field enforced
+anywhere, or purely descriptive.
+
+## 1. What exists today
+
+| Mechanism | Gates | Varies by | Enforced consistently? |
+|---|---|---|---|
+| `AUTH_OPTIONAL_SKILLS` | tool call (binary) | valid token, not identity | scattered across transports — see salesagent-3c4m |
```
ChrisHuie, line 13: `AUTH_OPTIONAL_SKILLS` does not exist at main — a grep for it returns nothing, so this row is a dead end for whoever picks the work up. The real declarations are `AUTH_OPTIONAL_TOOLS` (`src/core/mcp_auth_middleware.py:20`, consumed at `:45`) and, on the A2A side, `DISCOVERY_SKILLS` (`src/a2a_server/adcp_a2a_server.py:181`, consumed at `:1465`). The symbol does exist on the unmerged commit `1236ded5c`, so this row appears to describe that branch rather than main.

The "scattered across transports" point is right, and it is better served by naming all three declarations of the one policy — the two frozensets plus REST's per-route choice — because they currently disagree: `list_accounts` is exempt in both frozensets but uses `require_auth` at `src/routes/api_v1.py:475`.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+| `Product.allowed_principal_ids` | product visibility in `get_products` | buyer principal | **no — read-path only.** `create_media_buy` never re-checks it. **Filed as salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849), P0.** Not an isolated instance — see §1a: this is a recurring disease, not a single bug. |
+| Admin UI `User.role` (`admin`/`manager`/`viewer`) | admin routes | human user + tenant | **no** — two correct decorators exist (`require_auth(admin_only=...)`, `require_tenant_access()` in `src/admin/utils/helpers.py:259,299`), but several blueprints hand-roll their own inline `session.get("role")` check instead: `src/admin/blueprints/policy.py:29,146,149,226`, `inventory.py:562`, `core.py:273,277`. Same "one correct pattern, several hand-rolled duplicates" disease as the transport auth-policy scatter. **Not yet filed as a ticket — see §3.** Adjacent: **[GH #1805](https://github.com/prebid/salesagent/issues/1805)** — `require_tenant_access()` sits outside `log_admin_action`'s decorator, so a correct denial from the decorator itself goes unaudited. Different bug (audit gap, not a missing check), same decorator-stack neighborhood. |
```
ChrisHuie, line 16: Two problems here, and they are one defect: neither cited decorator implements `User.role` RBAC, so the inline checks are not "hand-rolled duplicates" of them.

`require_auth` (`src/admin/utils/helpers.py:259-297`) gates only on `is_super_admin(email)`; `require_tenant_access` gates tenant membership and is role-blind — an active `viewer` row passes it. `User.role` reaches no authorization decision anywhere: it flows into `session["is_tenant_admin"]`, which is written (`auth.py:656,719`) and never read, and `is_tenant_admin()` (`helpers.py:223`) has no call sites.

So the 20 inline `session.get("role")` checks are a weaker, separate mechanism. `session["role"]` is written at only three sites (`auth.py:550`, `auth.py:851`, `public.py:291`) and `"viewer"` at none of them, so the eight `role == "viewer"` denials are unreachable on every path.

The list is also 7 of 20. Missing: `policy.py:26,223`, `inventory.py:345,517,559`, `core.py:313,519`, plus two files absent entirely — `gam.py:135,348,711,763,848` and `operations.py:66`. Since §4 tracks this row as a child item, the list becomes the remediation scope; consider citing the count and the regeneration command instead of a list that reads as exhaustive.

Separately, `core.py:273,277` are role-based view dispatch on an unparameterized root route, not decorator substitutes — `require_tenant_access` wraps as `decorated_function(tenant_id, ...)` and structurally requires a `tenant_id` argument those two handlers do not have.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+| `SUPER_ADMIN_DOMAINS`/`_EMAILS` | cross-tenant admin access | human email domain | yes, routed through the same admin_only checks |
```
ChrisHuie, line 17: "yes, routed through the same admin_only checks" does not hold — there are two independent super-admin authorities on two different environment variables, with no code bridging them.

`is_super_admin()` reads the **plural** `SUPER_ADMIN_DOMAINS` (`src/admin/utils/helpers.py:169`, `src/core/config_loader.py:321`) and backs `require_auth(admin_only=True)`. Separately `auth.py:547-550` grants `session["role"] = "super_admin"` from `get_super_admin_domain()`, which reads the **singular** `SUPER_ADMIN_DOMAIN` (`src/core/domain_config.py:60`).

Under a singular-only config the two disagree: `is_super_admin()` is False (so the decorator denies) while every inline `role == "super_admin"` check passes. The domain-hijack guards at `src/admin/utils/domain_access.py:270,362` also key on the singular variable only.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+| Webhook HMAC signing | outbound message integrity | — | different category (authenticity, not access control) |
+| GAM adapter | — | — | **no native concept** — flattened to "tenant's service account can do whatever GAM allows" |
+| Creative approval | — | status transition, not identity | any principal owning the buy can act; not actor-gated |
+| `Account.account_scope` | — | — | pass-through AdCP spec field, not internally enforced |
+| `specialisms` (`get_adcp_capabilities`) | — | seller/tenant (declared, not buyer) | **no — purely descriptive.** See §2. |
+
+## 1a. The `allowed_principal_ids` gap is one instance of a recurring disease
+
+`#1849` is not isolated. The same shape — a scoping check that exists on the
+read path but is never re-checked on a write/action path, surfacing as a 200
```
ChrisHuie, line 28: This stated shape describes only #1849 of the four issues grouped below it. #1318 has no write path — it is a read path returning an empty list. #1702 has no check on either path. #1808 has no principal scoping anywhere, rather than a read-path check that is not re-checked.

A formulation that does fit all four: per-principal scoping implemented as a silent data filter, or omitted entirely, rather than an explicit authorization decision. The same over-narrow wording recurs at line 48 ("on non-read paths"), while two of the four members are read-path bugs.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+(sometimes with an empty body) instead of a 403 — recurs across resources:
+
+- **[GH #1318](https://github.com/prebid/salesagent/issues/1318)** (2026-05-18,
+  predates this investigation) — `MediaBuyRepository.get_by_principal` filters
+  silently by `principal_id`; a request for a media buy the caller doesn't own
+  returns an empty deliveries list, not a denial. The earliest known instance
+  of this exact pattern.
+- **[GH #1808](https://github.com/prebid/salesagent/issues/1808)** — MCP task
+  tools are tenant-scoped only; a sibling principal can read and terminalize
+  another principal's task. Partially closed by #1812 (`get_task`/
```
ChrisHuie, line 38: "Partially closed by #1812" is not accurate in two ways.

#1812 is `state=open`, `merged=false`, `mergeable_state=blocked` — it was open when this note was written and still is, so past tense reads as landed code.

Beyond merge state, the substance does not hold either: at main none of `get_task`, `complete_task`, `list_tasks`, or `WorkflowRepository.list_by_tenant` is principal-scoped. `get_task` (`src/core/tools/task_management.py:140-147`) runs `require_identity` → `require_tenant` → `require_principal_id` and then `get_by_step_id_or_raise(task_id)`; `require_principal_id` is a binary auth guard, not scoping, and `task.owner` is serialized at `:157` but never compared. `complete_task` (`:205-239`) will terminalize any task in the tenant.

Suggest: "Proposed in the open PR #1812; nothing has landed. At main all four are tenant-scoped only."

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+  `complete_task`); `list_tasks`/`list_by_tenant` still open.
+- **[GH #1702](https://github.com/prebid/salesagent/issues/1702)** — A2A
+  `tasks/get`'s in-memory task-map fast path has no identity check at all
+  (the durable path added by #1544 does check). Same subsystem as #1808, same
```
ChrisHuie, line 42: The parenthetical is not true at main. #1544 is `state=open`, `merged=false`, `mergeable_state=blocked`, and `resolve_durable_task_outcome` has zero hits in `src/` — there is no durable task read in the A2A server today, so there is no path that "does check".

The first half is confirmed: `on_get_task` → `_get_task_or_raise` → `self.tasks.get(task_id)` (`src/a2a_server/adcp_a2a_server.py:1042-1057`) with no identity; `_resolve_a2a_identity` is never called from that handler.

Worth adding while you are here: `on_cancel_task` (`:1059-1075`) has the same gap on a path that **mutates** state to `TASK_STATE_CANCELED`, so the read-only framing understates it.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+  underlying gap.
+
+None of these three are filed under `salesagent-0krw` — they were each found
+independently, in their own review threads, before this note existed. Listed
+here because §3's root cause (`ResolvedIdentity` carries no scoping beyond
+`tenant_id`; nothing enforces per-principal ownership on non-read paths) is
+exactly what all four bugs share. A structural fix — one ownership-check
```
ChrisHuie, line 49: This recommendation proposes building something the codebase already has twice, and both raise rather than filter:

- `_verify_principal` — `src/core/tools/media_buy_update.py:299`, called from `media_buy_update.py:403` and `performance.py:68` (also documented at `src/adapters/SECURITY_OVERVIEW.md:54`)
- `_require_account_access` — `src/core/helpers/account_helpers.py:98`, called at `:128` and `:182`

So the gap is adoption, not absence — which changes the work from "design a helper" to "extend an existing one and enumerate the paths that should call it".

The four also do not collapse to one remedy. #1808 fits the ownership shape. #1849 needs an allowlist-membership predicate at the create-path product lookup (`media_buy_create.py:2307-2312`) — there is no ownership relation to check. #1318 already scopes by `principal_id` in SQL (`src/core/database/repositories/media_buy.py:138-141`), so the change is filter-to-decision, which alters wire behavior and needs a partial-overlap policy call. #1702 must first resolve identity into `on_get_task` and give the in-memory map an owner dimension.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+four independent patches. Not scoped as its own epic yet.
+
+## 2. `specialisms` is a self-report, not a gate
+
+`specialisms` is a seller-wide (tenant-axis, not buyer-axis) array declared
+once — e.g. `src/core/tools/capabilities.py:100,272` hardcodes
+`specialisms=[AdcpSpecialism.sales_non_guaranteed]`. Grepped all of `src/`
+for every specialism value outside those two declaration sites — zero
+results. It's written once, serialized into the capabilities response, and
+never read anywhere else in the codebase.
```
ChrisHuie, line 61: "never read anywhere else in the codebase" is falsified by the test surface: `tests/unit/test_get_adcp_capabilities.py:179,198,247` reads it, and `:199` pins the wire value `data["specialisms"] == ["sales-non-guaranteed"]`. There is also a specialisms scenario in `tests/bdd/features/BR-UC-010-discover-seller-capabilities.feature` which is currently unbound (no `scenarios()` reference, no matching step defs).

The production-path conclusion holds — no production code reads it. Scoping the sentence to that ("no production code reads it") keeps the claim true, and the grep as stated covers `src/`, not the codebase.

Externally the field is load-bearing: per the pinned schema each specialism routes a storyboard bundle the compliance runner executes.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+
+Confirmed concretely: whether a buy requires manual approval is decided by
+`tenant_approval_required or adapter_approval_required`
+(`media_buy_create.py:2707-2709`) — completely disconnected from
```
ChrisHuie, line 65: Two refinements. The cited range should be `2706-2709` — `:2706` is where the left operand `tenant_approval_required` is defined; `:2709` alone is the assignment.

More substantively, this is not the whole decision. The flag is gated at `:2726` in combination with `not testing_ctx.dry_run` and `"create_media_buy" in manual_approval_operations`, and there is a second independent gate at `:3170` that routes to `requires_approval` when tenant or product auto-create is disabled. So a buy can require manual approval while `manual_approval_required` is False — worth saying, since the current text models a single two-operand decision.

The disconnection from `specialisms` is confirmed on both gates.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+`specialisms`. A tenant can declare only `sales-non-guaranteed` while its
+config forces every buy through guaranteed-style approval anyway, and
+nothing catches the divergence. `sync_governance`/`check_governance` don't
+even exist as tools in this codebase, so the `governance-*` specialisms
```
ChrisHuie, line 69: Both halves are true — neither tool is among the 16 registered at `src/core/main.py:351-366`, and `governance-*` specialisms do exist in the pin (3 of 22, each with a storyboard bundle). The inference is off in two ways though.

The more basic point is that this tenant declares no `governance-*` specialism at all, so there is no claim to diverge from — the missing tools are downstream of that.

And `governance-aware-seller` is not a governance-agent-only concern: in the pinned compliance tree it is `protocol: media-buy`, seller-side, with `required_tools: [sync_governance, create_media_buy]` — i.e. directly applicable to this agent. Also worth noting `governance_agents` binding data already flows through `sync_accounts` (`src/core/tools/accounts.py:255-269,629`).

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+aren't just unenforced — there's nothing to enforce.
+
+**Risk**: a tenant's declared specialisms and its actual runtime behavior
+can silently diverge. Low urgency (no financial exposure, unlike
+salesagent-7kwq) but a real compliance-integrity gap — the AAO compliance
```
ChrisHuie, line 74: The mechanism is the inverse of "only as honest as this self-report" — the badge is not taken on trust. Declared specialisms are the routing field: each maps to a storyboard bundle the compliance runner executes, and the badge issues only when they pass, with a degraded→revoked path on sustained failure.

So over-declaration is caught at the runner. The silent case is the opposite: a specialism you implement but omit is never graded ("No applicable tracks found"). The concrete local gap is that we declare `sales-non-guaranteed` while two of its bundled scenarios are not green and no required CI job runs the storyboards — which is a stronger version of the point being made here.

Terminology: the mark is "AAO Verified".

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+badge is only as honest as this self-report.
+
+## 3. Per-tool ACL by principal: doesn't exist, and isn't spec-mandated
+
+Confirmed: `ResolvedIdentity` (`src/core/resolved_identity.py:24-44`) carries
+only `principal_id`, `tenant_id`, `tenant`, `auth_token`, `protocol` — no
```
ChrisHuie, line 80: `ResolvedIdentity` carries seven fields, not five. The two omitted are `testing_context` (`:36`) and `account_id` (`:37`), both predating this branch.

`account_id` matters to the argument: it is scope-shaped and is consumed as an idempotency scope key at `media_buy_create.py:2104`, which cuts against the §1a root-cause sentence that identity "carries no scoping beyond `tenant_id`". Suggest listing all seven and distinguishing "no roles and no capability scopes" from "no scoping at all" — the former is what the code supports.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+roles or scopes. Once a token is valid, that principal can call every tool
+any other valid principal can call. There is no mechanism to vary
+tool/skill *access* by caller.
+
+But this is **not spec-mandated**. Read the pinned 3.1 schema
+(`get-adcp-capabilities-response.json`) directly: `capabilities` is a single
+global declaration, identical to every caller, no per-principal field
+anywhere. AdCP's actual per-buyer variance mechanism is **data-catalog
```
ChrisHuie, line 88: The premise is right but the conclusion does not follow, and this is the one I would most want changed, because the section will be cited to decline exactly the work the spec standardizes.

AdCP 3.1.1 does define a per-caller, per-task ACL — just not on the surface read here. The pinned SDK ships `3.1/core/account-authorization.json` with `allowed_tasks`, `field_scopes`, `scope_name`, `read_only`, and this on `allowed_tasks`: "Absence of a task from this list MUST be interpreted as 'not permitted' — invoking an absent task MUST return SCOPE_INSUFFICIENT." The capabilities document says so itself — "Per-caller authorization is NOT reported here" — and points to the `authorization` object on `sync_accounts` / `list_accounts` entries.

The refutation is already vendored here: `tests/fixtures/adcp_schemas_pinned/enums/error-code.json:56-58` defines `SCOPE_INSUFFICIENT`, `READ_ONLY_SCOPE` and `FIELD_NOT_PERMITTED`, and `:122` describes the first as "the task is not in the caller's `allowed_tasks` for this account".

Also, `allowed_principal_ids` appears in **0** pinned schema files — it is internal (`src/core/schemas/product.py:128`), so it should not be described as "AdCP's actual per-buyer variance mechanism". The spec's term for the catalog-variance behavior is the "account/caller authorization predicate" on `get_products`.

Populating `authorization` is not required for 3.x, so conformance today is fine by omission — but "not spec-mandated" is the wrong reason. Worth noting our `list_accounts` could not carry it as written: `src/core/schemas/account.py` narrows the SDK's `list[AccountWithAuthorization]` to `list[Account]`, and none of the three codes is emitted anywhere in `src/`.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+implements that correctly in principle (product.py's filter), just doesn't
+enforce it past the read path (§1, salesagent-7kwq).
+
+### Choke-point mapping, per transport
+
+- **MCP: one already exists.** `mcp.add_middleware(MCPAuthMiddleware())`
+  (`src/core/main.py:174`) runs `on_call_tool` before every tool call — a
+  per-principal-per-tool check could slot in here with no restructuring.
+- **A2A: partial.** One dispatcher (`_handle_explicit_skill`), but auth is
+  checked at ~10 independent raise sites feeding it, only one of which
```
ChrisHuie, line 100: The two numbers are right — there are exactly 10 auth raise sites and exactly one builds the two-layer envelope (`:587`, via `build_two_layer_error_envelope` at `:589`) — but "independent raise sites feeding it" misdescribes most of them.

Four are centralized in one helper, `_resolve_a2a_identity` (`:308,322,326,329`), so they are not independent. And four others (`:1107,1170,1251,1309`, the `tasks/pushNotificationConfig/*` handlers) never reach the dispatcher at all. Those four are verbatim copies of the same missing-token raise, so they are the duplicated shape a centralization effort would actually consolidate — which is a more useful thing for this paragraph to point at.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+  builds the correct two-layer envelope (see salesagent-3c4m, R3-4/R3-32).
+- **REST: none.** 13 routes each independently pick `require_auth` vs
+  `resolve_auth` as a per-route `Depends` default
```
ChrisHuie, line 103: The count is 12, not 13 — `grep -cE '^@router\.(get|post|put|patch|delete)'` on `src/routes/api_v1.py` returns 12 at both main and this branch's base (4 `resolve_auth`, 8 `require_auth`). The `224-486` range is right, and "no shared dependency all routes pass through" is confirmed: the `APIRouter` at `:50` declares no `dependencies=`.

One qualification on "REST: none" — `UnifiedAuthMiddleware` (`src/core/auth_middleware.py:23`, registered `src/app.py:480`) is an app-wide ASGI middleware every request traverses. It extracts the token into `AuthContext` without resolving identity, so the seam exists and carries no identity today. "Extend that middleware" rather than "build a choke point" is the smaller change.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+per-buyer *tool* ACL (as opposed to policy-by-skill) would be an **additive
+axis on top of that**, not a replacement, and would need REST's missing
+choke point built regardless of whether the ACL layer itself gets built.
+
+**Not currently scheduled.** Legitimate future improvement, not required by
+spec, not blocking anything currently in flight.
+
+## 4. Tracking
+
+- salesagent-7kwq / [GH #1849](https://github.com/prebid/salesagent/issues/1849) — P0, the confirmed exploitable gap (§1, `allowed_principal_ids`).
```
ChrisHuie, line 123: The `salesagent-*` identifiers do not resolve from this repository, which makes §4 — the section a reader goes to for next steps — non-actionable for anyone else.

`.beads/issues.jsonl` is tracked and holds 547 records, all with 3-character base suffixes. The six ids cited in this file (`0krw`, `7kwq`, `3c4m`, `1zq3`, `0krw.1`, `0krw.2`) are 4-character and match nothing there, nor in `.beads/beads.db`. Three of the five rows in this section have no GitHub fallback, and §3's entire target shape hangs off `3c4m`.

Either give those rows GitHub-resolvable ids, or state once that they belong to a private tracker so a reader stops looking.

On this line specifically: #1849 carries only a `bug` label and no priority marker, and its exploit path is derived from the code rather than reproduced — so "P0, the confirmed exploitable gap" overstates what the issue records. #1318's title does carry `[P1 security]`, so a reader will look for the equivalent marker and not find one.

ChrisHuie (commented):
Reviewed at `1cc18d57c` against current `main` (`eccc45a76`). The branch is 8 commits behind, so claims were graded on current main rather than this PR's base. Graded as a factual artifact: 15 `path:line` citations, 4 categorical greps, 12 issue-state claims, one spec reading. Pin derived at runtime: spec 3.1.1, SDK adcp 6.6.0.

The `allowed_principal_ids` finding in §1 holds on both halves — the read filter at `src/core/tools/products.py:361-393` and the unguarded create-path lookup at `src/core/tools/media_buy_create.py:2307-2312`. The MCP choke-point claim at §3 is also accurate (`src/core/main.py:174`; all 16 tools register through `_register_tool`, and the exempt-tool set parameterizes `require_valid_token` rather than bypassing the middleware). File hygiene has no issues. Comments are inline; the two that would change what a reader does next are §3's spec conclusion and §1a's structural-fix recommendation.

**CI first, since it is not your content.** The three failures (E2E, Security Audit, pip-audit) are stale-base only. `main` bumped `aiohttp>=3.14.0→>=3.14.3` (CVE-2026-59881/69243/69244) and `cryptography>=46.0.7→>=50.0.0` (CVE-2026-69247) after this branch's merge-base; pip-audit's "4 known vulnerabilities in 2 packages" matches those two packages exactly, and #1966/#1950/#1941 pass all three. A merge of `main` should clear it with no content change, and that is also what `BLOCKED` is reporting.

**Issue-state sweep.** Every issue this note lists as open is still open — no drift there. Two references do need changing: #1812 and #1544 are both open and unmerged but are credited in the past tense (inline). #1838 is now closed unmerged, superseded by #1868 which merged 2026-08-11; this file never mentions #1868, so the lineage as written leads to a never-merged PR.

**A few smaller things not worth their own thread:**

- Line 22, `Account.account_scope` — "pass-through" does not describe it: it is never written, the column is always NULL, and it is omitted from the `DBAccount(...)` kwargs at `src/core/tools/accounts.py:619-631`. The spec also makes it a seller-assigned result the agent does not choose, which is an odd fit for a column graded on "Enforced consistently?". The real shortfall is that we never assign it.
- Line 34, "the earliest known instance" cannot be checked by a reader. 292 pre-#1318 issue bodies scanned without finding an earlier one, so "the earliest instance filed on GitHub" would be both true and verifiable. Closest prior art is #1128, which #1808 classifies as a different vulnerability class.
- Line 91, "product.py's filter" is ambiguous — the filter is in `src/core/tools/products.py`, and `src/core/schemas/product.py` also exists and does not contain it.
- Line 82, "every tool any other valid principal can call" is false cross-tenant: `brand_manifest_policy` varies `get_products` reachability (`src/core/tools/products.py:197`). Scoping it to same-tenant would make it exact.
- §1a reads as an exhaustive list of the recurrence; `CreativeRepository.get_by_principal` (`src/core/database/repositories/creative.py:99-124`) is a byte-for-byte sibling of the #1318 shape and is not in it.
- §1's inventory omits a live per-principal mechanism: `AgentAccountAccess` (`src/core/database/models.py:869-893`) with `grant_access()` (`src/core/database/repositories/account.py:252-266`), called from `src/core/tools/accounts.py:636`. That is the substrate the spec's `authorization` object attaches to, which is partly why §3's "no mechanism" reads broader than intended.
- Line 56, `capabilities.py:100,272` has already drifted to `101,273` on main — #1868 landed an import above it, 13 days after writing.

**Two structural suggestions.** That last drift is worth generalizing: nothing validates this file's citations. `check-docs-links` matches `^(docs/.*\.md|[^/]+\.md)$`, so `.claude/notes/` is out of scope, and it skips external `http(s)` links regardless; `path:line` references are not links in any case. Since the note already dates itself, naming the commit it was verified against would let a later reader tell drift from error.

Second, it would be worth reconciling with the two existing docs on this ground: `docs/test-obligations/business-rules.md:62` already owns the `allowed_principal_ids` visibility invariant, and `docs/security.md:127-129` states principals "cannot access other tenants or principals", which §1a contradicts for non-read paths. Right now a reader can find all three and not know which is current.

**Not checked:** no tests, type check, or suite run — this is a source-and-issue-state reading. Compliance storyboards for the capabilities/ACL surface were not read, so whether any step grades `specialisms`, `authorization` or `account_scope` is unverified. The unbound BDD scenarios above were determined statically. Post-3.1.1 spec state was not graded. Issue and PR states were pulled 2026-08-18 and should be re-checked when the corrections are applied.

.claude/notes/acl-authorization-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+
+Not access-control-decision bugs — the underlying auth/ACL logic here is
+correct; the defect is in how a decision gets translated to a wire response.
+Listed because they surfaced from the same general area (auth path
+correctness) and are easy to conflate with the ACL-enforcement findings
+above:
+
+- [GH #1859](https://github.com/prebid/salesagent/issues/1859) — unauthenticated MCP/A2A calls to protected tools return HTTP 200 instead of 401 (transport never maps the correctly-raised `AdCPAuthenticationError` to a status code).
+- [GH #1861](https://github.com/prebid/salesagent/issues/1861) — DB/infra errors get caught and misclassified as auth-deny decisions; `google_callback` fails open on one such path.
+- [GH #1743](https://github.com/prebid/salesagent/issues/1743) — a rejected credential is classified `AUTH_MISSING` (retryable) instead of `AUTH_INVALID` (terminal), telling the agent to retry a revoked token.
```
ChrisHuie, line 139: This drops the mitigation the issue itself carries. #1743 argues the behavior is deliberate, documented and pinned across all three transports, that the pinned `enumDescription` scopes both codes to the `Authorization` header, and that `tests/integration/test_auth_suggestion_parity.py` asserts the current behavior — framing it as a spec-reading question about a credential header the spec does not define, not a confirmed divergence. #1859 and #1861 need no such qualifier; this one does.

Also, `AUTH_MISSING` is `correctable` in this repo's recovery classes, not "retryable".

PR #1899, opened 2026-08-10
fix: serialize MCP structured_content via model_dump across remaining tool wrappers

numarasSigmaSoftware:
## Summary

`fastmcp.tools.tool.ToolResult` passes `structured_content` through `pydantic_core.to_jsonable_python()`, which bypasses `AdCPBaseModel`'s `exclude_none=True` default entirely. A2A and REST call `.model_dump()` explicitly and correctly omit unset optional fields; any MCP wrapper that passed the raw response model straight to `ToolResult` instead leaked every unset optional field as an explicit `null` on MCP only — a real cross-transport wire-shape inconsistency, not a cosmetic issue.

This mirrors the same fix already applied to `get_products` (#1266) and `get_media_buy_delivery` (#1575), extended to the remaining wrappers that still passed the raw model.

- `list_accounts`, `sync_accounts` (`src/core/tools/accounts.py`)
- `get_adcp_capabilities` (`src/core/tools/capabilities.py`)
- `list_creative_formats` (`src/core/tools/creative_formats.py`)
- `list_creatives` (`src/core/tools/creatives/listing.py`)
- `sync_creatives` (`src/core/tools/creatives/sync_wrappers.py`)
- `get_media_buys` (`src/core/tools/media_buy_list.py`)
- `update_performance_index` (`src/core/tools/performance.py`)
- `list_authorized_properties` (`src/core/tools/properties.py`)
- `get_signals`, `activate_signal` (`src/core/tools/signals.py`)

Each site was empirically verified before fixing — constructed a minimal real instance of the actual response type and diffed `pydantic_core.to_jsonable_python()` output against `.model_dump(mode="json")`. The worst case (`get_adcp_capabilities`) leaked 30 explicit-null fields on MCP that A2A/REST correctly omit.

**`create_media_buy` / `update_media_buy` were checked and intentionally left unchanged.** Both return a `TaskResultEnvelope` subclass (`CreateMediaBuyResult` / `UpdateMediaBuyResult`) whose `@model_serializer(mode="wrap")` is baked into the model's core schema and internally calls `.model_dump()` on the nested domain response regardless of how the outer object is serialized — so raw-pass already produces byte-identical output to an explicit `model_dump()` call. Verified empirically (including the `replayed=True` variant), not just by inspection.

## Test plan

- [x] Added `tests/unit/test_mcp_wrapper_structured_content_wire_shape.py` — one test per fixed wrapper, each mocking `_impl` to return a real minimal response with optional fields left unset, asserting those fields are absent from `result.structured_content`
- [x] Mutation-tested one case (reverted the fix, confirmed the test fails with the leaked-null shape; restored, confirmed green)
- [x] `ruff format`, `ruff check`, `mypy` (281 files): pass
- [x] Duplication guard: unchanged baseline
- [x] Full unit suite: 5366 passed, 9 skipped, 26 xfailed, 0 failed


KonstantinMirin:
Heads-up before this goes further — #1868 (open, out of draft, in review) already lands this exact fix. Same root (`ToolResult` → `pydantic_core.to_jsonable_python()` bypassing `AdCPBaseModel.exclude_none`), same one-line fix (`structured_content=response.model_dump(mode="json")`), applied to the same 13 `src/core/tools/*` wrappers with the same call counts — just routed through a helper named `mcp_result` (in `_mcp.py`) instead of `build_mcp_tool_result`. The two will hard-conflict on every one of those files.

Since #1868 is further along and closes the schema-pin issues (#1843/#1710/#1881/#1888/#1889) that surfaced these leaks in the first place, I'd suggest we converge on it and close this as superseded. If there's a wrapper or field here that #1868 misses, call it out and I'll make sure it's covered there. Thanks for chasing the same bug — want to avoid the double effort.

PR #2075, opened 2026-08-25
fix: enforce tenant access on publisher partner routes

gagliathebuilder:
## Summary

Focused authorization fix extracted from issues surfaced during review of #1848.

Adds the existing canonical `@require_tenant_access(api_mode=True)` guard to all five tenant-scoped publisher-partner routes currently on `main`:

- `GET /tenant/<tenant_id>/publisher-partners`
- `POST /tenant/<tenant_id>/publisher-partners`
- `DELETE /tenant/<tenant_id>/publisher-partners/<partner_id>`
- `POST /tenant/<tenant_id>/publisher-partners/sync`
- `GET /tenant/<tenant_id>/publisher-partners/<partner_id>/properties`

This is intentionally a small, standalone subset fix rather than being folded into the larger #1848 channel-disclosure work.

## Behavior

- Anonymous requests receive JSON `401`
- Authenticated users without active membership in the requested tenant receive JSON `403`
- Rejected mutation requests do not alter publisher-partner state
- Active same-tenant members continue to access the routes normally

## Scope

No handler-body changes.
No PATCH route.
No `supported_channels` or other #1848 feature changes.

## Verification

- `tests/admin/test_publisher_partners_auth.py` — 15 passed
- `tests/integration/test_mock_adapter_publisher_sync.py` — 4 passed
- `make quality` — formatting, Ruff, mypy, duplication checks passed
- Unit suite — 5,653 passed, 9 skipped, 26 xfailed

Related: #1848

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/admin/blueprints/publisher_partners.py
+++ b/src/admin/blueprints/publisher_partners.py
@@ -14,6 +14,7 @@ from adcp.exceptions import AdagentsNotFoundError, AdagentsTimeoutError, Adagent
 from flask import Blueprint, Response, jsonify, request
 from sqlalchemy import select
 
+from src.admin.utils import require_tenant_access
 from src.core.config import get_config
 from src.core.database.database_session import get_db_session
 from src.core.database.models import PublisherPartner, Tenant
@@ -25,6 +26,7 @@ publisher_partners_bp = Blueprint("publisher_partners", __name__)
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners", methods=["GET"])
+@require_tenant_access(api_mode=True)
 def list_publisher_partners(tenant_id: str) -> Response | tuple[Response, int]:
     """List all publisher partners for a tenant."""
     try:
@@ -85,6 +87,7 @@ def list_publisher_partners(tenant_id: str) -> Response | tuple[Response, int]:
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners", methods=["POST"])
+@require_tenant_access(api_mode=True)
 def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
     """Add a new publisher partner."""
     try:
@@ -164,6 +167,7 @@ def add_publisher_partner(tenant_id: str) -> Response | tuple[Response, int]:
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners/<int:partner_id>", methods=["DELETE"])
+@require_tenant_access(api_mode=True)
 def delete_publisher_partner(tenant_id: str, partner_id: int) -> Response | tuple[Response, int]:
     """Delete a publisher partner."""
     try:
@@ -185,6 +189,7 @@ def delete_publisher_partner(tenant_id: str, partner_id: int) -> Response | tupl
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners/sync", methods=["POST"])
+@require_tenant_access(api_mode=True)
 def sync_publisher_partners(tenant_id: str) -> Response | tuple[Response, int]:
     """Sync verification status for all publisher partners."""
     try:
@@ -473,6 +478,7 @@ def sync_publisher_partners(tenant_id: str) -> Response | tuple[Response, int]:
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners/<int:partner_id>/properties", methods=["GET"])
+@require_tenant_access(api_mode=True)
 def get_publisher_properties(tenant_id: str, partner_id: int) -> Response | tuple[Response, int]:
     """Get properties for a specific publisher (fetched fresh from adagents.json)."""
     try:
--- a/tests/admin/test_publisher_partners_auth.py
+++ b/tests/admin/test_publisher_partners_auth.py
@@ -1,0 +1,263 @@
+"""Authorization regression tests for publisher-partner admin JSON routes.
+
+All five tenant-scoped routes must return JSON 401/403 from
+``require_tenant_access(api_mode=True)``. Rejected mutations must not
+alter stored partners. Authorized same-tenant members (active ``User``
+row, not the super-admin/test-mode bypass) continue to succeed.
+
+Uses factory-boy factories per ``tests/CLAUDE.md`` — no inline
+``session.add()`` / ``get_db_session()`` in test bodies.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+from adcp.exceptions import AdagentsNotFoundError
+from flask.testing import FlaskClient
+from sqlalchemy import func, select
+from sqlalchemy.orm import Session
+
+from src.admin.app import create_app
+from src.core.database.models import PublisherPartner
+from tests.factories import PublisherPartnerFactory, TenantFactory, UserFactory
+
+app = create_app()
+
+pytestmark = [pytest.mark.admin, pytest.mark.requires_db]
+
+CREATE_PAYLOAD = {"publisher_domain": "unauthorized-add.example", "display_name": "Should Not Persist"}
+
+# (id, method, path suffix relative to /publisher-partners, JSON body or None)
+_ROUTES: list[tuple[str, str, str, dict[str, str] | None]] = [
+    ("list", "GET", "", None),
+    ("create", "POST", "", CREATE_PAYLOAD),
+    ("delete", "DELETE", "/{partner_id}", None),
+    ("sync", "POST", "/sync", None),
+    ("properties", "GET", "/{partner_id}/properties", None),
+]
+
+
+@pytest.fixture
+def client() -> Iterator[FlaskClient]:
+    """Flask test client with CSRF disabled."""
+    app.config["TESTING"] = True
+    app.config["WTF_CSRF_ENABLED"] = False
+    app.config["SESSION_COOKIE_PATH"] = "/"
+    with app.test_client() as test_client:
+        yield test_client
+
+
+def _member_session(client: FlaskClient, tenant_id: str, email: str) -> None:
+    """Authenticated session for a normal tenant member.
+
+    Omits ``test_user`` / ``test_user_role`` so ``require_tenant_access``
+    uses the ``User`` membership lookup instead of the test-mode bypass.
+    """
+    with client.session_transaction() as sess:
+        sess["authenticated"] = True
+        sess["user"] = {"email": email, "is_super_admin": False}
+        sess["email"] = email
+        sess["tenant_id"] = tenant_id
+
+
+def _partners_url(tenant_id: str, suffix: str = "") -> str:
+    return f"/tenant/{tenant_id}/publisher-partners{suffix}"
+
+
+def _dispatch(client: FlaskClient, method: str, url: str, payload: dict[str, str] | None):
+    if method == "GET":
+        return client.get(url)
+    if method == "POST":
+        return client.post(url, json=payload or {})
+    if method == "DELETE":
+        return client.delete(url)
+    raise AssertionError(f"unexpected method {method}")
+
+
+def _partner_count(session: Session, tenant_id: str) -> int:
+    return session.scalar(select(func.count()).select_from(PublisherPartner).filter_by(tenant_id=tenant_id)) or 0
+
+
+def _partner_snapshot(session: Session, partner_id: int) -> tuple[Any, ...] | None:
+    partner = session.get(PublisherPartner, partner_id)
+    if partner is None:
+        return None
+    return (
+        partner.id,
+        partner.tenant_id,
+        partner.publisher_domain,
+        partner.display_name,
+        partner.is_verified,
+        partner.sync_status,
+        partner.sync_error,
+        partner.last_synced_at,
+    )
+
+
+def _seed_target_tenant(factory_session: Session) -> tuple[str, int, int, tuple[Any, ...] | None]:
+    """Create a tenant with one partner. Returns ids plus a stored-data snapshot."""
+    tenant = TenantFactory()
+    partner = PublisherPartnerFactory(tenant=tenant, publisher_domain="existing.example")
+    tenant_id = tenant.tenant_id
+    partner_id = partner.id
+    count = _partner_count(factory_session, tenant_id)
+    snapshot = _partner_snapshot(factory_session, partner_id)
+    return tenant_id, partner_id, count, snapshot
+
+
+def _assert_partners_unchanged(
+    factory_session: Session,
+    tenant_id: str,
+    partner_id: int,
+    count: int,
+    snapshot: tuple[Any, ...] | None,
+) -> None:
+    factory_session.expire_all()
+    assert _partner_count(factory_session, tenant_id) == count
+    assert _partner_snapshot(factory_session, partner_id) == snapshot
+    added = factory_session.scalars(
+        select(PublisherPartner).filter_by(tenant_id=tenant_id, publisher_domain=CREATE_PAYLOAD["publisher_domain"])
+    ).first()
+    assert added is None
+
+
+def _route_url(tenant_id: str, suffix_template: str, partner_id: int) -> str:
+    return _partners_url(tenant_id, suffix_template.format(partner_id=partner_id))
+
+
+class TestPublisherPartnersAnonymousDenied:
+    """Unauthenticated callers receive JSON 401 and cannot mutate partners."""
+
+    @pytest.mark.parametrize("route_id,method,suffix,payload", _ROUTES, ids=[r[0] for r in _ROUTES])
+    def test_anonymous_receives_json_401(
+        self,
+        client: FlaskClient,
+        factory_session: Session,
+        route_id: str,
+        method: str,
+        suffix: str,
+        payload: dict[str, str] | None,
+    ) -> None:
+        tenant_id, partner_id, count, snapshot = _seed_target_tenant(factory_session)
+
+        response = _dispatch(client, method, _route_url(tenant_id, suffix, partner_id), payload)
+
+        assert response.status_code == 401
+        assert response.get_json() == {"error": "Authentication required"}
+        _assert_partners_unchanged(factory_session, tenant_id, partner_id, count, snapshot)
+
+
+class TestPublisherPartnersOtherTenantDenied:
+    """Authenticated members of another tenant receive JSON 403 and cannot mutate."""
+
+    @pytest.mark.parametrize("route_id,method,suffix,payload", _ROUTES, ids=[r[0] for r in _ROUTES])
+    def test_other_tenant_receives_json_403(
+        self,
+        client: FlaskClient,
+        factory_session: Session,
+        route_id: str,
+        method: str,
+        suffix: str,
+        payload: dict[str, str] | None,
+    ) -> None:
+        tenant_id, partner_id, count, snapshot = _seed_target_tenant(factory_session)
+        other_tenant = TenantFactory()
+        other_user = UserFactory(tenant=other_tenant)
+        _member_session(client, other_tenant.tenant_id, other_user.email)
+
+        response = _dispatch(client, method, _route_url(tenant_id, suffix, partner_id), payload)
+
+        assert response.status_code == 403
+        assert response.get_json() == {"error": "Access denied"}
+        _assert_partners_unchanged(factory_session, tenant_id, partner_id, count, snapshot)
+
+
+class TestPublisherPartnersAuthorizedMember:
+    """Active same-tenant User membership is enough; super-admin bypass is not used."""
+
+    def test_get_list_returns_partners(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        partner = PublisherPartnerFactory(tenant=tenant, publisher_domain="listed.example")
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        response = client.get(_partners_url(tenant.tenant_id))
+
+        assert response.status_code == 200
+        body = response.get_json()
+        assert body["total"] == 1
+        assert body["partners"][0]["id"] == partner.id
+        assert body["partners"][0]["publisher_domain"] == "listed.example"
+
+    def test_post_creates_partner(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        response = client.post(
+            _partners_url(tenant.tenant_id),
+            json={"publisher_domain": "created.example", "display_name": "Created"},
+        )
+
+        assert response.status_code == 201
+        body = response.get_json()
+        assert body["publisher_domain"] == "created.example"
+        factory_session.expire_all()
+        stored = factory_session.scalars(
+            select(PublisherPartner).filter_by(tenant_id=tenant.tenant_id, publisher_domain="created.example")
+        ).first()
+        assert stored is not None
+        assert stored.display_name == "Created"
+
+    def test_delete_removes_partner(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        partner = PublisherPartnerFactory(tenant=tenant, publisher_domain="doomed.example")
+        partner_id = partner.id
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        response = client.delete(_partners_url(tenant.tenant_id, f"/{partner_id}"))
+
+        assert response.status_code == 200
+        assert response.get_json() == {"message": "Publisher deleted successfully"}
+        factory_session.expire_all()
+        assert factory_session.get(PublisherPartner, partner_id) is None
+
+    def test_sync_is_not_auth_rejected(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        response = client.post(_partners_url(tenant.tenant_id, "/sync"))
+
+        assert response.status_code not in (401, 403)
+        assert response.status_code == 200
+        assert response.get_json() == {"message": "No publishers to sync"}
+
+    def test_get_properties_is_not_auth_rejected(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        partner = PublisherPartnerFactory(tenant=tenant, publisher_domain="props.example")
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        with (
+            patch(
+                "src.admin.blueprints.publisher_partners.get_tenant_url",
+                return_value="https://agent.example",
+            ),
+            patch(
+                "src.admin.blueprints.publisher_partners.fetch_adagents",
+                side_effect=AdagentsNotFoundError("missing"),
+            ),
+        ):
+            response = client.get(_partners_url(tenant.tenant_id, f"/{partner.id}/properties"))
+
+        assert response.status_code == 200
+        body = response.get_json()
+        assert body["is_authorized"] is False
+        assert body.get("error") != "Authentication required"
+        assert body.get("error") != "Access denied"
--- a/tests/integration/test_mock_adapter_publisher_sync.py
+++ b/tests/integration/test_mock_adapter_publisher_sync.py
@@ -9,6 +9,7 @@ to show empty property/tag lists.
 """
 
 import pytest
+from flask import session
 from sqlalchemy import select
 
 from src.core.database.database_session import get_db_session
@@ -21,9 +22,24 @@ from src.core.database.models import (
 )
 
 
+def _authorize_direct_sync_call(tenant_id: str) -> None:
+    """Populate a super-admin test-mode session for direct view-function calls."""
+    session["authenticated"] = True
+    session["user"] = {"email": "test@example.com", "is_super_admin": True}
+    session["email"] = "test@example.com"
+    session["test_user"] = "test@example.com"
+    session["test_user_role"] = "super_admin"
+    session["test_user_name"] = "Test User"
+    session["test_tenant_id"] = tenant_id
+
+
 @pytest.mark.requires_db
 class TestMockAdapterPublisherSync:
     """Test that mock adapter publisher sync creates properties and tags."""
+
+    @pytest.fixture(autouse=True)
+    def _enable_test_mode(self, monkeypatch):
+        monkeypatch.setenv("ADCP_AUTH_TEST_MODE", "true")
 
     @pytest.fixture
     def mock_tenant(self, integration_db):
@@ -105,6 +121,7 @@ class TestMockAdapterPublisherSync:
         app = create_app()
 
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
             with patch("src.admin.blueprints.publisher_partners.get_config", return_value=mock_config):
                 with patch(
                     "src.admin.blueprints.publisher_partners.get_tenant_url",
@@ -139,6 +156,7 @@ class TestMockAdapterPublisherSync:
         app = create_app()
 
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
             with patch("src.admin.blueprints.publisher_partners.get_config", return_value=mock_config):
                 with patch(
                     "src.admin.blueprints.publisher_partners.get_tenant_url",
@@ -181,6 +199,7 @@ class TestMockAdapterPublisherSync:
         app = create_app()
 
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
             with patch("src.admin.blueprints.publisher_partners.get_config", return_value=mock_config):
                 with patch(
                     "src.admin.blueprints.publisher_partners.get_tenant_url",
@@ -213,6 +232,7 @@ class TestMockAdapterPublisherSync:
 
         # Run sync twice
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
             with patch("src.admin.blueprints.publisher_partners.get_config", return_value=mock_config):
                 with patch(
                     "src.admin.blueprints.publisher_partners.get_tenant_url",


tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -21,10 +22,25 @@
 )
 
 
+def _authorize_direct_sync_call(tenant_id: str) -> None:
```
KonstantinMirin, line 25: This helper writes the production super-admin test-mode session so the direct view-function call gets past the new guard. `src/admin/utils/helpers.py:330-337` returns `f(tenant_id, ...)` on `test_user_role == "super_admin"` **before** the `User` membership lookup this PR adds, so these four tests now pass with authorization disabled on the route the PR secures. Verified: insert an unconditional `return jsonify({"error": "Access denied"}), 403` right after that test-mode branch and `pytest tests/integration/test_mock_adapter_publisher_sync.py` is still 4 passed.

It is also a near-verbatim copy of `tests/harness/admin_accounts.py:165 _auth_integration`, already diverged: `diff <(sed -n '167,175p' tests/harness/admin_accounts.py) <(sed -n '26,33p' tests/integration/test_mock_adapter_publisher_sync.py)` shows the dropped `sess["tenant_id"]`.

Dispatch `POST /tenant/<tenant_id>/publisher-partners/sync` through the Flask test client with an ordinary active-member session (the `_member_session` shape in `tests/admin/test_publisher_partners_auth.py:54`, which never touches `ADCP_AUTH_TEST_MODE`) and delete this helper. Complete sweep, 7 sites: `:12`, `:25-33`, `:40-42`, `:124`, `:159`, `:202`, `:235`.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -21,10 +22,25 @@
...
+    session["test_tenant_id"] = tenant_id
+
+
 @pytest.mark.requires_db
 class TestMockAdapterPublisherSync:
     """Test that mock adapter publisher sync creates properties and tags."""
 
+    @pytest.fixture(autouse=True)
+    def _enable_test_mode(self, monkeypatch):
+        monkeypatch.setenv("ADCP_AUTH_TEST_MODE", "true")
```
KonstantinMirin, line 42: This autouse fixture turns on the production authorization bypass for the whole class. `ADCP_AUTH_TEST_MODE` is live production code, which is why `tests/unit/test_test_auth_production_guard.py` exists to fence it. Delete the fixture once the four tests dispatch through the route with a member session.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -9,6 +9,7 @@
 """
 
 import pytest
+from flask import session
```
KonstantinMirin, line 12: `from flask import session` at module scope is the only occurrence in `tests/` (`grep -rn "^from flask import.*session" tests/`), and it shadows the name that `with get_db_session() as session:` binds at `:49`, `:94`, `:137`, `:172`, `:211`, and `:245` in this same file — `session["authenticated"] = True` and `session.scalars(select(...))` read as if they act on the same object. Delete the import along with `_authorize_direct_sync_call`; set session state through `client.session_transaction() as sess` if anything still needs it.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -105,6 +121,7 @@ def test_sync_creates_property_tag(self, mock_tenant, publisher_partner):
         app = create_app()
 
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
```
KonstantinMirin, line 124: Site 1 of 4. Replace this block with `client.post(f"/tenant/{mock_tenant}/publisher-partners/sync")` on a Flask test client carrying an active-member session, and keep only the assertions. The `patch(...)` targets are module-level and work unchanged under client dispatch.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -139,6 +156,7 @@ def test_sync_creates_authorized_property(self, mock_tenant, publisher_partner):
...
```
KonstantinMirin, line 159: Site 2 of 4 of the same repair: dispatch through the route with a member session instead of calling the view function past the bypass.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -181,6 +199,7 @@ def test_sync_property_has_verified_status(self, mock_tenant, publisher_partner)
...
```
KonstantinMirin, line 202: Site 3 of 4 of the same repair: dispatch through the route with a member session instead of calling the view function past the bypass.

tests/integration/test_mock_adapter_publisher_sync.py
```diff
@@ -213,6 +232,7 @@ def test_sync_is_idempotent(self, mock_tenant, publisher_partner):
 
         # Run sync twice
         with app.test_request_context():
+            _authorize_direct_sync_call(mock_tenant)
```
KonstantinMirin, line 235: Site 4 of 4 of the same repair: dispatch through the route with a member session instead of calling the view function past the bypass. Converting a subset leaves this finding open.

src/admin/blueprints/publisher_partners.py
```diff
@@ -25,6 +26,7 @@
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners", methods=["GET"])
+@require_tenant_access(api_mode=True)
```
KonstantinMirin, line 29: The decorator is the correct seam and this line is right. The problem is that it is repeated at `:29`, `:90`, `:170`, `:192`, and `:481` in a blueprint where every route is `/<tenant_id>/publisher-partners...` with `api_mode=True` and there is no exempt route, so "this route is tenant-scoped" is a fact re-asserted at every site. Append an undecorated PATCH route here and `pytest tests/admin/test_publisher_partners_auth.py tests/unit/test_architecture_workflow_tenant_isolation.py` is still 16 passed — the class the PR exists to close stays open.

Make the blueprint own it: extract the body of `decorated_function` (`src/admin/utils/helpers.py:304-376`) into `check_tenant_access(tenant_id: str, *, api_mode: bool) -> Response | tuple[Response, int] | None`, leave the existing decorator as a two-line wrapper so all 146 adopters and both error literals are unchanged, then add `@publisher_partners_bp.before_request` calling it with `request.view_args["tenant_id"]`. These five decorator lines then come off.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+from tests.factories import PublisherPartnerFactory, TenantFactory, UserFactory
+
+app = create_app()
+
+pytestmark = [pytest.mark.admin, pytest.mark.requires_db]
+
+CREATE_PAYLOAD = {"publisher_domain": "unauthorized-add.example", "display_name": "Should Not Persist"}
+
+# (id, method, path suffix relative to /publisher-partners, JSON body or None)
+_ROUTES: list[tuple[str, str, str, dict[str, str] | None]] = [
```
KonstantinMirin, line 35: `_ROUTES` restates by hand the same five routes the blueprint decorates, so route #6 requires remembering two lists and forgetting this one means nothing fails. Derive it from `app.url_map` (rules whose endpoint starts with `publisher_partners.`) and assert the set is non-empty, so the parametrization grows with the blueprint.

The field meanings also live in the comment on `:34`, which is where a type belongs. Declare `class RouteCase(NamedTuple): route_id: str; method: str; suffix: str; payload: dict[str, str] | None` and parametrize on the record (`ids=lambda c: c.route_id`), which also removes the `route_id` argument threaded through `:135` and `:157` that neither test body uses.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+def client() -> Iterator[FlaskClient]:
+    """Flask test client with CSRF disabled."""
+    app.config["TESTING"] = True
+    app.config["WTF_CSRF_ENABLED"] = False
+    app.config["SESSION_COOKIE_PATH"] = "/"
+    with app.test_client() as test_client:
+        yield test_client
+
+
+def _member_session(client: FlaskClient, tenant_id: str, email: str) -> None:
```
KonstantinMirin, line 54: Omitting the `test_user*` keys so the real `User` membership lookup runs is the right call, and it is the shape `tests/integration/test_mock_adapter_publisher_sync.py:25` should have used. But the session key set is a private input contract of `require_tenant_access` (`src/admin/utils/helpers.py:299-380`) and it now has 16 hand-written writers across `tests/`, two of them added by this diff. Rename a key in the decorator and every copy stays green against a shape production no longer uses.

The abstraction that exists is one flavor short: `AdminAccountEnv.authenticate()` offers only the super-admin test-mode flavor, the one that skips the membership lookup, which is why this file hand-rolled the member shape and the other file reached for the bypass. Add the member flavor to the harness (`authenticate(tenant_id, *, role="member" | "super_admin")`) with a thin `tests/utils/admin_auth.py::login_as_member(client, ...)` for non-BDD callers, spell the keys once with a docstring pointing at `helpers.py:299`, make `_auth_integration` a consumer of it, and route both in-diff sites through it.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+from adcp.exceptions import AdagentsNotFoundError
+from flask.testing import FlaskClient
+from sqlalchemy import func, select
+from sqlalchemy.orm import Session
+
+from src.admin.app import create_app
+from src.core.database.models import PublisherPartner
+from tests.factories import PublisherPartnerFactory, TenantFactory, UserFactory
+
+app = create_app()
```
KonstantinMirin, line 28: A module-level `create_app()` plus the local `client` fixture at `:44-52` has no e2e counterpart, so session-cookie handling, `SESSION_COOKIE_PATH`, proxy and `SCRIPT_NAME` handling, the domain redirect at `src/admin/app.py:211`, and blueprint mounting on the real WSGI stack are never exercised by an authorization fix. `tests/harness/admin_accounts.py` already owns app construction, client lifecycle, auth, URL building, and method dispatch per mode (`_setup_integration` `:134` / `_setup_e2e` `:144`, `_auth_integration` `:165` / `_auth_e2e` `:177`, `_url` `:208`, `_get`/`_post_form`/`_post_json` `:245-257`) and this module retypes all five — the 10th module in `tests/admin/` to build its own app at import time.

Lift the mode-generic half of `AdminAccountEnv` into `tests/harness/admin_base.py::AdminBlueprintEnv`, make `AdminAccountEnv` subclass it with no behavior change, add `PublisherPartnersEnv(AdminBlueprintEnv)`, and express anonymous-denied / other-tenant-denied / rejected-mutation-leaves-state-intact / member-succeeds as `tests/bdd/features/BR-ADMIN-PUBLISHER-PARTNERS.feature`, the way `BR-ADMIN-ACCOUNTS.feature:113 @T-ADMIN-ACCT-009` does for accounts. Drop this line and the local fixture in favor of the env or a fixture in `tests/admin/conftest.py`.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+        assert factory_session.get(PublisherPartner, partner_id) is None
+
+    def test_sync_is_not_auth_rejected(self, client: FlaskClient, factory_session: Session) -> None:
+        tenant = TenantFactory()
+        user = UserFactory(tenant=tenant)
+        _member_session(client, tenant.tenant_id, user.email)
+
+        response = client.post(_partners_url(tenant.tenant_id, "/sync"))
+
+        assert response.status_code not in (401, 403)
```
KonstantinMirin, line 237: `assert response.status_code not in (401, 403)` is subsumed by `:238 assert response.status_code == 200` — no edit to `src/` makes this line fail while the next one passes. Delete it and rename the test to what it verifies (`test_member_sync_returns_sync_result`), so the negative phrasing stops inviting negative-space assertions.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+                "src.admin.blueprints.publisher_partners.fetch_adagents",
+                side_effect=AdagentsNotFoundError("missing"),
+            ),
+        ):
+            response = client.get(_partners_url(tenant.tenant_id, f"/{partner.id}/properties"))
+
+        assert response.status_code == 200
+        body = response.get_json()
+        assert body["is_authorized"] is False
+        assert body.get("error") != "Authentication required"
```
KonstantinMirin, line 262: These two assertions cannot fail. They sit after `:259 assert response.status_code == 200`, `require_tenant_access` emits those two strings only with 401 or 403 (`src/admin/utils/helpers.py:339, 363`), and the handler's own 200 branches return handler-specific strings (`src/admin/blueprints/publisher_partners.py:526, 547, 549, 551`). Verified: delete `@require_tenant_access(api_mode=True)` from `publisher_partners.py:481` and this test plus `test_sync_is_not_auth_rejected` are still 2 passed.

Delete `:262-263` and rename the test (`test_member_get_properties_reports_unauthorized`); the 403 behavior for this route is already covered non-vacuously by the `properties` parametrization row. Prefer deletion over weakening.

tests/admin/test_publisher_partners_auth.py
```diff
@@ -0,0 +1,263 @@
...
+    if method == "DELETE":
+        return client.delete(url)
+    raise AssertionError(f"unexpected method {method}")
+
+
+def _partner_count(session: Session, tenant_id: str) -> int:
+    return session.scalar(select(func.count()).select_from(PublisherPartner).filter_by(tenant_id=tenant_id)) or 0
+
+
+def _partner_snapshot(session: Session, partner_id: int) -> tuple[Any, ...] | None:
```
KonstantinMirin, line 85: `tuple[Any, ...]` describes an eight-field record whose fields this function names positionally (`id`, `tenant_id`, `publisher_domain`, `display_name`, `is_verified`, `sync_status`, `sync_error`, `last_synced_at`). It is the value the "rejected mutation changed nothing" assertion compares, so a failure prints two unlabeled 8-tuples and a reordering here is invisible to `_assert_partners_unchanged`. Same problem at `:101`, where `_seed_target_tenant`'s four-element return carries its meaning in a docstring and is unpacked positionally at two sites, and at `:71`, where `_dispatch` has no return annotation while every other `def` in this module is fully annotated.

These are internal test shapes with no AdCP protocol counterpart, so declare `class PartnerSnapshot(NamedTuple)` with the eight named fields and `class SeededTenant(NamedTuple)`, and annotate `_dispatch(...) -> TestResponse` (`werkzeug.test.TestResponse`).

src/admin/blueprints/publisher_partners.py
```diff
@@ -185,6 +189,7 @@ def delete_publisher_partner(tenant_id: str, partner_id: int) -> Response | tupl
 
 
 @publisher_partners_bp.route("/<tenant_id>/publisher-partners/sync", methods=["POST"])
+@require_tenant_access(api_mode=True)
```
KonstantinMirin, line 192: This view declares `-> Response | tuple[Response, int]` on the next line and the decorator discards it: `require_tenant_access` (`src/admin/utils/helpers.py:297`) has no annotations and `mypy.ini` sets `disallow_untyped_decorators = False`, so `mypy` reveals `sync_publisher_partners` as `Any`. That erasure is why `tests/integration/test_mock_adapter_publisher_sync.py` can call `.get_json()` on what is actually a `(Response, int)` tuple with no type error.

Annotating the views harder cannot help; fix it at the root with `ParamSpec` + `TypeVar` + `Concatenate[str, P]` on `require_tenant_access` and on `decorated_function`. About four lines, no behavior change, and it restores the return contract for all 146 adopters including the five added here.

KonstantinMirin (commented):
## Review — PR #2075

Before the detail, here is the whole picture in four parts: where the codebase stands, what this PR does, what I need to see before approving, and what the real fix looks like. The six items after that are the specifics.

### 1. Where the codebase stands today

Walking the live `app.url_map` there are **179 rules carrying `<tenant_id>`. 146 go through `require_tenant_access`; 33 did not.** The 33 sort into five groups, and the sorting is the whole point:

| group | count | what enforces tenancy |
|---|---|---|
| **This PR converts these** | **5** | `require_tenant_access` after this diff |
| Deliberately public | 5 | nothing, correctly — `auth.tenant_login`, `auth.tenant_google_auth`, `auth.gam_authorize`, `oidc.login`, `oidc.test_initiate` |
| API-key authenticated | 9 | `require_sync_api_key` (6), `require_tenant_management_api_key` (3) |
| Super-admin only | 1 | `require_auth(admin_only=True)` — `core.reactivate_tenant` |
| `require_auth()` only, hand-rolled check in the handler body | 7 | an inline `session.get("role")` / `session.get("tenant_id")` 403 |
| `require_auth()` only, **no tenant check anywhere** | 10 | nothing |

That last row is live cross-tenant exposure on `main`: `api.revenue_chart_api`, `api.get_tenant_products`, `api.get_product_suggestions`, `policy.rules`, and the six `gam_reporting_api` endpoints. `require_auth` checks `"user" in session` and nothing else — it never reads `tenant_id` (`src/admin/utils/helpers.py:259-294`) — and `gam_reporting_api.py:59` defines a third local copy of it that also only checks for a session. Any logged-in user of any tenant can pass another tenant's id to those ten and read its products, revenue and GAM reporting. They are out of scope for this PR and are recorded in the Notes.

One thing worth saying plainly: I could only produce that table by writing a throwaway script that builds the app, walks `url_map` and cross-references the AST for decorators. **Nothing in the repo can answer "which tenant routes are unprotected."** Neither of us knows this number from reading the code, and that is the finding underneath all six below.

### 2. What this PR does

It adds `require_tenant_access(api_mode=True)` to all 5 publisher-partner routes. The production half is right and I would not change it: the canonical seam, 5 of 5 routes, no handler-body edits, no second producer of the 401/403 contract. It converts 5 of the 33.

What it does not do is change the rule that produced the 33. Authorization stays opt-in — a route is public by default and becomes protected when an author remembers to type one line. That is why a sixth route added to this blueprint tomorrow ships unguarded and nothing goes red. I verified that: appending an undecorated PATCH route to the blueprint leaves the suite green.

### 3. What I need to see before approving

Three things, all about knowing the fix works and keeps working:

1. **A test that walks `app.url_map`** and asserts every `<tenant_id>` route either enforces membership or sits in a small, explicitly-reasoned exempt set (finding 1, Option B, with runnable code). This is what makes route #180 fail CI on the day it is written, and what replaces my ad-hoc script with something the suite knows.
2. **Fix the sync test that switches off authorization** (finding 2). As written, `tests/integration/test_mock_adapter_publisher_sync.py` enables `ADCP_AUTH_TEST_MODE` and forges a super-admin session, so its four tests pass even when the guard denies every request. One of the two files in this PR neutralizes the other.
3. **Cover the `is_active=False` rejection** (finding 3). The guard rejects three ways and the new suite tests two. A deactivated employee keeps access to all five routes today and nothing fails.

### 4. What the proper fix looks like

Deny by default at the boundary: a blueprint `before_request` hook runs `check_tenant_access` for every request, and the genuinely public routes carry an explicit marker saying so. Then "unguarded" stops being ambiguous — a public route is one that *declares* it is public, and a forgotten route fails closed instead of serving another tenant's data. Finding 1 Option A sets out the three steps.

That is a larger piece of work and it should not hold up this PR. It wants its own issue. The `url_map` test in point 3 above is the bridge: it gives you the guarantee now, and it keeps working unchanged once the `before_request` hook lands.

### The six items below, in one line each

1. **Enforcement is a per-view opt-in** — the class this PR exists to close stays open; Options A and B above.
2. **The sync tests switch on the production auth bypass** — verified: they pass with the guard denying everyone.
3. **The new suite states the guard's behavior by hand** — one rejection path untested, three assertions that cannot fail.
4. **The admin test env exists for one blueprint, so this forks a weaker in-process copy** — no e2e counterpart, no Gherkin.
5. **The auth-session shape has no owner** — written twice more here, one copy already diverged from its source.
6. **Known shapes carry no type contract** — the untyped decorator erases all ~150 views to `Any`, plus anonymous tuple records.

Items 1, 2 and 3 are the three above. Items 4, 5 and 6 are real and I would like them, but they are about how the tests are built rather than whether this change is correct — land them here or in their own PR, your call.

### Should fix

**1. Tenant enforcement is a per-view opt-in, so the class this PR exists to close stays open.**

Every route on `publisher_partners_bp` is `/<tenant_id>/publisher-partners...` and every one is `api_mode=True`. The blueprint is entirely tenant-scoped, with no exempt route. Expressing that as five identical decorator lines (`src/admin/blueprints/publisher_partners.py:29, 90, 170, 192, 481`) makes "this route is tenant-scoped" a fact re-asserted at every site. The same five-route list is then typed out a second time as `_ROUTES` (`tests/admin/test_publisher_partners_auth.py:35-42`), so route #6 requires remembering two lists: forget the first and the route is world-readable, forget the second and nothing fails.

Verified: append an undecorated PATCH route to the blueprint and run the new module plus the sibling tenant-isolation guard.

```
pytest tests/admin/test_publisher_partners_auth.py \
       tests/unit/test_architecture_workflow_tenant_isolation.py  -> 16 passed
```

The same holds repo-wide, which is what makes this the review's lead item rather than a note about one blueprint — see the table in the overview: 179 tenant-scoped rules, 146 guarded, 10 with no tenant check of any kind. Nothing in the code separates the deliberate cases from the forgotten ones, and no test knows the count:

```
grep -rn "before_request" src/admin/blueprints/   -> no blueprint-level hook anywhere
grep -rln "url_map" tests/                        -> one unrelated file; nothing walks the routes
```

That is in a repo which already encodes this exact "someone forgot the required decorator" class as a structural guard 104 times elsewhere. This is the case where adding a mechanism is the right answer: the design is sound, and what is missing is this repo's own regression seam for the class the PR was opened to close.

There are two ways to fix this and they are not equivalent in size. Both keep `helpers.py` the single producer of the 401/403 bodies.

**Option A — move enforcement to the boundary (the proper fix, and a bigger piece of work).**

1. Extract the body of `require_tenant_access`'s `decorated_function` (`src/admin/utils/helpers.py:304-376`) into a plain callable `check_tenant_access(tenant_id: str, *, api_mode: bool) -> Response | tuple[Response, int] | None` that returns `None` on grant and the rejection response otherwise. The existing decorator becomes a two-line wrapper over it, so all 146 current adopters are unchanged and no second error literal is written.
2. Add `@publisher_partners_bp.before_request` calling `check_tenant_access(request.view_args["tenant_id"], api_mode=True)`. A non-`None` return short-circuits the dispatch, so a new route on this blueprint cannot ship unguarded. The five per-view decorators then come off.
3. Repeat per blueprint, with the genuinely public routes carrying an explicit `@public_route` marker rather than an absence.

This is the design that actually closes the class, and I am not asking for it in this PR. It wants its own issue.

**Option B — the minimum I would not merge without: a test that walks `app.url_map`.**

Enumerate reality instead of intent. For every rule whose path contains `<tenant_id>`, assert the view function carries `require_tenant_access` (or is in a short, named exempt set), and assert the exempt set is exactly the routes that are deliberately public:

```python
# Every entry needs a reason. This set is the ONLY place a tenant-scoped route is
# allowed to skip membership enforcement, and it may only shrink.
EXEMPT = {
    # public: these ARE the login flow, there is no session yet
    "auth.tenant_login", "auth.tenant_google_auth", "auth.gam_authorize",
    "oidc.login", "oidc.test_initiate",
    # authenticated by API key instead of a user session
    *SYNC_API_ENDPOINTS,            # require_sync_api_key
    *TENANT_MANAGEMENT_ENDPOINTS,   # require_tenant_management_api_key
    # super-admin only, no tenant membership involved
    "core.reactivate_tenant",
}

def test_every_tenant_scoped_route_enforces_membership():
    app = create_app()
    unguarded = sorted(
        rule.endpoint
        for rule in app.url_map.iter_rules()
        if "<tenant_id>" in str(rule)
        and rule.endpoint not in EXEMPT
        and not getattr(app.view_functions[rule.endpoint], "_enforces_tenant_access", False)
    )
    assert not unguarded, f"tenant-scoped routes with no membership check: {unguarded}"
```

`require_tenant_access` sets `decorated_function._enforces_tenant_access = True` (one line), which is what makes this checkable without parsing source. Under Option A the `before_request` hook marks the whole blueprint instead and the same test keeps working unchanged.

Written today this test fails, listing the 17 routes that carry only `require_auth()` — the 10 with no tenant check and the 7 that hand-roll one in the body. That is the correct initial state, and the honest way to land it is to fix the 10 in a follow-up PR and add the 7 self-checking routes to `EXEMPT` with a comment, or better, convert them so the inline 403s disappear. Do not seed `EXEMPT` with the 10: that would convert a live cross-tenant read into a documented, passing test, and the exempt set is a ratchet that may only shrink.

Two things this buys that the current suite does not have. Route #180 fails CI on the day it is written, by the author who wrote it. And the count stops depending on someone running an ad-hoc script.

Whichever you pick, derive `_ROUTES` (`tests/admin/test_publisher_partners_auth.py:35-42`) from `app.url_map` rather than retyping it, so the per-route sweep grows with the blueprint.

**2. The sync tests satisfy the new guard by switching on the production authorization bypass.**

`tests/integration/test_mock_adapter_publisher_sync.py` reaches `sync_publisher_partners` by calling the view function inside `app.test_request_context()`. No request is issued, so the URL map, the blueprint registration, and the decorator as registered are never exercised. When the guard landed, those four calls started returning `(Response, 401)`. The repair was `_authorize_direct_sync_call` (`:25-33`), which writes seven keys into the `flask.session` proxy including `test_user_role = "super_admin"`, plus a class-wide autouse `ADCP_AUTH_TEST_MODE=true` (`:40-42`). In `src/admin/utils/helpers.py:330-337` that combination returns `f(tenant_id, ...)` before the `User` membership lookup this PR exists to add.

Verified with two mutations against the head tree:

```
# insert an unconditional `return jsonify({"error": "Access denied"}), 403` in
# require_tenant_access, right after the test-mode branch, so the guard denies everyone:
pytest tests/integration/test_mock_adapter_publisher_sync.py  -> 4 passed

# delete the _enable_test_mode fixture (:40-42) and the four _authorize_direct_sync_call calls:
pytest tests/integration/test_mock_adapter_publisher_sync.py  -> 3 failed, 1 passed (401)
```

The first result is the one that matters: these tests pass with tenant authorization denying every request. They cannot observe this PR's change in either direction, and they are now coupled to the key list of a production escape hatch that `tests/unit/test_test_auth_production_guard.py` exists to fence off.

Root cause: `sync_publisher_partners` (`src/admin/blueprints/publisher_partners.py:191-478`) is about 290 lines of sync policy, session work, and ORM construction inside a Flask view, so a test that wants the sync behavior has exactly one entry point and it is the view function. Once that entry point was guarded, the cheapest repair was to defeat the guard rather than to issue a request. The extraction of that orchestration into a service is a separate boundary (see Notes), but the repair here does not need it.

Fix: dispatch `POST /tenant/<tenant_id>/publisher-partners/sync` through the Flask test client with an ordinary active-member session, the shape this PR's own `tests/admin/test_publisher_partners_auth.py:54 _member_session` establishes, which never touches `ADCP_AUTH_TEST_MODE`. The `patch(...)` targets are module-level and work unchanged under client dispatch. Then delete `_authorize_direct_sync_call` (`:25-33`), the autouse `_enable_test_mode` fixture (`:40-42`), and the `from flask import session` import at `:12`, which is the only module-scope `flask.session` import in `tests/` and shadows the name `with get_db_session() as session:` binds at `:49`, `:94`, `:137`, `:172`, `:211`, `:245` in the same file. This is a complete sweep or it is not done. 7 sites: `:12`, `:25-33`, `:40-42`, `:124`, `:159`, `:202`, `:235`.

**3. The new module states the guard's behavior by hand instead of deriving it from the guard's decision, so three assertions cannot fail and one rejection path is untested.**

`require_tenant_access` rejects three ways: no session, no matching `User` row for the URL tenant, and `is_active=False` on that row (`src/admin/utils/helpers.py:358`). The module enumerates two of them. `UserFactory` sets `is_active = True` (`tests/factories/user.py:27`) and no test overrides it, so a deactivated member of the target tenant never reaches any of the five routes.

```
# remove is_active=True from the User select in helpers.py:358, then:
pytest tests/admin/test_publisher_partners_auth.py  -> 15 passed
```

A deactivated employee regains access to every publisher-partner route and this regression suite says nothing.

The same framing produces negative-space assertions on the two routes where a member is admitted. `:237 assert response.status_code not in (401, 403)` is immediately followed by `:238 assert response.status_code == 200`, which subsumes it. `:262-263 assert body.get("error") != "Authentication required"` / `!= "Access denied"` sit after `:259 assert response.status_code == 200`, and `require_tenant_access` only emits those two strings with 401 or 403 (`helpers.py:339, 363`), while the handler's own 200 branches return handler-specific strings (`publisher_partners.py:526, 547, 549, 551`).

```
# delete @require_tenant_access(api_mode=True) from publisher_partners.py:481, then:
pytest tests/admin/test_publisher_partners_auth.py::TestPublisherPartnersAuthorizedMember  -> 2 passed
```

Fix: derive the reject axis from the predicate. Add one parametrization row with `UserFactory(tenant=tenant, is_active=False)` pointed at the target tenant, asserting the same JSON 403 body and the same state invariance. Delete `:237` and `:262-263` and rename `test_sync_is_not_auth_rejected` / `test_get_properties_is_not_auth_rejected` to what they verify, so the negative phrasing stops inviting negative-space assertions; the sibling tests in the same class already show the honest form (`body["total"] == 1`, `body["publisher_domain"] == ...`, plus a database read-back). While you are there, `_assert_partners_unchanged` is applied to the `list` and `properties` rows, which are GET handlers with no write path, so carry a mutating flag on the route rows so it runs only where a mutation is possible.

**4. The admin test environment exists for one blueprint, so the second one forks a weaker, in-process-only copy of it.**

`tests/admin/test_publisher_partners_auth.py` builds a Flask app at module import (`:28 app = create_app()`), mutates that global's `TESTING`/`WTF_CSRF_ENABLED`/`SESSION_COOKIE_PATH` from a local `client` fixture (`:44-52`), hand-writes the authenticated session (`:54`), builds URLs (`:67`, `:128`), and switches on the HTTP method (`:71`). Every one of those five responsibilities already has an owner, per mode, in `tests/harness/admin_accounts.py`: `_setup_integration` (`:134`) / `_setup_e2e` (`:144`), `authenticate()` → `_auth_integration` (`:165`) / `_auth_e2e` (`:177`), `_url()` (`:208`), `_get`/`_post_form`/`_post_json` (`:245-257`), and `_AdminResponse` (`:28`) normalizing the two response types so steps never learn which surface is live.

The consequence is that a module-level `create_app()` has no e2e counterpart: session-cookie handling, `SESSION_COOKIE_PATH`, proxy and `SCRIPT_NAME` handling, the domain redirect at `src/admin/app.py:211`, and blueprint mounting on the real WSGI stack are never exercised by an authorization fix. And there is no Gherkin for any of it — `grep -rni "publisher.partner" tests/bdd/ tests/harness/` returns no feature, no env, no step module, while the exact analogue ships for the other blueprint at `tests/bdd/features/BR-ADMIN-ACCOUNTS.feature:113` (`@T-ADMIN-ACCT-009 @auth`, "Unauthenticated access is denied").

Root cause: `AdminAccountEnv` was written as a standalone class rather than on a shared base, so the second admin surface has nothing to inherit and the path of least resistance is to retype client lifecycle, auth, URL building, and method dispatch in the test module. The missing `AdminBlueprintEnv` is the defect, not the new file. "There is no publisher-partners env yet" is a reason to build it.

Fix as one move: lift the mode-generic half of `AdminAccountEnv` (`_AdminResponse`, mode detection `:91-93`, `__enter__`/`__exit__`, `_setup_integration`/`_setup_e2e`, `authenticate`/`clear_auth`, `_get`/`_post_form`/`_post_json`) into `tests/harness/admin_base.py::AdminBlueprintEnv`; make `AdminAccountEnv` subclass it with no behavior change; add `PublisherPartnersEnv(AdminBlueprintEnv)` exposing `list/create/delete/sync/properties` and a `partner_snapshot()`. Then `tests/bdd/features/BR-ADMIN-PUBLISHER-PARTNERS.feature` carries the four behaviors the PR body already states in prose (anonymous is denied, a member of another tenant is denied, a rejected mutation leaves stored partners unchanged, an active member succeeds), with Given/When/Then naming no Flask, no session key, and no status code as wire mechanics. Delete the module-level `create_app()` and the local `client` fixture in favor of the env or a fixture in `tests/admin/conftest.py`; that file owns `factory_session` (`:19`) and `ui_client` (`:86`) but never grew the plain "real app, real database, CSRF off" client that 10 modules now each supply themselves. The parametrized pytest rows can stay as the fast per-route sweep, but they are not the bar. One caveat: the admin BDD family is currently pinned to `AdminAccountEnv(mode="integration")` at `tests/bdd/conftest.py:3466`, so the new env's e2e path has to be genuinely reachable or the single-surface gap reappears one layer up.

**5. The admin auth-session shape has no owner, and this diff writes it twice more — one copy has already diverged from its source.**

`require_tenant_access` (`src/admin/utils/helpers.py:299-380`) is the only reader of this session shape, and it branches on `test_user`, `test_tenant_id`, `test_user_role == "super_admin"`, then `session["user"]` as dict-or-string. The diff adds `_member_session` (`tests/admin/test_publisher_partners_auth.py:54`, 4 keys) and `_authorize_direct_sync_call` (`tests/integration/test_mock_adapter_publisher_sync.py:25`, 7 keys). The second is a near-verbatim copy of `tests/harness/admin_accounts.py:165 _auth_integration` and is already out of step with it in the commit that created it:

```
diff <(sed -n '167,175p' tests/harness/admin_accounts.py) \
     <(sed -n '26,33p'   tests/integration/test_mock_adapter_publisher_sync.py)
# key-for-key identical except the dropped sess["tenant_id"]
```

The duplication ratchet did not notice. `pylint --enable=R0801` at head still reports the same single pair (`tests/admin/test_authorized_properties.py:62-72` matching `tests/harness/admin_accounts.py:167-178`), `.duplication-baseline` `tests: 73` is unchanged, and neither file this PR touches appears in the 73 hits — the new copy escapes the six-line similarity window because it writes `session[...]` instead of `sess[...]` and dropped a key.

Root cause: the abstraction that exists is one flavor short in the wrong direction. `AdminAccountEnv.authenticate()` offers only the super-admin test-mode flavor, the one that skips the membership lookup. There is no owned way to say "authenticate as an ordinary active member", which is why the PR hand-rolled that shape in one file and reached for the bypass in the other. Fix: add the member flavor to the harness as `AdminBlueprintEnv.authenticate(tenant_id, *, role="member" | "super_admin")` on the base from finding 4, with the member path writing `authenticated`/`user`/`email`/`tenant_id` and no `test_user*` keys and the e2e path performing a real login, plus a thin `tests/utils/admin_auth.py::login_as_member(client, ...)` for non-BDD callers, with the key set spelled once and a docstring pointing at `helpers.py:299`. Make `_auth_integration` a consumer of it. Route both in-diff sites through it (2 of 2). Done alongside findings 2 and 4 this costs almost nothing: the second site disappears entirely and the first moves into the env.

**6. Known shapes on both sides of the diff carry no type contract.**

All five views declare `-> Response | tuple[Response, int]` (`publisher_partners.py:30, 91, 171, 193, 482`), and the seam this PR adopts discards it. `require_tenant_access` (`src/admin/utils/helpers.py:297`) has no annotations, and `mypy.ini` sets `disallow_untyped_decorators = False`:

```
printf 'from src.admin.blueprints.publisher_partners import sync_publisher_partners\nreveal_type(sync_publisher_partners)\n' > /tmp/rt.py
mypy --config-file=mypy.ini /tmp/rt.py   # Revealed type is "Any"
```

That erasure is why `tests/integration/test_mock_adapter_publisher_sync.py` can call `.get_json()` on what is actually a `(Response, int)` tuple with no type error — it is load-bearing for finding 2 staying invisible. Annotating the views harder cannot work; the erasure happens in the decorator. Fix it at the root with `ParamSpec` + `TypeVar` + `Concatenate[str, P]` on `require_tenant_access` and the same on `decorated_function`. About four lines, no behavior change, and it restores the return contract for all 146 adopters including the five added here.

The same missing-contract problem appears in the new test records. `_partner_snapshot(...) -> tuple[Any, ...] | None` (`:85`) is a fixed eight-field record (`id, tenant_id, publisher_domain, display_name, is_verified, sync_status, sync_error, last_synced_at`), and it is the value the "rejected mutation changed nothing" assertion compares, so a failure prints two unlabeled 8-tuples and a reordering inside `_partner_snapshot` is invisible to `_assert_partners_unchanged`. `_seed_target_tenant(...) -> tuple[str, int, int, tuple[Any, ...] | None]` (`:101`) carries its meaning in a docstring and is unpacked positionally at two sites. `_ROUTES` (`:34-42`) carries its field meanings in a `# (id, method, path suffix…, JSON body or None)` comment, which is the tell that a type is missing, and threads `route_id` through both `@parametrize` signatures (`:135`, `:157`) where neither body uses it. `_dispatch` (`:71`) has no return annotation while every other `def` in the module is fully annotated. These are internal test shapes with no AdCP protocol counterpart, so `NamedTuple` is the right concrete type: `PartnerSnapshot`, `SeededTenant`, and `RouteCase` with named fields, parametrized on the record (`ids=lambda c: c.route_id`), which also removes the unused `route_id` argument. Annotate `_dispatch(...) -> TestResponse` (`werkzeug.test.TestResponse`).

- Seventeen sibling tenant-scoped routes carry only `require_auth()` at head, which checks `"user" in session` and never reads `tenant_id`. Ten have no tenant check of any kind: `src/admin/blueprints/api.py` (`revenue-chart`, `products`, `products/suggestions`), `src/admin/blueprints/policy.py::rules`, and the six endpoints in `src/adapters/gam_reporting_api.py`, which defines its own third copy of `require_auth` at `:59` that also only checks for a session. Any logged-in user of any tenant can pass another tenant's id to these and read its products, revenue and GAM reporting. The other seven — `inventory.py` (`orders`, `check-inventory-sync`, `analyze-ad-server`), `operations.py::reporting`, and `policy.py` (`index`, `update`, `review_task`) — hand-roll an inline `session.get("role")` / `session.get("tenant_id")` 403 in the handler body, a second producer of the contract `require_tenant_access` owns. All of this is out of scope here: the diff touches none of those files and the PR was right to keep itself to the publisher-partners blueprint. They are the natural payload of finding 1's mechanism once it exists. File a follow-up issue and self-assign it, scoped to "audit every `<tenant_id>` route for membership enforcement, convert the `require_auth()`-only routes, delete the inline 403s, and collapse the three `require_auth` implementations into one". I checked #1848: it is OPEN, has no assignees, and is titled "feat: Property channel disclosure on PublisherPartner (Phase 1)", so it does not establish a deferral home for this.
- The sync orchestration inside the view (`src/admin/blueprints/publisher_partners.py:191-478`) is the root of finding 2's entry-point problem, but extracting it is a separate boundary: it is pre-existing handler-body code, the PR's stated scope excludes handler-body changes, and the debt is already recorded (`tests/unit/test_architecture_no_raw_select.py:203-207` allowlists all five functions). Finding 2's fix is fully achievable without it. The extraction, a `PublisherPartnerSyncService` returning a typed result with the view reduced to identity plus `jsonify`, needs its own filed, self-assigned issue.
- `require_tenant_access` logs at INFO on every call (`src/admin/utils/helpers.py:305-308`), a level louder than the DEBUG used elsewhere in the same function, and now fires on five more JSON routes. Out of scope because the diff does not touch `helpers.py` and the log prints key names, not values.

PR #2141, opened 2026-08-28
chore: enhance coding efficiency guidelines in CLAUDE.md

patmmccann:
Added guidelines for efficient coding practices and bug fixes from ponytail repo


ChrisHuie (approved):

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+Question complex requests: "Do you actually need X, or does Y cover it?"
+Pick the edge-case-correct option when two stdlib approaches are the same size, lazy means less code, not the flimsier algorithm.
+Mark deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic): comment naming the ceiling and upgrade path.
+Not lazy about: understanding the problem (read it fully and trace the real flow before picking a rung, a small diff you don't understand is just laziness dressed up as efficiency), input validation at trust boundaries, error handling that prevents data loss, security, accessibility, the calibration real hardware needs (the platform is never the spec ideal, a clock drifts, a sensor reads off), anything explicitly requested. Lazy code without its check is unfinished: non-trivial logic leaves ONE runnable check behind, the smallest thing that fails if the logic breaks (an assert-based demo/self-check or one small test file; less frameworks, less fixtures). Trivial one-liners need no test.
```
KonstantinMirin, line 561: This sentence states the verification bar a second time, lower, two lines above `### Test Integrity Policy — ZERO TOLERANCE`.

- "an assert-based demo/self-check" is never collected: `pytest.ini:3-4` pins `testpaths = tests` and `python_files = test_*.py`, and `:31` lists `scripts` under `norecursedirs`. Break the production logic such a check covers and `make quality` stays green.
- "ONE runnable check" caps coverage at one path where `tests/CLAUDE.md:107` parametrizes every BDD scenario across all four transports.
- "less frameworks, less fixtures" instructs the shape the guards reject: `CLAUDE.md:310-311` marks factory fixtures MANDATORY, `tests/CLAUDE.md:129,173` make factory-boy mandatory, and `tests/unit/test_architecture_bdd_no_dict_registry.py` fails raw-dict Given steps.
- "Trivial one-liners need no test" exempts the diffs `CLAUDE.md:533` and `:535` cover unconditionally, three and five lines above. A one-line change to a status enum, an error code, or a response field is a wire-contract change covered by the graded contract at `dist/compliance/3.1.1/**`.

Delete these clauses. Test-sufficiency policy has one home, `tests/CLAUDE.md`; if you want a sentence here, make it a pointer to it, the way `CLAUDE.md:500-504` points at the Error Verification Policy.

Same line: "the calibration real hardware needs (the platform is never the spec ideal, a clock drifts, a sensor reads off)" has no referent here. `grep -rni "sensor\|calibration" --include="*.py" src tests` returns 0. Replace that tail with this system's non-negotiables: tenant isolation and auth on every handler, AdCP compliance for client-facing models, a single Alembic head, wire-envelope error contracts, per-item error handling.

KonstantinMirin (commented):
## Review — PR #2141

**Overview** — This PR has no closing issue, and the description names a source ("from ponytail repo") and an action rather than a problem, so nothing says whether it worked. Read charitably, the goal is to steer agents toward reuse-first minimal changes and root-cause bug fixes, which is a goal this repo already holds. The change appends 26 lines of another project's document into `### Quality Rules` instead of amending the sections that already own each subject, so `CLAUDE.md` now answers three questions two different ways: how much code to write, how much test is enough, and how to mark a known ceiling in source. `CLAUDE.md` is the normative spec every agent reads before writing code, so an agent that finds two answers cites the newer, weaker, nearer one.

The whole review is one move: revert `CLAUDE.md:536-561`, then add back only the rules this repo does not already state, as edits inside the sections that own them. The items below are the specific conflicts that move has to resolve, each named with its owning section.

### Should fix

**1. The verification bar is stated a second time, lower, two lines above the ZERO TOLERANCE policy**

`CLAUDE.md:561` ends with:

> Lazy code without its check is unfinished: non-trivial logic leaves ONE runnable check behind, the smallest thing that fails if the logic breaks (an assert-based demo/self-check or one small test file; less frameworks, less fixtures). Trivial one-liners need no test.

Every clause conflicts with a rule this repo enforces:

- "an assert-based demo/self-check" is a check CI never runs. `pytest.ini:3-4` pins `testpaths = tests` and `python_files = test_*.py`, and `pytest.ini:31` lists `scripts` under `norecursedirs`. Put an assert-based self-check in `scripts/` or an `if __name__ == "__main__":` block, break the production logic it checks, and `make quality` stays green because the file is never collected.
- "ONE runnable check" sets the ceiling at one path where `tests/CLAUDE.md:107` parametrizes every BDD scenario across all four transports. Change a tool's error emission so the MCP wrapper drops the `errors` array while `_impl` still returns it: a single check written against `env.call_impl(...)` passes, and the four-transport scenario is what fails.
- "less frameworks, less fixtures" instructs the shape the guards reject. `CLAUDE.md:310-311` marks factory fixtures MANDATORY, `tests/CLAUDE.md:129` and `:173` make factory-boy factories mandatory, and `tests/unit/test_architecture_bdd_no_dict_registry.py` fails Given steps that build raw dicts. An agent cannot satisfy both rules in one diff.
- "Trivial one-liners need no test" exempts the diffs that `CLAUDE.md:533` ("AdCP compliance test for all client-facing models") and `:535` ("Roundtrip test required for any operation using `apply_testing_hooks()`") cover unconditionally, three and five lines above it. Adding one field to a client-facing model or one `apply_testing_hooks()` call site is a one-line diff. In this codebase the one-line diff is usually the wire-contract diff: a status enum value, an error code, a response field, all covered by the graded contract at `dist/compliance/3.1.1/**` under the pinned AdCP 3.1.1 (`docs/adcp-spec-version.md`).

Root: test-sufficiency policy is being restated instead of referenced. `tests/CLAUDE.md` is the single home for it, and `CLAUDE.md:500-504` (Error Verification) is the done-right example in this same file: one sentence, one pointer, one home.

Delete the sufficiency clauses. If you want a sentence here, make it a pointer: behavior changes are verified by a transport-blind BDD scenario across all four transports, errors through `assert_envelope_shape` on `wire_error_envelope`, successes through `wire_field`/`wire_dict`, data through factory-boy factories — see `tests/CLAUDE.md`.

**2. "How much code to write" now has two authorities that disagree, and the new one supplies the excuse the DRY invariant bans**

`CLAUDE.md:29-37` is the Non-Negotiable Invariant on this subject, and it says, verbatim:

> **NEVER** cite "avoid over-engineering" or "keep it simple" to justify leaving duplicated logic in place

The new lines hand an agent that citation, 500 lines later, with no force marker and no `**Enforced by:**` line:

- `:536` "The best code is the code never written."
- `:553` "No abstractions that weren't explicitly requested."
- `:556` "Deletion over addition. Boring over clever. Fewest files possible."
- `:557` "Shortest working diff wins."

The caveat on `:553` ("allow shared helpers and existing architecture abstractions") permits reusing an abstraction that exists and says nothing about creating the missing one, which is the only case the invariant and `.duplication-baseline` (`{"src": 35, "tests": 71, "scripts": 0}`) actually govern. Three other sections in this file are subordinate to a size ladder as written:

- The layer contract. The repository method, the `_impl`, the per-transport wrappers and the structural guard are not abstractions a ticket requests; `CLAUDE.md:16` requires all of them for every new tool. Under "no unrequested abstractions / fewest files / shortest working diff", the compliant way to add a media-buy write is `get_db_session()` and `session.add()` inside the `_impl`, which is the block `CLAUDE.md:222-234` labels WRONG and `:236` forbids outright.
- The Spec-Grounding Gate at `CLAUDE.md:133-139`. For protocol behavior, what gets built comes from the pinned spec section plus the graded storyboard step, not from the ticket or the diff size. A ladder whose first rung is YAGNI (`:540`) and whose last tie-break is "shortest working diff wins" gives an agent a reason to omit a required-but-unrequested field or graded branch.
- Pattern 1 at `CLAUDE.md:146-160`, MANDATORY: protocol shapes come from the pinned `adcp` SDK by inheritance through the `Library*` alias, never restated. The ladder puts "Does the standard library already do this?" (`:542`) and "Can this be one line?" (`:545`) above "Does an already-installed dependency solve it?" (`:544`) and never names the SDK, so the rung an agent stops at for an AdCP payload is a hand-rolled dict or a hand-listed enum.

Also on this subject: `:541` is a fifth statement of the reuse half of the DRY invariant (`CLAUDE.md:8`, `:33`, `:67`, `:147`, `:751-753` already state it), and `:549` ("Bug fix = root cause… grep every caller… fix the shared function once") is the one clause that matches this repo's policy — which is exactly why it should amend `CLAUDE.md:29-59` rather than restate it in a different voice.

Root: the change appends a competing statement of a policy the document already owns. Delete `:536`, `:541`, `:549`, `:553`, `:556`, `:557` from `### Quality Rules`. Add to `CLAUDE.md:29-59`, under the force marker and `**Enforced by:**` line that section already carries, the two things the invariant does not yet say: the reuse-first rungs, and the precedence rule — when the shortest working diff is a second copy of a logical operation, the extraction wins and the baseline goes down, and the layer split is required output rather than an unrequested abstraction. Add one rung above stdlib for the ladder that survives: an AdCP protocol shape or protocol behavior is decided by Pattern 1 and the Spec-Grounding Gate, not by this ladder.

**3. `ponytail:` forks the in-source marker convention and opens a debt channel nothing counts**

`CLAUDE.md:560` tells agents to mark a deliberate simplification with a `ponytail:` comment. `grep -rIn ponytail` over the whole tree returns exactly one hit: that line. It has no definition, no guard, and no issue reference.

`CLAUDE.md:121` already defines the marker for this job and states the requirement in the negative on purpose: `# FIXME(#<gh-issue>)`, "reference a GitHub issue/PR number, never a local beads id (beads ids don't resolve for outside contributors)". There are 67 `FIXME(#` sites in `src/` and `tests/`, and `tests/unit/test_guards_no_beads_ids.py` exists to keep non-resolvable local markers out of this tree. `ponytail:` is a local token that resolves to nothing for an outside contributor.

Two faults, not one. The convention is forked, so one job has two spellings and neither a grep nor a guard covers both. And the rule sanctions parking new debt in the PR that creates it, with none of the three properties every deferral here has: a registry that enumerates it, a count that can be shown to fall, and a shrink-only direction (`CLAUDE.md:120`). Read with "Shortest working diff wins" three lines above, the cheapest compliant path through the new text is defer-with-a-comment.

Delete `:560`. If the ceiling-and-upgrade-path idea is worth keeping, add it as a clause to the existing bullet at `CLAUDE.md:121`, using the same `# FIXME(#<gh-issue>)` spelling with the ceiling and upgrade path in the comment body and the issue filed and open, so one grep and one guard cover both cases.

**4. The block sits under `## Testing Guidelines`, and the mirrored copy in `.claude/rules/patterns/testing-patterns.md` now disagrees**

`### Quality Rules` (`CLAUDE.md:531`) is a four-bullet test list under `## Testing Guidelines` (`:490`). The diff adds a fifth bullet at `:536` and then 25 more lines of general coding policy as unlabeled prose, ending at `:561`, immediately above `### Test Integrity Policy — ZERO TOLERANCE` at `:563`. Section headings are the only scoping mechanism this file has, so both readings are available and both are wrong. Scoped to tests, "Deletion over addition" and "Fewest files possible" sit directly above the policy whose rule 1 is "NEVER skip, ignore, deselect, or exclude failing tests", and "fewest files" works against the per-file caps in `CLAUDE.md:532` and `tests/unit/test_architecture_behavioral_mock_cap.py`. Scoped globally, general architecture policy sits 250 lines from the architecture patterns section at `CLAUDE.md:144` and 500 lines from the invariant it modifies.

`### Quality Rules` is mirrored into `.claude/rules/patterns/testing-patterns.md:54-58`, which still carries the original four bullets. Ten sections mirror this way. An agent whose context loads the rules file now reads a different quality bar than one reading `CLAUDE.md`.

Two more things go with placement. The new rules carry no bindingness marker (`MANDATORY`/`NEVER`/`MUST`) and no `**Enforced by:**` line, while `CLAUDE.md:37`, `:163`, `:202`, `:293`, and `:532` each name their mechanism — so an agent hitting the contradictions above has no tie-break signal. And `CLAUDE.md:540-547` (the ladder) and `:553-561` (the ten rules) carry no list markers, so consecutive bare lines collapse into two run-on paragraphs when rendered; `:551` is a plain `Rules:` where `:119`, `:157`, `:282`, `:288`, and `:341` are `**Rules:**`.

Move the surviving text out of `## Testing Guidelines`, leave `### Quality Rules` byte-identical to its mirror, and update the mirror in the same commit. Render the ladder as a numbered list (it is ordered and stops at the first rung that holds), the rules as `-` bullets, and the label as `**Rules:**`.

**5. Rules that name things this codebase does not have**

`CLAUDE.md:543` adds a rung for "a native platform feature". `CLAUDE.md:561` tells agents not to be lazy about "the calibration real hardware needs (the platform is never the spec ideal, a clock drifts, a sensor reads off)". `grep -rni "sensor\|calibration\|native platform" --include="*.py" src tests` returns 0, and `grep -rn "native platform"` matches only `CLAUDE.md:543`. This is a Python ad-serving backend; its tiers are the standard library, an installed dependency, and the pinned `adcp` SDK.

Related: `:541` tells an agent to reuse "the helper, util, or pattern that's already here". This tree carries 67 `FIXME(#…)`-marked violation sites and 103 guard files whose allowlists enumerate code that is known to be wrong. An agent that reads those sites as the house pattern clones them, and the cheapest resolution of the resulting guard failure is a new allowlist entry.

Delete the "native platform feature" rung and the hardware clause. If a "not lazy about" list survives, populate it with this system's non-negotiables: tenant isolation and auth on every handler, AdCP compliance for client-facing models, a single Alembic head, wire-envelope error contracts, and per-item error handling so one bad row does not stop a batch. Scope the reuse rung to the sanctioned seam (repository, helper, factory, harness env) and say that a site carrying `FIXME(#…)` or listed in a guard allowlist is debt and must not be copied.

**6. State the problem the PR fixes**

There is no closing issue, and the description names a source and an action. Write the failure mode you observed in this repo that the new rules prevent, so a reviewer can check whether the change achieved it.

- Out of scope: nothing checks that `CLAUDE.md` and `.claude/rules/patterns/*.md` stay in step. The mirror spans 10 sections and is maintained by hand. Building that guard is a separate tooling boundary this diff does not otherwise touch, so it is not this PR's job. It is the reason the fork in item 4 would go unnoticed, which is why item 4 itself belongs here.
- Out of scope: `.claude/rules/workflows/bug-reporting.md:39-41` says "Don't refactor surrounding code (separate PR)", which contradicts `CLAUDE.md:34` ("refactor duplicated code… is a **bug fix**, not an 'improvement'"). That contradiction predates this diff and lives in a file the diff does not touch. If item 2 moves the root-cause rule into `CLAUDE.md:29-59`, pointing `bug-reporting.md:41` at that section is the natural follow-up.
- Out of scope: `CLAUDE.md:370-407` and `:611-626` are near-verbatim copies of `.claude/rules/patterns/code-patterns.md:5-59`. The diff does not touch those lines. Noted because the doc set already has two homes for "how to write code here", and this PR adds to a third.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+Does an already-installed dependency solve it? Use it.
+Can this be one line? Make it one line.
+Only then: write the minimum code that works.
+The ladder runs after you understand the problem, not instead of it: read the task and the code it touches, trace the real flow end to end, then climb.
+
+Bug fix = root cause, not symptom: a report names a symptom. Grep every caller of the function you touch and fix the shared function once — one guard there is a smaller diff than one per caller, and patching only the path the ticket names leaves a sibling caller still broken.
+
+Rules:
+
+No abstractions that weren't explicitly requested. Caveat: allow shared helpers and existing architecture abstractions.
```
KonstantinMirin, line 553: In this repo the repository method, the `_impl`, the per-transport wrappers, and the structural guard are not abstractions a ticket requests — `CLAUDE.md:16` requires all of them for every new tool. The caveat permits reusing an abstraction that already exists and says nothing about creating the missing one, which is the case the DRY invariant at `CLAUDE.md:29-37` and `.duplication-baseline` govern.

Read with `:556-557`, this line's literal instruction for a media-buy write is `get_db_session()` plus `session.add()` inside the `_impl`: one file, fewer lines, no new abstraction — the block `CLAUDE.md:222-234` labels WRONG and `:236` forbids.

It also reads against Pattern 1 (`CLAUDE.md:146-160`, MANDATORY): a `Product(LibraryProduct)` subclass is an abstraction nobody requested, and hand-building the dict is the shorter diff.

Delete the rule here and state the boundary once inside `CLAUDE.md:29-59`: new code belongs in the layer that owns it even when a smaller diff exists elsewhere, and an AdCP protocol shape comes from the `adcp` SDK through the `Library*` alias.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+No new dependency if it can be avoided.
+No boilerplate nobody asked for.
+Deletion over addition. Boring over clever. Fewest files possible.
```
KonstantinMirin, line 556: `CLAUDE.md:35` says, verbatim, "**NEVER** cite 'avoid over-engineering' or 'keep it simple' to justify leaving duplicated logic in place". These two lines supply that citation: extraction adds a file and lengthens the diff, so under "Fewest files possible" and "Shortest working diff wins" the duplicated version is the compliant one, and `.duplication-baseline` (`{"src": 35, "tests": 71, "scripts": 0}`) stops falling.

Layer separation here deliberately spends files: one behavior is a schema change, an `_impl`, an MCP wrapper, an A2A raw function, a REST route, a repository method, and a structural guard. The hedge on `:557` gestures at the right criterion but never names it.

Delete both lines and amend `CLAUDE.md:29-59` instead: the smallest change at the layer that owns the behavior, and when the shortest working diff is a second copy of a logical operation, the extraction wins and the baseline goes down.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+Shortest working diff wins, but only once you understand the problem. The smallest change in the wrong place isn't lazy, it's a second bug.
+Question complex requests: "Do you actually need X, or does Y cover it?"
+Pick the edge-case-correct option when two stdlib approaches are the same size, lazy means less code, not the flimsier algorithm.
+Mark deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic): comment naming the ceiling and upgrade path.
```
KonstantinMirin, line 560: `grep -rIn ponytail` over the tree returns exactly one hit: this line. The token has no definition, no guard, and no issue reference.

`CLAUDE.md:121` already defines the marker for this job and states the requirement in the negative on purpose: `# FIXME(#<gh-issue>)`, "reference a GitHub issue/PR number, never a local beads id (beads ids don't resolve for outside contributors)". There are 67 `FIXME(#` sites in `src/` and `tests/`, and `tests/unit/test_guards_no_beads_ids.py` exists to keep non-resolvable local markers out of the tree.

The rule also sanctions parking new debt in the PR that creates it, with none of the three properties every deferral here has: a registry that enumerates it, a count that can be shown to fall, and a shrink-only direction (`CLAUDE.md:120`). With "Shortest working diff wins" three lines above, the cheapest compliant path is defer-with-a-comment.

Delete this line. If the ceiling and upgrade path are worth recording, add that clause to `CLAUDE.md:121` using the existing spelling, with the issue filed and open.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
 - AdCP compliance test for all client-facing models
 - Test YOUR code, not Python built-ins
 - Roundtrip test required for any operation using `apply_testing_hooks()`
+- You are a lazy senior developer. Lazy means efficient, not careless. The best code is the code never written.
```
KonstantinMirin, line 536: This bullet opens 26 lines of general coding policy inside `## Testing Guidelines → ### Quality Rules`, whose other four bullets are all test rules and each name their mechanism ("pre-commit enforces", "AdCP compliance test", "Roundtrip test required"). This one names none, carries no `MANDATORY`/`NEVER`/`MUST` marker and no `**Enforced by:**` line, and is the only bullet in the file that states a persona rather than a rule.

"The best code is the code never written" also answers the same question as `CLAUDE.md:31-37` with the opposite polarity, 500 lines away, and neither points at the other.

`### Quality Rules` is mirrored into `.claude/rules/patterns/testing-patterns.md:54-58`, which still carries the original four bullets — 10 sections mirror this way, so an agent loading the rules file now reads a different quality bar. Drop this bullet, keep `### Quality Rules` byte-identical to its mirror, and put the surviving policy in `CLAUDE.md:29-59` or `.claude/rules/patterns/code-patterns.md`.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+
+Before writing any code, stop at the first rung that holds:
+
+Does this need to be built at all? (YAGNI)
+Does it already exist in this codebase? Reuse the helper, util, or pattern that's already here, don't re-write it.
```
KonstantinMirin, line 541: Reusing a helper is right. Reusing a "pattern that's already here" is the argument the allowlist discipline blocks. This tree carries 67 `FIXME(#…)`-marked violation sites and 103 guard files whose allowlists enumerate code that is known to be wrong: raw `select()` outside repositories, `model_dump()` in `_impl`, `session.query()`, dict registries in BDD Given steps. An agent that reads those sites as the house pattern clones them, and the cheapest resolution of the resulting guard failure is a new allowlist entry, which `CLAUDE.md:120` forbids.

Scope the rung to the sanctioned seam — repository, helper, factory, harness env — and say that a site carrying `FIXME(#…)` or listed in a guard allowlist is debt and must not be copied. This rung is also the fifth statement of the reuse half of the DRY invariant (`CLAUDE.md:8`, `:33`, `:67`, `:147`, `:751-753`); fold it into `CLAUDE.md:29-59` rather than adding a sixth home.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+Does the standard library already do this? Use it.
+Does a native platform feature cover it? Use it.
```
KonstantinMirin, line 543: "Native platform feature" names no tier this repo has. `grep -rn "native platform"` matches only this line. The tiers here are the standard library, an already-installed dependency, and the pinned `adcp` SDK.

The SDK rung is the one that matters and it is missing: `CLAUDE.md:146-160` requires protocol types to come from `adcp` by inheritance through the `Library*` alias, MANDATORY, never restated. As ordered, an agent climbing this ladder stops at `:542` ("standard library") or `:545` ("one line") and hand-rolls the wire shape as a dict, a `TypedDict`, or a hand-listed enum — the shortest diff every time, and the one construction this repo forbids.

Delete this rung and add one above stdlib: an AdCP protocol shape or protocol behavior is decided by Pattern 1 (`:146-160`) and the Spec-Grounding Gate (`:133-139`), not by this ladder.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+Does an already-installed dependency solve it? Use it.
+Can this be one line? Make it one line.
+Only then: write the minimum code that works.
+The ladder runs after you understand the problem, not instead of it: read the task and the code it touches, trace the real flow end to end, then climb.
+
+Bug fix = root cause, not symptom: a report names a symptom. Grep every caller of the function you touch and fix the shared function once — one guard there is a smaller diff than one per caller, and patching only the path the ticket names leaves a sibling caller still broken.
```
KonstantinMirin, line 549: This is the one clause in the block that matches this repo's policy, which is why it should amend the owning section rather than become a third home for the rule. `CLAUDE.md:33-36` and the Decision Tree at `:759-765` already require it, and `.claude/rules/workflows/bug-reporting.md:41` gives the opposite answer ("Don't refactor surrounding code (separate PR)").

The justification also picks the fix location by diff size. When the callers are the MCP wrapper, the A2A raw function, and the REST route, the nearest shared function is often a transport-side helper, and putting the guard there is genuinely the smaller diff while being the transport-parity violation `test_transport_agnostic_impl.py` and `test_architecture_boundary_completeness.py` catch.

Move the sentence into `CLAUDE.md:29-59` and state the criterion as ownership: fix at the lowest layer that owns the behavior — `_impl` for a business rule, the repository for a data-access defect — never in one transport wrapper.

CLAUDE.md
```diff
@@ -533,6 +533,32 @@ def test_something():
...
+
+Rules:
```
KonstantinMirin, line 551: Two formatting problems that change what the text means. `Rules:` is plain where `CLAUDE.md:119`, `:157`, `:282`, `:288`, and `:341` all use `**Rules:**`, which is how an agent scopes a rules block to its section. And the ten rules below (`:553-561`), like the seven rungs above (`:540-547`), carry no list markers, so consecutive bare lines collapse into one paragraph when rendered — the ladder loses the ordering that "stop at the first rung that holds" depends on.

Use `**Rules:**` for the label, `-` bullets for the rules, and a numbered list for the ladder.

Issue #2104, opened 2026-08-26
test setup: every xdist worker collects the whole suite, so collection is O(items x workers) behind a hard barrier


KonstantinMirin:
**Testing setup only — no production code changes.**

## Problem

Under xdist every worker collects the **entire** suite, then runs its slice. Collection cost is therefore `O(items × workers)` when it should be `O(items)`, and it sits on a **hard barrier**: xdist schedules nothing until every worker reports collection finished (`xdist/dsession.py::worker_collectionfinish` — "if the scheduler indicates collection is finished (i.e. all initial nodes have submitted their collections)").

Measured on `noetis-b-vm` (40 cores) running `bdd_inprocess` alone, 7223 items, via the per-worker profile in `tests/_worker_profile.py`:

| workers | import/worker | collect/worker | **barrier before any test runs** |
|---|---|---|---|
| 16 | 26.1 s | 93.4 s | **119.5 s** |
| 24 | 28.9 s | 101.6 s | **130.5 s** |
| 32 | 37.3 s | 122.8 s | **160.1 s** |
| 40 | 45.1 s | 146.6 s | **191.6 s** |

Each worker spends ~100–150 s collecting 7223 items in order to run roughly 200–450 of them.

Aggregate overhead as a share of all CPU spent climbs with worker count — 28% at 16, 35% at 24, 41% at 32, **44% at 40**. At 40 workers the box burns 5863 s of collection and 1802 s of import to run 9939 s of tests.

## This caps every other parallelism issue

#2007, #2008, #2009 and #2010 all **add workers**, and each new worker pays the barrier above. A worker-count sweep (`bdd_inprocess` alone) went 390 s at 24 workers to 459 s at 32 and 478 s at 40 — adding workers past ~24 made it *slower*. Note the wall-clock numbers are noisy because the box is shared (repeat samples at N=24 ranged 390–691 s); the per-worker barrier table above is the reliable signal, and it is monotone.

Three effects compound, all measured:

1. **Per-worker startup grows with N** — 119.5 s → 191.6 s. Workers contend to start.
2. **Collection grows superlinearly** — 2.5× the workers costs 3.9× the total collection.
3. **The same tests get slower** — an identical 7223 items cost 4816 s of test time at 16 workers and 9939 s at 40.

## The mechanism already exists in this repo

[...]

PR #2158, opened 2026-08-31
fix(test): stop the ui suite deleting the in-flight results directory, and add a per-worker timing profile

KonstantinMirin:
Adds a per-process timing profile for the test run, and fixes the bug it
immediately exposed: **the ui suite was deleting every other suite's results
mid-run, on every run.**

## The bug

`pytest-playwright` defaults `--output` to `test-results` and registers a
session-scoped autouse fixture that `shutil.rmtree`s it:

```python
@pytest.fixture(scope="session", autouse=True)
def delete_output_dir(pytestconfig):
    output_dir = pytestconfig.getoption("--output")   # default: "test-results"
    if os.path.exists(output_dir):
        shutil.rmtree(output_dir)
```

`[testenv:ui]` runs `pytest tests/ui/` with cwd `/app`, so that resolved to
`/app/test-results` — the same directory `run_all_tests.sh:181` uses for
`$RESULTS_DIR`. And `depends = e2e` guarantees unit, admin and e2e have all
finished by the time ui starts.

Caught by sampling the directory every 3 s during a full run:

    19:05:59   19 records  [unit x16, unit-main, admin-main, e2e-main]
    19:07:30    1 record   [ui-main]

The directory's own `Birth` timestamp equalled the ui session start — it was
removed and recreated, not emptied. Three consecutive runs lost the same three
suites; the survivors were always exactly those finishing after ui.

Per-suite JSON reports escaped only by accident: tox writes them to
`{toxworkdir}` and `run_all_tests.sh` copies them in *after* tox exits. Anything
written into `$RESULTS_DIR` *during* a run was destroyed.

This is the defect `run_all_tests.sh:405-410` already worked around without
naming — a second `mkdir -p` described as "insurance against `$RESULTS_DIR`
having been removed or never created for any reason".

**Fix:** `--output={envtmpdir}/playwright-artifacts`, so playwright's cleanup can
only reach its own scratch directory.

## The profile

`tests/_worker_profile.py` records, per process: `import` (interpreter + plugin
and conftest imports), `collect`, `tests` (that process's own test durations),
and `idle` (scheduling gaps and the tail while other workers still have work).
Off unless `PYTEST_WORKER_PROFILE` names a directory.

It exists because "wall minus busy" is a subtraction, not a measurement: it
lumps interpreter start, application import, collection and end-of-run idle into
one number and cannot say which dominates.

Records are keyed `<suite>-<workerid>.json`. The suite component is
load-bearing — xdist worker ids restart at `gw0` in every suite and `tox -p`
points them all at one directory, so without it suites overwrite each other.
Measured before the fix: two consecutive runs of an unchanged tree reported a
10/6 and then an 8/8 split between the only two suites whose records survived.

`summarise()` groups by suite rather than averaging across them, and reports the
xdist controller separately from the workers — it collects like they do but
executes nothing, so folding it in understates per-worker test time.

## What it is already good for

    suite            procs  import  collect   tests
    bdd-inprocess-s1    12    95.8    142.4   456.5
    integration         24    94.3     35.4   374.8
    unit                24     0.3     99.5   134.4

That decomposition is what shows the in-network pole is 61% test execution and
19% collection — see #2104 for where that leads.

## Review response

Four items from @agentzyx and @ChrisHuie, each with its own commit.

**The profile could fail a suite whose tests all passed** (`dbd0815bd`). The
`mkdir` + `write_text` pair at `pytest_sessionfinish` was unguarded, and an
exception out of that hook is an unhandled error. Measured: a session with all
six tests passing and a clean json-report exited 1 with no summary line. Now
`try` / `except OSError`, printing to stderr — `OSError` and not `Exception`, so
`PermissionError`, `FileExistsError`, `IsADirectoryError` and `ENOSPC` are all
caught while a defect in the record dict stays loud. A diagnostic must never
change the outcome of the thing it measures.

**The xdist controller was recording every worker's test time** (`d207406ba`).
`pytest_runtest_logreport` fires on the controller for every report a worker
sends back, so the controller's `tests_s` was the whole suite's execution time
attributed to a process that executes nothing. `record_test_duration` now takes
the report and returns early when `report.worker_id` is set — the attribute
xdist attaches on serialization to the controller.

**The module comment claimed a measurement it had not made** (`12128994b`).
`_T_IMPORT` is bound at module import, so everything before it — interpreter
start, plugin discovery — is invisible to the profile. The comment now states
that gap instead of implying the number covers it.

**`summarise()` had no caller anywhere in the tree** (`9bc244207`). The records
were written on every in-network run and nothing read them. Now
`scripts/ci/report_worker_profile.py` prints the per-suite table from
`run_all_tests.sh`, into the run log where CI and agents already look:

    Worker profile (per suite):
      suite                  wrk  collected  startup/wrk   tests_s   idle_s  slowest
      probe                    2          6          0.1       1.6      0.0      1.0

It gates on exactly one condition: a suite whose workers disagree on what they
collected. Under `--dist load` every worker collects the whole suite, so a
disagreement is a self-contradiction in the data — a collection or sharding
defect — and checking it needs no baseline. It deliberately sets **no** threshold
on startup or idle: a threshold is a ratchet, a ratchet needs a baseline, and a
baseline nobody agreed on turns an observation into a gate that fails for
reasons no one chose.

One more commit (`a584c6c09`) cherry-picks the `repo-invariants` hook to the
commit stage only. At pre-push the hook's file list is the whole pushed range,
so a push carrying a merge asks you to fix beads-id citations in files the
branch never touched — here, `tests/unit/test_creative_agent_connection_alias.py`
from #1802. It is the same commit as on #2189; whichever merges first, the other
drops it.

## Verification

`REMOTE_TEST_HOST=noetis-b-vm cassini run`, and `make quality` at 6582 passed /
0 failed after the review commits. The ui change is a path. The profile is
default-on for the in-network path (`run_all_tests.sh` names its directory),
which is why the write guard above is the ordinary case rather than a corner.


tests/conftest.py
```diff
@@ -804,6 +809,35 @@ def benchmark(request):
...
+def pytest_collection_finish(session):
+    _worker_profile.on_collection_finish(len(session.items))
+
+
+def pytest_runtestloop(session):
+    _worker_profile.on_loop_start()
+
+
+def pytest_runtest_logreport(report):
+    _worker_profile.record_test_duration(getattr(report, "duration", 0.0))
```
agentzyx: **Item 1 — the xdist controller's record carries the whole suite's test time.**

**Context:** under `-n`, the controller re-emits every worker's report through this hook; a report deserialized from a worker carries its originating worker id (`xdist/remote.py:286`). The controller also does not collect — `DSession.pytest_collection` returns `True` (`xdist/dsession.py:102-105`) and, registered at configure time, runs ahead of the rootdir conftest under pluggy's LIFO firstresult, so `on_collection_start` and `on_collection_finish` never fire there.

**Problem:** `tests/unit/test_adcp_contract.py` plus `tests/unit/test_run_all_tests_contract.py`, 105 tests, `TOX_ENV_NAME=unit`:

```
-n 2                                       -n 4
unit-gw0   tests_s 1.352                    sum(4 workers) tests_s  2.191
unit-gw1   tests_s 1.222                    unit-main      tests_s  2.191
sum        tests_s 2.574                    <- ratio 1.0
unit-main  tests_s 2.574   <- ratio 1.0
unit-main  collected 0  import_s 0.0  collect_s 0.0
```

The controller's `tests_s` is the suite's entire test time at both worker counts, so a raw read of the directory gives 5.148s against a true 2.574s. `summarise()` does exclude `main`, but it has no caller (item 4) and `run_all_tests.sh:199` now writes this directory into `$RESULTS_DIR` on every in-network run, so the raw records are the only artifact anyone actually receives.

The same subtraction corrupts `idle_s` at `_worker_profile.py:133`. The controller executed nothing, so its true idle is its whole loop, but the suite's test time is subtracted from it: measured `idle_s 4.567` against a `6.759`s loop, understated by exactly the `2.191` it never ran. Once execution rather than collection dominates a suite, summed test time exceeds the controller's loop duration and `max(0.0, ...)` flattens it to `0.0` instead — reproduced in a high-utilisation run.

The rationale at `_worker_profile.py:163-166` is inverted in both halves: the controller does not "collect like they do" (`collected = 0`) and it does not leave test time out. So `summary["controller_collect_s"]` at `:212` is `0.0` by construction — measured `0.0` for both suites in a real two-suite directory — and carries no information. `controller_wall_s` is the one that does.

**Fix** (applied and re-run, not just proposed):

```diff
 def pytest_runtest_logreport(report):
-    _worker_profile.record_test_duration(getattr(report, "duration", 0.0))
+    # On the controller every worker's report is re-emitted here, carrying its
+    # originating worker id; counting those makes this record the whole suite.
+    if getattr(report, "worker_id", None) is None:
+        _worker_profile.record_test_duration(getattr(report, "duration", 0.0))
```

```diff
-        "idle_s": round(max(0.0, (finish - loop_start) - _test_seconds), 3),
+        "idle_s": round((finish - loop_start) - _test_seconds, 3),
```

With both applied: controller `tests_s` becomes `0.0`, controller `idle_s` becomes its whole loop (`7.822` at `-n 2`, `6.727` at `-n 4`), a naive sum over the directory equals the true value, no worker residual goes negative once the clamp is gone (worker `idle_s` ranged `0.02`–`0.11`), 105 passed at both worker counts, and both files stay `ruff format`/`ruff check` clean.

Worth noting the choice this leaves open: with the guard the controller still writes an all-zero record, which is honest but arguably better omitted, or marked with an explicit role. And whichever discriminator you pick — `report.worker_id`, or `hasattr(config, "workerinput")` / `pluginmanager.hasplugin("dsession")` at session finish, both of which I confirmed identify the controller — it keys on xdist behaviour under a floor-only pin (`pytest-xdist>=3.5.0`), so an xdist upgrade can silently restore the double-count. That is the concrete reason item 4's test is worth having.

tests/_worker_profile.py
```diff
@@ -0,0 +1,214 @@
...
+from __future__ import annotations
+
+import json
+import os
+import time
+from pathlib import Path
+from typing import Any
+
+# Import time is the earliest point this plugin can observe. Everything before it
+# (interpreter boot, pytest's own bootstrap) is invisible and is reported
```
agentzyx: **Item 3 — this sentence is not implemented.**

**Problem:** the comment says interpreter boot and pytest's own bootstrap are "reported separately as the gap against the process's own start time where available". Nothing measures process start. A grep of the module for `psutil`, `create_time`, `/proc`, `boot_time`, `os.times`, `clock_gettime`, `process_time` and `start_time` returns zero hits, and the twelve keys the module writes — `suite`, `worker`, `collected`, `import_s`, `collect_s`, `tests_s`, `idle_s`, `worker_wall_s`, and the four `summarise()` fields — contain no such gap. `worker_wall_s` is `finish - _T_IMPORT`, i.e. measured from module import.

Since the argument for the module is that "wall minus busy" lumps interpreter start, application import, collection and idle into one number, an unimplemented claim to have split interpreter start back out reproduces that ambiguity inside the tool built to remove it: anyone comparing `slowest_worker_wall_s` against suite wall charges interpreter boot and every plugin imported before the rootdir conftest to scheduling idle. That matters directly for #2104, where the per-worker startup figure is the quantity being weighed against worker count.

**Fix:** measure it from `/proc/self/stat` field 22 against boot time — Linux-only, which "where available" already hedges, and the in-network box is the only place the number matters — or state plainly that it is unmeasured and outside `worker_wall_s`. Note `psutil` is not currently a dependency, so the `Process().create_time()` route would add one for this alone.

tests/_worker_profile.py
```diff
@@ -0,0 +1,214 @@
...
+        "import_s": round(collect_start - _T_IMPORT, 3),
+        "collect_s": round(collect_finish - collect_start, 3),
+        "tests_s": round(_test_seconds, 3),
+        "idle_s": round(max(0.0, (finish - loop_start) - _test_seconds), 3),
+        "worker_wall_s": round(finish - _T_IMPORT, 3),
+    }
+    (out / f"{record['suite']}-{record['worker']}.json").write_text(json.dumps(record), encoding="utf-8")
+
+
+def summarise(directory: str | Path) -> dict[str, Any]:
```
agentzyx: **Item 4 — nothing reads the records, and nothing tests the code that reads them.**

**Problem:** `summarise` has zero references outside this module across every `.py`, `.sh`, `.md`, `.ini`, `.toml`, `.yml`, `.yaml` and `.cfg` file in the tree. Every other mention of the profile is a writer (the five `tests/conftest.py` hooks) or a path. No test file references the module at all. So as of `a26d54e0` the profile writes a record directory into `$RESULTS_DIR` on every in-network run and nothing in the repo consumes it.

That output is not idle, though — which is what makes this worth doing now rather than later. #2104's table of per-worker import/collect and "barrier before any test runs" at 16/24/32/40 workers is this module's output (the barrier column is `startup_per_worker_s` at `:207` — 26.1 + 93.4 = 119.5 checks out), and its acceptance criteria name the profile as the instrument: "Keep the per-worker profile reporting, so the before/after is measured rather than asserted." Four other issues are gated on those numbers. The reading is a manual step, and item 1 is precisely a case where the raw records give twice the right answer.

The ungraded invariants are the ones this branch's own history says were already wrong once: grouping by suite (`b03587ca` fixed a 10/6-then-8/8 disagreement between two runs of an unchanged tree), `collected_values` as a set replacing the `workers[0]["collected"]` assumption "that answered for records it never read", the controller split, and the `parallel or records` fallback at `:178`. I verified the first two work today — two suites in one directory group correctly, and the controller stays out of `tests_s_total` — which is the argument for pinning them rather than against: a regression is silent because nothing executes this code, and that is why the 10/6 disagreement survived two full runs before anyone noticed.

**Fix:** a `tests/unit/test_worker_profile.py` over synthetic record directories — two suites in one directory, a controller plus workers, workers whose `collected` disagree (`collected_per_worker is None`), and a serial single-`main` suite. Pure stdlib, no infrastructure, and it is what would catch an xdist upgrade breaking the item 1 discriminator. Then give the profile one reader — a `scripts/` entry point, or a line in `run_all_tests.sh` printing `summarise("$RESULTS_DIR/worker-profile")` — so the default-on write has something on the other end and #2104's numbers come from code rather than a REPL.

agentzyx (commented):
​

tests/conftest.py
ChrisHuie: **Item 1 — the xdist controller's record carries the whole suite's test time.**

**Context:** under `-n`, the controller re-emits every worker's report through this hook; a report deserialized from a worker carries its originating worker id (`xdist/remote.py:286`). The controller also does not collect — `DSession.pytest_collection` returns `True` (`xdist/dsession.py:102-105`) and, registered at configure time, runs ahead of the rootdir conftest under pluggy's LIFO firstresult, so `on_collection_start` and `on_collection_finish` never fire there.

**Problem:** `tests/unit/test_adcp_contract.py` plus `tests/unit/test_run_all_tests_contract.py`, 105 tests, `TOX_ENV_NAME=unit`:

```
-n 2                                       -n 4
unit-gw0   tests_s 1.352                    sum(4 workers) tests_s  2.191
unit-gw1   tests_s 1.222                    unit-main      tests_s  2.191
sum        tests_s 2.574                    <- ratio 1.0
unit-main  tests_s 2.574   <- ratio 1.0
unit-main  collected 0  import_s 0.0  collect_s 0.0
```

The controller's `tests_s` is the suite's entire test time at both worker counts, so a raw read of the directory gives 5.148s against a true 2.574s. `summarise()` does exclude `main`, but it has no caller (item 4) and `run_all_tests.sh:199` now writes this directory into `$RESULTS_DIR` on every in-network run, so the raw records are the only artifact anyone actually receives.

The same subtraction corrupts `idle_s` at `_worker_profile.py:133`. The controller executed nothing, so its true idle is its whole loop, but the suite's test time is subtracted from it: measured `idle_s 4.567` against a `6.759`s loop, understated by exactly the `2.191` it never ran. Once execution rather than collection dominates a suite, summed test time exceeds the controller's loop duration and `max(0.0, ...)` flattens it to `0.0` instead — reproduced in a high-utilisation run.

The rationale at `_worker_profile.py:163-166` is inverted in both halves: the controller does not "collect like they do" (`collected = 0`) and it does not leave test time out. So `summary["controller_collect_s"]` at `:212` is `0.0` by construction — measured `0.0` for both suites in a real two-suite directory — and carries no information. `controller_wall_s` is the one that does.

**Fix** (applied and re-run, not just proposed):

```diff
 def pytest_runtest_logreport(report):
-    _worker_profile.record_test_duration(getattr(report, "duration", 0.0))
+    # On the controller every worker's report is re-emitted here, carrying its
+    # originating worker id; counting those makes this record the whole suite.
+    if getattr(report, "worker_id", None) is None:
+        _worker_profile.record_test_duration(getattr(report, "duration", 0.0))
```

```diff
-        "idle_s": round(max(0.0, (finish - loop_start) - _test_seconds), 3),
+        "idle_s": round((finish - loop_start) - _test_seconds, 3),
```

With both applied: controller `tests_s` becomes `0.0`, controller `idle_s` becomes its whole loop (`7.822` at `-n 2`, `6.727` at `-n 4`), a naive sum over the directory equals the true value, no worker residual goes negative once the clamp is gone (worker `idle_s` ranged `0.02`–`0.11`), 105 passed at both worker counts, and both files stay `ruff format`/`ruff check` clean.

Worth noting the choice this leaves open: with the guard the controller still writes an all-zero record, which is honest but arguably better omitted, or marked with an explicit role. And whichever discriminator you pick — `report.worker_id`, or `hasattr(config, "workerinput")` / `pluginmanager.hasplugin("dsession")` at session finish, both of which I confirmed identify the controller — it keys on xdist behaviour under a floor-only pin (`pytest-xdist>=3.5.0`), so an xdist upgrade can silently restore the double-count. That is the concrete reason item 4's test is worth having.

tests/_worker_profile.py
```diff
@@ -0,0 +1,214 @@
...
+        global _test_seconds
+        _test_seconds += seconds or 0.0
+
+
+def on_session_finish() -> None:
+    if not enabled:
+        return
+    _mark("finish")
+    out = Path(_OUT_DIR)
+    out.mkdir(parents=True, exist_ok=True)
```
ChrisHuie: **Item 2 — an unwritable profile path fails a suite whose tests all passed.**

**Context:** `a26d54e0` made the profile default-on at `run_all_tests.sh:199` for the in-network path, and that path is taken by `ci.yml:379` (`run_all_tests.sh bdd_e2e`) and `:456` (`run_all_tests.sh storyboard`), not only by a local sweep.

**Problem:** `on_session_finish()` writes with no exception guard, so an I/O failure propagates out of `pytest_sessionfinish` as an unhandled traceback. Runs of `tests/unit/test_run_all_tests_contract.py` at this head, each reaching `[100%]` with all 6 tests passing and a `--json-report` of `{'passed': 6, 'total': 6, 'collected': 6}`:

- profile dir under a non-writable parent → `PermissionError: [Errno 13] Permission denied` from this `mkdir`, **pytest exit 1**, no summary line.
- profile path occupied by a regular file → `FileExistsError: [Errno 17]`, **pytest exit 1**.
- control, identical command with a writable directory → **exit 0**.

The report is still written, so the run lands exactly in the case `run_all_tests.sh:535-560` has to reason about: a nonzero exit where every suite report is clean. The permission variant is not hypothetical here — a bind-mounted path created by a different uid is the failure documented at `run_all_tests.sh:249-250` and `:451-456`, and a full volume produces the same class.

**Fix** (applied and re-run):

```diff
     out = Path(_OUT_DIR)
-    out.mkdir(parents=True, exist_ok=True)
+    try:
+        out.mkdir(parents=True, exist_ok=True)
+    except OSError:
+        return
```

plus the same guard around the `write_text` at `:136`. With both in place the non-writable-parent run goes to **exit 0**, 6 passed, no traceback, and the happy path still writes `unit-main.json`. A measurement should not be able to fail the run it measures.

tests/_worker_profile.py
ChrisHuie: **Item 3 — this sentence is not implemented.**

**Problem:** the comment says interpreter boot and pytest's own bootstrap are "reported separately as the gap against the process's own start time where available". Nothing measures process start. A grep of the module for `psutil`, `create_time`, `/proc`, `boot_time`, `os.times`, `clock_gettime`, `process_time` and `start_time` returns zero hits, and the twelve keys the module writes — `suite`, `worker`, `collected`, `import_s`, `collect_s`, `tests_s`, `idle_s`, `worker_wall_s`, and the four `summarise()` fields — contain no such gap. `worker_wall_s` is `finish - _T_IMPORT`, i.e. measured from module import.

Since the argument for the module is that "wall minus busy" lumps interpreter start, application import, collection and idle into one number, an unimplemented claim to have split interpreter start back out reproduces that ambiguity inside the tool built to remove it: anyone comparing `slowest_worker_wall_s` against suite wall charges interpreter boot and every plugin imported before the rootdir conftest to scheduling idle. That matters directly for #2104, where the per-worker startup figure is the quantity being weighed against worker count.

**Fix:** measure it from `/proc/self/stat` field 22 against boot time — Linux-only, which "where available" already hedges, and the in-network box is the only place the number matters — or state plainly that it is unmeasured and outside `worker_wall_s`. Note `psutil` is not currently a dependency, so the `Process().create_time()` route would add one for this alone.

tests/_worker_profile.py
ChrisHuie, line 198: **Item 4 — nothing reads the records, and nothing tests the code that reads them.**

**Problem:** `summarise` has zero references outside this module across every `.py`, `.sh`, `.md`, `.ini`, `.toml`, `.yml`, `.yaml` and `.cfg` file in the tree. Every other mention of the profile is a writer (the five `tests/conftest.py` hooks) or a path. No test file references the module at all. So as of `a26d54e0` the profile writes a record directory into `$RESULTS_DIR` on every in-network run and nothing in the repo consumes it.

That output is not idle, though — which is what makes this worth doing now rather than later. #2104's table of per-worker import/collect and "barrier before any test runs" at 16/24/32/40 workers is this module's output (the barrier column is `startup_per_worker_s` at `:207` — 26.1 + 93.4 = 119.5 checks out), and its acceptance criteria name the profile as the instrument: "Keep the per-worker profile reporting, so the before/after is measured rather than asserted." Four other issues are gated on those numbers. The reading is a manual step, and item 1 is precisely a case where the raw records give twice the right answer.

The ungraded invariants are the ones this branch's own history says were already wrong once: grouping by suite (`b03587ca` fixed a 10/6-then-8/8 disagreement between two runs of an unchanged tree), `collected_values` as a set replacing the `workers[0]["collected"]` assumption "that answered for records it never read", the controller split, and the `parallel or records` fallback at `:178`. I verified the first two work today — two suites in one directory group correctly, and the controller stays out of `tests_s_total` — which is the argument for pinning them rather than against: a regression is silent because nothing executes this code, and that is why the 10/6 disagreement survived two full runs before anyone noticed.

**Fix:** a `tests/unit/test_worker_profile.py` over synthetic record directories — two suites in one directory, a controller plus workers, workers whose `collected` disagree (`collected_per_worker is None`), and a serial single-`main` suite. Pure stdlib, no infrastructure, and it is what would catch an xdist upgrade breaking the item 1 discriminator. Then give the profile one reader — a `scripts/` entry point, or a line in `run_all_tests.sh` printing `summarise("$RESULTS_DIR/worker-profile")` — so the default-on write has something on the other end and #2104's numbers come from code rather than a REPL.

ChrisHuie (commented):
# Review — #2158 (verified at head `b03587ca`, review state pulled 2026-09-02T16:35Z)

The `ui` diagnosis holds, and I checked it against the installed package rather than its docs: `pytest_playwright.py:462-464` defaults `--output` to `test-results`, and `delete_output_dir` at `:68-79` is `@pytest.fixture(scope="session", autouse=True)` calling `shutil.rmtree(output_dir)`. The fix's blast radius is complete — `pytest-playwright` is only in the `ui-tests` extra (`pyproject.toml:66-68`), `extras = ui-tests` appears once in `tox.ini` (`:140`), and `pytest tests/ui/` has exactly one invocation in the tree (`tox.ini:163`), so no other suite loads the plugin and no other env needs the flag. Nothing in `tests/ui/` references the output path, so relocating it cannot break an assertion there, and the fixture now cleans only its own scratch directory.

The surrounding claims check out too. `/app/$RESULTS_DIR` resolves through the `tests` service's `- .:/app` bind mount (`docker-compose.e2e.yml:189`, `:356-357`). Both readers of `$RESULTS_DIR` use non-recursive `*.json` globs (`scripts/check_truncated_reports.py:36`, `run_all_tests.sh:539`), so the new `worker-profile/` subdirectory is inert to them. `TOX_ENV_NAME` needs no `pass_env` entry and cannot be shadowed — `tox/tox_env/api.py:418` assigns it after `_load_pass_env` (`:401`) and after the `set_env` loop (`:409-411`). And the suite-qualified record name does what it claims: running two `TOX_ENV_NAME`s into one directory, `summarise()` groups them separately with no collision and correctly keeps the controller out of `tests_s_total`.

One thing the description undersells. The retention prune that keeps the last ten run directories lives only in `run_all_tests_host.sh:27`, and the rmtree took `test-results` itself rather than its contents — so every in-network run also destroyed all previously archived run directories, not just the in-flight ones.

Four items, all in `tests/_worker_profile.py`, posted as threads on the lines they concern. Each was reproduced against this head in a clean checkout with the real `tests/conftest.py`, and for the first two I applied the proposed fix and re-ran rather than only proposing it.

1. The xdist controller's record carries the whole suite's test time — the persisted directory double-counts, and `idle_s` is wrong by the same amount.
2. An unwritable profile path fails a suite whose tests all passed, with the profile now on by default.
3. The one sentence bounding what the profile cannot see is not implemented.
4. Nothing reads the records, and nothing tests the code that reads them — while #2104's acceptance criteria depend on that output being read correctly.

Items 1 and 2 are defects in numbers and exit codes. Items 3 and 4 are what makes them durable: the module's output already appears in #2104 as the evidence for capping worker counts, and no mechanism goes red if it drifts.

---

**Ran:** clean detached checkout at this head (`HEAD` asserted, tree clean, added lines confirmed present). Provenance — `tests/_worker_profile.py` exists neither at the merge base (`1d93e372`) nor on current `main`; `git log -S` puts items 1, 3, 4 and the unguarded write in `bb2143e6`, and the default-on flip in `a26d54e0`. Behaviour — `tests/unit/test_adcp_contract.py` + `tests/unit/test_run_all_tests_contract.py` (105 tests) under `-n 2` and `-n 4` with `PYTEST_WORKER_PROFILE` and `TOX_ENV_NAME` set, records compared worker-against-controller, before and after the item 1 and 2 patches; the same slice under a non-writable parent, an occupied path, and a writable control; two `TOX_ENV_NAME`s through `summarise()`; `ruff format --check` and `ruff check` on both patched files. Sources — `pytest-playwright==0.7.2`, `xdist/dsession.py`, `xdist/remote.py`, `tox/tox_env/api.py`, `configparser` parse of the head `tox.ini`; tree-wide greps for callers, readers and tests of the module.

**Not run:** `tox -e ui` and any full in-network run — the `--output` change is verified from the plugin source and the parsed `tox.ini`, not by executing the ui suite. No access to the remote box, so the 63-record sweep, the nine-name forwarding tuple and the sampled 19-records-to-1 timeline are taken as measured, unreproduced. Slices ran against the repo's existing virtualenv rather than a tox-built one; these items touch `tests/` only. Did not check whether the CI workspaces retain the profile directory. `mypy` covers `src/` only, so the new module is not type-checked by the Type Check job.

tests/conftest.py
KonstantinMirin: Both diffs applied, in two commits.

`d207406ba` takes the `report.worker_id` discriminator, in the form you suggested — `record_test_duration` now receives the report and returns early when the id is set.

`4af6e0a33` removes the `max(0.0, ...)` clamp, which the earlier commit had missed. Your argument for it holds independently of the controller: a worker runs its tests one at a time, so their summed durations fit inside the loop containing them, and the residual therefore cannot legitimately go negative. A clamp on a value that cannot legitimately go negative can only ever hide an accounting defect — and the controller double-count was exactly such a defect, rendered as a plausible `0.0` rather than the negative number that would have exposed it.

The same commit drops `controller_collect_s`, for the reason you gave: `DSession.pytest_collection` returns `True` and, registered at configure time, runs ahead of the rootdir conftest under pluggy's LIFO firstresult, so the collection hooks never fire on the controller and the field is `0.0` by construction. A field that is always the same constant reports nothing. `controller_wall_s` stays.

On the choice you left open — I kept the controller's record rather than omitting it, because `controller_wall_s` is derived from that record and it is the one field on it that carries information. Omitting the record would take that with it.

On the xdist pin: agreed, and that is why `tests/unit/test_worker_profile_write_guard.py` pins the discriminator directly — one test asserts a re-emitted worker report is not counted, another that the process's own report is. An upgrade that changes when `worker_id` is attached fails those rather than silently restoring the double-count.

tests/_worker_profile.py
KonstantinMirin: Fixed in `dbd0815bd`, with one deliberate difference from the suggested diff.

The guard wraps both the `mkdir` and the `write_text` in a single `try`, and prints to stderr rather than returning silently. A silent return makes a missing record indistinguishable from a suite that never ran, which is the failure mode this directory exists to rule out — you should be able to tell "the profile could not be written" from "nothing measured this".

`OSError` rather than `Exception`, so `PermissionError`, `FileExistsError`, `IsADirectoryError` and `ENOSPC` are all caught, while a defect in the record dict above stays loud instead of being hidden behind a guard meant for a read-only disk.

Three tests in `tests/unit/test_worker_profile_write_guard.py` cover it: a non-writable parent, a path occupied by a regular file, and the happy path still writing the record — the last one because a guard that turns the writer into a no-op would also pass the first two.

Your framing is the one I took as the rule: a diagnostic must not change the outcome of the thing it measures.

tests/_worker_profile.py
KonstantinMirin: Took the second option, in `12128994b`: the comment now states that the pre-import gap is unmeasured and outside `worker_wall_s`, instead of claiming to report it.

I did not measure it. `/proc/self/stat` would be Linux-only, and while you are right that the in-network box is where the number matters, the module also runs on macOS dev machines — a field present on one platform and absent on the other makes `summarise`'s output platform-dependent for the sake of one number. `psutil` would add a dependency for this alone.

So the limitation stands as you described it, now stated rather than implied: `startup_per_worker_s` is import plus collect measured from module import, and anything before that — interpreter boot, plugin discovery — is charged to neither startup nor idle, it is simply outside the record. Anyone comparing `slowest_worker_wall_s` against suite wall should read the remainder as orchestration *plus* that gap.

If you want it measured for #2104 rather than documented, say so and I will do it as a follow-up — it is a real number and I would rather add it deliberately than fold it into this PR.

tests/_worker_profile.py
KonstantinMirin: Both halves done.

**The reader** (`9bc244207`): `scripts/ci/report_worker_profile.py`, called from `run_all_tests.sh` after the reports are collected, printing the per-suite table into the run log where CI and agents already look:

```
Worker profile (per suite):
  suite                  wrk  collected  startup/wrk   tests_s   idle_s  slowest
  probe                    2          6          0.1       1.6      0.0      1.0
```

It gates on exactly one condition: a suite whose workers disagree on what they collected. Under `--dist load` every worker collects the whole suite, so a disagreement is a self-contradiction in the data — a collection or sharding defect — and checking it needs no baseline.

It sets no threshold on startup or idle. A threshold is a ratchet, a ratchet needs a baseline, and a baseline nobody has agreed on turns an observation into a gate that fails for reasons no one chose. The numbers get printed; only the contradiction fails.

**The tests** (`4af6e0a33`): `tests/unit/test_worker_profile_summarise.py`, seven tests over synthetic directories, covering the four cases you listed — two suites in one directory, controller plus workers, workers whose `collected` disagree, and a serial single-`main` suite — plus per-worker startup against slowest wall, the agreeing case, and the empty directory. Pure stdlib, no infrastructure.

Your reason for wanting them is the one that convinced me: these invariants were already wrong once and nothing noticed. The 10/6-then-8/8 disagreement survived two full runs precisely because no code executed this path.

Three more in `test_report_worker_profile.py` cover the reader's gate, and five in `test_worker_profile_write_guard.py` cover the write guard and the item 1 discriminator.

PR #2167, opened 2026-09-01
fix: resolve the tenant's own AI configuration once, and say when ranking is skipped

RATO-tdunkel:
A seller who configures AI in the Admin UI gets no AI product ranking. Products come back
in catalog order, the response says nothing about it, and the only trace is a `debug` log
line. The symptom a publisher reports is "the ranking prompt has no effect".

`_get_products_impl` called the AI factory with no arguments, so `is_ai_enabled()` and
`create_model()` saw only the platform-level environment key (`GEMINI_API_KEY` and
friends, via `get_platform_defaults()`). The Admin UI writes to `tenants.ai_config`. The
two never met. `ai_config` was also absent from both tenant projections
(`TenantContext`, `serialize_tenant_to_dict`), so the value could not have reached the
call even if it were passed on.

The first version of this PR passed `tenant.get("ai_config")` into the two factory calls.
Review found that this wrote the resolution rule a fourth time, converted one of the two
AI consumers in the function it edited, and left the silent-skip half of the problem
untouched. The change below is the rework.

## The fix

`TenantAIConfig` gains the two class methods that own the rule, in the module that owns
the model (`src/services/ai/config.py`):

- **`coerce(value, *, source=...)`** — the one parse of a stored AI configuration value.
  The factory's three parse sites and `naming.py` route through it.
- **`from_tenant(tenant)`** — answers "given a tenant, what is its AI configuration?"
  once. Accepts the ORM row, `TenantContext`/`LazyTenantContext`, and the plain tenant
  dict; the field names are identical across all three and only the access mechanism
  differs, which is what `_read_tenant_field` handles.

Resolution order is `ai_config` carrying a non-empty `api_key`, then the legacy
`gemini_api_key` column, then `None`. Both consumers in `_get_products_impl` — ranking
and the policy check — go through it, as do `naming.py` and both AI-availability sites in
`admin/blueprints/creatives.py`.

Usability, not truthiness, is the contract. `admin/blueprints/settings.py:558-561` writes
`{"provider": ..., "model": ...}` with no `api_key` when the seller leaves the key field
blank, and flashes "AI features will be disabled" as it does. Returning that row as a real
configuration opened the policy gate for those tenants and ran `check_brief_compliance` on
the *operator's* platform credential — a live LLM call, and a BLOCKED verdict there raises
`AdCPPolicyViolationError`, a buyer-visible rejection the seller never asked for.

Two further defects were found while reworking and are fixed at the same seam:

- **The platform-key fallback was provider-blind.** `create_model` resolved provider and
  api_key from independent expressions, so a tenant configured for `anthropic` with no key
  of its own got an `AnthropicModel` holding the platform's *Google* credential — the
  operator's Gemini key sent to `api.anthropic.com`. `get_products` is auth-optional
  discovery, so an anonymous brief reaches it. The platform provider, model and key now
  travel together: they fill in only when the resolved provider is the same vendor
  (`factory.py:130-136`).
- **The legacy key was paired with the platform's model.** With `PYDANTIC_AI_MODEL` set to
  an Anthropic model, a tenant on the legacy `gemini_api_key` column got a `GoogleModel`
  named `claude-sonnet-4-…`, and every call 404'd. `from_tenant` pins
  `DEFAULT_GOOGLE_MODEL` with the provider — the same pin `policy_check_service.py:67`
  hard-codes at its own call site.

### 1. The policy consumer in the same function still read `gemini_api_key`

Fixed. Both consumers resolve through `TenantAIConfig.from_tenant`, the deprecated
`gemini_api_key=` kwarg is gone from the `PolicyCheckService` call, and the disabled
reason is provider-neutral.

The gate itself needed more than a substitution. `from_tenant(...) is not None` is the
policy gate; `factory.is_ai_enabled(...)` is the ranking gate. They are deliberately
different questions: ranking may fall back to a deployment-wide platform key (existing,
relied-upon behaviour — a `GEMINI_API_KEY` ranks for every tenant with a ranking prompt),
while policy screening on the operator's credential for a tenant the UI called disabled is
the defect described above.

### 2. Nothing owned tenant-to-`TenantAIConfig` resolution

Fixed as described under **The fix**. The four divergent copies the review tabulated are
now one call each. The absent-test divergence is resolved in favour of falsiness, so an
empty `ai_config` dict falls through to the legacy column instead of shadowing it;
`naming.py` was the lone `is None` site.

Not adopted: typing `TenantContext.ai_config` as `TenantAIConfig | None` and coercing at
the ORM boundary. `from_tenant` reads the tenant in whichever shape it arrives, so the
projections stay as they are and the two two-line plumbing hunks remain (they are what
carry the column into the dict `_get_products_impl` receives). Retyping the column model
is a separate change with its own migration and admin-write surface.

### 3. A malformed `ai_config` row aborted the call — but not with a 500

Fixed, and **the reviewer's diagnosis of the symptom was wrong**. `pydantic.ValidationError`
subclasses `ValueError`, and `normalize_to_adcp_error` (`src/core/exceptions.py:1360`)
maps every `ValueError` to `AdCPValidationError` at all three transport boundaries. So the
buyer did not get a 500. Reproduced against the tree:

```
$ uv run python -c "...TenantAIConfig.model_validate({'settings':{'temperature':9}})..."
raised     : ValidationError | isinstance ValueError: True
normalized : AdCPValidationError
error_code : VALIDATION_ERROR
recovery   : correctable
field      : settings.temperature
status_code: 400
```

That is worse than a 500 for the buyer, not better: a 400 with `recovery: "correctable"`
tells an autonomous agent to fix its request and resend, and names `settings.temperature`
— a field that does not exist in the `get_products` request schema (verified against the
pinned `get-products-request.json`, whose 18 top-level properties contain no `settings`).
The buyer is told to correct something it never sent and cannot reach.

`coerce` fixes this at the parse boundary rather than by widening a caught tuple, because
two failure modes arrive there and they need opposite answers:

- **Malformed stored data** (a dict or list from the JSONB column that fails the model) —
  the seller's row is broken, the buyer's request is not. Degrades to a bare
  `TenantAIConfig()` with one WARNING, and `from_tenant` then declines to return it, so it
  falls through to the legacy column and then to `None`. It is never returned as if it
  were the tenant's configuration. Returning it let the factory fill in the *platform's*
  provider and key, so a seller who deliberately chose Anthropic had the buyer's brief and
  product catalog sent to Google.
- **A caller passing the wrong Python type** (a `str`, an `int`) — `json_type.py` raises
  `TypeError` for anything that is not a dict or list, so these shapes cannot come from
  the column. They come from a caller passing the wrong argument. That raises `TypeError`
  naming the parameter and the offending type.

This is also why `tests/unit/test_naming_parameter_bug.py`'s assertion is **restored**. It
had been rewritten from `pytest.raises(AttributeError)` to expect "one WARNING", which
pinned a caller bug as an acceptable degradation — the naming call then ran on somebody
else's AI configuration and returned a plausible name. The exception type moved
(`TypeError` at the parse boundary rather than `AttributeError` several frames deeper),
but "a caller mistake raises" is the contract and it is back.

### 4. The tests did not reach the changed behaviour

Fixed, and re-verified by mutation rather than by inspection.

A new local feature, `tests/bdd/features/BR-UC-001-ai-ranking-tenant-config.feature`,
grades BR-RULE-005 across all four transports. Both scenarios remove the platform key in
their Given step (`tests/conftest.py` sets it suite-wide, and `is_ai_enabled` returns true
on that key alone), so the tenant's own configuration is the only AI configuration in
play and neither outcome is reachable without reading it. Wiring:

- `UC-001` gains an `ENV_ROUTES` row (`tests/bdd/conftest.py:3829`) over the existing
  `ProductEnv`. It had a harness all along; no row claimed the bucket, so a `@T-UC-001-*`
  scenario xfailed as "No harness wired for UC-001". `EXPECTED_WIRED_ROUTES` in
  `test_architecture_measurement_floors.py` moves 25 → 26 to match.
- The two `ProductEnv` setup methods that cannot be realized over e2e are declared in
  `EXPECTED_UNSUPPORTED_DECLARATIONS` with their reasons at the env methods, not as a
  nodeid ledger entry or an xfail tag: the platform `GEMINI_API_KEY` lives in the live
  server's own process environment (and is cached there by `get_factory`'s `lru_cache`),
  and scripted relevance scores have no live-server injection surface.

The ranking call's two free-text arguments are now graded. `rank_products_async` takes the
seller's ranking prompt as `custom_prompt` and the buyer's brief as `brief`; the prompt
builder concatenates both, so swapping them changes only the wording sent to the model and
no wire assertion can see it. Mutation-verified before the fix: with the two exchanged in
production, the BDD scenarios and the whole of `test_get_products_impl_coverage.py` stayed
green. The harness stub now checks which text arrived as which argument, and the unit test
asserts `rank.assert_awaited_once_with(...)` rather than `await_count == 1`.

`tests/integration/test_tenant_utils.py` gains the `ai_config` leg, so the two projection
hunks fail on revert.

### 5. The silent skip, and the `errors[]` advisory

Fixed. This is the buyer-visible part of the change: `GetProductsResponse.errors[]` is now
populated where it was hard-coded to `None`.

The ranking block moved out of `_get_products_impl` into `_rank_products_with_ai`, which
appends exactly one advisory on every path where ranking was configured and did not
happen. Owning the failure paths inside that function, rather than in a `try` at the call
site, is what makes "every" true: the old handler logged a warning and dropped through
with advisories untouched, so a provider 401/429/5xx returned a plausible-looking
catalog-order list with `errors=None`.

Two codes, because the two conditions need opposite advice:

- **`CONFIGURATION_ERROR`** (`recovery: terminal`) when nothing usable resolved, or when
  `create_model` refuses the seller's configuration. No retry at any backoff supplies a
  missing API key.
- **`SERVICE_UNAVAILABLE`** (`recovery: transient`) when the ranking call did not
  complete. pydantic-ai's `ModelHTTPError`, `UnexpectedModelBehavior`,
  `UsageLimitExceeded` and `UserError` all subclass `RuntimeError`, so every provider auth
  failure, 429, timeout and malformed response lands here.

`recovery` is derived from the code by `_ADVISORY_RECOVERY` rather than passed alongside
it: a call site free to pair them is a call site free to tell a buyer to retry a missing
API key. The `cause` clause never interpolates `str(exc)` — the message is wire-visible to
an anonymous buyer, and provider exception text carries request URLs, response bodies and
key fragments. The exception class name is enough to classify; the full text goes to the
seller's log.

**Where this departs from the review.** The review said "Do not use `CONFIGURATION_ERROR`:
in 3.1.1 it is `terminal` and sellers MUST flip transport failure markers, which is the
wrong shape for a response that still returns products." The wire-placement clause in the
pinned `enums/error-code.json` reads in full: "**Wire placement.** The deployment cannot
produce a success artifact, so sellers MUST flip transport-level failure markers (HTTP
5xx, MCP `isError: true`, A2A `failed`)…". Its antecedent does not hold here — the
deployment *can* produce a success artifact, and does: the catalog comes back. The
in-repo precedent is already merged and does exactly this:
`src/core/tools/media_buy_list.py:566-591` emits `CONFIGURATION_ERROR` with
`recovery: "terminal"` on `GetMediaBuysResponse.errors[]` of a completed, data-bearing
response. The review's underlying point is taken where it applies: the transient
condition, which is the one that actually happens in production, gets
`SERVICE_UNAVAILABLE`, and the two conditions are split rather than sharing one code.

A platform-specific code was considered and rejected. `core/error.json` does declare the
vocabulary open ("senders MAY emit codes outside that set"), so `PRODUCT_RANKING_UNAVAILABLE`
would be spec-legal. This repo is deliberately stricter: `verify_feature_error_codes.py`
states the policy — "the AdCP spec accepts arbitrary code strings, but our scenarios are
derived from the pinned spec and must stay on the standard vocabulary" — and
`tests/unit/test_architecture_error_code_compliance.py` enforces it by AST-scanning
`Error(code=...)` sites in `src/core/tools/`, which is where this advisory is built. No
canonical code expresses "an optional enhancement was unavailable, seller-side, on an
otherwise successful response"; `UNSUPPORTED_FEATURE` is `correctable`, which tells the
buyer to fix something only the seller can. `CONFIGURATION_ERROR` carries the correct
recovery classification, and recovery is what the spec tells receivers to act on. The
vocabulary gap is worth raising upstream; inventing a local code to paper over it is not
this PR's call to make.

**A2A reported `success=false` for every advisory-carrying response.**
`_stamp_a2a_protocol_fields` derived `success = not bool(errors)`, and
`_handle_get_products_skill` (`adcp_a2a_server.py:1673`) runs every `get_products`
response through it at line 1704. Before this
change `errors` was always `None`, so `success` was always `True`; with the advisory, a
tenant would report `success=false` on every `get_products` call. Pinned
`core/protocol-envelope.json` separates the two questions, so `_response_failed` now keys
on the `severity` marker rather than on emptiness. An entry marked `severity: "warning"`
is by the schema's own definition not a failure; anything else — including every fatal
`errors[]` this codebase emits, none of which carries the marker — still reports
`success=False`. The predicate is deliberately not keyed on the tool or the code, so
`get_media_buys`' `_omitted_row_advisory` and the `property_list_unsupported_advisories`
on `create_media_buy`/`update_media_buy` become one-keyword fixes rather than a per-tool
branch here.

The `score < 0.1` exclusion is deliberately **not** reported on `errors[]`: the pinned
schema assigns score-threshold reporting to `filter_diagnostics` and calls it
observability.

## Spec grounding for the `errors[]` advisory

Pinned version: `adcp==6.6.0`, targeting AdCP spec **3.1.1** (`pyproject.toml:10`;
`docs/adcp-spec-version.md`). All quotes verified against
`.venv/lib/python3.12/site-packages/adcp/_schemas/3.1/`.

**The channel exists on this response.** `media-buy/get-products-response.json`
declares `errors` among its 17 top-level properties, described as "Task-specific errors
and warnings (e.g., product filtering issues)", `$ref`-ing `core/error.json`. The
response has no top-level `required` array, so `errors[]` is optional and a completed
response may carry it.

**Non-fatal warnings ride the payload, never the envelope.**
`core/protocol-envelope.json`, on `properties.adcp_error`: "a fatal task failure SHOULD
populate both this envelope-level field AND the payload's `errors[]` array … Non-fatal
warnings populate ONLY `payload.errors[]` with `severity: warning` — the envelope MUST NOT
carry `adcp_error` for non-failures." The envelope's own examples carry the contrast:
`examples[2].data.payload.errors[0].severity` is `"warning"`.

**`severity` is not a declared field of `core/error.json`.** Stated plainly because it
matters: the error item declares `code`, `message`, `field`, `suggestion`, `retry_after`,
`issues`, `details`, `recovery`, `source`, `sdk_id`, with `required: ["code", "message"]`
and `additionalProperties: true`. `severity` is admitted by `additionalProperties`, named
by the envelope schema as the marker for a non-fatal payload error, and carried by that
schema's own examples in exactly this position. That is the whole basis for emitting it.

**The code vocabulary is open.** `core/error.json`, on `code`: "The error-code vocabulary
is open: `error.code` is wire-typed `string` (not a closed enum), the standard codes
published in `enums/error-code.json` are documentary, and senders MAY emit codes outside
that set… Receivers MUST decode unknown codes — treat the response as well-formed, read
`error.recovery` for the recovery classification, and fall back to `transient` when
`recovery` is absent." This is why `recovery` is stated on the wire rather than left to be
inferred from the enum.

**`enumMetadata` for the two codes used** (`enums/error-code.json`):

| Code | `recovery` | `suggestion` |
|---|---|---|
| `CONFIGURATION_ERROR` | `terminal` | surface to a human at the seller — the buyer cannot resolve a seller-side deployment misconfiguration and MUST NOT auto-retry |
| `SERVICE_UNAVAILABLE` | `transient` | retry with exponential backoff |

The `CONFIGURATION_ERROR` `enumDescription` also carries the message-content rule the
advisory follows: sellers "MUST NOT include credentials, connection strings, or stack
traces — the message is wire-visible to the buyer".

**Storyboard clause: ungraded.** No conformance phase grades this. Evidence:
`docs/test-obligations/storyboard-wireability.yaml` enumerates every storyboard step;
16 of them call `get_products`, and every graded check on those steps is
`response_schema` or `field_present` on `products[…]`. `path: errors…` appears in no step
in either obligation map, and the strings `ranking`, `relevance` and `unranked` appear
zero times across both files. The behaviour is graded here instead, by the BDD scenarios
in finding 4.

## What is deliberately not in this PR

Each of these is a real site with the same root cause, outside the boundary of this diff:

- **Three adapter call sites still read `tenant.gemini_api_key` directly** (#2223) and pass it as
  a bare string into the naming path: `src/adapters/google_ad_manager.py:645`,
  `src/adapters/mock_ad_server.py:817`, `src/adapters/gam/managers/workflow.py:167`. They
  work — `coerce` accepts the legacy path — but they ask the question themselves instead
  of calling `from_tenant`, so a tenant on `ai_config` alone gets platform naming.
- **`src/services/setup_checklist_service.py:730` and `:1116`** (#2224) both compute
  `gemini_configured = bool(tenant.gemini_api_key)`, so the setup checklist reports "AI not
  configured" for a tenant configured through `ai_config`. Same shape as the
  `creatives.py` inconsistency fixed here, in a service this diff does not otherwise touch.
- **`src/services/policy_check_service.py:63-70`** (#2225) keeps its own hand-rolled
  `TenantAIConfig(provider="gemini", model="gemini-2.0-flash", api_key=...)` construction
  behind the deprecated `gemini_api_key=` kwarg. `products.py` no longer takes that path,
  but the shim is still reachable from other callers and duplicates what `from_tenant`
  now owns. Removing it means auditing the remaining callers of the deprecated kwarg.
- **Three existing advisories do not carry `severity: "warning"`**, so they keep reporting
  `success=false` over A2A: `src/core/tools/media_buy_list.py:582` and the two attached by
  `src/services/targeting_capabilities.py:255`. This is their pre-existing behaviour, not a
  regression from this change — `_response_failed` only moves the marker for entries that
  explicitly declare themselves advisory, precisely so that "is this non-fatal?" stays a
  per-advisory ruling rather than something inferred from the code or the tool. Each needs
  one keyword and its own decision.
- **`ai_config["api_key"]` is stored in plaintext.** (#2226) `src/core/database/models.py:98,102`
  comments the column as carrying `api_key (encrypted)`; `src/admin/blueprints/settings.py`
  writes it in clear text and nothing decrypts on read. This diff neither reads nor writes
  an encryption path. Flagged in the original description and repeated here because it is
  still true: the comment and the write disagree, and one of them has to move.

## CI

The Quality Gate failure on this PR was **branch staleness, not a defect in the change**.
`check_ruff_complexity_count.py::check_baseline_not_raised` compares the local
`.ruff-complexity-baseline` against `origin/main`'s and fails when any key is higher —
"The ruff complexity baseline may only shrink." The PR head (`f78a60cc`) branched from
`1d93e372` and does not contain `f9e82da6`, which lowered the baseline:

```
$ git show refs/pull/2167/head:.ruff-complexity-baseline
{ "C901": 182, "PLR0912": 133, "PLR0915": 107 }
$ git show origin/main:.ruff-complexity-baseline
{ "C901": 177, "PLR0912": 130, "PLR0915": 102 }
$ git merge-base --is-ancestor f9e82da6 refs/pull/2167/head; echo $?
1
```

All three keys higher than main, all three failing the raise guard. Fixed by rebasing onto
`main`; the hook is clean on the rebased tree (`C901: 177 (unchanged)`, `PLR0912: 130
(unchanged)`, `PLR0915: 102 (unchanged)`).

`.duplication-baseline` `tests` drops 71 → 69: two duplicate test blocks were collapsed
while reworking, and the ratchet auto-lowers.

## Changes

| File | Change |
|---|---|
| `src/services/ai/config.py` | `TenantAIConfig.coerce` / `.from_tenant`, `_read_tenant_field`, `DEFAULT_GOOGLE_MODEL` paired with the provider |
| `src/services/ai/factory.py` | Three parse sites route through `coerce`; platform provider/model/key fill in only for the same vendor |
| `src/core/tools/products.py` | Ranking extracted to `_rank_products_with_ai`; advisory builder; both AI consumers resolve through `from_tenant` |
| `src/a2a_server/adcp_a2a_server.py` | `_response_failed` keys the `success` marker on `severity`, not on `errors[]` emptiness |
| `src/admin/blueprints/creatives.py` | `_tenant_ai_enabled` helper; the availability flag and the review impl now agree |
| `src/core/utils/naming.py` | Routes through `coerce`; drops its own `is None` absent-test |
| `src/core/tenant_context.py`, `src/core/utils/tenant_utils.py` | Carry `ai_config` in both tenant projections |

Tests: a new four-transport BDD feature for BR-RULE-005 plus its steps and binder, a new
unit module for the admin AI-review availability flag, the `ai_config` leg in
`test_tenant_utils.py`, the restored assertion in `test_naming_parameter_bug.py`, and the
two structural-guard ledgers (`EXPECTED_WIRED_ROUTES`,
`EXPECTED_UNSUPPORTED_DECLARATIONS`) updated in the same change.

## Verification

```
$ uv run pytest -q -p no:randomly tests/unit/test_get_products_impl_coverage.py \
    tests/unit/test_ai_service_factory.py tests/unit/test_naming_parameter_bug.py \
    tests/unit/test_ai_review.py tests/unit/test_admin_creative_ai_review_availability.py \
    tests/unit/test_a2a_success_field_consolidation.py \
    tests/unit/test_architecture_measurement_floors.py \
    tests/unit/test_architecture_e2e_rest_escape_hatches.py
158 passed, 7 skipped, 1 warning in 2.74s

$ uv run pytest -q -p no:randomly tests/bdd/test_uc001_ai_ranking_tenant_config.py \
    tests/integration/test_get_products_behavioral.py tests/integration/test_tenant_utils.py \
    tests/integration/test_a2a_response_compliance.py     # against PostgreSQL
69 passed, 59 warnings in 12.41s
```

Each behavioural decision above has a test that fails when the decision is reverted. The
ones worth naming, because they were silently green before: the ranking argument order
(finding 4), the both-set precedence branch of `from_tenant`, the provider-scoped platform
fallback, and the A2A `success` marker on an advisory-carrying response.

### DIFF
--- a/src/core/tenant_context.py
+++ b/src/core/tenant_context.py
@@ -56,6 +56,7 @@ class TenantContext(BaseModel):
     approval_mode: str = "require-human"  # BR-RULE-037: creative approval mode
     account_approval_mode: str | None = None  # BR-RULE-060: account approval mode (auto|credit_review|legal_review)
     gemini_api_key: str | None = None
+    ai_config: dict[str, Any] | None = None  # provider/model/api_key; supersedes gemini_api_key
     creative_review_criteria: str | None = None
     brand_manifest_policy: str = "require_auth"
     advertising_policy: dict[str, Any] | None = None
@@ -117,6 +118,7 @@ class TenantContext(BaseModel):
             approval_mode=tenant.approval_mode or "require-human",
             account_approval_mode=tenant.account_approval_mode,
             gemini_api_key=tenant.gemini_api_key,
+            ai_config=tenant.ai_config,
             creative_review_criteria=tenant.creative_review_criteria,
             brand_manifest_policy=tenant.brand_manifest_policy or "require_auth",
             advertising_policy=safe_json_loads(tenant.advertising_policy, None),
--- a/src/core/tools/products.py
+++ b/src/core/tools/products.py
@@ -663,8 +663,17 @@ async def _get_products_impl(
             from src.services.ai.factory import get_factory
 
             factory = get_factory()
-            if factory.is_ai_enabled():
-                model = factory.create_model()
+            # Pass the tenant's own AI configuration. Without it the factory falls
+            # back to an empty TenantAIConfig and sees only the platform-level key
+            # from the environment, so a tenant that configured its key in the Admin
+            # UI got no ranking at all — and the miss is logged at debug level, so it
+            # fails silently. Mirrors the fallback in src/core/utils/naming.py.
+            tenant_ai_config = tenant.get("ai_config")
+            if not tenant_ai_config and tenant.get("gemini_api_key"):
+                tenant_ai_config = {"provider": "gemini", "api_key": tenant.get("gemini_api_key")}
+
+            if factory.is_ai_enabled(tenant_ai_config):
+                model = factory.create_model(tenant_ai_config=tenant_ai_config)
                 agent = create_ranking_agent(model)
 
                 # Convert products to dicts for ranking
--- a/src/core/tools/products.py
+++ b/src/core/tools/products.py
@@ -13,7 +13,7 @@ from typing import Annotated, Any, cast
 from adcp import FormatId, ProductFilters
 from adcp import GetProductsRequest as GetProductsRequestGenerated
 from adcp import Product as LibraryProduct
-from adcp.types import BrandReference, ContextObject, PropertyListReference
+from adcp.types import BrandReference, ContextObject, Error, PropertyListReference
 from fastmcp.server.context import Context
 from pydantic import Field
 
@@ -24,6 +24,7 @@ from src.core.exceptions import (
     AdCPAdapterError,
     AdCPAuthenticationError,
     AdCPAuthorizationError,
+    AdCPConfigurationError,
     AdCPError,
     AdCPPolicyViolationError,
     AdCPValidationError,
@@ -39,6 +40,7 @@ from src.core.testing_hooks import AdCPTestContext
 from src.core.tool_context import ToolContext
 from src.core.transport_helpers import resolve_identity_from_context
 from src.core.validation_helpers import adcp_validation_boundary, safe_parse_json_field
+from src.services.ai.config import TenantAIConfig
 from src.services.policy_check_service import PolicyCheckService, PolicyStatus
 
 logger = logging.getLogger(__name__)
@@ -150,6 +152,193 @@ def filter_products_by_property_list(
     return [p for p in products if should_include_product_for_property_list(p, allowed_properties)]
 
 
+# The recovery classification each advisory code carries on the wire, mirroring the
+# `enumMetadata` block of the pinned `enums/error-code.json`. Derived from the code
+# rather than passed alongside it: recovery is a PROPERTY of the code, and a call site
+# free to pair them itself is a call site free to tell a buyer to retry a missing API
+# key (or to give up on a rate limit).
+#
+# * CONFIGURATION_ERROR -> "terminal" / "surface to a human at the seller — the buyer
+#   cannot resolve a seller-side deployment misconfiguration and MUST NOT auto-retry".
+# * SERVICE_UNAVAILABLE -> "transient" / "retry with exponential backoff".
+#
+# recovery is stated on the wire rather than left to be inferred from the enum because
+# pinned `core/error.json` makes the code vocabulary OPEN ("senders MAY emit codes
+# outside that set ... read `error.recovery` for the recovery classification"): a client
+# that does not consult the enum still classifies these correctly.
+_ADVISORY_RECOVERY: dict[str, str] = {
+    "CONFIGURATION_ERROR": "terminal",
+    "SERVICE_UNAVAILABLE": "transient",
+}
+
+
+def _unranked_products_advisory(tenant_id: str, *, code: str, cause: str) -> Error:
+    """The advisory that stands in for AI ranking a tenant asked for but did not get.
+
+    Ranking is configured (``product_ranking_prompt`` is set) but did not happen, so
+    ``products[]`` comes back in catalog order. Silence here is the bug: the buyer sees
+    a plausible-looking list and has no way to tell it was never ranked. One builder,
+    two callers — the shape, the field pointer and the ``PRODUCT_RANKING_UNAVAILABLE``
+    marker are identical; only the CODE and the ``cause`` clause differ, because the two
+    conditions need opposite advice:
+
+    * **Nothing resolved / nothing usable resolved** — CONFIGURATION_ERROR. The fault is
+      entirely on the seller's side of the wire (the buyer's request was valid) and no
+      retry, at any backoff, supplies a missing API key. This mirrors the ruling at
+      ``src/core/tools/media_buy_list.py::_omitted_row_advisory``.
+    * **The ranking call itself did not complete** — SERVICE_UNAVAILABLE. A provider
+      429, a 5xx, a timeout, an unexpected model response: pydantic-ai raises
+      ``ModelHTTPError`` / ``UnexpectedModelBehavior`` / ``UsageLimitExceeded`` /
+      ``UserError``, all of which subclass ``RuntimeError``. Telling the buyer
+      "terminal — MUST NOT auto-retry" for a rate limit is wrong advice; the pinned
+      enumDescription for SERVICE_UNAVAILABLE is "Seller service is temporarily
+      unavailable. Retry with exponential backoff. Recovery: transient."
+
+    The ``cause`` clause never interpolates ``str(exc)``. This message is wire-visible to
+    an anonymous buyer (``get_products`` is auth-optional discovery), and a provider
+    exception's text can carry request URLs, response bodies and key fragments; the
+    pinned enumDescription for CONFIGURATION_ERROR states sellers "MUST NOT include
+    credentials, connection strings, or stack traces — the message is wire-visible to
+    the buyer". The exception class name is enough to classify; the full text goes to
+    the seller's log.
+
+    It rides on ``payload.errors[]`` inside a SUCCESS envelope, never as an envelope
+    ``adcp_error``: pinned ``core/protocol-envelope.json`` states that non-fatal warnings
+    populate ONLY ``payload.errors[]`` with ``severity: warning`` and that the envelope
+    MUST NOT carry ``adcp_error`` for non-failures, and pinned
+    ``get-products-response.json`` requires ``errors[]`` only when ``status == "failed"``
+    while describing it as "Task-specific errors and warnings", so a completed response
+    may carry it. Ungraded: no phase in the pinned storyboards grades a get_products
+    ranking advisory.
+    """
+    return Error(  # structural-guard: advisory unranked-catalog notice in GetProductsResponse.errors[]
+        code=code,
+        recovery=_ADVISORY_RECOVERY[code],
+        # severity is not a field of the pinned core/error.json, whose
+        # additionalProperties is true; it is the marker the envelope schema names for a
+        # non-fatal payload error, and the envelope's own examples carry it exactly here.
+        # It is also what keeps the A2A envelope reporting success=True for an advisory
+        # (see AdCPRequestHandler._stamp_a2a_protocol_fields).
+        severity="warning",
+        message=(
+            f"PRODUCT_RANKING_UNAVAILABLE: seller {tenant_id!r} has AI product ranking "
+            f"configured but {cause}, so these products are returned unranked (catalog "
+            f"order), not ordered by relevance to the brief."
+        ),
+        field="products[]",
+    )
+
+
+async def _rank_products_with_ai(
+    products: list[Product],
+    tenant: Any,
+    ranking_prompt: str,
+    brief_text: str,
+    advisories: list[Error],
+) -> list[Product]:
+    """Order products by AI-scored relevance to the brief, dropping the irrelevant ones.
+
+    Returns the products unchanged when ranking cannot run, appending EXACTLY ONE
+    advisory to ``advisories`` on every such path — the unusable-configuration one and
+    the call-did-not-complete one alike. Owning the failure paths here, rather than in a
+    ``try`` at the call site, is what makes that "every" true: the caller's handler
+    logged a warning and dropped through with ``advisories`` untouched, so a provider
+    401/429/5xx returned a plausible-looking catalog-order list with ``errors=None`` —
+    precisely the silent failure the advisory exists to end, on the paths most likely to
+    happen in production.
+
+    Extracted from ``_get_products_impl`` so the ranking decision reads in one place —
+    and so this addition does not deepen a function already over every complexity
+    threshold.
+    """
+    from src.services.ai.agents.ranking_agent import create_ranking_agent, rank_products_async
+    from src.services.ai.factory import get_factory
+
+    tenant_id = tenant["tenant_id"]
+    factory = get_factory()
+    # TenantAIConfig.from_tenant owns "given a tenant, what is its AI configuration?" —
+    # ai_config first, the legacy gemini_api_key column second, None when neither is set.
+    # Reading only the platform environment key here meant a tenant that configured its
+    # key in the Admin UI got no ranking at all.
+    tenant_ai_config = TenantAIConfig.from_tenant(tenant)
+
+    # is_ai_enabled(), not `tenant_ai_config is not None`: the platform-key fallback is
+    # existing, relied-upon behaviour for ranking — a deployment-wide GEMINI_API_KEY
+    # ranks for every tenant that set a ranking prompt.
+    if not factory.is_ai_enabled(tenant_ai_config):
+        logger.warning(
+            "[GET_PRODUCTS] AI ranking is configured for tenant %s but no usable AI configuration "
+            "resolved (tenant ai_config/gemini_api_key: %s; platform environment key: absent). "
+            "Returning products unranked.",
+            tenant_id,
+            "present but unusable" if tenant_ai_config is not None else "absent",
+        )
+        advisories.append(
+            _unranked_products_advisory(
+                tenant_id,
+                code="CONFIGURATION_ERROR",
+                cause="no usable AI configuration resolved for it",
+            )
+        )
+        return products
+
+    try:
+        model = factory.create_model(tenant_ai_config=tenant_ai_config)
+        ranking_result = await rank_products_async(
+            agent=create_ranking_agent(model),
+            custom_prompt=ranking_prompt,
+            brief=brief_text,
+            products=products,
+        )
+    except (AdCPConfigurationError, ImportError, RuntimeError, OSError) as e:
+        # Two conditions, opposite advice, one handler so neither can be added without
+        # its advisory:
+        #
+        # * AdCPConfigurationError is the factory REFUSING this seller's configuration.
+        #   `is_ai_enabled` can report a coherent provider/model/key and `create_model`
+        #   still decline the combination (a provider with no explicit-API-key
+        #   integration, where the string form would authenticate with whatever ELSE is
+        #   in the environment). It extends Exception, not RuntimeError, so nothing here
+        #   caught it: a seller-side misconfiguration 500'd an auth-OPTIONAL discovery
+        #   call for every anonymous buyer. Same fault as the branch above, same code.
+        # * Everything else is the call not completing. pydantic-ai's ModelHTTPError,
+        #   UnexpectedModelBehavior, UsageLimitExceeded and UserError all subclass
+        #   RuntimeError, so every provider auth failure, 429, timeout and malformed
+        #   response lands here. That is transient, not a deployment misconfiguration.
+        #
+        # `str(e)` is logged for the seller and deliberately kept OFF the wire (see
+        # `_unranked_products_advisory`).
+        code, cause = (
+            ("CONFIGURATION_ERROR", "its AI configuration cannot build a model")
+            if isinstance(e, AdCPConfigurationError)
+            else ("SERVICE_UNAVAILABLE", "the ranking call did not complete")
+        )
+        logger.warning(
+            "[GET_PRODUCTS] AI ranking did not run for tenant %s (%s). Returning products unranked. Cause: %s",
+            tenant_id,
+            cause,
+            e,
+        )
+        advisories.append(_unranked_products_advisory(tenant_id, code=code, cause=f"{cause} ({type(e).__name__})"))
+        return products
+
+    # Build a map of product_id -> (score, reason); products the agent did not score sort last.
+    ranking_map = {r.product_id: (r.relevance_score, r.reason) for r in ranking_result.rankings}
+    ranked = sorted(products, key=lambda p: ranking_map.get(p.product_id, (0.0, ""))[0], reverse=True)
+    # Drop very low relevance. Deliberately NOT reported on errors[]: the pinned schema
+    # assigns score-threshold reporting to filter_diagnostics and calls it observability,
+    # not error reporting.
+    ranked = [p for p in ranked if ranking_map.get(p.product_id, (0.0, ""))[0] >= 0.1]
+
+    for r in ranking_result.rankings:
+        logger.info(f"[AI_RANKING] {r.product_id}: score={r.relevance_score:.2f}, reason={r.reason}")
+    logger.info(
+        f"[GET_PRODUCTS] AI ranking applied: {len(ranking_result.rankings)} products ranked, "
+        f"{len(ranked)} products above threshold"
+    )
+    return ranked
+
+
 async def _get_products_impl(
     req: GetProductsRequestGenerated, identity: ResolvedIdentity | None
 ) -> GetProductsResponse:
@@ -229,15 +418,41 @@ async def _get_products_impl(
         policy_disabled_reason = "disabled_by_tenant"
         logger.info(f"Policy checks disabled for tenant {tenant['tenant_id']}")
     else:
-        # Get tenant's Gemini API key for policy checks
-        tenant_gemini_key = tenant.get("gemini_api_key")
-        if not tenant_gemini_key:
-            # No API key - cannot run policy checks
+        # Resolve the tenant's own AI configuration. `TenantAIConfig.from_tenant` returns
+        # a config only when the tenant has its OWN usable credential — ai_config with a
+        # non-empty api_key first, the legacy gemini_api_key column second, None
+        # otherwise — so `is not None` here reads as "this seller has a credential of its
+        # own". That is the same question the pre-PR gate asked as
+        # `tenant.get("gemini_api_key")`, generalised to the column the Admin UI actually
+        # writes: reading only gemini_api_key meant a tenant that configured AI in the UI
+        # had its policy checks silently skipped even with "enabled" ticked.
+        #
+        # Usability is load-bearing, not incidental. src/admin/blueprints/settings.py
+        # stores {"provider": ..., "model": ...} with NO api_key when the seller leaves
+        # the key blank, and tells them "AI features will be disabled" as it does. If
+        # that row came back as a configuration, this gate would open and
+        # check_brief_compliance would make a live LLM call on the OPERATOR's platform
+        # credential — and a BLOCKED verdict raises AdCPPolicyViolationError below, a new
+        # buyer-visible rejection for a tenant the UI said was switched off.
+        #
+        # The gate is NOT factory.is_ai_enabled(): is_ai_enabled counts the platform
+        # GEMINI_API_KEY from the environment, so gating on it would silently switch
+        # policy checks on for every tenant on any deployment that sets that variable.
+        # Gating on the tenant's own configuration widens the gate exactly as far as the
+        # bug requires and no further.
+        tenant_ai_config = TenantAIConfig.from_tenant(tenant)
+        if tenant_ai_config is None:
+            # No AI configuration - cannot run policy checks
             policy_result = None
-            policy_disabled_reason = "no_gemini_api_key"
-            logger.warning(f"Policy checks enabled but no Gemini API key configured for tenant {tenant['tenant_id']}")
+            policy_disabled_reason = "no_ai_configuration"
+            logger.warning(f"Policy checks enabled but no AI configuration for tenant {tenant['tenant_id']}")
         else:
-            policy_service = PolicyCheckService(gemini_api_key=tenant_gemini_key)
+            # tenant_ai_config=, not the deprecated gemini_api_key=: that parameter pins
+            # provider/model instead of honouring the tenant's own, and its _UNSET
+            # sentinel treats an explicit None as "hard-disable AI". Leaving it unset
+            # keeps the intended branch — configuration honoured, platform defaults
+            # filling any field the tenant did not set.
+            policy_service = PolicyCheckService(tenant_ai_config=tenant_ai_config)
 
             # Use advertising_policy settings for tenant-specific rules
             tenant_policies = advertising_policy if advertising_policy else {}
@@ -653,63 +868,20 @@ async def _get_products_impl(
         eligible_products = filtered_products
 
     # AI-powered product ranking (when tenant has product_ranking_prompt configured)
+    advisories: list[Error] = []
     product_ranking_prompt = tenant.get("product_ranking_prompt")
     if product_ranking_prompt and brief_text and eligible_products:
-        try:
-            from src.services.ai.agents.ranking_agent import (
-                create_ranking_agent,
-                rank_products_async,
-            )
-            from src.services.ai.factory import get_factory
-
-            factory = get_factory()
-            # Pass the tenant's own AI configuration. Without it the factory falls
-            # back to an empty TenantAIConfig and sees only the platform-level key
-            # from the environment, so a tenant that configured its key in the Admin
-            # UI got no ranking at all — and the miss is logged at debug level, so it
-            # fails silently. Mirrors the fallback in src/core/utils/naming.py.
-            tenant_ai_config = tenant.get("ai_config")
-            if not tenant_ai_config and tenant.get("gemini_api_key"):
-                tenant_ai_config = {"provider": "gemini", "api_key": tenant.get("gemini_api_key")}
-
-            if factory.is_ai_enabled(tenant_ai_config):
-                model = factory.create_model(tenant_ai_config=tenant_ai_config)
-                agent = create_ranking_agent(model)
-
-                # Convert products to dicts for ranking
-                # Run AI ranking
-                ranking_result = await rank_products_async(
-                    agent=agent,
-                    custom_prompt=product_ranking_prompt,
-                    brief=brief_text,
-                    products=eligible_products,
-                )
-
-                # Build a map of product_id -> (score, reason)
-                ranking_map = {r.product_id: (r.relevance_score, r.reason) for r in ranking_result.rankings}
-
-                # Sort products by relevance score (highest first)
-                # Products not in ranking_map get score 0
-                eligible_products.sort(
-                    key=lambda p: ranking_map.get(p.product_id, (0.0, ""))[0],
-                    reverse=True,
-                )
-
-                # Filter out products with very low relevance (score < 0.1)
-                eligible_products = [p for p in eligible_products if ranking_map.get(p.product_id, (0.0, ""))[0] >= 0.1]
-
-                # Log the ranking results
-                for r in ranking_result.rankings:
-                    logger.info(f"[AI_RANKING] {r.product_id}: score={r.relevance_score:.2f}, reason={r.reason}")
-
-                logger.info(
-                    f"[GET_PRODUCTS] AI ranking applied: {len(ranking_result.rankings)} products ranked, "
-                    f"{len(eligible_products)} products above threshold"
-                )
-            else:
-                logger.debug("[GET_PRODUCTS] AI ranking configured but AI not enabled (no API key)")
-        except (ImportError, RuntimeError, OSError) as e:
-            logger.warning(f"Failed to apply AI product ranking: {e}. Returning unranked products.")
+        # No handler here on purpose: `_rank_products_with_ai` owns every failure path so
+        # that each one emits its advisory. A `try` here could only log and drop through
+        # with `advisories` empty — which is how a provider 429 returned an unranked list
+        # with `errors=None`.
+        eligible_products = await _rank_products_with_ai(
+            products=eligible_products,
+            tenant=tenant,
+            ranking_prompt=product_ranking_prompt,
+            brief_text=brief_text,
+            advisories=advisories,
+        )
 
     # Annotate pricing options with adapter support (AdCP PR #88)
     # Do this BEFORE serialization to avoid reconstruction issues
@@ -759,7 +931,7 @@ async def _get_products_impl(
     # but mypy still needs cast() due to list invariance in static typing
     resp = GetProductsResponse(
         products=cast(list[LibraryProduct], eligible_products),
-        errors=None,
+        errors=advisories or None,
         context=req.context,
     )
 
--- a/src/core/utils/tenant_utils.py
+++ b/src/core/utils/tenant_utils.py
@@ -53,6 +53,7 @@ def serialize_tenant_to_dict(tenant: Tenant) -> dict[str, Any]:
         "account_approval_mode": tenant.account_approval_mode,
         "supported_billing": safe_json_loads(tenant.supported_billing, None),
         "gemini_api_key": tenant.gemini_api_key,
+        "ai_config": tenant.ai_config,
         "creative_review_criteria": tenant.creative_review_criteria,
         "brand_manifest_policy": tenant.brand_manifest_policy,
         "advertising_policy": safe_json_loads(tenant.advertising_policy, None),
--- a/tests/unit/test_get_products_impl_coverage.py
+++ b/tests/unit/test_get_products_impl_coverage.py
@@ -512,6 +512,72 @@ class TestAIRankingDisabled:
         assert result.products[1].product_id == "prod-b"
 
 
+class TestAIRankingUsesTenantConfig:
+    """The ranking path must consult the tenant's own AI configuration.
+
+    Intent: a tenant that configures its API key in the Admin UI (stored in
+    tenants.ai_config) gets AI ranking. Calling the factory without that config
+    makes it fall back to the platform-level environment key, so such a tenant
+    silently got no ranking — the miss is only logged at debug level.
+    """
+
+    @pytest.mark.asyncio
+    async def test_tenant_ai_config_is_passed_to_factory(self):
+        """The tenant's ai_config reaches both is_ai_enabled and create_model."""
+        ai_config = {"provider": "google", "model": "gemini-flash-lite-latest", "api_key": "tenant-key"}
+        tenant = _make_tenant()
+        tenant["product_ranking_prompt"] = "rank by relevance"
+        tenant["ai_config"] = ai_config
+        identity = _make_identity(principal_id="user-1", tenant_id="test-tenant", tenant=tenant)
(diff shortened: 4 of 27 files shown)


src/core/tools/products.py
```diff
@@ -663,8 +663,17 @@ async def _get_products_impl(
...
-            if factory.is_ai_enabled():
-                model = factory.create_model()
+            # Pass the tenant's own AI configuration. Without it the factory falls
+            # back to an empty TenantAIConfig and sees only the platform-level key
+            # from the environment, so a tenant that configured its key in the Admin
+            # UI got no ranking at all — and the miss is logged at debug level, so it
+            # fails silently. Mirrors the fallback in src/core/utils/naming.py.
+            tenant_ai_config = tenant.get("ai_config")
+            if not tenant_ai_config and tenant.get("gemini_api_key"):
+                tenant_ai_config = {"provider": "gemini", "api_key": tenant.get("gemini_api_key")}
```
KonstantinMirin: This is the third hand-rolled copy of the rule "tenant `ai_config`, else synthesise one from the legacy `gemini_api_key`, else nothing" (`src/core/utils/naming.py:99-104`, `src/admin/blueprints/creatives.py:1076-1082`, and a fourth inside `src/services/policy_check_service.py:63-70`). They have already diverged: `naming.py:100` treats only `None` as absent while line 672 treats any falsy value as absent, so the comment above claiming "Mirrors the fallback in `src/core/utils/naming.py`" points a future maintainer at a different rule. The literal `"gemini"` also bypasses `canonicalize_google_provider()` / `CANONICAL_GOOGLE_PROVIDER`, which `src/services/ai/config.py:17-29` exports for exactly this question.

Give the rule an owner in the module that owns the model: `TenantAIConfig.from_tenant(tenant) -> TenantAIConfig | None` in `src/services/ai/config.py`, returning the model rather than a hand-built dict. Then route both consumers in this function through it (here and `products.py:232-240`).

KonstantinMirin (commented):
## Review — PR #2167

**Overview** — The PR fixes a real defect at the right layer: `_get_products_impl` now passes the tenant's own AI configuration to the ranking factory, so a tenant whose key lives in `tenants.ai_config` gets ranked products, and the change sits in the shared core so all four transports behave the same. The shape stops short in two ways. The description diagnoses the missing abstraction ("mirrored from `src/core/utils/naming.py`, which already solves the same problem correctly") and then writes the rule a third time instead of extracting it; and it converts one of the two AI consumers inside the function it edits, so the same tenant now gets AI ranking with its own key while its brief goes unscreened by policy checks in the same request. Name one owner for resolving a tenant's AI configuration, and route both consumers through it, rather than adding a fourth hand-rolled copy.

### Should fix

**1. Partial sweep: the policy consumer in the same function still reads only `gemini_api_key`**

`_get_products_impl` has two consumers of the tenant's AI configuration. This PR converts one.

```python
# src/core/tools/products.py:233-240  (unconverted)
tenant_gemini_key = tenant.get("gemini_api_key")
if not tenant_gemini_key:
    policy_result = None
    policy_disabled_reason = "no_gemini_api_key"
    logger.warning(f"Policy checks enabled but no Gemini API key configured for tenant {tenant['tenant_id']}")
else:
    policy_service = PolicyCheckService(gemini_api_key=tenant_gemini_key)

# src/core/tools/products.py:671-676  (converted)
tenant_ai_config = tenant.get("ai_config")
...
if factory.is_ai_enabled(tenant_ai_config):
```

For the tenant this PR exists for, `policy_result` stays `None`, so `check_brief_compliance` never runs and the product eligibility filter at `products.py:612` is skipped. The reason string names `gemini_api_key`, a field that is no longer the source of truth, and the kwarg the call site passes is marked `DEPRECATED - Use tenant_ai_config instead` by its own callee (`src/services/policy_check_service.py:52,58`). Taking that legacy path is also what pins `model="gemini-2.0-flash"` at `policy_check_service.py:67`, so the same tenant gets a different model for policy checks than for ranking.

Root: there is no seam to route a consumer through, so fixing the call site you are looking at is the cheapest move and the sibling consumer forks. Both consumers are inside the function this diff edits, so both are in scope here: `src/core/tools/products.py:232-240` and `src/core/tools/products.py:671-673`, 1 of 2 converted. Resolve once at the top of `_get_products_impl`, pass `PolicyCheckService(tenant_ai_config=...)`, drop the deprecated kwarg, and make the disabled reason provider-neutral (`"no_ai_configuration"`).

**2. Nothing owns tenant to `TenantAIConfig` resolution, so the rule is copied and the value travels untyped**

`src/services/ai/config.py` owns `TenantAIConfig`, the provider vocabulary (`CANONICAL_GOOGLE_PROVIDER`, `canonicalize_google_provider`) and half the migration rule (`uses_legacy_gemini_api_key`, called from `src/admin/blueprints/settings.py:561,628`). Nothing answers "given a tenant, what is its AI config?", so every consumer improvises it, and the copies have already diverged:

| Site | Absent test | Provider | Model |
|---|---|---|---|
| `src/core/utils/naming.py:99-104` | `is None` | literal `"gemini"` | unset |
| `src/admin/blueprints/creatives.py:1076-1082` | falsy | literal `"gemini"` | unset |
| `src/core/tools/products.py:671-673` (added here) | falsy | literal `"gemini"` | unset |
| `src/services/policy_check_service.py:63-70` | n/a | literal `"gemini"` | pinned `gemini-2.0-flash` |

The comment added at `products.py:670` says "Mirrors the fallback in `src/core/utils/naming.py`", but the guard differs: `naming.py:100` treats only `None` as absent, `products.py:672` treats any falsy value as absent. A future maintainer following that comment reads the wrong precedent. R0801 cannot see a three-line copy (`min-similarity-lines = 6`, `pyproject.toml:229`), so a green `make quality` is not evidence the duplication is acceptable, and raising `.duplication-baseline` is not the resolution.

The same missing owner produces the type and carrier problems in this diff. `TenantContext.ai_config` is declared `dict[str, Any] | None` (`src/core/tenant_context.py:59`) in a model whose docstring says it "Replaces the fragile `dict[str, Any]` tenant representation with a typed, validated Pydantic model", while `TenantAIConfig` (`src/services/ai/config.py:33`) is documented as "the structure of the `ai_config` JSON column on the Tenant table". Because the value travels as a dict, `products.py:673` hand-assembles `{"provider": ..., "api_key": ...}` with no key checking and the factory re-runs `TenantAIConfig.model_validate` at `src/services/ai/factory.py:109,224,248`, two of them per ranking request. `ai_config` is also the only JSON field in either tenant projection that skips the sibling `safe_json_loads` treatment (6 of 6 and 7 of 7 elsewhere in `tenant_context.py:100-126` and `tenant_utils.py:36-62`). And `ai_config` carries `api_key` and `logfire_token` (`src/services/ai/config.py:46,49`), so adding it to `serialize_tenant_to_dict` puts a credential into the dict that `src/core/audit_logger.py:360` hands whole to `get_slack_notifier`, which reads three keys (`src/services/slack_notifier.py:39-48`), and into the process-wide tenant ContextVar behind `get_current_tenant()`.

Make one move: add the resolver to the module that owns the model and have it return the model.

```python
# src/services/ai/config.py
class TenantAIConfig(BaseModel):
    @classmethod
    def from_tenant(cls, tenant) -> "TenantAIConfig | None":
        """ai_config, else the legacy gemini_api_key, else None.
        Accepts the ORM row, TenantContext, or the tenant dict."""
```

It decides the absent test once, applies `canonicalize_google_provider` once, and returns `TenantAIConfig`. Then type `TenantContext.ai_config` as `TenantAIConfig | None` and coerce at the ORM boundary. The project's own rule points at the column: "all new columns should specify a model - bare JSONType (no model) is legacy" (`src/core/database/json_type.py:6-8,22-47`), and `ai_config` is bare at `src/core/database/models.py:99`. JSONB storage is unchanged and `TenantAIConfig` is all-optional with `extra="ignore"`, so no data migration. With resolution on demand, the AI layer fetches its own configuration and neither plumbing hunk in this diff is needed, which also removes the credential from the ambient tenant dict. `naming.py:99-104`, `creatives.py:1076-1082` and `policy_check_service.py:63-70` each collapse to the same one-line call once the seam exists; if you do not sweep them here, file a follow-up issue and self-assign it naming those three sites.

**3. A malformed `ai_config` row turns the whole `get_products` call into a 500**

The ranking block is deliberately fail-open: it catches `(ImportError, RuntimeError, OSError)` at `src/core/tools/products.py:712` and logs "Returning unranked products". That tuple was written when the factory only saw platform environment config. This diff routes tenant-stored JSON into `factory.is_ai_enabled(...)` and `factory.create_model(...)`, both of which call `TenantAIConfig.model_validate` (`factory.py:109,224`), and `pydantic.ValidationError` is a `ValueError`, so it escapes:

```
cd /srv/ws/salesagent/main-pr2167 && .venv/bin/python -c "
from src.services.ai.factory import AIServiceFactory
try: AIServiceFactory().is_ai_enabled({'provider':'google','api_key':'k','settings':{'temperature':9}})
except Exception as e: print(type(e).__name__, isinstance(e,(ImportError,RuntimeError,OSError)))"
# -> ValidationError False
```

`settings` is round-tripped opaquely by `src/admin/blueprints/settings.py:565-566` and never re-validated on write, and the column is bare JSONB the database cannot constrain. One row aborts the whole request on all four transports, where the surrounding code promises unranked products. This is the unit of work, not distrust of the schema: an optional feature's configuration should degrade that feature, not the call. Parsing at the ORM boundary (finding 2) removes the case; while the column stays bare, name `ValidationError` in the caught tuple at line 712.

**4. The tests do not reach the changed behavior; three of four production hunks stay green when reverted**

Both new tests set `mock_factory.is_ai_enabled.return_value = False` (`tests/unit/test_get_products_impl_coverage.py:537,564`), so `create_model(tenant_ai_config=...)` at `products.py:676` never runs. Verified mutation: revert line 676 to `factory.create_model()` and both tests pass (`pytest tests/unit/test_get_products_impl_coverage.py -k TenantConfig` reports 2 passed). The docstring at line 526 claims the config "reaches both is_ai_enabled and create_model"; the second is unreachable by construction. Both tests build their tenant with `_make_tenant()` (lines 42-50), a plain dict they write `ai_config` into, so `TenantContext.from_orm_model` and `serialize_tenant_to_dict` are never imported and both plumbing hunks stay green when deleted. `grep -rn "ai_config" tests/ --include=*.py` shows no test calls either function with the field. The factory double is a bare `MagicMock()` with no `spec`, so `assert_called_once_with(ai_config)` asserts this call site's positional punctuation rather than `AIServiceFactory.is_ai_enabled`'s contract.

Root: the tests are written at the mock boundary and never enter the production tenant resolution path (`LazyTenantContext._resolve` to `get_tenant_by_id` to `serialize_tenant_to_dict` to `TenantContext.from_dict`, where `from_dict` filters to declared model fields at `tenant_context.py:141-142`, which is exactly what makes both plumbing hunks load-bearing).

The observable is on the wire. In AdCP 3.1.1, brief mode is "publisher curates products from the brief" (`dist/docs/3.1.1/media-buy/task-reference/get_products.mdx`) and the curated set is `products[]` (`dist/schemas/3.1.1/media-buy/get-products-response.json`). `tests/bdd/features/BR-UC-001-discover-available-inventory.feature:549-584` already describes the outcome (BR-RULE-005 INV-1/2/4: below-threshold excluded, sorted descending, unranked on failure) and collects zero tests: no `scenarios()` binder, no step definitions, and `UC-001` has no `ENV_ROUTES` row, so it would auto-xfail with "No harness wired" even if bound. `pytest tests/bdd --collect-only -q | grep -ci "rank\|inv-005"` returns 0. The harness exists: `get_products` has a live four-transport route (`UC-GET-PRODUCTS`, `tests/bdd/conftest.py:3720`) bound by `tests/bdd/test_get_products_inventory_profile.py` over `ProductEnv`.

Write the scenario transport-blind: Given a tenant whose AI configuration carries its own API key, And the tenant has a product ranking prompt, When the Buyer Agent requests products with a brief, Then the products are ordered by relevance and the below-threshold product is absent, plus the legacy-field sibling. Seed the tenant through an env method (`TenantFactory(ai_config=..., product_ranking_prompt=...)`) with a `realize_e2e` counterpart, since `ProductEnv.EXTERNAL_PATCHES["ranking_factory"]` (`tests/harness/product.py:75`) is in-process only and the e2e leg would pass without exercising anything. Keep `dispatch_request` as the sole writer of `ctx["result"]` and read the outcome with `wire_field(ctx, "products")`. That one scenario fails on both mutations above. Do not register the tag in an xfail or `_E2E_*_INTERNAL_TAGS` set to make the e2e leg pass. Add the `ai_config` leg to `tests/integration/test_tenant_utils.py:92-120`, the in-repo template for a field added to both tenant representations, so the plumbing hunks fail on revert. If you keep a fast local check, collapse the two copied test bodies into one `@pytest.mark.parametrize` over the four resolution branches (`ai_config` only, legacy only, both so `ai_config` wins, neither so `None`) against the finding-2 seam, with `create_autospec(AIServiceFactory)`. The both-set precedence branch is unpinned today: swap the two lines at `products.py:671-673` and the suite stays green.

**5. The silent skip the PR names as half the problem is unchanged, and this diff widens it**

`src/core/tools/products.py:710` still reads `logger.debug("[GET_PRODUCTS] AI ranking configured but AI not enabled (no API key)")`. It is the `else` of the `if` this diff rewrote at line 675, so this diff changes that branch. Its sibling for the identical situation at line 238 logs at `warning` with the tenant id, and the success path of the same block logs at INFO (lines 703-708), so the skip is the only outcome of the block that is invisible in normal logs. The PR description states this as broken ("Nothing in the response or at INFO level indicates that ranking was skipped, which makes this hard to diagnose from the outside") and the diff does not touch it.

The diff also widens the unreported case: `ai_config` tenants now enter the ranking block for the first time, so the `score < 0.1` exclusion at line 701 starts dropping products for a class of tenants that never reached it, with nothing distinguishing "no matching inventory" from "curation excluded them". The response is built with a hard-coded `errors=None` at line 762, eleven lines away, and AdCP 3.1.1 declares the channel: `dist/schemas/3.1.1/media-buy/get-products-response.json` has `errors` ("Task-specific errors and warnings (e.g., product filtering issues)") and `incomplete`, with the partial-success shape (`status: "completed"` plus advisory `errors[]`) documented at `dist/docs/3.1.1/building/by-layer/L3/error-handling.mdx:786-820`. The repo already does this at `src/core/tools/media_buy_list.py:177-184`.

Raise line 710 to `logger.warning` with the tenant id and which resolution failed, and add an advisory `errors[]` entry at the response construction, graded as a wire field in the finding-4 scenario. Do not use `CONFIGURATION_ERROR`: in 3.1.1 it is `terminal` and sellers MUST flip transport failure markers (`error-handling.mdx:519`), which is the wrong shape for a response that still returns products; the `code` vocabulary is open (`dist/schemas/3.1.1/core/error.json`). If diagnosability is not going to land in this PR, drop the claim from the problem statement so the description and the diff agree.

- Out of scope: `TenantContext.from_orm_model` (`src/core/tenant_context.py:100-126`) and `serialize_tenant_to_dict` (`src/core/utils/tenant_utils.py:36-62`) are two hand-maintained field-by-field projections of the same ORM row, which is why adding one column cost two edits here. Unifying them is a separate architectural boundary whose work lives outside this diff. Noted only because the finding-2 fix makes both edits unnecessary rather than duplicating them.
- Out of scope: `src/core/database/models.py:97-103` comments `ai_config.api_key` as encrypted while `src/admin/blueprints/settings.py:547-577` writes it in plain text. You raised this in the PR description; the diff neither reads nor writes the encryption path, so it is a separate boundary. No issue number is cited anywhere in this PR or its comment history, so there is no deferral to verify. File a follow-up issue and self-assign it, naming that scope: reconcile the comment with the plaintext write by fixing one or the other.

src/core/tools/products.py
```diff
@@ -663,8 +663,17 @@ async def _get_products_impl(
...
+
+            if factory.is_ai_enabled(tenant_ai_config):
```
KonstantinMirin: This line and line 676 route tenant-stored JSON into the factory, and both call `TenantAIConfig.model_validate` (`src/services/ai/factory.py:109,224`). `pydantic.ValidationError` is a `ValueError`, so it is not caught by the deliberate fail-open `except (ImportError, RuntimeError, OSError)` at line 712, and one non-conforming `ai_config` row turns the whole `get_products` call into a 500 on all four transports where the surrounding code promises unranked products.

Reproduce:
```
.venv/bin/python -c "
from src.services.ai.factory import AIServiceFactory
try: AIServiceFactory().is_ai_enabled({'provider':'google','api_key':'k','settings':{'temperature':9}})
except Exception as e: print(type(e).__name__, isinstance(e,(ImportError,RuntimeError,OSError)))"
# -> ValidationError False
```
`settings` is round-tripped opaquely by `src/admin/blueprints/settings.py:565-566` and never re-validated, and the column is bare JSONB. Parse at the ORM boundary so the request path never validates raw tenant JSON; while the column stays bare, name `ValidationError` in the caught tuple at line 712.

src/core/tenant_context.py
```diff
@@ -56,6 +56,7 @@ class TenantContext(BaseModel):
     approval_mode: str = "require-human"  # BR-RULE-037: creative approval mode
     account_approval_mode: str | None = None  # BR-RULE-060: account approval mode (auto|credit_review|legal_review)
     gemini_api_key: str | None = None
+    ai_config: dict[str, Any] | None = None  # provider/model/api_key; supersedes gemini_api_key
```
KonstantinMirin, line 59: `TenantAIConfig` (`src/services/ai/config.py:33`) is documented as "the structure of the `ai_config` JSON column on the Tenant table", and this module's own docstring says it "Replaces the fragile `dict[str, Any]` tenant representation with a typed, validated Pydantic model". Typing the field `dict[str, Any] | None` gives callers no key or value checking, which is why `src/core/tools/products.py:673` hand-assembles the mapping and the factory re-runs `TenantAIConfig.model_validate` at `factory.py:109,224` twice per ranking request.

Declare it `ai_config: TenantAIConfig | None` and coerce once at the ORM boundary. The project's own type states the rule: "all new columns should specify a model - bare JSONType (no model) is legacy" (`src/core/database/json_type.py:6-8,22-47`), and `ai_config` is bare at `src/core/database/models.py:99`. JSONB storage is unchanged and `TenantAIConfig` is all-optional with `extra="ignore"`, so no data migration.

src/core/utils/tenant_utils.py
```diff
@@ -53,6 +53,7 @@ def serialize_tenant_to_dict(tenant: Tenant) -> dict[str, Any]:
         "account_approval_mode": tenant.account_approval_mode,
         "supported_billing": safe_json_loads(tenant.supported_billing, None),
         "gemini_api_key": tenant.gemini_api_key,
+        "ai_config": tenant.ai_config,
```
KonstantinMirin, line 56: `ai_config` carries `api_key` and `logfire_token` (`src/services/ai/config.py:46,49`), and this function's output is handed whole to the Slack notifier (`src/core/audit_logger.py:360` -> `get_slack_notifier(tenant_config=...)`, which reads three keys per `src/services/slack_notifier.py:39-48`) and populates the process-wide tenant ContextVar behind `get_current_tenant()`. It is also the only JSON field here that skips the `safe_json_loads` treatment the other 7 fields use.

Have the AI layer resolve its own configuration on demand instead: with `TenantAIConfig.from_tenant(...)` in `src/services/ai/config.py`, this hunk and the `TenantContext` hunk are both unnecessary and the credential stays inside `src/services/ai`. This line is also untested: deleting it leaves the whole unit suite green, because the new tests hand-build the tenant dict.

tests/unit/test_get_products_impl_coverage.py
```diff
@@ -512,6 +512,72 @@ async def test_ai_not_enabled_returns_unranked_products(self):
...
+        tenant = _make_tenant()
+        tenant["product_ranking_prompt"] = "rank by relevance"
+        tenant["ai_config"] = ai_config
+        identity = _make_identity(principal_id="user-1", tenant_id="test-tenant", tenant=tenant)
+        req = _make_request()
+
+        mock_factory = MagicMock()
+        # Keep AI off so the test stops at the factory calls it is asserting on,
+        # rather than needing a working agent.
+        mock_factory.is_ai_enabled.return_value = False
```
KonstantinMirin: Forcing `is_ai_enabled` to `False` means `create_model(tenant_ai_config=...)` at `src/core/tools/products.py:676` never executes, so half of the fix is unreachable by construction while the docstring at line 526 claims the config "reaches both is_ai_enabled and create_model". Verified mutation: revert line 676 to `factory.create_model()` and `pytest tests/unit/test_get_products_impl_coverage.py -k TenantConfig` still reports 2 passed. Swapping the two branches at `products.py:671-673` also stays green, so the `ai_config`-wins-over-legacy precedence is unpinned.

The assertion at line 551 compounds it: `mock_factory = MagicMock()` has no `spec`, so `assert_called_once_with(ai_config)` asserts this call site's positional punctuation rather than `AIServiceFactory.is_ai_enabled`'s contract. Grade the observable the PR names (product order and threshold filtering with ranking on), and build the double with `create_autospec(AIServiceFactory, instance=True)`.

tests/unit/test_get_products_impl_coverage.py
```diff
@@ -512,6 +512,72 @@ async def test_ai_not_enabled_returns_unranked_products(self):
...
+    silently got no ranking — the miss is only logged at debug level.
+    """
+
+    @pytest.mark.asyncio
+    async def test_tenant_ai_config_is_passed_to_factory(self):
+        """The tenant's ai_config reaches both is_ai_enabled and create_model."""
+        ai_config = {"provider": "google", "model": "gemini-flash-lite-latest", "api_key": "tenant-key"}
+        tenant = _make_tenant()
+        tenant["product_ranking_prompt"] = "rank by relevance"
+        tenant["ai_config"] = ai_config
```
KonstantinMirin: `_make_tenant()` (lines 42-50) returns a plain dict that the test writes `ai_config` into, so neither `TenantContext.from_orm_model` (`src/core/tenant_context.py:121`) nor `serialize_tenant_to_dict` (`src/core/utils/tenant_utils.py:56`) runs. Both plumbing hunks stay green when deleted, and they are load-bearing on the real path: production resolves through `LazyTenantContext._resolve` -> `get_tenant_by_id` -> `serialize_tenant_to_dict` -> `TenantContext.from_dict`, and `from_dict` filters to declared model fields at `tenant_context.py:141-142`.

The repo has the template for this change shape: `tests/integration/test_tenant_utils.py:92-120` (`test_account_approval_mode_round_trips_through_tenant_context`) grades a newly plumbed field through the ORM model, the serializer, `from_orm_model`, `from_dict` and `ctx.get()`. Add the `ai_config` leg there, and seed the tenant with `TenantFactory(ai_config=...)` so the assertion fails when either hunk is reverted.

RATO-tdunkel:
The four CodeQL log-injection alerts (CWE-117) on this branch are fixed in b26c2b3c.

All four are mine — `git blame` puts them in 6082528b and 40c16930 — so they are new
findings, not pre-existing lines that happened to land in the diff.

Two interpolate a tenant identifier. The other two are the ones that matter: the ranking
failure logs the provider exception, whose text carries the remote response body verbatim,
and `TenantAIConfig.coerce` logs a `ValidationError`, whose message embeds the
seller-stored JSON it rejected and renders over several lines. The exception text was
already deliberately kept off the buyer's wire for exactly that reason, and then written
to the log unescaped — an inconsistency worth naming rather than quietly patching.

The repo already had the primitive, private to `webhook_validator.py`, and it is wider
than the reflex fix: `str.splitlines` treats ten characters as a line boundary and
`urlsplit` strips three of them, so escaping `\n` and `\r` alone leaves seven ways through
(VT, FF, the file/group/record separators, NEL, U+2028, U+2029). It is now
`src/core/utils/log_safe.py`, used by both callers instead of copied, and it escapes
rather than deletes so an operator can see that a value tried to break the line.

Worth flagging about the grading, since it is the kind of thing this review has rightly
pushed on: the first mutation run over the five call sites caught only three. Both
tenant-identifier sites were held by nothing — reverting either `log_safe` call left the
suite green. One test now drives both in a single request (a tenant carrying a ranking
prompt AND an enabled advertising policy trips both notices), and the assertion that a
record is a single line moved into the shared `_assert_one_products_warning` helper, so it
covers its six call sites rather than one. Second run: five of five.

`make quality` exits 0 locally (6654 passed); the duplication, complexity and type-ignore
baselines are unchanged.

RATO-tdunkel:
Could a maintainer approve the workflow runs on this branch when convenient? The runs on
the newest commit (`27bc5cd7`) sit at `action_required`, so only the three
`pull_request_target` checks have executed.

State, so the approval is worth the click:

- The full suite passed on the previous commit (`b26c2b3c`): 30 checks green, including
  Quality Gate, Unit, all Integration shards, BDD, E2E, Storyboard Conformance and CodeQL.
- Two commits since then close the CodeQL log-injection alerts (CWE-117) this branch
  introduced. The second one exists because CodeQL caught a path the first fix missed —
  the label `from_tenant` builds from the tenant id, which I had taken for an internal
  constant without reading where it is assembled. CodeQL was right and I was not.
- `make quality` exits 0 locally on `27bc5cd7` (6656 passed), with the duplication,
  complexity and type-ignore baselines unchanged.
- I cannot confirm alert 1151 is resolved until CodeQL runs on `27bc5cd7`. The fix is
  verified locally and mutation-tested, but that is not the same thing, so I would rather
  say so than imply the alerts are closed.

Two red checks are worth discounting in advance: **Security Audit** and **pip-audit** both
report the same 8 advisories on `httpx2` / `httpcore2` 2.3.0. They are not from this
branch — `main`'s own Security workflow has been failing since 2026-09-09 (`99846481`),
and this PR touches neither `pyproject.toml` nor `uv.lock`. A bump wants its own change,
not least because `httpx` carries the egress seam #1802 just introduced. Happy to file an
issue for it if that is useful rather than already tracked.

No rush on the review itself — the approval is the only thing blocking the evidence.

RATO-tdunkel:
#1721 landed while this was in review, and it changes the ground this PR stands on. Before
I spend the rework in the wrong direction, I would rather ask how you want it handled.

What it does to this branch:

- **It deletes the file this PR's unit coverage lives in.**
  `tests/unit/test_get_products_impl_coverage.py` is gone, along with
  `test_get_products_response_str.py` and `test_products_transport_wrappers.py`. That is
  where the four-branch resolution test, the advisory tests and last week's log-injection
  regressions all sit. The rebase stops on that conflict immediately.
- **That looks deliberate rather than incidental**, and I read it as the right call: the
  new `tests/CLAUDE.md` says a direct `_impl` call "verifies nothing on the wire" and that
  `Transport.IMPL` is removed. So I am not looking to restore the file.
- `_get_products_impl` is rewritten, down to the signature (`PublicIdentity`).
- `src/core/errors/codes.py` now owns `recovery`, `suggestion` and `message` for every
  code. This PR's `_ADVISORY_RECOVERY` table is now a duplicate of that, so it should go
  and the advisory should read from the one table.

What survives untouched is the part your review pushed for: the four-transport BDD module
and the two integration tests over the real `LazyTenantContext -> serialize_tenant_to_dict
-> from_dict` chain. Both are wire-based, which is the direction #1721 moves in.

My proposed plan, if you agree:

1. Rebase onto current `main`.
2. Drop `_ADVISORY_RECOVERY`; read recovery from `src/core/errors/codes.py`.
3. Re-apply the `products.py` changes onto the rewritten function.
4. Keep the BDD and integration legs as they are.
5. Do **not** recreate the deleted unit file. Where a deleted test graded something the
   BDD leg does not yet cover, add the scenario rather than a new `_impl` test.

Two things I would like a steer on:

- Is that the shape you want, or would you rather this be closed and reopened fresh
  against the new `main`? The diff will look quite different either way, and if you prefer
  a clean PR I would rather do that than force a large rebase through review history.
- Step 5 is the one I am least sure about. Some of the deleted coverage was seam-level
  rather than wire-level — for example that a malformed `ai_config` row degrades instead
  of raising, which is a `TenantAIConfig.coerce` property, not a `get_products` property.
  Should that kind of thing live as a plain unit test on the owning module, or do you want
  it graded through a scenario too?

Separately, the CI runs on `27bc5cd7` are still at `action_required`, so the last two
commits (the CodeQL log-injection fixes) have never been verified by CI. That approval is
still the only thing blocking that evidence.

Issue #2160, opened 2026-08-31
update_media_buy(paused=true) does not advance revision — BR-RULE-291 INV-5 requires every state-changing write to bump


numarasSigmaSoftware:
## Summary

A bare `update_media_buy(paused=true)` (or `paused=false`) does not advance the media buy's
`revision`. Pausing flips `is_paused` — a state change the buyer observes on the next read —
so per **BR-RULE-291 INV-5** it must bump the counter.

## Spec grounding

`tests/bdd/features/BR-UC-019-query-media-buys.feature` INV-5, verbatim:

> every successful state-changing write increments revision by at least 1

The pinned `update-media-buy-response.json` types `revision` as the buy's optimistic-
concurrency token: *"Use this value in subsequent update_media_buy requests intended to
change state."* A pause changes state, so a buyer that paused and then sends its last-known
token would be comparing against a value the pause did not move — a lost-update window.

## Why the graded scenario does not catch it

`T-UC-019-inv-291-5` is live (UC-019 has a `_detect_uc` branch; the scenario passes, not
xfailed). But its step is generic — *"one successful update_media_buy lands"* — and the
harness satisfies it with a **field-changing** update, which does bump. No graded scenario
drives a **bare pause** specifically, so the pause instance of INV-5 is an uncovered case
rather than a failing test.

## Mechanism

`_update_media_buy_impl` routes a pause through the adapter action
`pause_media_buy`/`resume_media_buy` (`src/core/tools/media_buy_update.py`, `req.paused`
branch). That path performs the adapter call but never calls `update_fields`/`update_status`
on the repository, so it never reaches `_bump_revision`. Verified pre-existing: the same
routing is present at `26783202f` (the current main), before any revision-enforcement work.

## Fix sketch

Route the pause/resume state change through the repository so the write and the bump share a
transaction, as every other state-changing update already does — or, minimally, bump on the
`is_paused` transition. Add a UC-019 scenario that drives a bare pause and asserts the
revision advanced, so INV-5's pause case is graded rather than incidentally covered.

Found while implementing the supplied-token CONFLICT enforcement (a separate change); that
work enforces a *supplied* token but does not change how a pause writes, so this is out of
its scope.

numarasSigmaSoftware:
[...]

Issue #2128, opened 2026-08-27
media_buy_update raw-session writes are untracked, and aliased imports hide them from the raw-select guard


KonstantinMirin:
Found by auditing every bare `# FIXME:` marker in `src/` and `tests/` against the issue tracker. This site matched nothing.

## The debt

`src/core/tools/media_buy_update.py:391` takes `session = uow.session` inside `_update_media_buy_impl`. The block below is dominated by **writes**, not reads:

    session.add(ObjectWorkflowMapping(...))   :614, :1382
    session.delete(...)                       :926, :1130
    session.flush()                           :958, :1194, :1230

## Why #1119 does not cover it

#1119 (*migrate remaining raw `select()` calls to repository pattern*) is scoped to `select()` and its acceptance is "`ALLOWLIST` in `test_architecture_no_raw_select.py` is empty". Two things follow:

1. It says nothing about `session.add`/`delete`/`flush`, which is most of what this block does.
2. **This file is absent from that allowlist entirely — and not because it is clean.** The guard discovers ORM names from `models.py`, and this module imports them aliased:

       from ... import CreativeAssignment as DBAssignment   # :66
       from ... import Product as DBProduct                 # :74

   so `select(DBAssignment)` / `select(DBProduct)` at `:910`, `:1079`, `:1122`, `:1142` are invisible to a name-based detector.

**So #1119 can be closed in full — allowlist empty, acceptance met — while this file is never touched.** That is the finding worth more than the debt itself: a guard whose membership is decided by import spelling, with a spelling it cannot see.

## Fix

Route the writes through `WorkflowRepository` / the relevant repositories. Separately, resolve ORM names through the module`s `ImportFrom` asnames in `test_architecture_no_raw_select.py` so aliased spellings are visible — otherwise this class of hole recurs silently in any module that renames on import.

PR #2053, opened 2026-08-21
feat: persisted media-buy revision counter + confirmed_at emission

numarasSigmaSoftware:
## What

The AdCP media-buy optimistic-concurrency fields, and nothing else:

- **`revision`** — a persisted monotonic counter (`media_buys.revision`), bumped on every
  successful mutation, emitted by `create_media_buy`, `update_media_buy` and `get_media_buys`.
- **`confirmed_at`** — the instant the seller committed to running the buy, written once
  (`approved_at` on the manual-approval path, `created_at` on synchronous auto-approve).
- **CONFLICT enforcement** — a supplied token that does not match the persisted revision
  rejects with `CONFLICT`, and that gate runs *before* the terminal-state gate so the buyer
  gets refetch-and-retry rather than "gone".

37 files, +1057 src / +2148 tests.

## Why this PR exists

This is the counter extracted from #1544. That PR grew to 261 files by folding in eight
unrelated workstreams; the counter itself was reviewed and found settled, so it is lifted
here to land on its own. The riders have their own issues and will land separately:
#2043 (approval finalizer / lease subsystem), #2044 (validation-taxonomy migration),
#2045 (UC-005 type-extension contract).

## Spec grounding

Pinned **AdCP 3.1.1** (`adcp==6.6.0`). `create-media-buy-response.json` `oneOf[0].required`
includes `confirmed_at` and `revision`; `update-media-buy-response.json` requires `revision`;
the `get-media-buys-response.json` item requires both. `confirmed_at` is typed
`["string","null"]` — nullable but required — so it is widened back to nullable via subclass
while `revision` stays a non-null `int` with `ge=1`.

`revision` is typed `{"type": "integer", "minimum": 1}` under draft-07, where `integer`
matches any number with a zero fractional part — so `7.0` is **valid** and is accepted on
every transport, while `7.5`, `"7"` and booleans are rejected.

`revision` is graded by no 3.1.1 conformance storyboard (ungraded). `confirmed_at` is graded
presence-only at IO-signing by the `sales-guaranteed` storyboard.

## Grading

Per-transport integration wire tests (a2a/mcp/rest) pin the emitted envelope, and each of the
three load-bearing behaviours is mutation-verified — reverting it reddens its own grader and
nothing else:

- reverting the revision bump → 7 tests fail
- reversing the CONFLICT/terminal precedence → all three transports fail
- rejecting whole-number floats again → all three transports fail

## Known gaps, stated up front

[...]

PR #2173, opened 2026-09-01
fix: enforce the media-buy revision optimistic-concurrency token in update_media_buy

numarasSigmaSoftware:
## What

`update_media_buy` enforces the AdCP `revision` optimistic-concurrency token: a supplied
`revision` that does not match the persisted value is rejected with `CONFLICT`, and the
comparison is made atomically with the write.

## Why

main already persists `revision`, bumps it on mutation, and emits it — and
`UpdateMediaBuyRequest.revision` carries the pinned spec text verbatim:

> When provided, sellers **MUST reject the update with `CONFLICT`** if the media buy's
> current revision does not match, and **MUST enforce that comparison atomically with the
> write.**

`_build_update_request` plumbs the value into the request. `_update_media_buy_impl` never
read it. So today a buyer can send a correct token, receive `200`, and have the update
applied over a concurrent write — the field is advertised but the guarantee is not honoured.

## The design: the token is a condition on the write, not an operation beside it

The comparison is a filter on the single advance the repository already owns, not a separate
locked read-then-write:

```
UPDATE media_buys SET revision = revision + 1, <fields...>
WHERE tenant_id = :t AND media_buy_id = :id AND (:expected IS NULL OR revision = :expected)
```

`rowcount == 1` succeeds; `rowcount == 0` re-reads to distinguish a missing buy
(`MEDIA_BUY_NOT_FOUND`) from a mismatch (`CONFLICT`, naming both versions). It is atomic in
one statement, so it needs no held row lock and does not wait on #1644; it runs before any
adapter call because the write does. A mutating request advances the parent revision by
**exactly one** (tokened or not); a no-op advances by zero. This retires the two-phase claim,
the seven manual claim sites, the prepayment ledger, and the lock-timeout branch the earlier
compare-beside-the-write shape needed.

## What this adds

- The value contract: strict `int >= 1`; whole-number floats accepted and normalised
  (A2A delivers integers as doubles); numeric strings, booleans, `1.5`, non-finite, and an
  explicit `null` rejected as `VALIDATION_ERROR` (a schema-value violation — the same class
  and code as a malformed `idempotency_key` on this model). The bound is read once from the
  pinned schema (`update_media_buy_revision_schema`) and feeds every site.
- The CONFLICT error shape, with the expected/current version pair in `details`.
- The atomic conditional advance above, ordered so `CONFLICT` is raised before any adapter
  side effect.
- One shared seam for the value contract and presence on all transports: the request is
  built from the payload AS SENT (`raw_wire_payload`), so `UpdateMediaBuyRequest` answers
  "did the buyer send `revision`?" (and the explicit-null rejection) in one place — the
  per-transport gates (`validate_revision_wire_value`, the MCP compat middleware's
  `_gate_revision_argument`, the REST `model_fields_set` read, the A2A hand-list) are gone.
  Routing the A2A handler through the shared builder also forwards every field
  (`idempotency_key`, `currency`, `pacing`, flight dates, `ext`), addressing the
  `update_media_buy` leg of #1885.
- Wire-level oracles grading emitted `revision`, the CONFLICT pair, and every accept/reject
  input on mcp/rest/a2a; the UC-003 revision scenarios dispatch as raw wire parameters and
  are graded on the wire.

## Spec grounding (ungraded)

The `revision` value contract is grounded in the pinned request schema
`dist/schemas/3.1.1/media-buy/update-media-buy-request.json` (`revision`:
`{"type": "integer", "minimum": 1}`). Choosing `VALIDATION_ERROR` for a forbidden value
matches the value/range taxonomy and the sibling `idempotency_key` rule on this same model.
The AdCP 3.1.1 conformance storyboard does **not** grade `revision`
(`grep -rn "revision" dist/compliance/3.1.1 --include=*.yaml` returns only narrative and a
sample body), so this behaviour is ungraded; the citation is recorded here and in the two
revision test-module docstrings per the CLAUDE.md Spec-Grounding Gate.

## Grading

The UC-003 revision scenarios previously marked expected-to-fail (citing #1607 and #1885)
now pass and are graduated in this branch, graded on the wire. The atomicity is proven by a
race test: two independent connections + a `threading.Barrier` produce one success and one
`CONFLICT`; removing the conditional `WHERE revision = :expected` clause reddens it (a
demonstrated lost update).

## Known limits, stated up front

- **`paused=true` bumping and package-only writes.** Under the single-advance design a
  mutating request advances the parent revision exactly once via the backstop
  `advance_revision`, which covers the pause and package/targeting/creative-assignment
  paths in the same transaction. The raw session writes at `media_buy_update.py`
  (`FIXME(#2128)`) are not relocated behind repository methods in this PR — the correctness
  guarantee rests on the single in-transaction advance, not on each write bumping; the
  relocation remains #2128's scope. #2160 (`paused` bump semantics) is pre-existing.
- **Deferred-update apply time has no owner.** The manual-approval submission branch compares
  the token without spending it (the update is not applied yet); no deferred-update executor
  exists in the tree to re-compare at apply time. Tracked as a follow-up.
- **A2A emits these integers as JSON doubles** (`2.0` where mcp/rest emit `2`). Schema-valid
  under draft-07 and deliberately not normalised; the wire tests assert the type per transport.
- #1644 (nested `get_db_session()` ending the transaction) is a separate architectural
  boundary; finding #1's design is chosen so it does not wait on it.

## Relationship to #2053

This is the enforcement layer from #2053, rebased onto main. #1941 landed that PR's
substrate (the migrations, columns, counter, backfill, and status vocabulary) on a different
design; this branch keeps only what main still lacks — the conditional advance and its
transport/validator surface. Full context on #2053, which this supersedes.


KonstantinMirin (commented):
## Review — PR #2173

**Overview** — #1607 requires `update_media_buy` to reject a stale `revision` with `CONFLICT` and to enforce that comparison atomically with the write (`dist/schemas/3.1.1/media-buy/update-media-buy-request.json`, `revision`). The wire-facing half lands: a stale token answers a schema-shaped `CONFLICT` on MCP, REST, and A2A read off real wire bytes, and the two-connection `threading.Barrier` race test is a real oracle. The atomicity requirement is not met in general, and the PR says so: the guarantee holds only for a single-package update with no inline-creative sync. That gap comes from a design choice rather than a missing case — the comparison is built as an operation beside the write instead of a condition on the write, so it needs a row lock the flow cannot hold, and the two-phase split, the seven manual claim sites, the prepayment ledger, and the lock-timeout branch are all compensation for that. Reconsider the shape before adding anything else to it; most of the items below disappear when it changes.

### Should fix

#### 1. The revision comparison is an operation beside the write, not a condition on it

`compare_and_set_revision` takes its own row lock and its own advance, so `_update_media_buy_impl` has to carry that lock by hand from the comparison to whichever write runs. What that costs today:

- Two comparison methods with the same body: `assert_revision_matches` (`src/core/database/repositories/media_buy.py:681`) and `compare_and_set_revision` (`:725`). The first documents itself as "deliberately NOT sufficient on its own".
- Seven hand-written call sites in one function: `src/core/tools/media_buy_update.py:474`, `:707`, and the `_claim_revision()` closure at `:490` invoked at `:831`, `:986`, `:1398`, `:1481`, `:1495`, each with a comment explaining why that branch is the last point where the lock still holds. An eighth write branch will not call it, and nothing will say so.
- `_prepaid_revision_advance` (`repositories/media_buy.py:74`, consumed at `:536`, recorded at `:790`), a per-request credit whose only job is to cancel the second advance this design creates. Its own docstring records that the credit "can be undone underneath us ... Measured on two consecutive updates in one process, that goes both ways". `MediaBuyRepository` is now the only one of 12 repositories that carries request state.
- `_lock_row_for_revision_check` (`:631`) with `set_config('lock_timeout')`, a `55P03` branch, and `AdCPRevisionConflictError.unobserved`, a second `CONFLICT` that carries no `current_version`. Nothing executes that branch: `grep -rn "55P03\|lock_timeout" tests/` returns nothing, and `tests/unit/test_adcp_contract.py:3685` constructs the exception directly instead of reaching it. `SET LOCAL lock_timeout` is transaction-scoped, so the 3 s cap stays in force for every later write in the same request, where nothing translates `55P03` into an AdCP error.

Two consequences the buyer sees:

- `_claim_revision()` at `:831` runs after `adapter.update_media_buy()` returns, so a `CONFLICT` on the pause and resume branch is raised after the ad server has already been changed. "Reject the update with CONFLICT" is supposed to mean nothing was applied.
- Sending a token changes how far the counter moves. `paused=True` with no token writes nothing and leaves `revision` where it was; the same request with a matching token advances it through `:831`, and `:1495` advances a request that wrote nothing at all. `tests/integration/test_media_buy_revision.py:173` asserts `after > before` for that tokened pause, so the divergence is asserted as correct, while the real "a token changes nothing about how far the counter moves" oracle at `:178` is applied only to the budget branch, where it holds.

**Root cause.** The token is modelled as a second write (compare, advance, then reconcile) whose correctness depends on a Postgres row lock surviving to the write. It cannot survive, because nested `get_db_session()` and `CreativeUoW` blocks share the scoped session and end the transaction mid-request (#1644, pre-existing). This design builds on that fault instead of routing around it.

**Fix:** Make the token a filter on the advance the repository already owns. `_bump_revision` is documented as the sole owner of the advance, with 6 in-repository call sites. Give `MediaBuyRepository.update_status` and `update_fields` an `expected_revision: int | None = None` and emit one statement, `UPDATE media_buys SET revision = revision + 1, ... WHERE tenant_id = :t AND media_buy_id = :id AND revision = :expected`, with `rowcount == 0` triggering a re-read and `AdCPRevisionConflictError.mismatch`. That is atomic in one statement, so it needs no held lock and does not wait on #1644, it runs before any adapter call because the write does, every `CONFLICT` names both versions, and no future write branch can skip it because there is no other way to write. Pass `req.revision` from `_update_media_buy_impl` into the write it already makes, and delete `assert_revision_matches`, `compare_and_set_revision`, `_claim_revision` and its seven call sites, `_prepaid_revision_advance` and its `isinstance` guard, `_lock_row_for_revision_check`, `_REVISION_LOCK_TIMEOUT_MS`, the `55P03` branch, and `AdCPRevisionConflictError.unobserved` in the same pass. The raw session writes at `media_buy_update.py:429` (`FIXME(#2128)`) move behind repository methods in the same change, because the new guarantee holds only if every write goes through the bump point.

**Done when:** `grep -n "_claim_revision\|assert_revision_matches\|compare_and_set_revision\|_prepaid_revision_advance\|lock_timeout\|55P03\|unobserved" src/core/tools/media_buy_update.py src/core/database/repositories/media_buy.py src/core/exceptions.py` returns nothing; `test_racing_updates_with_the_same_token_produce_one_win_and_one_conflict` still passes; and a test asserts that a tokened `paused=true` and an untokened `paused=true` advance `revision` by the same amount, in the shape of `test_media_buy_revision.py:178`.

#### 2. Four boundaries each answer "did the buyer send `revision`?", and #1885's hand-list grew by one name

One rule, four enforcement sites, three spellings of the same question:

```
src/core/mcp_compat_middleware.py:29,40   _REVISION_TOOL = "update_media_buy"; "revision" in arguments
src/a2a_server/adcp_a2a_server.py:2090    "revision" in params
src/routes/api_v1.py:375                  "revision" in body.model_fields_set
src/core/schemas/_base.py:2487            _normalize_revision, a fourth application
```

`mcp_compat_middleware.py` is the schema-driven envelope-tolerance layer; it now knows one tool name and one field name. `api_v1.py:375` is the only read of `model_fields_set` in `src/`. To keep the raw value alive for the REST gate, the body field lost its type contract: `revision: RawRevision = None` where `RawRevision = Annotated[Any, ...]` (`api_v1.py:133`). And `validate_revision_wire_value(*, present: bool, value)` makes `present=True, value=<nothing sent>` a representable state.

The comment at `mcp_compat_middleware.py:83` says the arguments dict "is the last point at which what the buyer actually sent is still visible". The tree one module away says otherwise. `raw_wire_payload` is this codebase's answer to "what did the buyer literally send", and it is already plumbed on all three transports: `src/core/mcp_auth_middleware.py:60`, `src/routes/rest_compat_middleware.py:59` into `api_v1.py:310,324`, `src/a2a_server/adcp_a2a_server.py:1505,1602`, consumed by `create_media_buy` at `src/core/tools/media_buy_create.py:2223,4723`.

The same edit closes #1885 on the A2A leg. `adcp_a2a_server.py:2077` builds an `UpdateMediaBuyRequest` from five hand-picked keys and then uses only `media_buy_id` and `paused`; `:2093-2104` forwards a second hand-list of ten keys to a tool that rebuilds the request through `_build_update_request` (`media_buy_update.py:1795`) anyway. Measured against `_build_update_request`'s 17 parameters, the A2A hand-list still drops `flight_start_date`, `flight_end_date`, `currency`, `pacing`, `daily_budget`, `ext`, and `idempotency_key` — so BR-RULE-215 INV-6 ("revision and idempotency_key are evaluated independently") cannot hold on A2A, for the same reason `revision` did not. The xfail reason this PR deleted named the remedy in as many words: route the handler through `media_buy_update._build_update_request`, which already forwards every field. Adding `revision` to the hand-list is the opposite move. `tests/bdd/conftest.py:840` records "This work closes #1885 as well as #1607"; the diff does not support that, and the PR body says "partially addresses".

**Root cause.** Request construction has two owners: a builder in the business layer, and a per-transport re-derivation in each handler. Every new request field has to be remembered in three places, and the failure mode is silent field loss. Presence is then re-derived at each of those places because the typed model is built before anything has read the raw payload.

**Fix:** Thread `raw_wire_payload` into `update_media_buy_raw` and `_build_update_request` the way `create_media_buy_raw` already does — REST through the route's existing `raw_json_body` dependency, A2A by widening the `if skill_name == "create_media_buy"` case at `adcp_a2a_server.py:1602` to every handler, MCP by reading the ctx state in the wrapper as `media_buy_create.py:4723` does. Let `UpdateMediaBuyRequest`'s `mode="before"` validator answer `"revision" in payload` and apply the value rule. If a transport genuinely cannot carry the mapping, use a sentinel-typed parameter (`revision: int | AbsentType = ABSENT`) so presence travels inside the value instead of beside it. Delete `_REVISION_TOOL`, `_gate_revision_argument`, the three `present=` derivations, the `Any`-typed REST field, and both A2A hand-lists; keep only the handler's legacy `updates.packages` pre-processing. Do not generalize the middleware to "any tool with a revision"; that keeps per-field knowledge in the layer whose whole value is having none. While you are in `_base.py`, annotate `_normalize_revision(cls, values: Any) -> Any` to match `normalize_legacy_geo` at `:1541`. Correct `tests/bdd/conftest.py:840` to state what the diff does, and update `update_media_buy_raw`'s docstring at `media_buy_update.py:1785`, which still tells callers the stale-token CONFLICT check is an ungraded gap.

**Done when:** `grep -rn "validate_revision_wire_value\|_REVISION_TOOL\|model_fields_set" src/` returns nothing, `grep -n "params.get(" src/a2a_server/adcp_a2a_server.py` shows no field hand-list in `_handle_update_media_buy_skill`, and an A2A test asserts that `idempotency_key` sent over A2A reaches the tool.

#### 3. The graduated BDD rows never reach a transport, and the new protocol behavior has no scenario

The two below-minimum rows graduated at `tests/bdd/conftest.py:842` do not dispatch. The When step builds the model in the test process:

```python
# tests/bdd/steps/domain/uc003_update_media_buy.py:804
req = UpdateMediaBuyRequest(**update_kwargs)
except AdCPError as e:
    ctx["error"] = e
    return          # <- returns before dispatch_request
```

`UpdateMediaBuyRequest(media_buy_id='mb', paused=True, revision=0)` raises `AdCPInvalidRequestError`, which subclasses `AdCPError`, so the step returns and `ctx["result"]` stays `None`. `_assert_error_outcome` (`uc002_create_media_buy.py:1317`) then falls back to asserting on the in-process exception under three transport labels. Delete all three boundary gates this PR added and the rows stay green. What flipped them from XFAIL to XPASS is the exception class changing inside the test process, not seller behavior. The conftest's stated reason, that the PR "dropped the field's `ge=1`", is contradicted by `_base.py:2363` (`Field(ge=1)`) and by this PR's own `test_ge_constraint_is_inherited_not_redeclared`.

`T-UC-003-partition-revision[wrong_type "7"]` is dormant: `Given the request revision is set to "7"` matches neither `{revision:d}` nor the `<not provided>` literal, so it falls into the non-strict `StepDefinitionNotFoundError` auto-xfail. A typed, strict, `ref=`-bearing registry entry became an unmonitored one. The remedy is written out in the file itself — `given_request_revision_absent` (`uc003_update_media_buy.py:2367`) exists precisely because an Examples literal an int parser cannot match "failed on StepDefinitionNotFoundError and the suite's non-strict auto-xfail absorbed it". The harness for the quoted form ships in this diff: `MediaBuyDualEnv.call_via` takes raw flat kwargs and `test_update_media_buy_revision_validation_wire.py:44` already sends the string `"7"` on three transports.

Five behaviors this PR introduces have no scenario at all and are tested only through `_update_media_buy_impl` with patched internals: `CONFLICT` outranking the terminal-state gate, compare-without-spend on submission, the dry-run compare, and the per-branch claim placement (`test_media_buy_revision.py:222,245,305,337,361`, with `mock.patch.object`). `:489` asserts `adapter_spy.assert_not_called()`, in-process state no wire path can observe. The explicit null, the whole-number float, the version pair, and the context echo live in a bespoke module pinned to `[MCP, REST, A2A]`, so the BDD lane's fourth transport is never asked. Two accept-side tests assert only `not result.is_error` (`test_update_media_buy_revision_validation_wire.py:112,121`), so they cannot tell a honoured token from one that was validated and discarded: make `validate_revision_wire_value` return `None` for a whole-number float and `test_accepts_a_whole_number_float` still passes.

**Root cause.** The typed model is constructed on the test side of the boundary, so any input the model refuses can never reach the seller and the scenario tests the harness instead of the wire. Protocol-contract tests then got a second home with its own transport list rather than going into the feature file that already owns UC-003 and already runs four transports.

**Fix:** Give the UC-003 When step a raw-payload branch that hands `update_kwargs` to the transport unconverted — `tests/bdd/steps/domain/local_constraint_relaxations.py:114` is the shape to copy, including its `then_refused_on_wire` that refuses to fall back to `ctx["error"]`. Add the quoted-value Given. Re-point the two below-minimum rows and the `wrong_type` row at it. Move the explicit null, the whole-number float, the version pair, the context echo, and the post-write revision into `BR-UC-003-update-media-buy.feature`, verified with `assert_wire_error` and `wire_dict` on all four transports. Re-express the pre-adapter fail-fast on something a transport can observe, such as the buy being unchanged on a subsequent read, instead of `adapter_spy.assert_not_called()`. Do not register the resulting e2e variants in any `*_E2E_*_INTERNAL_TAGS` set. Keep the `_impl` tests as the fast inner loop. Replace the two hand-rolled `wire is not None` guards (`test_update_media_buy_revision_validation_wire.py:180` and the sibling in `test_media_buy_revision.py`) with `result.require_wire()` (`tests/harness/transport.py:428`).

**Done when:** `make check-dormant` reports no dormant row under `@T-UC-003-partition-revision`; deleting `_gate_revision_argument` from `mcp_compat_middleware.py` makes the below-minimum rows fail; and `grep -n "adapter_spy.assert_not_called\|mock.patch.object" tests/integration/test_media_buy_revision.py` returns only lines inside tests that do not claim to verify buyer-visible behavior.

#### 4. Two structural guards now check an empty population

The diff removes the last three `cause=... scope=... ref=...` typed xfail reasons from `tests/bdd/`, and in the same PR replaces the anti-vacuity floor `typed_seen >= 3` with `typed_seen == declared` (`tests/unit/test_architecture_bdd_xfail_reason_tokens.py:296`). That assertion now holds at `0 == 0`, with the violations loop running zero times. Named mutation: make `_is_xfail_call` return `False` unconditionally and the guard still passes. Confirm with `python3 -c "import ast,pathlib; print(sum(1 for p in pathlib.Path('tests/bdd').rglob('*.py') for n in ast.walk(ast.parse(p.read_text())) if isinstance(n,ast.Constant) and isinstance(n.value,str) and n.value.startswith('cause=')))"` — 0 at head, 3 on main.

`TestTheRoutingOutcome::test_a_per_transport_reason_keeps_its_mcp_and_rest_rows` (`:441`) asserts that 8 rows survive collection because they carry `scope=per-transport`. This PR deleted those markers, so `conftest.py:3059` (`if strict_xfails and not per_transport`) never deselects them — they now survive whatever the predicate does. The test's own comment already warns about this population: "8 of those rows carry no strict xfail at all and survive either way".

**Root cause.** The population these guards check is production markers, which are designed to be removed when the gap closes, so the guard stops being able to fail once the work succeeds. The response here was to relax the enforcement so the diff could pass, which moves the ratchet the wrong way.

**Fix:** Pick one. Either give the parser guard a population it owns (one synthetic scenario carrying a strict `cause=... scope=per-transport ref=...` reason, collected in a temporary bdd session, so an inverted predicate fails regardless of the real tree), or delete the guard together with the `cause=`/`scope=`/`ref=` taxonomy and the `scope=per-transport` deselection exemption at `conftest.py:3019-3062`, which now has no user. Prefer deleting over keeping a test that reports coverage it does not have. Restoring the numeric floor is not the fix; it would block the next legitimate graduation.

**Done when:** inverting the `parsed.scope == "per-transport"` predicate at `tests/bdd/conftest.py:3057` makes at least one test in `test_architecture_bdd_xfail_reason_tokens.py` fail, or that file and the exemption are both gone.

#### 5. The repository decides the buyer-facing error code, and the exception logs from its constructor

`repositories/media_buy.py:711` and `:766` answer "no row for (tenant, id)" with `AdCPInvariantViolationError`, which maps to `SERVICE_UNAVAILABLE` with `recovery="transient"` (`src/core/exceptions.py:1202`). That tells a buyer to back off and retry a media buy that is gone or was never theirs, and `tests/integration/test_media_buy_repository_writes.py:649` asserts it for a cross-tenant id. The same class already answers this correctly at `:93` — `get_by_id_or_raise` raises `AdCPMediaBuyNotFoundError`, and its docstring says it exists to collapse this guard into one place.

`AdCPInvariantViolationError.__init__` calls `logger.error(..., stack_info=True)` (`exceptions.py:1221`), so constructing the exception emits the line whether or not anyone raises it. Run `python -c "from src.core.exceptions import AdCPInvariantViolationError; AdCPInvariantViolationError('row vanished')"` and you get an ERROR with a stack trace for an exception that was never raised. The PR's own `test_invariant_violation_logs_the_diagnostic` demonstrates it. This is a second error channel that `record_boundary_error` (`src/core/tool_error_logging.py:168`) does not know about, and that function owns error observability and its severity policy across MCP, A2A, and REST.

**Root cause.** Two established boundaries were bypassed. `get_by_id_or_raise` owns which AdCP code absence gets; `record_boundary_error` owns error observability. The logging call sits in `__init__` because the exception discards its diagnostic (`super().__init__(_INVARIANT_VIOLATION_MESSAGE)`), leaving the constructor as the only place it can go.

**Fix:** Route both lookups through `get_by_id_or_raise` so absence has one answer in the class and no second answer can be written, and keep `AdCPInvariantViolationError` for states the schema genuinely excludes. Update `test_other_tenant_cannot_compare_and_set` to assert `MEDIA_BUY_NOT_FOUND`. Carry the diagnostic as a non-wire attribute on the exception, kept out of `to_dict` and `details`, and let `record_boundary_error` emit it, so construction has no side effect. While you are in that file, hoist the function-local `from src.core.exceptions import ...` at `repositories/media_buy.py:106` and `:230` into the module-level import this diff added at `:28`, so the file has one import style.

**Done when:** `grep -n "AdCPInvariantViolationError" src/core/database/repositories/media_buy.py` returns nothing, and `logger.error` no longer appears in `AdCPInvariantViolationError.__init__`.

#### 6. The pinned `minimum` is hand-written four times, and the two published schemas already disagree

`minimum: 1` and `ge=1` appear at `src/core/schemas/_base.py:2336` (`RawRevision`'s `WithJsonSchema`), `:2363` (`RevisionValidator`'s `Field(ge=1)`), `:2375` (the buyer-facing message "at least 1"), and `src/core/tools/media_buy_update.py:1673` (the MCP wrapper parameter). `src/core/schemas/_pinned_fields.py:192` already exposes `revision_minimum()`, which reads the SDK's bundled pin, and production already uses it at `src/core/tools/media_buy_list.py:718`. `_base.py:590` states the rule verbatim: a bound written in a second place is a bound that can diverge from the pin.

The two published interfaces have already diverged, and the new guard cannot see it:

```
MCP  tool schema:  {"default": null, "minimum": 1,   "type": "integer"}
REST OpenAPI:      {"type": "integer", "minimum": 1.0}
```

`default: null` is not an integer, and it is a value this PR's own gate rejects with `INVALID_REQUEST` — a generated client that sends the published default gets an error. `tests/unit/test_adcp_contract.py:3746,3758` compares against a hand-written `1`, and `1 == 1.0` in Python, so the guard passes both ways and would keep passing if the pin moved.

**Fix:** Add a `@cache`d accessor beside `revision_minimum()` that reads `update-media-buy-request.json`, build one published schema fragment from it, and feed `RawRevision`, the MCP wrapper annotation, `Field(ge=...)`, and the error message from that one value. Annotate the MCP parameter with `RawRevision` rather than repeating the fragment. Point the guard at the accessor and add an assertion that MCP and REST publish the same fragment and that neither publishes a `default`. Under finding 2 the wrapper's `WithJsonSchema` has no reason to exist at all. Separately, drop the hand-passed `recovery=` arguments in `test_update_media_buy_revision_validation_wire.py` — `assert_wire_error` already defaults to the pinned enum's classification (`tests/harness/transport.py:463`), so passing `recovery="transient"` restates a pinned fact in the test.

**Done when:** `grep -rn '"minimum": 1\|ge=1' src/core/schemas/_base.py src/core/tools/media_buy_update.py` returns no `revision` site, and a test asserts the MCP and REST published `revision` fragments are equal.

#### 7. One model answers one class of schema violation with two error codes, and nothing records whether the storyboard covers it

`_validate_revision` raises `AdCPInvalidRequestError` (`INVALID_REQUEST`), while `_check_idempotency_key` on the same `UpdateMediaBuyRequest` raises `VALIDATION_ERROR` through `validate_idempotency_key_shape`. Both are pinned-enum values and both are `correctable`, so the wire code is not wrong — but a buyer gets two codes for one failure class on one request. `AdCPInvalidRequestError`'s own docstring (`src/core/exceptions.py:684`) says the storyboard defines the code per operation, so callers must use the class graded for that scenario. Nothing grades this one: `grep -rn "revision" dist/compliance/3.1.1 --include=*.yaml` returns only narrative and a sample body. `CLAUDE.md:135` requires the spec citation plus the storyboard step that grades it, or an explicit "ungraded" note, and #1607's criterion 5 asked for the same.

**Fix:** Pick one code for "the buyer sent a field value the pinned schema does not admit" on this model and apply it to both fields, with the reason recorded next to the choice. Add the grading note to the PR description and to the two new test module docstrings, citing `dist/schemas/3.1.1/media-buy/update-media-buy-request.json` and stating that the 3.1.1 conformance storyboard does not grade this behavior.

**Done when:** `grep -n "VALIDATION_ERROR\|INVALID_REQUEST" src/core/schemas/_base.py` shows one code for the `UpdateMediaBuyRequest` field-value violations, and both new test modules carry the ungraded note.

#### 8. The new tests re-derive read and seed helpers that an existing module owns

`tests/helpers/media_buy_write_seam.py` exists because the "did the write bump revision by exactly one" assertion "was asserted in six places, in five spellings ... with three different answers to the question that actually decides whether the assertion is real: WHICH session to read the row back through". This diff adds spellings seven and eight: `read_revision` (`tests/integration/test_media_buy_revision.py:136`) copies that session reasoning into a fresh docstring, and `test_a_field_writing_update_advances_the_revision_exactly_once` (`:178`) rewrites by hand the exact-delta assertion `assert_status_move_carried_bookkeeping(..., bumps=1)` already encodes. `tests/integration/test_media_buy_repository_writes.py:607,640,699` add three more inline re-reads.

The same shape appears in the seeding block: the four-statement ritual (`setup_media_buy_data()` + `MediaBuyFactory(...)` + `env._commit_factory_data()` + `env._seeded_media_buy_id = ...`) plus the same REST-URL comment appears at `test_update_media_buy_revision_validation_wire.py:53` and `test_media_buy_revision.py:515`, the second carrying two `# noqa: SLF001` waivers. The three-transport list is spelled out twice as `WIRE_TRANSPORTS` and `_WIRE_TRANSPORTS`, with a comment at `test_media_buy_revision.py:498` naming the duplication.

**Root cause.** `MediaBuyDualEnv` has no public way to seed an existing media buy for the update path, so every caller builds one, and the read-side owner exists but is not reachable from these tests without going through it.

**Fix:** Use `read_media_buy_state(TENANT_ID, media_buy_id)` and the existing exact-delta assertion from `tests/helpers/media_buy_write_seam.py:52,82`; if the revision-only case needs a narrower entry point, add `assert_revision_advanced(before, after, bumps=1)` to that module so there is still one owner. Add `MediaBuyDualEnv.seed_existing_media_buy(media_buy_id, *, status="active", revision=None)` in `tests/harness/media_buy_dual.py` that creates the row, commits the factory data, and sets the routing id, then delete both copies and both `SLF001` waivers. Put the three-transport list on `tests/harness/transport.py` as a module constant.

**Done when:** `grep -rn "_commit_factory_data\|_seeded_media_buy_id" tests/integration/test_media_buy_revision.py tests/integration/test_update_media_buy_revision_validation_wire.py` returns nothing, and `read_revision` is gone.

- The A2A `2.0` fork pinned by `WIRE_NUMBER_TYPE` (`test_update_media_buy_revision_validation_wire.py:159`) is forced by the wire, not chosen: A2A payloads leave through `Part(data=_dict_to_value(...))` and `_dict_to_value` builds a `google.protobuf.struct_pb2.Value` (`src/a2a_server/adcp_a2a_server.py:190`), whose only numeric kind is a double. Both forms are draft-07 conformant and the type check is the only assertion that can see the fork appear or disappear. Keep it as it is.
- #1644 (nested `get_db_session()` ending the transaction) is a separate architectural boundary and its repair is out of this diff. Finding 1's remedy is chosen so it does not wait on it. #2160 (`paused=true` does not advance `revision`) and #2128 (raw session writes bypass the parent bump) are pre-existing on main and out of scope as behavior. All three are OPEN with no assignee, so none of them establishes an owned deferral for the residual of #1607's atomicity requirement. #1885 is OPEN and assigned to KonstantinMirin, not to you, so it does not support the "closes #1885" claim either. File and self-assign whatever residual you intend to carry out of this PR, naming the exact scope.
- Deferred-update apply time has no owner. The submission branch compares the token without spending it (`src/core/tools/media_buy_update.py:700-712`) on the argument that the update is not applied yet, but nothing re-compares when it is applied, and no deferred-update executor exists in the tree. Out of scope because the missing executor is a boundary this diff neither adds nor touches. File and self-assign a follow-up for "re-compare `revision` when an approved deferred update is applied" rather than leaving the correctness argument in a code comment.

src/core/tools/media_buy_update.py
```diff
@@ -439,6 +439,64 @@ def _update_media_buy_impl(
...
+            # the whole of what a simulation owes the buyer.
+            #
+            # Called at most once per request, at the LAST point before the first
+            # persistent write of whichever branch runs, so the row lock it takes is
+            # still held when that write commits. Once is what keeps the advance count
+            # at one: this prepays a single advance and the following write draws on the
+            # prepayment instead of taking its own (see ``_bump_revision``).
+            claimed_revision = False
+
+            def _claim_revision() -> None:
```
KonstantinMirin: This closure and its `claimed_revision` latch are invoked by hand at six branch sites in this function (`:831`, `:986`, `:1398`, `:1481`, `:1495`), each with a comment explaining why that branch is the last point where the row lock still holds. A seventh write branch will not call it and nothing will fail.

Make the token a filter on the advance the repository already owns instead: give `MediaBuyRepository.update_status` and `update_fields` an `expected_revision: int | None = None` and emit `UPDATE media_buys SET revision = revision + 1, ... WHERE tenant_id = :t AND media_buy_id = :id AND revision = :expected`, raising `AdCPRevisionConflictError.mismatch` on `rowcount == 0` after a re-read. That is atomic in one statement, needs no held lock, and cannot be skipped. Then delete this closure, `assert_revision_matches`, `compare_and_set_revision`, and `_prepaid_revision_advance`.

src/core/tools/media_buy_update.py
```diff
@@ -760,6 +824,11 @@ def _update_media_buy_impl(
                     )
                     return UpdateMediaBuyResult(response=error_response, status=AdcpTaskStatus.failed.value)
                 else:
+                    # Phase 2: the adapter is done, so this is the last point before the
+                    # pause/resume's persistent effect on the row is read back and
+                    # reported. Deliberately after the adapter call, not before it: the
+                    # adapter can open a nested session and drop the lock.
+                    _claim_revision()
```
KonstantinMirin: This claim runs after `adapter.update_media_buy()` has returned, so a `CONFLICT` raised here is raised after the ad server has already been changed. The pinned schema (`dist/schemas/3.1.1/media-buy/update-media-buy-request.json`) says sellers MUST reject the update with `CONFLICT`, which means nothing was applied.

A conditional `UPDATE` inside the repository write orders the comparison before every adapter call by construction, because the write does.

src/core/tools/media_buy_update.py
```diff
@@ -1410,7 +1487,12 @@ def _update_media_buy_impl(
                     logger.warning("GAM sync NOT implemented - GAM still has old dates")
 
             # Create ObjectWorkflowMapping to link media buy update to workflow step
-            # This enables webhook delivery when the update completes
+            # This enables webhook delivery when the update completes.
+            # Phase 2 backstop: a request whose branches wrote nothing to the media buy
+            # row still lands this mapping, and a honoured token is still spent exactly
+            # once -- so an accepted update always leaves the buyer a fresh token,
+            # exactly as the single pre-split comparison did.
+            _claim_revision()
```
KonstantinMirin: This advances `revision` for a request whose branches wrote nothing to the media buy row. Combined with `:831`, whether the buyer sent a token now changes how far the counter moves: `paused=True` with no token leaves `revision` alone, and the same request with a matching token advances it.

`tests/integration/test_media_buy_revision.py:173` asserts `after > before` for that tokened pause, so the divergence is asserted as correct, while the real "a token changes nothing about how far the counter moves" oracle at `:178` is applied only to the budget branch. Under a conditional `UPDATE` only real writes advance and this backstop has nothing to do.

src/core/database/repositories/media_buy.py
```diff
@@ -57,6 +67,11 @@ class MediaBuyRepository:
     def __init__(self, session: Session, tenant_id: str) -> None:
         self._session = session
         self._tenant_id = tenant_id
+        #: ``{media_buy_id: revision}`` — the value ``compare_and_set_revision`` advanced
+        #: a buy TO in this unit of work. The next write to that buy spends the advance
+        #: instead of taking its own, but only while the row still carries the recorded
+        #: value — see ``_bump_revision``.
+        self._prepaid_revision_advance: dict[str, int] = {}
```
KonstantinMirin: This makes `MediaBuyRepository` the only one of 12 repositories that carries request state, and it exists only to cancel the second advance that `compare_and_set_revision` takes. The `_bump_revision` docstring records that the credit "can be undone underneath us ... Measured on two consecutive updates in one process, that goes both ways" - a correctness mechanism whose input is nondeterministic.

If the token is a `WHERE revision = :expected` clause on the write the repository already makes, there is no second advance to cancel and this dict, its `isinstance` guard at `:536`, and the record at `:790` all go away.

src/core/database/repositories/media_buy.py
```diff
@@ -576,6 +628,168 @@ def update_status(
...
+        the query says rather than on a refresh the ORM is not asked for.
+
+        The lock timeout keeps a contended request from pinning a connection behind
+        whoever holds the row. Timing out means the current revision was never
+        observed, which is a conflict the buyer can retry -- not a server fault.
+        """
+        # Postgres SET does not accept bind parameters; set_config(..., is_local=true)
+        # does, and is likewise reverted when this transaction ends.
+        self._session.execute(
+            text("SELECT set_config('lock_timeout', :timeout, true)"),
```
KonstantinMirin: `set_config('lock_timeout', ..., is_local=true)` is transaction-scoped, so this 3 s cap stays in force for every later write in the same request, and nothing translates a `55P03` raised by those writes into an AdCP error.

The `55P03` branch below and `AdCPRevisionConflictError.unobserved` are also unreachable from any test: `grep -rn "55P03\|lock_timeout" tests/` returns nothing, and `tests/unit/test_adcp_contract.py:3685` constructs the exception directly instead of reaching this code. A conditional `UPDATE` takes no lock, so there is nothing to time out and this whole helper goes away.

src/core/database/repositories/media_buy.py
```diff
@@ -576,6 +628,168 @@ def update_status(
...
+        clause. See ``_update_media_buy_impl``.
+
+        Raises:
+            AdCPRevisionConflictError: the current revision differs from
+                ``expected_revision``.
+            AdCPInvariantViolationError: the row is gone.
+        """
+        media_buy = self.get_by_id(media_buy_id)
+        if media_buy is None:
+            raise AdCPInvariantViolationError(
```
KonstantinMirin: `AdCPInvariantViolationError` maps to `SERVICE_UNAVAILABLE` with `recovery="transient"`, so a buyer whose media buy is gone or was never theirs is told to back off and retry. `tests/integration/test_media_buy_repository_writes.py:649` asserts that for a cross-tenant id.

This class already answers absence correctly at `:93`: `get_by_id_or_raise` raises `AdCPMediaBuyNotFoundError`, and its docstring says it exists to collapse this guard into one place. Route both this site and `:766` through it, and update the cross-tenant test to assert `MEDIA_BUY_NOT_FOUND`.

src/core/mcp_compat_middleware.py
```diff
@@ -9,19 +9,39 @@
...
 from src.core.request_compat import deep_strip_to_schema, normalize_request_params, strip_unknown_params
+from src.core.schemas import validate_revision_wire_value
 from src.core.tool_error_logging import _translate_to_tool_error, record_boundary_error
 
 logger = logging.getLogger(__name__)
 
+#: The only AdCP tool whose request carries an optimistic-concurrency ``revision``.
+#: Named rather than applied to every tool so a stray key on some other tool keeps
+#: whatever handling that tool already has.
+_REVISION_TOOL = "update_media_buy"
```
KonstantinMirin: This is the schema-driven envelope-tolerance layer, and it now knows one tool name and one field name. The comment at `:83` says the arguments dict "is the last point at which what the buyer actually sent is still visible", but `raw_wire_payload` already answers that question and is already plumbed on MCP (`src/core/mcp_auth_middleware.py:60`, read in the wrapper at `src/core/tools/media_buy_create.py:4723`).

Thread the raw payload into `_build_update_request` and let `UpdateMediaBuyRequest`'s `mode="before"` validator answer `"revision" in payload` once, then delete `_REVISION_TOOL` and `_gate_revision_argument`. Generalizing this to "any tool with a revision" keeps per-field knowledge in the layer whose value is having none.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -2083,6 +2083,12 @@ async def _handle_update_media_buy_skill(self, parameters: dict, identity: Resol
                 context=params.get("context"),
             )
 
+        # The buyer's optimistic-concurrency token, gated by the shared value contract.
+        # `params` is a plain dict, so this is the one place that can still tell an
+        # omitted `revision` from an explicitly-supplied null — the pinned request
+        # schema gives the field no null arm, and further in both read as None.
+        revision = validate_revision_wire_value(present="revision" in params, value=params.get("revision"))
```
KonstantinMirin: Third site answering "did the buyer send `revision`?", in a third spelling (`mcp_compat_middleware.py:40` uses `"revision" in arguments`, `src/routes/api_v1.py:375` uses `body.model_fields_set`). `raw_wire_payload` is already built at `:1505` and already forwarded to `create_media_buy` at `:1602`.

Widen that `if skill_name == "create_media_buy"` case to every handler and hand the mapping to `_build_update_request`, so presence is answered once inside the request model.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -2093,6 +2099,7 @@ async def _handle_update_media_buy_skill(self, parameters: dict, identity: Resol
             packages=params.get("packages"),
             push_notification_config=params.get("push_notification_config"),
             context=params.get("context"),
+            revision=revision,
```
KonstantinMirin: This adds one more name to the hand-list #1885 exists to delete. The xfail reason this PR removed named the remedy: route the handler through `media_buy_update._build_update_request`, which already forwards every field.

Measured against `_build_update_request`'s 17 parameters, this list still drops `flight_start_date`, `flight_end_date`, `currency`, `pacing`, `daily_budget`, `ext`, and `idempotency_key`, so BR-RULE-215 INV-6 (revision and idempotency_key evaluated independently) cannot hold on A2A, for the same reason `revision` did not. Replace both hand-lists with one call that forwards `params`, keeping only the legacy `updates.packages` pre-processing.

src/routes/api_v1.py
```diff
@@ -126,10 +126,11 @@ class UpdateMediaBuyBody(SalesAgentBaseModel):
...
     #
-    # The seller does not yet ACT on it — the stale-token CONFLICT check is a separate,
-    # still-xfailed gap (BR-RULE-215 partitions). Accepting it is transport parity, not
-    # a claim that concurrency is enforced.
-    revision: int | None = None
+    # Typed RawRevision (Any) rather than ``int | None`` so the wire value survives to
+    # the shared gate in the route below. A narrower annotation coerces first — Pydantic
+    # would read the string "7" as 7 here while a plain-dict boundary would not, which
+    # is one rule on paper and three behaviours on the wire.
+    revision: RawRevision = None
```
KonstantinMirin, line 136: The REST body field gives up its type contract (`RawRevision` is `Annotated[Any, ...]`) so the raw value survives to the route's gate below. That is the boundary being in the wrong place: the raw payload is already available on this route as `raw_wire_payload` (`:310,324`).

Forward the raw mapping to `_build_update_request` and restore `revision: int | None = None` here.

src/routes/api_v1.py
```diff
@@ -367,6 +368,14 @@ async def update_media_buy(media_buy_id: str, body: UpdateMediaBuyBody, identity
         push_notification_config = to_push_notification_config(body.push_notification_config)
         context = to_context_object(body.context)
         reporting_webhook = to_reporting_webhook(body.reporting_webhook)
+    # The buyer's optimistic-concurrency token, gated by the shared value contract.
+    # ``model_fields_set`` is the only thing here that still knows whether the buyer
+    # sent the key at all: an omitted `revision` and an explicitly-supplied null both
+    # arrive as None, and the pinned request schema gives the field no null arm.
+    revision = validate_revision_wire_value(
```
KonstantinMirin: This is the only read of `model_fields_set` in `src/`, and it is the fourth place that decides whether `revision` was present. The signature it feeds, `validate_revision_wire_value(*, present: bool, value)`, makes `present=True, value=<nothing sent>` a representable state.

Pass `raw_wire_payload` (already a parameter of this route at `:324`) into `update_media_buy_raw` and let the request model's `mode="before"` validator answer `"revision" in payload`.

src/core/schemas/_base.py
```diff
@@ -2304,6 +2310,93 @@ def validate_idempotency_key_shape(key: str | None) -> None:
...
+#: differently -- FastMCP turns the string "7" into 7 against an ``int | None``
+#: annotation, while a plain dict boundary hands "7" through untouched. The result
+#: would be one rule on paper and three behaviours on the wire. Carry the wire value
+#: under this alias, then run it through ``validate_revision_wire_value``.
+#:
+#: ``WithJsonSchema`` is schema-only: it publishes the buyer-facing contract (an integer
+#: >= 1) on generated interfaces without adding a validator, so the ``Any`` runtime
+#: carrier -- and the string/null distinction it preserves -- is unchanged. A revision
+#: of 0 still reaches ``validate_revision_wire_value`` and is rejected there.
+type RawRevision = Annotated[Any, WithJsonSchema({"type": "integer", "minimum": 1})]
```
KonstantinMirin: `minimum: 1` is written by hand here, at `:2363` (`Field(ge=1)`), at `:2375` (the error message), and at `src/core/tools/media_buy_update.py:1673`. `src/core/schemas/_pinned_fields.py:192` already exposes `revision_minimum()` reading the SDK's bundled pin, and `media_buy_list.py:718` already uses it. `:590` in this file states the rule: a bound written in a second place is a bound that can diverge from the pin.

Add a cached accessor beside `revision_minimum()` for the request schema, build one published fragment from it, and feed all four sites from that value.

src/core/schemas/_base.py
```diff
@@ -2390,6 +2483,26 @@ def unwrap_and_parse(cls, values):
 
         return values
 
+    @model_validator(mode="before")
+    @classmethod
```
KonstantinMirin, line 2495: New `mode="before"` validator with no parameter or return annotation, in a file where the siblings all declare one: `normalize_legacy_geo(cls, values: Any) -> Any` (`:1541`), `remove_invalid_fields(cls, values: dict) -> dict` (`:1940`), `_validate_package_update_shape(cls, data: Any) -> Any` (`:2239`).

Annotate it `def _normalize_revision(cls, values: Any) -> Any:`. This validator is also where the presence rule belongs once the raw payload reaches the model.

src/core/tools/media_buy_update.py
```diff
@@ -1582,7 +1664,14 @@ async def update_media_buy(
...
     ext: dict[str, Any] | None = None,  # AdCP ExtensionObject for custom fields
     idempotency_key: Annotated[str | None, Field(description="Idempotency key for retry safety")] = None,
-    revision: Annotated[int | None, Field(description="Expected current revision (optimistic concurrency)")] = None,
+    revision: Annotated[
+        int | None,
+        Field(description="Expected current revision (optimistic concurrency)"),
+        # Schema-only: publish the buyer-facing contract (integer >= 1) on the MCP tool
+        # schema. Not a validator — revision=0 still reaches _validate_revision (the shared
+        # gate) and is rejected there with INVALID_REQUEST, same as REST/A2A.
+        WithJsonSchema({"type": "integer", "minimum": 1}),
```
KonstantinMirin: Fourth hand-written copy of the pinned `minimum`, and this one publishes a contract the server rejects. The registered MCP tool advertises `{"default": null, "minimum": 1, "type": "integer"}` while REST publishes `{"type": "integer", "minimum": 1.0}`. `null` is not an integer, and a generated client that sends the published default gets `INVALID_REQUEST` from `validate_revision_wire_value`.

Annotate this parameter with `RawRevision` rather than repeating the fragment, build that fragment from `_pinned_fields`, and suppress the `default` on the published schema.

src/core/exceptions.py
```diff
@@ -1121,6 +1121,107 @@ class AdCPIdempotencyExpiredError(AdCPConflictError):
...
+    ``error-code.json`` enumMetadata, and a conformance oracle grades that match.
+    ``INTERNAL_ERROR`` is deliberately not used -- it is absent from the pinned enum
+    and would only be remapped to ``SERVICE_UNAVAILABLE`` anyway.
+
+    The diagnostic is logged in ``__init__`` rather than at the raise sites so that
+    no raise site can drop it, and so the buyer-facing message stays fixed.
+    """
+
+    def __init__(self, diagnostic: str, **kwargs: Any) -> None:
+        logger.error("AdCP invariant violated: %s", diagnostic, stack_info=True)
```
KonstantinMirin: Logging in `__init__` means constructing the exception emits the ERROR, whether or not anyone raises it. `python -c "from src.core.exceptions import AdCPInvariantViolationError; AdCPInvariantViolationError('row vanished')"` prints an ERROR with a stack trace for an exception that was never raised, and the new `test_invariant_violation_logs_the_diagnostic` relies on exactly that.

This is also a second error channel that `record_boundary_error` (`src/core/tool_error_logging.py:168`) does not know about, and that function owns severity policy across MCP, A2A, and REST. Carry the diagnostic as a non-wire attribute kept out of `to_dict` and `details`, and let `record_boundary_error` emit it.

tests/bdd/conftest.py
```diff
@@ -823,80 +823,38 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
...
-                        strict=True,
-                    )
-                )
+        # Graduated (#1885, a2a, 4 rows — the same 4 CONFLICT rows on a2a): its reason
+        # said implementing CONFLICT ALONE would leave these red, because
+        # _handle_update_media_buy_skill rebuilt the request from hand-listed fields and
+        # `revision` was in none of them, so the token never reached the tool. That drop
+        # is fixed too — the handler now gates and forwards `revision`
+        # (adcp_a2a_server.py, `validate_revision_wire_value`) — which is why these xpass
+        # rather than staying red. This work closes #1885 as well as #1607.
```
KonstantinMirin: The diff does not close #1885. `_handle_update_media_buy_skill` still rebuilds the request from two hand-listed field subsets (`adcp_a2a_server.py:2077` and `:2093-2104`), which is the bypass the issue exists to delete. `revision` was the field that exposed it. The PR body says "partially addresses", and #1885 is assigned to KonstantinMirin, not to you.

This registry is a graded artifact, so a false "closes" retires the remaining bypass sites silently. Correct the comment to state what the diff does.

tests/bdd/conftest.py
```diff
@@ -823,80 +823,38 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
...
+        #
+        # Graduated ([order R3]): the two revision-below-minimum rows (partition
+        # below_min "0", boundary "revision 0 (below minimum 1)") now XPASS on every
+        # transport, so no explicit xfail is applied to them any more. This PR gates the
+        # wire-supplied `revision` at the transport boundary (validate_revision_wire_value
+        # -> AdCPInvalidRequestError) and dropped the field's `ge=1`, so a below-minimum
+        # token now reaches the seller and is answered with the INVALID_REQUEST the
+        # scenario asserts -- instead of being blocked by a construction-time `ge=1`
```
KonstantinMirin: These two rows do not reach a transport. The When step builds the model in the test process (`tests/bdd/steps/domain/uc003_update_media_buy.py:804`), `UpdateMediaBuyRequest(revision=0)` raises `AdCPInvalidRequestError` which subclasses `AdCPError`, and the step returns before `dispatch_request`. `_assert_error_outcome` then asserts on the in-process exception under three transport labels. Delete all three boundary gates this PR added and these rows stay green.

The stated reason is also wrong: `_base.py:2363` still carries `Field(ge=1)`, as this PR's own `test_ge_constraint_is_inherited_not_redeclared` asserts. Give the When step a raw-payload branch (`tests/bdd/steps/domain/local_constraint_relaxations.py:114` is the shape) and re-point these rows at it.

tests/bdd/conftest.py
```diff
@@ -823,80 +823,38 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
...
+        # ValidationError that never reached the wire. Reverting the boundary's
+        # INVALID_REQUEST emission reddens these rows, which is what makes the graduation
+        # real rather than vacuous.
+        #
+        # The wrong_type "7" row still XFAILs, for a DIFFERENT and true reason: its Gherkin
+        # step `Given the request revision is set to "7"` has no step definition, so the
+        # StepDefinitionNotFoundError auto-xfail hook (pytest_runtest_makereport) routes it
+        # -- a missing-step-definition harness gap, not a production gap. No explicit marker
```
KonstantinMirin: A typed, strict, `ref=`-bearing registry entry became an unmonitored auto-xfail. The remedy is already written in the steps file: `given_request_revision_absent` (`tests/bdd/steps/domain/uc003_update_media_buy.py:2367`) exists precisely because an Examples literal an int parser cannot match "failed on StepDefinitionNotFoundError and the suite's non-strict auto-xfail absorbed it".

Write the quoted-value Given the same way. The harness ships in this diff: `MediaBuyDualEnv.call_via` takes raw flat kwargs and `tests/integration/test_update_media_buy_revision_validation_wire.py:44` already sends the string `"7"` on three transports.

tests/unit/test_architecture_bdd_xfail_reason_tokens.py
```diff
@@ -244,9 +278,25 @@ def test_every_cause_prefixed_reason_under_tests_bdd_parses(self):
...
+            f"{_conftest_under_grade()}, the conftest this guard imports the parser from — "
+            f"the scan is looking somewhere else, so its verdict covers nothing"
+        )
+        # Anti-vacuity, part 2: two independent censuses of the same tree agree.
+        # Deliberately NOT a floor. Typed reasons are meant to graduate, and a
+        # hardcoded minimum turns the last graduation into a guard failure; what
+        # must never happen is the xfail-call walk losing sight of a typed reason
+        # the tree still holds. When the last one graduates both counts read 0 and
+        # this passes, which is the correct answer.
+        assert typed_seen == declared, (
```
KonstantinMirin: This replaces the anti-vacuity floor `typed_seen >= 3` in the same PR that removed the last three typed reasons from `tests/bdd/`, so it now holds at `0 == 0` with the violations loop running zero times. Named mutation: make `_is_xfail_call` return `False` unconditionally and this still passes.

`TestTheRoutingOutcome::test_a_per_transport_reason_keeps_its_mcp_and_rest_rows` at `:441` has the same problem: the 8 rows it counts no longer carry a strict xfail, so `conftest.py:3059` keeps them whatever the predicate does. Either give this guard a population it owns (one synthetic scenario with a strict `cause=... scope=per-transport ref=...` reason, collected in a temp bdd session) or delete the guard together with the taxonomy and the `scope=per-transport` exemption at `conftest.py:3019-3062`, which now has no user.

tests/unit/test_adcp_contract.py
```diff
@@ -3570,5 +3579,184 @@ def test_v36_product_in_get_products_response(self):
...
+        """The registered update_media_buy MCP tool publishes revision as a bounded
+        integer, not a nullable integer with no lower bound."""
+        import asyncio
+
+        from src.core.main import mcp
+
+        tool = asyncio.run(mcp.get_tool("update_media_buy"))
+        revision = tool.parameters["properties"]["revision"]
+        assert revision.get("type") == "integer", revision
+        assert revision.get("minimum") == 1, revision
```
KonstantinMirin: This guard compares against a hand-written `1`, so it is the fifth copy of the pinned bound rather than a check on it. It also cannot see the divergence that already exists: MCP publishes `{"default": null, "minimum": 1, "type": "integer"}` and REST publishes `{"type": "integer", "minimum": 1.0}`, and `1 == 1.0` in Python.

Assert both interfaces against the `_pinned_fields` accessor, assert the two published fragments are equal to each other, and assert neither publishes a `default`.

tests/integration/test_media_buy_repository_writes.py
```diff
@@ -576,3 +578,123 @@ def test_bulk_create_tenant_isolation(self, tenant_a, tenant_b, principal_a, pri
...
+        with pytest.raises(AdCPInvariantViolationError):
+            with MediaBuyUoW(tenant_a) as uow:
+                uow.media_buys.compare_and_set_revision("mb_cas_never_existed", expected_revision=1)
+
+    def test_other_tenant_cannot_compare_and_set(self, tenant_a, tenant_b, principal_a):
+        """The locking re-read carries the same tenant filter as every other query."""
+        with MediaBuyUoW(tenant_a) as uow:
+            uow.media_buys.create(make_media_buy(tenant_a, principal_a, "mb_cas_iso"))
+
+        with pytest.raises(AdCPInvariantViolationError):
```
KonstantinMirin: This asserts that asking for another tenant's media buy answers `SERVICE_UNAVAILABLE` with `recovery="transient"`, which tells the buyer to retry something that will never succeed. The correct answer already exists in the same class: `get_by_id_or_raise` (`repositories/media_buy.py:93`) raises `AdCPMediaBuyNotFoundError`.

Route the lookup through it and change this assertion to `AdCPMediaBuyNotFoundError`.

tests/integration/test_media_buy_revision.py
```diff
@@ -0,0 +1,546 @@
...
+
+    Deliberately not another update_media_buy call: not every update action bumps the
+    counter (a bare pause does not), so driving the setup through the tool would make
+    the fixture depend on which action happens to write.
+    """
+    with MediaBuyUoW(TENANT_ID) as uow:
+        uow.media_buys.update_fields(media_buy_id, order_name="moved on by another writer")
+
+
+def read_revision(media_buy_id: str) -> int:
```
KonstantinMirin: `tests/helpers/media_buy_write_seam.py` already owns this decision. Its docstring says the obligation "was asserted in six places, in five spellings ... with three different answers to the question that actually decides whether the assertion is real: WHICH session to read the row back through". This is spelling seven, with the session reasoning copied into a fresh docstring.

Use `read_media_buy_state(TENANT_ID, media_buy_id)` (`media_buy_write_seam.py:52`) and, for the exact-delta check at `:178`, `assert_status_move_carried_bookkeeping(..., bumps=1)` (`:82`). If the revision-only case needs a narrower entry point, add `assert_revision_advanced(before, after, bumps=1)` to that module so there is still one owner.

tests/integration/test_media_buy_revision.py
```diff
@@ -0,0 +1,546 @@
...
+        before = read_revision("mb_rev_match")
+
+        result = _update_media_buy_impl(
+            req=UpdateMediaBuyRequest(media_buy_id="mb_rev_match", paused=True, revision=before),
+            identity=_identity(),
+        )
+
+        assert result.response.media_buy_id == "mb_rev_match"
+        after = read_revision("mb_rev_match")
+        assert after > before, "an honoured token must be spent -- the revision has to move"
```
KonstantinMirin, line 206: This asserts that a tokened `paused=True` advances `revision`. An untokened `paused=True` does not, because the pause path never reaches the repository, so this test asserts the divergence as correct: whether the buyer sends a token changes how far the counter moves.

`:178` states the right invariant ("supplying a token changes nothing about how far the counter moves") but applies it only to the budget branch. Once the token is a `WHERE revision = :expected` clause on the write, only real writes advance, and this assertion becomes an equality against the untokened advance.

tests/integration/test_media_buy_revision.py
```diff
@@ -0,0 +1,546 @@
...
+            assert exc_info.value.wire_error_code == "CONFLICT"
+            assert exc_info.value.details == {
+                "resource_id": mbid,
+                "expected_version": before,
+                "current_version": before + 1,
+            }
+            # The protection is the phase-1 fail-fast, not an atomic claim: the stale token
+            # was rejected before the adapter was even acquired. Dropping the phase-1 check
+            # lets execution reach get_adapter (and the adapter call), which is the gap.
+            adapter_spy.assert_not_called()
```
KonstantinMirin: `adapter_spy.assert_not_called()` is in-process state that no wire path can observe, and it is the only assertion covering "CONFLICT is raised before any adapter side effect", the part of #1607 that matters.

Re-express it on something a transport can see, such as the media buy being unchanged on a subsequent read, and move the scenario into `BR-UC-003-update-media-buy.feature` so it runs on all four transports.

tests/integration/test_media_buy_revision.py
```diff
@@ -0,0 +1,546 @@
...
+    with MediaBuyDualEnv() as env:
+        tenant, principal, _product, _pricing = env.setup_media_buy_data()
+        MediaBuyFactory(
+            tenant=tenant,
+            principal=principal,
+            media_buy_id=media_buy_id,
+            status="active",
+            revision=_WIRE_SEED_REVISION,
+        )
+        env._commit_factory_data()  # noqa: SLF001 — the harness's factory/session flush seam
```
KonstantinMirin: Second copy of the four-statement seeding block from `tests/integration/test_update_media_buy_revision_validation_wire.py:53`, with the same REST-URL comment reworded and two `# noqa: SLF001` waivers. The comment at `:498` names the duplication of the transport table as well.

The waivers are the signal: `MediaBuyDualEnv` has no public way to seed an existing buy for the update path. Add `MediaBuyDualEnv.seed_existing_media_buy(media_buy_id, *, status="active", revision=None)` in `tests/harness/media_buy_dual.py` that creates the row, commits the factory data, and sets the routing id, then delete both copies and both waivers. Put the three-transport list on `tests/harness/transport.py`.

tests/integration/test_update_media_buy_revision_validation_wire.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_accepts_a_whole_number_float(self, integration_db, transport):
+        """2.0 is schema-valid under draft-07 `integer` and must not be rejected.
+
+        A2A delivers integers as doubles, so refusing this form would refuse a
+        conformant buyer on one transport and accept it on the others.
+        """
+        with MediaBuyDualEnv() as env:
+            _seed(env)
+            result = _update(env, transport, revision=1.0)
```
KonstantinMirin, line 135: This asserts only `not result.is_error`, so it cannot tell a honoured token from one that was validated and then discarded. Make `validate_revision_wire_value` return `None` for a whole-number float and this test still passes. `test_accepts_a_plain_integer` at `:117` has the same gap.

Assert the emitted `revision` the way `test_successful_update_emits_the_advanced_revision` does, so the accept path proves the token reached the comparison.

tests/integration/test_update_media_buy_revision_validation_wire.py
```diff
@@ -0,0 +1,234 @@
...
+        the response model -- a model that serializes correctly in-process proves
+        nothing about what crossed the wire.
+        """
+        with MediaBuyDualEnv() as env:
+            _seed(env)
+            result = _update(env, transport, revision=1)
+
+        assert not result.is_error, result.wire_error_envelope
+        wire = result.wire_response
+        assert wire is not None, f"{transport} captured no success wire body to grade"
```
KonstantinMirin: This re-derives the guard `result.require_wire()` already provides (`tests/harness/transport.py:428`); `test_media_buy_revision.py` carries a second copy.

Call `result.require_wire()` and delete both. Also drop the hand-passed `recovery=` arguments at `:86`, `:102`, and the CONFLICT assertions below: `assert_wire_error` already defaults to the pinned enum's classification for the code (`transport.py:463`), so passing it restates a pinned fact in the test.

Issue #2181, opened 2026-09-02
fix: complete AdCP 3.1 version negotiation on MCP (echo + VERSION_UNSUPPORTED + wire UC-010)


numarasSigmaSoftware:
## Summary
Follow-up to #1512, which (Tier 1) made the MCP compat middleware **tolerate** the AdCP
version-envelope request fields `adcp_version`/`adcp_major_version` (strips them before
FastMCP's TypeAdapter rejects them). MCP currently **tolerates-but-ignores** — it never
reads the client's declared version, so the response-side negotiation the spec describes is
unimplemented on MCP.

Per `docs/test-obligations/storyboard-issue-map.yaml` (`universal/version-negotiation.yaml`,
`coverage: partial`): "#1512 covers accepting adcp_version/adcp_major_version on the wire.
The storyboard additionally grades VERSION_UNSUPPORTED on cross-major mismatch and same-major
downshift, which no issue tracks."

## Scope (all graded by `dist/compliance/3.1.1/universal/version-negotiation.yaml`)
1. **Read + echo** the client's `adcp_version` and echo the served release on responses;
   advertise `adcp.supported_versions` on `get_adcp_capabilities`. SHOULD at 3.1, **blocking at 3.2**.
   (A2A/REST already read `adcp_version`; MCP does not — `src/core/mcp_compat_middleware.py`
   strips it. Consider `NormalizationResult.declared_version` so all three transports share one seam.)
2. **`VERSION_UNSUPPORTED`** on cross-major mismatch, same-major downshift, and disagreeing
   dual-emit (`adcp_version` major ≠ `adcp_major_version`). Untracked today.
3. **Bind `BR-UC-010-discover-seller-capabilities.feature`** — its version-negotiation
   scenarios are written and `@source`-pinned to adcp v3.1.1 but have no `scenarios()` binding,
   so 0 are collected. Same dormancy class as #1973 (UC-001).

## Related
#1512 (Tier 1), #1955 (A2A supported_majors renders [3.0] vs [3]), #1958 (landing footer
version), #1973 (UC-001 BDD dormant).

numarasSigmaSoftware:
[...]

Issue #1969, opened 2026-08-13
MCP tools reject unknown request fields, but every AdCP request schema sets additionalProperties: true


KonstantinMirin:
## Summary

Every AdCP request schema our MCP tools implement sets `additionalProperties: true`, so an unknown request field is **schema-valid and must be tolerated**. We reject it.

Counted over the pinned 3.1.1 bundle:

```
request schemas: 116 | additionalProperties: {True: 104, False: 3, unset: 9}
```

The 3 strict ones are `core/pagination-request.json` and the two `trusted-match` schemas — none of them belong to a tool we register. For **all 16 registered MCP tools**, the schema says tolerate.

## Why we reject

FastMCP builds each tool's input schema from its Python signature and validates arguments with pydantic before the tool body runs. Anything not in the signature is rejected with:

```
VALIDATION_ERROR: Unexpected keyword argument
```

This is the same mechanism as #1512 and #1193, but those were about fields the spec *names*. This issue is the residue: fields the spec does not name, which `additionalProperties: true` nonetheless requires us to accept.

## Why this matters beyond conformance

It defeats forward compatibility exactly where the project already intends to have it. `CLAUDE.md` Critical Pattern #7 says production runs `extra="ignore"` (forward compatible) and dev/CI runs `extra="forbid"` (strict). That policy is implemented on our Pydantic models — but the MCP boundary never reaches them, because FastMCP rejects at the signature first. So a buyer on a newer AdCP release sending a field we have not adopted yet gets a hard failure from a seller that believes it is forward compatible.

## Not graded, and that is itself worth noting

No storyboard in the pinned tree grades unknown-request-field handling. `universal/schema-validation.yaml` — the obvious candidate, and one this repo has cited for it before — grades response conformance and temporal constraints instead. So this gap is invisible to the conformance runner and will not show up in the ledger; it has to be caught by reading the schema.

## Suggested fix

Mirror the existing environment-based policy at the MCP registration chokepoint (`src/core/main.py::_register_tool`, which already applies `accepts_spec_request_fields`): tolerate unknown fields when `ENVIRONMENT=production`, stay strict otherwise, matching `get_pydantic_extra_mode()`. That keeps typo detection in dev/CI — the only real argument for rejecting — while making the deployed agent behave as the schema requires.

## Provenance

[...]

PR #1546, opened 2026-07-06
feat: AdCP 3.1.1 version negotiation, envelope tolerance, and bearer normalization (#1512)

numarasSigmaSoftware:
## What

Makes the Sales Agent interoperable with clients pinned to the repository's
AdCP **3.1.1** contract and closes the transport/security gaps found while
reviewing that boundary.

**Scope note (2026-07-30):** the durable idempotency subsystem that briefly
lived on this branch (universal read-tool idempotency, downstream-mutation
claims/reconciliation, migrations `a4d7e8c91f20` + `b5c8f1d20a37`, A2A
task/webhook durability) has been split out to draft PR #1689, its designated
home, so this PR stays at the surface reviewers have already covered. The
revert commit on this branch removes exactly those three commits; nothing else
changed, and none of that work is lost.

### Version negotiation, envelopes, and authentication

- Derives the advertised AdCP release/major from the in-repo
  `ADVERTISED_ADCP_VERSIONS` constant, cross-checked against the pinned
  `adcp==6.6.0` SDK mapping by a loud guard.
- Resolves `adcp_version` and deprecated `adcp_major_version` consistently on
  MCP, A2A, and REST; unsupported pins return the real
  `VERSION_UNSUPPORTED` wire envelope with authoritative recovery details.
- Accepts standard AdCP envelope fields without discarding fields that the
  selected operation actually implements, and echoes request `context` from
  the real response wire body where the contract requires it.
- Normalizes bearer credentials through one parser on every ingress and pins
  AUTH-before-VERSION precedence without disclosing supported versions to an
  unauthenticated caller.

### Required mutation keys and honest capability advertisement

- Requires and validates `idempotency_key` on `create_media_buy`,
  `update_media_buy`, `sync_creatives`, and `sync_accounts` across MCP, A2A,
  REST, and discovery schemas. The shared boundary enforces the AdCP 3.1.1
  string, 16–255-character, and `^[A-Za-z0-9_.:-]{16,255}$` contract before
  business logic runs.
- Preserves the buyer's exact key; no transport fabricates a replacement for a
  missing or malformed key.
- Advertises seller-wide `adcp.idempotency.supported: false` with no replay TTL.
  `create_media_buy` retains its existing durable replay path (live and graded
  by the UC-002 replay scenario), but the other in-scope mutators can
  re-execute, so this PR does **not** claim seller-wide cached-response,
  conflict, or concurrency semantics. FIXME(#1607) tracks flipping the
[...]

PR #2182, opened 2026-09-02
fix: tolerate AdCP adcp_version/adcp_major_version negotiation fields on MCP (#1512)

### LINKED ISSUE #1512: fix: MCP tools reject AdCP adcp_version/adcp_major_version negotiation fields — agent uncallable by official AdCP SDK clients
## Summary
The MCP tool wrappers reject the `adcp_version` and `adcp_major_version` version-negotiation fields that every official AdCP SDK client injects into requests. Tool calls fail argument validation *before the handler runs*, so the agent is effectively uncallable by conformant AdCP clients — including the conformance runner.

## Impact
- Reproduced with **both** reference clients: Python `adcp` CLI (5.7.0, matched to the pin) and the TypeScript `@adcp/sdk` conformance runner (9.6.2).
- `get_adcp_capabilities` and `get_products` both fail. `list_tools` (discovery) works, but typed tool calls don't.
- The conformance runner can't get past capability discovery (the first storyboard step), so no behavioral conformance can be graded.

## Reproduction
Python:
```
uvx --from 'adcp==5.7.0' adcp http://<agent>/mcp/ --auth <token> get_adcp_capabilities
# Error: 1 validation error for call[get_adcp_capabilities]
#   adcp_version  Unexpected keyword argument  input_value='3.1-beta.3'
```
TypeScript (`@adcp/sdk` `testing.testCapabilityDiscovery`):
```
STEP get_adcp_capabilities -> ERROR
  mcp_error: 2 validation errors for call[get_adcp_capabilities]
    adcp_major_version  Unexpected keyword argument  input_value=3
    adcp_version        Unexpected keyword argument  input_value='3.1'
```
Server log:
```
Error validating tool 'get_adcp_capabilities'
ValidationError: call[get_adcp_capabilities] adcp_version Unexpected keyword argument
```

## Root cause
MCP tools are registered as FastMCP tools with explicit parameter signatures (e.g. `get_adcp_capabilities(protocols, ctx)`), so FastMCP auto-generates strict argument validation that rejects any extra field. The AdCP SDK client's envelope enricher injects `adcp_version`/`adcp_major_version` on every request for version negotiation; the SDK's own request models tolerate this via `extra="allow"`, but the FastMCP wrappers don't.

## Suggested fix
Tolerate/ignore the `adcp_version` and `adcp_major_version` envelope fields on incoming MCP tool calls centrally — e.g. in `src/core/mcp_compat_middleware.py` (the existing "universal request normalization" layer) — rather than adding params to every wrapper. Restores interop with the official SDK clients and unblocks the conformance runner.

## Environment
- Revision: `upstream/main` @ `a6d96cb6`; AdCP spec pin **3.1.0-beta.3** (`adcp==5.7.0`)
- Reproduced from Python `adcp` 5.7.0 and TS `@adcp/sdk` 9.6.2

numarasSigmaSoftware:
## Summary
Fixes #1512. Official AdCP SDK clients (Python `adcp` CLI, TS `@adcp/sdk` conformance
runner) inject the version-envelope fields `adcp_version` and `adcp_major_version` on
**every** request. The MCP tool wrappers rejected them at FastMCP's TypeAdapter *before the
handler runs* (`VALIDATION_ERROR`), so the agent was uncallable by conformant clients and the
conformance runner couldn't get past capability discovery — but only in dev/CI: the existing
compat middleware stripped unknown fields **production-only** (`if is_production()`), so the
dev/CI environment (where tests and the conformance runner run) still rejected them.

This strips **only** those two named, spec-defined fields, **unconditionally** (all
environments), in the shared MCP compat middleware — leaving the general production-only
unknown-field strip (and its "fail loudly in dev on genuinely unknown fields" behavior)
untouched. On A2A, tolerance is per-skill, not blanket: skills that pluck named params from the dict
(`get_products`, `create_media_buy`, `get_media_buy_delivery`) accept both fields, but the two
that `model_validate` the whole params dict — `get_media_buys` (`adcp_a2a_server.py:2112`) and
`update_performance_index` (`:2175`) — reject both in dev/CI (`extra="forbid"`, fields
undeclared). REST tolerates `adcp_version` (a declared field on the request bodies) but rejects
`adcp_major_version` in dev/CI for the same reason. These A2A gaps and the REST gap are the same
env-asymmetry this PR fixes for MCP; they are out of scope here and tracked under #2207 (which subsumes #2184). This PR
brings MCP to parity for request acceptance; full A2A/REST parity is #2207.

One current limitation: the strip removes the fields by key without validating their shape, so a *malformed* pin (wrong type, a patch-precision string like `3.1.1`, an out-of-range major) is silently accepted rather than rejected with `VALIDATION_ERROR` as the pinned schema requires. Rejecting malformed pins on the MCP path is tracked in #2194.

## Spec grounding (pinned AdCP 3.1.1, `adcp==6.6.0`)
| Claim | Authoritative artifact |
|---|---|
| `adcp_version`/`adcp_major_version` are request-envelope fields on every request | `dist/schemas/3.1.1/core/version-envelope.json`, `allOf`-composed into every request schema |
| Server MUST accept them; MUST honor `adcp_major_version` through 3.x | `dist/schemas/3.1.1/core/version-envelope.json`; `dist/docs/3.1.1/reference/versioning.mdx` |
| Behavior is graded | `dist/compliance/3.1.1/universal/version-negotiation.yaml`; `docs/test-obligations/storyboard-issue-map.yaml` maps it to #1512 (`coverage: partial`) |

## Scope
Tier 1 — **request tolerance only**. The tolerate-but-ignore remainder (read + echo
`adcp_version`, advertise `supported_versions`, `VERSION_UNSUPPORTED` on cross-major, and
wiring the dormant `BR-UC-010` BDD scenarios) is deferred to **#2181**. The deferred
pieces are graded differently at 3.1.1: the presence/format checks (advertise
`supported_versions`, echo `adcp_version`, release-precision pattern) carry `severity: advisory`
with a `permanent_advisory` 3.2 promotion in `universal/version-negotiation.yaml`, so deferring
them costs nothing at 3.1.1. The `VERSION_UNSUPPORTED`-on-cross-major path is **not** advisory:
`universal/error-compliance.yaml`'s `unsupported_major_version` step (whose `VERSION_UNSUPPORTED` is `recovery: correctable` per the pinned `_schemas/3.1/manifest.json` — the storyboard's free-text `expected: fatal` is not the wire contract) carries
validations with no `severity:` key, which `storyboard-schema.yaml:1176` defaults to `required`
(fails the step, contributes to `steps_failed`) — so that path is blocking at 3.1.1 today. It
stays deferred to #2181, but as a real (blocking) conformance gap, not an advisory one. The hard-graded `adcp.major_versions` advertisement is already
emitted.

## Tests
- Unit `test_mcp_compat_middleware.py::TestMiddlewareStripsEnvelopeVersionFields` — the fields
  are stripped in dev AND production mode; a genuinely-unknown field is preserved (loud-in-dev intact).
- Unit `test_pydantic_schema_alignment.py::test_envelope_version_fields_match_pinned_schema` —
  oracle asserting `ENVELOPE_VERSION_FIELDS` equals the `properties` of the pinned
  `core/version-envelope.json` (schema directory derived from the installed `adcp` pin, not a
  `3.1` literal), so an upstream-added envelope field can't be silently stripped-and-rejected
  with no failing test.
- Integration `test_mcp_unknown_field_handling.py::test_version_envelope_fields_tolerated` —
  in-process via `env.call_via(Transport.MCP)` (dev mode, the failing environment), beside the
  negative control `test_unknown_field_rejected`: a `get_products` call carrying
  `adcp_version`/`adcp_major_version` succeeds, where a genuinely-unknown field is rejected on
  the same transport. Reverting the strip reddens it (the fields reach FastMCP's TypeAdapter).

`make quality` green; complexity/duplication ratchets unchanged (the Step-0 debug log was
extracted into `strip_envelope_version_fields()` to keep `on_call_tool` under the PLR0915
ceiling).


src/core/mcp_compat_middleware.py
```diff
@@ -49,8 +58,11 @@ async def on_call_tool(
             return await call_next(context)
 
         tool_name = context.message.name
-        normalized = dict(arguments)
-        modified = False
+        # Step 0: Strip AdCP version-envelope fields (all environments). Official SDK
+        # clients inject these on every request; FastMCP's TypeAdapter rejects them.
+        # Folded into initialization so on_call_tool adds no net statements (PLR0915).
+        normalized = {k: v for k, v in arguments.items() if k not in ENVELOPE_VERSION_FIELDS}
+        modified = bool(ENVELOPE_VERSION_FIELDS & arguments.keys())
```
ChrisHuie: **[SHOULD-FIX] B1 — this stage is silent, and in production it deletes a log line that exists today.**

I ran the real middleware at both revisions in both environments, capturing this module's logger:

| | production | dev |
|---|---|---|
| merge base | forwards `['brief']` **and logs** `WARNING Stripped unknown fields from get_products: adcp_major_version, adcp_version` | forwards all three (the rejection this PR fixes) |
| this head | forwards `['brief']`, **no log** | forwards `['brief']`, no log |

The other three stages all log (`request_compat.py:167`, and `:86-90` / `:115-120` in this file), and the rationale at `:73-77` is that unknown-field handling must stay visible. One `logger.debug` guarded on a non-empty intersection closes it — `debug` rather than `warning`, since every conformant client carries these fields on every request.

src/core/mcp_compat_middleware.py
```diff
@@ -49,8 +58,11 @@ async def on_call_tool(
             return await call_next(context)
 
         tool_name = context.message.name
-        normalized = dict(arguments)
-        modified = False
+        # Step 0: Strip AdCP version-envelope fields (all environments). Official SDK
+        # clients inject these on every request; FastMCP's TypeAdapter rejects them.
+        # Folded into initialization so on_call_tool adds no net statements (PLR0915).
+        normalized = {k: v for k, v in arguments.items() if k not in ENVELOPE_VERSION_FIELDS}
```
ChrisHuie: **[SHOULD-FIX] C1 — this drops the values instead of setting them aside.**

The pinned schema's own text is the authority: `adcp_major_version.description` reads "Servers MUST continue to honor this field through 3.x … Sellers validate against their supported major_versions and return `VERSION_UNSUPPORTED` if unsupported. When omitted, the seller assumes its highest supported version." And `versioning.mdx:90` requires `VERSION_UNSUPPORTED` when the two fields disagree at the major level.

After this line, a declared value is indistinguishable from an omitted one. Measured on the real MCP path at both revisions:

| payload | merge base | this head |
|---|---|---|
| `adcp_major_version: 99` | ERROR on `adcp_major_version` | **SUCCESS** |
| `adcp_version: "99.0"` | ERROR on `adcp_version` | **SUCCESS** |

So a buyer pinned to a version you can't speak now gets 3.1-shaped data believing it's 99-shaped; previously it got an error, with the wrong code.

To be precise, the value isn't unrecoverable — `src/core/mcp_auth_middleware.py:58-61` stashes `raw_wire_payload` before this middleware runs (`src/core/main.py:174-175` registers auth first). But that capture exists to be the idempotency payload-hash input, so reading version negotiation out of it couples two unrelated capabilities.

Two remedies, either a recorded decision rather than a silent one. A `set_state` beside the strip is about two lines. Or declare the two parameters on the wrapper signatures instead of stripping — I checked that on a FastMCP tool and it keeps `additionalProperties: False` while both accepting the fields **and** delivering them to the handler (`{'saw_version': '3.1', 'saw_major': 3}`), where the undeclared sibling raises the `Unexpected keyword argument` error from #1512. That's also the remedy #1193 and #2184 chose for this defect class, and the pinned SDK models it that way. Stripping can't generalize to `ext`, so this is the divergent option and wants a stated rationale.

src/core/request_compat.py
```diff
@@ -16,6 +16,12 @@
 logger = logging.getLogger(__name__)
 V25_SIGNALS: frozenset[str] = frozenset({"brand_manifest", "promoted_offerings", "campaign_ref"})
 
+# AdCP version-envelope fields (spec 3.1.1 core/version-envelope.json, allOf-composed
+# into every request schema). Official SDK clients inject these on every request; the
+# server MUST accept them. Stripped unconditionally on MCP (all environments) so they
+# never reach FastMCP's TypeAdapter (additionalProperties: false).
+ENVELOPE_VERSION_FIELDS: frozenset[str] = frozenset({"adcp_version", "adcp_major_version"})
```
ChrisHuie, line 23: **[SHOULD-FIX] D1 — this hand-transcribes a machine-readable artifact, and is the second copy of one concept.**

The members here are exactly the `properties` of the pinned SDK's `adcp/_schemas/3.1/core/version-envelope.json`, whose own description says the version semantics "live in exactly one place" — and nothing in the suite compares the two. The only `version-envelope` references under `tests/` are a URL literal at `tests/e2e/test_schema_validation_standalone.py:171` and a docstring line in the new test file.

Separately, `tests/unit/test_pydantic_schema_alignment.py:95` declares `_VERSION_FIELDS` with byte-identical members and 8 use sites; that file already imports `src.core.*`, so unification is a one-line import.

Three mutations over the four affected suites (159 passed at baseline, total stable each time): dropping a member gives `2 failed`, adding a stray member gives `11 failed` — but **a spec-side addition has no oracle at all**, which is this PR's own regression class. I wrote that check and ran both arms: asserting this constant equals the pinned schema's `properties`, with the version directory derived from the pin rather than a `3.1` literal, passes today and fails against a schema grown by one field.

Extracting a `strip_envelope_version_fields()` helper next to the constant also makes the call site one statement, which removes the need for the ratchet comment at `mcp_compat_middleware.py:63`.

tests/integration/test_mcp_version_envelope_tolerance.py
```diff
@@ -0,0 +1,53 @@
+"""MCP tolerance for AdCP version-envelope request fields.
+
+AdCP spec 3.1.1 composes core/version-envelope.json (adcp_version,
+adcp_major_version) into every request schema via allOf, and every official AdCP
+SDK client injects these fields on every request. FastMCP's TypeAdapter
+(additionalProperties: false) would reject them before the handler runs unless the
+MCP compat middleware strips them first. A2A and REST already tolerate these
+fields; this test pins the same tolerance for MCP over the real HTTP transport in
```
ChrisHuie: **[SHOULD-FIX] A2 — this docstring asserts a tolerance the code doesn't have, and contradicts the PR description.**

I drove the real `/api/v1/products` route through `RestCompatMiddleware` and FastAPI: `adcp_version` alone passes validation, `adcp_major_version` returns **HTTP 422 `adcp_major_version:extra_forbidden`**, and both together return 422 on `adcp_major_version`. `src/routes/rest_compat_middleware.py` only normalizes deprecated names and never strips, so nothing intercepts it earlier.

Your PR description states this REST gap correctly, so the description and this docstring disagree — and the docstring is the one that ships. The A2A half is true for `get_products` and false for the two skills named in the lead comment.

tests/integration/test_mcp_version_envelope_tolerance.py
```diff
@@ -0,0 +1,53 @@
...
+    @pytest.fixture
+    async def mcp_client(self, mcp_server, sample_tenant, sample_principal, sample_products):
+        """Create MCP client for testing with test data."""
+        headers = {"x-adcp-auth": sample_principal["access_token"]}
+        transport = StreamableHttpTransport(url=f"http://localhost:{mcp_server.port}/mcp/", headers=headers)
+        client = Client(transport=transport)
+
+        async with client:
+            yield client
+
```
ChrisHuie: **[SHOULD-FIX] G1 — this fixture is a clone, and the file sits in a costlier home than two that already exist.**

Diffed against `tests/integration/test_mcp_tool_roundtrip_minimal.py:30-39`: exactly one comment line and one blank line differ. `mcp_server` is `@pytest.fixture(scope="function")` (`tests/integration/conftest.py:576`) and spawns a server with `subprocess.Popen` (`:648`), so two test methods mean two server subprocesses.

Two better-suited homes already exist and neither was touched: `tests/integration/test_mcp_unknown_field_handling.py` (factories, `env.call_via(Transport.MCP)`, `assert_envelope_shape`) and `tests/harness/test_forward_compat_acceptance.py` (cross-transport, in-process, no DB). I checked that relocating wouldn't cost coverage — `McpDispatcher` dispatches via `Client(mcp)`, the full FastMCP pipeline, so the middleware still runs and you additionally get a `wire_error_envelope` assertion these tests don't make.

Either move the two scenarios into the first, or hoist `mcp_client` into `tests/integration/conftest.py` beside `mcp_server`.

tests/unit/test_mcp_compat_middleware.py
```diff
@@ -109,6 +109,68 @@ async def capturing_call_next(context):
...
+        with patch("src.core.config.is_production", return_value=False):
+            await middleware.on_call_tool(ctx, capturing_call_next)
+
+        assert captured_ctx is not None
+        assert captured_ctx is not ctx
+        forwarded = captured_ctx.message.arguments
+        assert "adcp_version" not in forwarded
+        assert "adcp_major_version" not in forwarded
+        # Genuinely-unknown field is NOT removed by this step — dev must still fail loudly.
+        assert forwarded["bogus_field"] == "keep-me"
```
ChrisHuie, line 143: **[SHOULD-FIX] F5 (site) — this line pins the behaviour the spec forbids, for every field but the two this PR fixes.**

Both 3.1.1 request schemas set `additionalProperties: true`, and `versioning.mdx:264` says "Readers MUST ignore unknown fields rather than erroring." Since `ENVIRONMENT` is set in neither `ci.yml` nor `tox.ini`, `is_production()` is false in CI and the production-only Step 2 never runs there — so this assertion pins dev/CI rejection of any additional property, which is #1512's own bug minus the two named fields.

Measured on the wire: a call carrying only `future_unknown_field` errors identically at both revisions, with `code=VALIDATION_ERROR`, `recovery=correctable`, `field=future_unknown_field`.

Not a fold-in — changing dev loud-failure is a real ergonomics trade-off, and no open issue covers it. Rationale for the deferral is in the lead comment.

tests/unit/test_mcp_compat_middleware.py
```diff
@@ -109,6 +109,68 @@ async def capturing_call_next(context):
...
+        assert forwarded["brief"] == "ads"
+
+    @pytest.mark.asyncio
+    async def test_envelope_version_fields_removed_in_production(self, middleware):
+        """The Step-0 strip is unconditional: envelope fields are removed in production too.
+
+        Pins the "all environments" contract against a future is_production() re-gate
+        (mirroring the is_production-gated Step 2 that sits right below it).
```
ChrisHuie: **[SHOULD-FIX] E1 — this test doesn't pin what its docstring claims, for two separate reasons.**

The docstring says it "pins the 'all environments' contract against a future `is_production()` re-gate".

Mutation one — re-gate Step 0 on `is_production()`: `15 passed` becomes `1 failed / 14 passed`, and the failure is `test_envelope_version_fields_removed_in_dev`. This test stays green under exactly the change it names. Structurally it must: in production a gated Step 0 and an unconditional Step 0 are indistinguishable.

Mutation two — revert Step 0 entirely, then run these same assertions with `_get_known_params` returning a resolvable set so Step 2 is live: **1 passed**. Its green today comes from `_make_context` setting `fastmcp_context = None` (line 31), which disables Step 2 — not from Step 0's presence.

I prototyped the replacement and ran it both directions: spy on `normalize_request_params` and assert it was called once with the envelope-free dict. Step 2 can't satisfy that, because it runs after Step 1. Result: 2 passed at this head, 2 failed with Step 0 reverted — including the production arm, which this test cannot grade.

Also worth noting: the dev test currently reddens on its `captured_ctx is not ctx` identity assertion at `:139` rather than on the field-absence assertions, so asserting absence first would make it fail on the contract.

ChrisHuie (commented):
Reviewed at fb9aaa934. 0 BLOCKER / 10 SHOULD-FIX. Seven findings are anchored to their lines as separate threads; the four below aren't tied to a line in this diff, so they're here.

**[SHOULD-FIX] A1 — the body's "A2A already tolerates both fields (it plucks named params from a dict)" is false for two routed skills.**
Ran the real composition (`normalize_request_params` → `model_validate`) across all four A2A skills that validate raw params. The normalizer keeps both fields — it translates deprecated names, it doesn't strip. With `ENVIRONMENT` unset, `GetMediaBuysRequest` (`src/a2a_server/adcp_a2a_server.py:2112`) and `UpdatePerformanceIndexRequest` (`:2175`) reject both fields; `CreateMediaBuyRequest` and `GetMediaBuyDeliveryRequest` accept them. Under `ENVIRONMENT=production` both become `extra=ignore` and accept, so it's a dev/CI-only gap, symmetric with the REST gap you track as #2184. Both skills are routed at `:1578` and `:1580`.

The claim is true for the tools this PR exercises and wrong unqualified. Fix the body, and either fold the strip into the shared seam at `:1523` or add these two skills to #2184.

**[SHOULD-FIX] A3 — "advisory at 3.1 (blocking at 3.2)" is right for what you defer and wrong for one part of it.**
In `universal/version-negotiation.yaml`, the three checks your Scope paragraph defers — `adcp.supported_versions` presence, envelope `adcp_version` presence, and its release-precision pattern — each carry `severity: advisory` with a `permanent_advisory` reason naming the 3.2 promotion. That half is exactly right, and `adcp.major_versions` there carries no severity (required through 3.x), which you already emit.

But in `universal/error-compliance.yaml`, step `unsupported_major_version` (`sample_request: adcp_major_version: 99`) has three validations — `error_code = VERSION_UNSUPPORTED`, `field_present context`, `field_value correlation_id` — and **none carries a `severity:` key**. `storyboard-schema.yaml:1176` defines validation-level severity as defaulting to `"required"`, where "a \"required\" validation failure fails the step and contributes to `steps_failed`". So the major-version `VERSION_UNSUPPORTED` path is blocking at 3.1.1 today, not advisory. Worth splitting that sentence, since it's what the deferral rests on.

**[SHOULD-FIX] F1 — the storyboard ledger records a cause that no longer applies; it needs re-attribution, not a re-seed.**
`tests/storyboard/known_failures.txt:54-57` attributes 11 `mcp::` entries to `VALIDATION_ERROR: Unexpected keyword argument` / #1512, and `:59` to the capability probe itself being rejected. That cause is gone — on the real MCP path the same call errors at the merge base and succeeds here.

I graded the conformance suite at both revisions against the same sha256-verified 3.1.1 bundle rather than inferring the consequence: `xpassed` is 0 on both sides and all three relevant entries stay `xfailed`, so **nothing graduates**. What's now false is the recorded reason — `:79` and `:80` return SUCCESS where their storyboards require `VERSION_UNSUPPORTED`. Re-attribute those and the `:59` note, plus `EXPECTED_LEDGER` at `tests/unit/test_storyboard_ledger_state.py:44`.

Two mechanics worth knowing. Entries load `xfail(strict=False)` (`tests/storyboard/conftest.py:44`), so a graduation would never have reddened anything. And getting any grading at all required `STORYBOARD_COMPLIANCE_DIR`/`STORYBOARD_SCHEMA_ROOT`: unset, `tests/storyboard/test_storyboard_conformance.py:344-354` emits a single `environment-not-configured` skip and returns early, so the job brings up the full stack, grades nothing, and goes green. Pre-existing and repo-level, not yours — but it's why "re-seed from an in-network run" isn't currently actionable.

On attribution, since the raw counts differ: over the 38 checks graded at both revisions, 29 `xfailed→xfailed`, 8 `skipped→skipped`, 1 `failed→failed` — **zero regressions from this PR**. Sixteen fail at head, but 15 were never collected at base, so they have no baseline and I'm not attributing them to you.

**[SHOULD-FIX] F5 — dev/CI still rejects every *other* additional property, and this PR newly pins that.**
`versioning.mdx:264` says "Readers MUST ignore unknown fields rather than erroring", for containers with `additionalProperties: true` — which both request schemas are. `is_production()` reads `ENVIRONMENT`, defaulting to `"development"` (`src/core/config.py:169`), and it's set in neither `ci.yml` nor `tox.ini`, so the production-only Step 2 never runs in CI. Measured: a call carrying only `future_unknown_field` errors identically at both revisions (`VALIDATION_ERROR` / `correctable` / `field=future_unknown_field`). So any additional property the conformance runner sends still errors in the exact environment you say it runs in. Changing dev loud-failure is a real trade-off rather than a fold-in, and no open issue covers it (#2184 is REST-only, #1193 is `ext`-only).

Two repo-level notes, neither yours. `check_code_duplication.py:69` delegates to a ratchet whose only failure mode is "count increased", with no `origin/main` comparison, and the committed `.duplication-baseline` is stale-high by 2 — the hook prints "tests/: 69 duplicate blocks (-2 fixed vs baseline 71)" and exits 0, which is why the fixture clone below tripped nothing. And `tests/harness/test_forward_compat_acceptance.py:217` allowlists `adcp_major_version` out of the dev-rejects case; that suite is green at this head and with Step 0 reverted alike, so nothing pins the acceptance this PR adds.

src/core/mcp_compat_middleware.py
numarasSigmaSoftware: Added a guarded `logger.debug` at the strip. To keep `on_call_tool` under the PLR0915 ceiling, the strip+log moved into `strip_envelope_version_fields()` next to the constant, so it now logs from `src.core.request_compat`. dd07198f, 2b9eb2e2.

src/core/mcp_compat_middleware.py
numarasSigmaSoftware: Added a comment at the strip: declared values are intentionally dropped for request tolerance; cross-major `VERSION_UNSUPPORTED` is deferred to #2181, with the `raw_wire_payload` stash noted as the recovery input if #2181 needs the declared version. Kept the strip (declare-on-signature would be a cross-wrapper change, out of scope here). da448a80.

src/core/request_compat.py
numarasSigmaSoftware: `tests/unit/test_pydantic_schema_alignment.py` now imports this constant as its `_VERSION_FIELDS` (one SSOT), and a new oracle asserts `ENVELOPE_VERSION_FIELDS` equals the `properties` of the pinned `core/version-envelope.json` (schema dir derived from the pin, no `3.1` literal) — so an upstream-added envelope field fails a test instead of being silently stripped-and-rejected. cc963a3e.

tests/integration/test_mcp_version_envelope_tolerance.py
numarasSigmaSoftware: This file is deleted — the scenarios were relocated in-process (see the G1 thread), so the misleading docstring is gone with it. 543c7c92.

tests/integration/test_mcp_version_envelope_tolerance.py
numarasSigmaSoftware: Relocated into `tests/integration/test_mcp_unknown_field_handling.py` via `env.call_via(Transport.MCP)` (in-process, full FastMCP pipeline, beside the `test_unknown_field_rejected` negative control), and deleted this file — the cloned fixture and subprocess-per-test are gone. 543c7c92, b620c069.

tests/unit/test_mcp_compat_middleware.py
numarasSigmaSoftware: Left as-is deliberately: dev-strict rejection of *undefined* (unnamed) fields is intended per CLAUDE.md Pattern #7. `additionalProperties: true` binds production acceptance (`extra="ignore"`, which holds), not dev-mode tolerance. #1969 raised exactly this and was closed as not-a-gap. This assertion pins that intended behavior — happy to reopen #1969 if you disagree with its resolution.

tests/unit/test_mcp_compat_middleware.py
numarasSigmaSoftware: The production arm now spies `normalize_request_params` and asserts it receives the envelope-free dict; reverting Step 0 reddens it (verified). Dev asserts reordered so a regression fails on the contract, not on the identity check. The docstring's own rationale was inverted on a re-read (asserting forwarded args would *not* be vacuous here) and is corrected — the real reason for the Step-1 spy is isolation from a live Step 2. 5479a855, 302b108f.

src/core/mcp_compat_middleware.py
```diff
@@ -49,8 +58,13 @@ async def on_call_tool(
...
         tool_name = context.message.name
-        normalized = dict(arguments)
-        modified = False
+        # Step 0: Strip AdCP version-envelope fields (all environments) via the helper.
+        # The declared values (incl. adcp_major_version) are intentionally dropped for
+        # request tolerance — this does NOT negotiate cross-major support. Emitting
+        # VERSION_UNSUPPORTED for a mismatched major is deferred to #2181; if that work
+        # needs the declared version it can read the raw payload from mcp_auth_middleware's
+        # "raw_wire_payload" state stash rather than reconstructing it here.
+        normalized, modified = strip_envelope_version_fields(tool_name, arguments)
```
KonstantinMirin, line 67: This is the only call site of `strip_envelope_version_fields`, but `normalize_request_params` on the next line has three: here, `src/a2a_server/adcp_a2a_server.py:1523`, and `src/routes/rest_compat_middleware.py:62`. The two envelope fields are `allOf`-composed by the pinned schema into every request, so acceptance cannot fork per transport. With `ENVIRONMENT` unset, `GetMediaBuysRequest` and `UpdatePerformanceIndexRequest` still reject both fields and `GetProductsBody` still rejects `adcp_major_version`.

Fold the removal into `normalize_request_params` (`src/core/request_compat.py:119`) and delete this stage, the import at `:23`, the comment above, and the four-stage docstring at `:34-38`. All three transports become tolerant with no new call site, and #2184 is no longer needed. Do it together with the `NormalizationResult` change on the helper's return, because a bare delete inside the seam silently serves v2-compat products to A2A and REST buyers that declared 3.1 (`adcp_a2a_server.py:1661`, `api_v1.py:255`).

KonstantinMirin (commented):
## Review — PR #2182

**Overview** — #1512 required that a request carrying the spec-named version-envelope fields is not rejected, so official SDK clients can call the agent at all, and its suggested fix names the central request-normalization layer by name. This PR accepts both fields on MCP, which is the symptom the issue names first, and `ENVELOPE_VERSION_FIELDS` with its pinned-schema oracle is the correct pin. The approach is where the problem is: tolerance was wired at one transport's edge instead of into `normalize_request_params`, the seam A2A and REST already call, and it was built by deleting the declared version rather than recording it. `adcp_major_version: 99` now returns SUCCESS with 3.1-shaped data on MCP where it previously returned an error, and nothing in the tree can fail on that.

### Your open question, answered

You asked whether to reopen #1969. No, don't reopen it, and F5 is out of this PR's frame. `CLAUDE.md:306-308` states dev/CI `extra="forbid"` and production `extra="ignore"` as the intended project contract, and #1969 is closed having raised exactly that question. Your assertion at `tests/unit/test_mcp_compat_middleware.py:147` asserts intended behavior, and your reading is right. The two envelope fields are a different case because the pinned schema names them in `dist/schemas/3.1.1/core/version-envelope.json` and `allOf`-composes them into every request, which is why #1512 is a real defect and #1969 was not. Nothing below re-litigates dev-mode strictness for genuinely undefined fields.

### Should fix

#### 1. Envelope handling belongs inside `normalize_request_params`, and it has to record the declared version instead of deleting it ↩ raised 2026-09-03, still open

These are two halves of one design decision, so they are one fix.

`src/core/request_compat.py` says "Shared by all transports (MCP, A2A, REST)" at `:5`, and `normalize_request_params` is the seam each transport calls exactly once: `mcp_compat_middleware.py:70`, `adcp_a2a_server.py:1523`, `rest_compat_middleware.py:62`. You put the new helper in that module but invoked it at `mcp_compat_middleware.py:67` only. With `ENVIRONMENT` unset:

```
GetMediaBuysRequest {'adcp_version': '3.1'}                REJECTED: adcp_version
GetProductsBody     {'brief': 'x', 'adcp_major_version': 3} REJECTED: adcp_major_version
```

The same four lines executed inside the seam make all three transports accept, with no new call site.

The reason the move is not safe as written is the second half. `strip_envelope_version_fields(...) -> tuple[dict[str, Any], bool]` returns "did anything go?" and cannot carry the value that went, so deletion is the only tolerance the stage can express. `NormalizationResult` (`request_compat.py:48-54`) is this layer's declared result and already carries `inferred_version`; `normalize_request_params` already removes `account_id`, `campaign_ref`, `brand_manifest` and `promoted_offerings` from `params` while recording what it learned on that result. The new stage copied the removal and dropped the recording. Two things follow:

- `adcp_major_version: 99` is now indistinguishable from an omitted field. `dist/schemas/3.1.1/core/version-envelope.json` says servers MUST honor `adcp_major_version` through 3.x and return `VERSION_UNSUPPORTED` when they cannot, and `dist/compliance/3.1.1/universal/error-compliance.yaml` checks `unsupported_major_version` with validations that carry no `severity:` key, which `storyboard-schema.yaml:1176` defaults to `required`. #2181 owns emitting the error. It does not own destroying the input the error is computed from.
- The declared `adcp_version` is live input to response shaping on the two transports you did not wire. `adcp_a2a_server.py:1661` reads it out of the post-seam dict and feeds `apply_version_compat` at `:1668`; `api_v1.py:255` does the same through `body.adcp_version`. Folding a bare delete into the seam silently serves v2-compat products to a buyer that declared 3.1:

```
3.1  seam v2compat= False  |  after strip v2compat= True
```

The comment at `mcp_compat_middleware.py:61-66` points the follow-up work at `mcp_auth_middleware`'s `raw_wire_payload` stash. That stash is the idempotency payload-hash input (`mcp_auth_middleware.py:56-61`, consumed through `canonical_payload_hash` at `media_buy_create.py:2308`) and it is built three different ways: FastMCP context state on MCP, raw bytes on `request.state` on REST (`rest_compat_middleware.py:59`), a deepcopy kwarg on A2A (`adcp_a2a_server.py:1505`). A negotiation built on it needs three different implementations. Delete that sentence and make it unnecessary.

**Fix:** add `declared_adcp_version: str | None = None` and `declared_adcp_major_version: int | None = None` to `NormalizationResult` (`src/core/request_compat.py:48-54`), remove the two fields from `params` and set those results inside `normalize_request_params` (`:119`), then delete `ENVELOPE_VERSION_FIELDS`'s "on MCP" wording, `strip_envelope_version_fields`, the import at `mcp_compat_middleware.py:23`, the Step-0 call and comment at `:61-67`, and the four-stage docstring at `:34-38`. Five sites, all in this diff:

1. `src/core/request_compat.py:119` — fold the removal and the recording in; restate the constant comment and docstring in protocol terms, since "on MCP" and "FastMCP's TypeAdapter" are false for the other two callers.
2. `src/core/mcp_compat_middleware.py:23, 61-67` — delete. The seam does this once the removal moves into it.
3. `src/a2a_server/adcp_a2a_server.py:1661` — read `compat_result.declared_adcp_version` threaded from `:1523` instead of `parameters.get("adcp_version")`. Without this the change is a silent v2-compat regression on A2A.
4. `src/routes/rest_compat_middleware.py:62-64` — the body is rewritten only `if result.translations_applied`, so the seam fix is a no-op on REST until the result reports that the params changed. Carry the declared value to the route the way this file already carries `raw_wire_payload` on `request.state`.
5. `src/routes/api_v1.py:255` — read the declared value from that carrier rather than `body.adcp_version`, so removing the field cannot fall back to the `"1.0.0"` default.

The `bool` disappears (callers key "modified" off the same result), the Step-0 log returns to the caller beside the other three stages, and the duplicate of `strip_unknown_params`'s filter-and-report shape at `request_compat.py:37-41` goes with it.

**Done when:** `grep -rn "strip_envelope_version_fields" src/` returns nothing; with `ENVIRONMENT` unset, `GetMediaBuysRequest`, `UpdatePerformanceIndexRequest` and `GetProductsBody` all accept both fields through their transports; `normalize_request_params("get_products", {"adcp_version": "3.1", "adcp_major_version": 3, "brief": "x"})` returns `params == {"brief": "x"}` with `declared_adcp_version == "3.1"` and `declared_adcp_major_version == 3`; and an A2A `get_products` call declaring `adcp_version: "2.5"` still returns v2-shaped products.

#### 2. No test can tell "the version was accepted" from "the version was thrown away", and the one wire test runs on one transport

Swap `"adcp_version": "3.1", "adcp_major_version": 3` for `"99.0"` / `99` in all three new unit tests, change no production code, and the suite is **16 passed**. That holds because `ENVELOPE_VERSION_FIELDS & params.keys()` never reads a value, so no assertion over the current design can distinguish the two outcomes. This is the same root as item 1: recording the declared version is what turns "mis-negotiated" into something a test can name.

The rest of the test coverage follows the shipped design rather than the contract:

- `tests/integration/test_mcp_unknown_field_handling.py:80` asserts `result.is_success`, which is `error is None and payload is not None` (`tests/harness/transport.py:371`). The harness ships `TransportResult.require_wire()` (`:428-449`) for the success side, and its docstring states the case it exists to catch: a success with no stashed wire body means the dispatch bypassed the real pipeline. The negative control three lines above already asserts the full envelope through `assert_envelope_shape` and `wire_error_envelope["errors"][0]["field"]`.
- A field the pinned schema puts on every request envelope is graded on MCP alone. The four-transport harness exists: `tests/bdd/features/BR-COMPAT-001-deprecated-field-normalization.feature` is wired by `tests/bdd/test_compat001_deprecated_field_normalization.py`, its steps are registered at `tests/bdd/conftest.py:83`, and it multiplies over a2a/mcp/rest and e2e_rest. `ProductEnv` already declares `MCP_TOOL`, `A2A_SKILL` and `REST_ENDPOINT`.
- `tests/unit/test_mcp_compat_middleware.py:195` asserts the `logger.debug` format string and counts every `debug` call on the whole `src.core.request_compat` logger, which also emits at `:86` and `:291`. A payload that reaches the brand_manifest branch at `:86` produces a second record and fails this test for an unrelated reason.

**Fix:** after item 1 lands, add one transport-blind scenario to BR-COMPAT-001 ("a Buyer Agent that declares its AdCP version is served normally"), dispatched through `dispatch_request` and asserted with `wire_dict` / `wire_field` (`tests/bdd/steps/_outcome_helpers.py:80, 92`). Do not tag it `@mcp`; that suppresses the parametrization at `tests/bdd/conftest.py:3218` and hides the gap #1512 named. Do not copy the in-process pattern at `tests/bdd/steps/domain/compat_normalization.py:31-75`. Keep the MCP integration case as the FastMCP TypeAdapter mechanic and assert it with `result.require_wire()["products"]`. Re-express the log assertion on `caplog.at_level(logging.DEBUG, logger="src.core.request_compat")`. Hoist the request payload, which is written out four times (`test_mcp_compat_middleware.py:126-130, :165, :183` and `test_mcp_unknown_field_handling.py:76-78`), into one shared constant.

Once the declared version is on `NormalizationResult`, wiring `BR-UC-010-discover-seller-capabilities.feature:967` and `:986` becomes possible; the emission itself stays in #2181. Today those two scenarios have no `tests/bdd/test_uc010_*.py`, so they are never collected and cannot object to the behavior change.

**Done when:** `grep -n "is_success" tests/integration/test_mcp_unknown_field_handling.py` no longer matches line 80; the new scenario appears in `make check-dormant` as wired and runs on a2a, mcp, rest and e2e_rest; and re-running the `99` mutation above fails at least one test.

#### 3. Three registries record a cause this PR removes ↩ raised 2026-09-03, still open

`tests/storyboard/known_failures.txt:56-57` states that all 11 `mcp::` entries carry `VALIDATION_ERROR: Unexpected keyword argument` and are #1512, and `:59` that the capability probe is itself rejected. This PR removes that cause on MCP. Two of those entries, `:79` `unsupported_major_version` and `:80` `unsupported_release_version`, now fail for the opposite reason: SUCCESS where the storyboard requires `VERSION_UNSUPPORTED`. Entries load `xfail(strict=False)` (`tests/storyboard/conftest.py:41-44`), so the change of cause does not change the color, and the written reason is the only record there is.

Three more statements in the same family:

- `tests/unit/test_storyboard_ledger_state.py:44-49` mirrors the claim, and that file's own rule at `:24-28` is to update the ledger file and `EXPECTED_LEDGER` in the same change.
- `known_failures.txt:141-144` reads "NOTE FOR WHOEVER LANDS #1546 / #1512 ... Remove them in that same change" for the 40 `mcp::security_transport::signed_requests::*` entries at `:146-185`, whose collection is predicated on the probe being rejected. #1546 is closed. Either those 40 entries go or the premise sentence at `:133-136` is false.
- `docs/test-obligations/storyboard-issue-map.yaml:37-43` routes `version-negotiation.yaml` to `[1512]` and says the remainder "no issue tracks". #2181 tracks it. `:102` predicts the gate closes "when #1546 lands".

Separately, the PR description's Scope paragraph cites `unsupported_major_version` as "recovery: fatal". That comes from the storyboard step's free-text `expected:` block. `dist/schemas/3.1.1/enums/error-code.json` gives `VERSION_UNSUPPORTED` `recovery: correctable`, and `assert_wire_error` defaults recovery to the pinned enum, so an implementation built from "fatal" emits a value the guarded helper rejects. Correct it so #2181 inherits the wire-correct classification.

**Fix:** correct the records in this PR. Re-attribute `known_failures.txt:54-59`, state on `:79` and `:80` that the declared major and release versions are now accepted and dropped so the required `VERSION_UNSUPPORTED` never fires (#2181), mirror the same text in `EXPECTED_LEDGER`, decide the `:141-144` instruction one way or the other, re-point the obligation map at #2181, and remove the two dead #1546 references. Add no entries, and remove none without a measured run. Do not register the now-silent `adcp_major_version: 99` success as a new xfail or ledger row.

**Done when:** `git diff 12df85c01194ab6673f456e1edb206f912f5178d...HEAD -- tests/storyboard/known_failures.txt tests/unit/test_storyboard_ledger_state.py docs/test-obligations/storyboard-issue-map.yaml` is non-empty, the entry count in `known_failures.txt` has not increased, and `grep -rn "1546" tests/storyboard/known_failures.txt docs/test-obligations/storyboard-issue-map.yaml` returns nothing.

#### 4. The new single source of truth has a second hand-written copy, and the one consumer you routed renames it

`ENVELOPE_VERSION_FIELDS` is the right constant, and the oracle that pins it to the schema is the right guard. Neither is doing its job yet:

- `tests/harness/test_forward_compat_acceptance.py:217` hand-lists `"adcp_major_version"` in the "names the seller KNOWS" allowlist, and `:48` hand-writes the same name into `FUTURE_TOP_LEVEL_FIELD`. When upstream adds a third envelope field, the oracle fails, the constant grows, the strip starts accepting the new field, and this allowlist still flags it. That file's own comment at `:213-216` records this having already happened once, for `adcp_major_version`.
- `tests/unit/test_pydantic_schema_alignment.py:34` imports the constant as `_VERSION_FIELDS`, so `grep -rn ENVELOPE_VERSION_FIELDS tests/` matches the import line and one assertion message, and no use site. This repo's underscore alias preserves the name and marks privacy only (`get_header_case_insensitive as _get_header_case_insensitive`, `src/core/auth.py:38`); renaming aliases here exist for real collisions, and there is none. Three comments at `:30-33`, `:101` and `:133-134` exist only to explain that the name is not the name.

**Fix:** import `ENVELOPE_VERSION_FIELDS` in `tests/harness/test_forward_compat_acceptance.py` and write the allowlist as `{...} | ENVELOPE_VERSION_FIELDS`, keying the `:48` payload entry off it too. Import it under its own name in the alignment suite and rename the use sites at `:92, :101, :145, :147, :467, :635, :818, :1180, :1220, :1576, :1683`, then delete the three explanatory comments. Keep the rationale prose in exactly one docstring; it is currently written four times in `src/` (`request_compat.py:19-22` and `:27-35`, `mcp_compat_middleware.py:35-38` and `:61-66`), and one copy is already wrong by scope.

**Done when:** `grep -rn "_VERSION_FIELDS" tests/unit/test_pydantic_schema_alignment.py` returns nothing, and `grep -n '"adcp_major_version"' tests/harness/test_forward_compat_acceptance.py` returns nothing.

- The round-2 items landed: the `debug` log at the strip, the schema-derived oracle, the production-arm spy on Step 1, and the relocation of the integration scenarios off the cloned subprocess fixture. The oracle reads the version directory from the installed `adcp` pin rather than a `3.1` literal, which is the pattern to keep. Extracting `strip_envelope_version_fields` to hold `on_call_tool` under the PLR0915 ceiling instead of raising `.ruff-complexity-baseline` is the right call on the ratchet, and `.duplication-baseline` `tests: 71 → 69` moves the right way.
- #2184 is open and unassigned and its title scopes it to REST (`fix(rest): accept adcp_major_version on request bodies (dev/CI parity)`), so it establishes no deferral for the A2A gaps in `get_media_buys` and `update_performance_index`. Item 1's fix removes the need for the ticket. If you keep the per-transport shape instead, file an issue naming those two skills and self-assign it. #2181 is open and assigned to you, so the deferral for echo, `supported_versions` and `VERSION_UNSUPPORTED` emission holds.
- `src/routes/api_v1.py:68-231` restates the AdCP request shape per endpoint and re-declares `adcp_version: str = "1.0.0"` on 11 bodies, then excludes it by string literal in four `model_dump(exclude={"adcp_version"})` calls at `:270, :285, :498, :509`. `"1.0.0"` is not a valid release-precision wire value at 3.1.1 (`dist/docs/3.1.1/reference/versioning.mdx:82`) and the default makes `needs_v2_compat()` true for every REST caller that omits the field. Out of scope: this is the REST request-model fork, a separate boundary the diff does not touch. Only the `get_products` reader at `:255` is in scope, as a consequence of item 1. Worth a filed, self-assigned issue.
- `apply_version_compat` runs at two transport edges (`adcp_a2a_server.py:1668`, `api_v1.py:255`) and is deliberately absent on MCP, pinned by `tests/unit/test_product_mcp_wrapper.py:80`. Out of scope: one response-shaping rule owned by two of three transports is a pre-existing boundary this diff does not touch. It becomes the natural follow-up once item 1 puts the declared version on the shared result.
- `tests/bdd/steps/domain/compat_normalization.py:31-75` implements the BR-COMPAT-001 When steps by calling `normalize_request_params` in process, so the three transport variants run the same in-process code with no wire. Out of scope: pre-existing and untouched. Noted so the scenario in item 2 is not written from it.

src/core/request_compat.py
```diff
@@ -16,6 +16,31 @@
...
+    clients inject these on every request; the server MUST accept them.
+
+    Returns a ``(cleaned dict, modified)`` tuple where ``modified`` is True iff any
+    envelope field was present. Emits a debug log naming the stripped fields.
+    """
+    stripped = ENVELOPE_VERSION_FIELDS & params.keys()
+    if stripped:
+        logger.debug("Stripped AdCP version-envelope fields from %s: %s", tool_name, ", ".join(sorted(stripped)))
+    cleaned = {k: v for k, v in params.items() if k not in ENVELOPE_VERSION_FIELDS}
+    return cleaned, bool(stripped)
```
KonstantinMirin, line 41: `tuple[dict[str, Any], bool]` says whether anything was removed and cannot say what value it carried, so deletion is the only tolerance this stage can express. `NormalizationResult` at `:48-54` is this layer's declared result and already carries `inferred_version`, and `normalize_request_params` already removes `account_id`, `campaign_ref`, `brand_manifest` and `promoted_offerings` from `params` while recording what it learned on that result. This stage copied the removal and dropped the recording.

Add `declared_adcp_version: str | None = None` and `declared_adcp_major_version: int | None = None` to `NormalizationResult`, set them inside `normalize_request_params`, and delete this helper. Every transport then receives the declared value from the result, #2181 reads a typed field instead of `mcp_auth_middleware`'s idempotency stash, and `adcp_major_version: 99` stops being indistinguishable from an omitted field. Note that `rest_compat_middleware.py:64` rewrites the body only `if result.translations_applied`, so the result has to report that the params changed or the fix is a no-op on REST.

src/core/mcp_compat_middleware.py
```diff
@@ -49,8 +58,13 @@ async def on_call_tool(
             return await call_next(context)
 
         tool_name = context.message.name
-        normalized = dict(arguments)
-        modified = False
+        # Step 0: Strip AdCP version-envelope fields (all environments) via the helper.
+        # The declared values (incl. adcp_major_version) are intentionally dropped for
+        # request tolerance — this does NOT negotiate cross-major support. Emitting
+        # VERSION_UNSUPPORTED for a mismatched major is deferred to #2181; if that work
+        # needs the declared version it can read the raw payload from mcp_auth_middleware's
```
KonstantinMirin, line 65: `raw_wire_payload` is the idempotency payload-hash input (`src/core/mcp_auth_middleware.py:56-61`, consumed through `canonical_payload_hash` at `src/core/tools/media_buy_create.py:2308`), and it is built three different ways: FastMCP context state here, raw bytes on `request.state` at `src/routes/rest_compat_middleware.py:59`, a deepcopy kwarg at `src/a2a_server/adcp_a2a_server.py:1505`. Version negotiation built on it needs three different implementations, and it makes negotiation depend on the idempotency capture.

Delete this recovery-route sentence and record the declared values on `NormalizationResult` instead, which every transport already receives from the shared seam.

src/core/request_compat.py
```diff
@@ -16,6 +16,31 @@
 logger = logging.getLogger(__name__)
 V25_SIGNALS: frozenset[str] = frozenset({"brand_manifest", "promoted_offerings", "campaign_ref"})
 
+# AdCP version-envelope fields (spec 3.1.1 core/version-envelope.json, allOf-composed
+# into every request schema). Official SDK clients inject these on every request; the
+# server MUST accept them. Stripped unconditionally on MCP (all environments) so they
```
KonstantinMirin, line 21: This module's header at `:5` reads "Shared by all transports (MCP, A2A, REST)", and its three neighbours (`normalize_request_params`, `strip_unknown_params`, `deep_strip_to_schema`) are specified transport-blind. "Stripped unconditionally on MCP" here, and "so they never reach FastMCP's TypeAdapter" in the docstring at `:29-32`, name one transport in a shared contract, which is how the wiring asymmetry reads as intentional to the next reader.

Restate both in protocol terms: the AdCP version envelope is accepted on every request and is not a tool parameter. Keep the rationale in one docstring; it is currently written four times in `src/` (here, `:27-35`, `mcp_compat_middleware.py:35-38` and `:61-66`).

tests/unit/test_mcp_compat_middleware.py
```diff
@@ -109,6 +109,96 @@ async def capturing_call_next(context):
             assert captured_ctx.message.arguments == {"brand": {"domain": "acme.com"}}
 
 
+class TestMiddlewareStripsEnvelopeVersionFields:
+    """AdCP version-envelope fields are stripped unconditionally (all environments)."""
+
+    @pytest.mark.asyncio
+    async def test_envelope_version_fields_removed_in_dev(self, middleware):
```
KonstantinMirin, line 117: Replace `"adcp_version": "3.1", "adcp_major_version": 3` with `"99.0"` / `99` in all three tests of this class, change no production code, and the file is 16 passed. The code under test never reads a value (`ENVELOPE_VERSION_FIELDS & params.keys()`), so this class asserts that two keys were deleted, not that a version the seller can serve was accepted.

Don't add a `99` case on top of the current design; it would only re-assert deletion. Record the declared values on `NormalizationResult` first, then assert on them here and on the wire in the BDD scenario.

tests/unit/test_mcp_compat_middleware.py
```diff
@@ -109,6 +109,96 @@ async def capturing_call_next(context):
...
+        async def capturing_call_next(context):
+            return None
+
+        with (
+            patch("src.core.config.is_production", return_value=False),
+            patch("src.core.request_compat.logger") as mock_logger,
+        ):
+            await middleware.on_call_tool(ctx, capturing_call_next)
+
+        mock_logger.debug.assert_called_once_with(
```
KonstantinMirin: This asserts the `logger.debug` format string and counts every `debug` call on the module-level `src.core.request_compat` logger, which also emits at `request_compat.py:86` and `:291`. A payload that reaches the brand_manifest branch at `:86` during the same call produces a second record and fails this test for a reason unrelated to the strip. Rewording the message also fails it, although the behavior it names has not changed.

Assert the emitted record instead: `caplog.at_level(logging.DEBUG, logger="src.core.request_compat")`, then check that both stripped field names appear in a DEBUG record for `get_products`.

tests/integration/test_mcp_unknown_field_handling.py
```diff
@@ -57,6 +57,28 @@ def test_unknown_field_rejected(self, integration_db):
...
+
+        with ProductEnv(tenant_id=TENANT_ID) as env:
+            _create_tenant_with_product()
+            result = env.call_via(
+                Transport.MCP,
+                brief="test ads",
+                adcp_version="3.1",
+                adcp_major_version=3,
+            )
+            assert result.is_success
```
KonstantinMirin: `is_success` is `error is None and payload is not None` (`tests/harness/transport.py:371`); it does not require that any wire body was captured. The harness ships `TransportResult.require_wire()` (`:428-449`) for the success side, and its docstring names the case it exists to catch: a success with no stashed body means the dispatch bypassed the real pipeline. The negative control at `:43-56` already asserts the full envelope through `assert_envelope_shape` and `wire_error_envelope["errors"][0]["field"]`.

Assert on `result.require_wire()["products"]` instead. The field is on every request envelope, so the behavior also needs a transport-blind scenario in `tests/bdd/features/BR-COMPAT-001-deprecated-field-normalization.feature`, wired by `tests/bdd/test_compat001_deprecated_field_normalization.py` and steps registered at `tests/bdd/conftest.py:83`, dispatched through `dispatch_request` and asserted with `wire_dict` / `wire_field`. Don't tag it `@mcp`; that suppresses the parametrization at `tests/bdd/conftest.py:3218`.

tests/unit/test_pydantic_schema_alignment.py
```diff
@@ -26,6 +26,12 @@
 from pydantic import BaseModel, ValidationError
 
 from src.core.exceptions import AdCPInvalidRequestError
+
+# SSOT: the version-envelope field set lives in src/core/request_compat.py. Import it
+# (aliased to the historical local name) rather than re-listing it here, so the MCP
+# strip and this alignment suite can never disagree. Set-equality with the pinned
+# schema is graded by test_envelope_version_fields_match_pinned_schema below.
+from src.core.request_compat import ENVELOPE_VERSION_FIELDS as _VERSION_FIELDS
```
KonstantinMirin: The rename means `grep -rn ENVELOPE_VERSION_FIELDS tests/` matches this line and one assertion message, and no use site, which is the opposite of what the constant was created for. This repo's underscore alias preserves the name and marks privacy only (`get_header_case_insensitive as _get_header_case_insensitive`, `src/core/auth.py:38`); renaming aliases here exist for real collisions, and there is none.

Import it under its own name and rename the use sites at `:92, :101, :145, :147, :467, :635, :818, :1180, :1220, :1576, :1683`, then delete the comment above and the two at `:101` and `:133-134` that exist to explain the alias. The second hand-written copy at `tests/harness/test_forward_compat_acceptance.py:48` and `:217` should import the constant too.

src/core/mcp_compat_middleware.py
numarasSigmaSoftware: Deferred to #2207 (self-assigned), which subsumes #2184 (closed). Measured the blast radius: folding into the seam plus the required A2A/REST reader rewiring is a 3-transport expansion of a focused MCP PR, not a net-shrink, and — as you note — a bare delete in the seam regresses A2A/REST v2-compat, so the safe version is the whole rewiring. #2207 owns it.

src/core/request_compat.py
numarasSigmaSoftware: Agreed — the `tuple[dict, bool]` can't carry the value, and recording `declared_adcp_version`/`declared_adcp_major_version` on `NormalizationResult` (which already carries `inferred_version`) is the right shape. That's the core of #2207; kept out of this focused PR for the scope reason on the seam thread.

src/core/mcp_compat_middleware.py
numarasSigmaSoftware: #2207 records the declared values on `NormalizationResult` instead of the `raw_wire_payload` stash — exactly to avoid the three-implementation coupling and the idempotency-capture dependency you describe. The recovery-route sentence goes away with the seam move there.

src/core/request_compat.py
numarasSigmaSoftware: The strip is genuinely MCP-only in this PR (the seam move is deferred), so "on MCP" is accurate for the current code. #2207 moves it into the shared seam and restates the constant/docstring in protocol terms at that point, and collapses the four rationale copies to one.

tests/unit/test_mcp_compat_middleware.py
numarasSigmaSoftware: The vacuity is rooted in not recording the value; recording it and asserting on it is #2207. Per your note I did not add a `99` case on the current design (it would only re-assert deletion). Hardened the independent parts in ad23b071 (caplog, require_wire, shared constant).

tests/unit/test_mcp_compat_middleware.py
numarasSigmaSoftware: Fixed in ad23b071: `caplog.at_level(logging.DEBUG, logger="src.core.request_compat")`, asserting both stripped field names appear in a DEBUG record for `get_products` — no longer a call-count on the module logger or the format string.

tests/integration/test_mcp_unknown_field_handling.py
numarasSigmaSoftware: Fixed in ad23b071: asserts `result.require_wire()["products"]` instead of `is_success`. The transport-blind BR-COMPAT-001 scenario rides with #2207, since it needs the recorded declared version.

tests/unit/test_pydantic_schema_alignment.py
numarasSigmaSoftware: Fixed in e3aa804a: imported `ENVELOPE_VERSION_FIELDS` under its own name, renamed all use sites, deleted the three alias-explanation comments. The second hand-written copy at `test_forward_compat_acceptance.py` now imports the constant too (b8c66f49).

PR #1547, opened 2026-07-06
fix: A2A top-level failures return failed Task with AdCP envelope, not JSON-RPC InternalError

numarasSigmaSoftware:
## Problem

The A2A server surfaced **application-layer failures on the JSON-RPC transport layer**, emitted webhooks that contradicted the AdCP terminal-state rules, leaked raw exception text (credentials/connection strings) onto the wire, and lost the buyer's task identity across the async workflow boundary (poll/cancel/webhook correlation).

## Spec Grounding (AdCP 3.1.1; prose tree rendered at `dist/docs/3.1.1/`)

- `building/operating/transport-errors.mdx` **"Layer Separation"**: application/task errors (including unknown/unsupported skills, caught *after* routing) belong in the task response body as a failed Task with the two-layer envelope; JSON-RPC/`MethodNotFound` is reserved for transport faults and infra-before-dispatch. The table at `transport-errors.mdx:11-18` lists `internal crash` under the transport layer, with no RFC keyword — an earlier revision of this description wrote "(SHOULD-level)", which the spec does not say, and described a split by WHERE the crash surfaces, which the code does not implement: `_dispatch_under_sanitize_seam` catches only `(AdCPError, ValueError, PermissionError)`, so an untyped skill-handler crash reaches the same transport-layer `InternalError` as a boundary crash. The inner seam's own docstring states this correctly.
- `building/operating/transport-errors.mdx` **"Security Considerations" § Seller Requirements**: error responses MUST NOT include hostnames, database/SQL text, credentials, or upstream internals; `suggestion` carries generic guidance only. This grounds the boundary sanitizer (`safe_adcp_error`) including its synthesized suggestion.
- `building/by-layer/L0/a2a-guide.mdx` **"Webhook Trigger Rules"**: no webhook is sent when the initial response is already terminal; webhooks fire only for non-terminal initial responses and later async transitions, which carry the full Task's artifacts. `canceled` is listed among the final states ("Cancellation confirmed").
- `tasks/get` / `tasks/cancel` semantics are **A2A-protocol-native** (A2A spec Task Management: `TaskNotFoundError`, `TaskNotCancelableError` for terminal tasks; SDK `default_request_handler` as reference cross-check) — AdCP prose defines no cancel contract of its own.
[...]

Issue #1588, opened 2026-07-11
Clarify AdCP Layer Separation for untyped internal crashes (transport vs task body)


ChrisHuie:
Follow-up from #1547 (re-review). The AdCP Layer Separation model is ambiguous for untyped internal crashes, and our A2A handler resolves that ambiguity implicitly. Track the decision and, if warranted, seek upstream clarification.

## Detail
AdCP 3.1.0-beta.3 `transport-errors.mdx` §"Layer Separation" lists **"internal crash"** under the **Transport** column → JSON-RPC `error` / A2A protocol error. But `on_message_send`'s top-level `except Exception` (`src/a2a_server/adcp_a2a_server.py:955-998`) routes **all** untyped exceptions — genuine handler bugs included — to a failed **Task body** as `SERVICE_UNAVAILABLE`, not JSON-RPC. So a server bug becomes indistinguishable on the wire from a legitimate application outcome (e.g. `UNSUPPORTED_FEATURE`), which is the distinction Layer Separation exists to preserve.

The current choice is defensible: the two-layer rule is SHOULD-level, "internal crash" arguably means connection/protocol-level, and A2A clients get a terminal task state plus an extractable envelope. But #1547's in-code comment/PR body assert this as a clean "MUST → task body," which overstates the spec for the untyped case. The typed application path (`AdCPCapabilityNotSupportedError` → `UNSUPPORTED_FEATURE`/correctable) is unambiguously conformant and is the graded/tested path.

## Contract
- Decide and document whether untyped internal crashes should route to the task body (current) or to JSON-RPC per the literal table; if the latter, branch on `isinstance(e, AdCPError)` at the outer handler.
- Narrow the in-code comment to typed/application failures and label the untyped-crash routing an explicit decision.
- If the spec is genuinely underspecified here, raise a clarification upstream (adcontextprotocol/adcp).

Distinct from #1574 (cross-transport execution of the typed scenario once the transport-aware harness lands).

numarasSigmaSoftware:
Decision from the #2183 / #2185 review (pinned AdCP 3.1.1):

Use a strict typed/untyped split at protocol boundaries:

- A typed `AdCPError` is an expected application failure and belongs in the task/tool response as the two-layer AdCP error envelope.
[...]

PR #2183, opened 2026-09-02
fix: route A2A failures to the correct protocol layer (#2070)

### LINKED ISSUE #2070: A2A failed-Task error envelopes are built, attached, then discarded

numarasSigmaSoftware:
## Summary

Fixes #2070.

The A2A `on_message_send` boundary now applies a strict typed/untyped split:

- Expected, typed `AdCPError` application failures return a failed A2A Task carrying the standard two-layer AdCP error envelope.
- Unexpected, untyped exceptions are logged internally and surface as a sanitized JSON-RPC `-32603 Internal error`. Their exception text and internals are not sent to the buyer, and the tentative lifecycle Task/push configuration is discarded.
- Existing `A2AError` transport failures continue to pass through unchanged.
- The four task-push-notification configuration methods are unchanged because they have no Task body in which to carry an application error.

This corrects the earlier implementation, which normalized arbitrary crashes to retryable `SERVICE_UNAVAILABLE` Task failures.

## Spec grounding

Pinned AdCP 3.1.1, `docs/building/operating/transport-errors.mdx`, "Layer Separation":

- application/task failures belong in the task response body;
- internal crashes belong on the transport/protocol error path;
- raw implementation details must not be exposed to clients.

No AdCP 3.1.1 conformance storyboard directly grades this A2A routing distinction, so the behavior is pinned with focused unit and real HTTP wire tests.

## Scope

This remains a focused amendment to the #2070 A2A boundary fix extracted from closed PR #1547.

The explicit-skill dispatcher still normalizes arbitrary exceptions into task envelopes. That broader repository-wide policy change is intentionally not bundled here and remains tracked by the canonical issue #1588 (#2185 was closed as a duplicate). MCP transport behavior also remains separate because the currently pinned Python MCP v1 stack swallows `McpError` at a lower boundary; fixing that safely requires an SDK/migration decision.

## Tests

- Typed `AdCPError` → failed Task with its original wire code and recovery metadata.
- Untyped exception → sanitized JSON-RPC `-32603`, with no exception text, AdCP envelope, or retained lifecycle Task.
- Existing A2A transport errors still pass through.
- Real HTTP transport contract test verifies the serialized JSON-RPC response.

`make quality` is green: **6577 passed, 39 skipped, 32 xfailed**. Focused A2A tests: **65 passed**.

### DIFF
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -833,7 +833,7 @@ class AdCPRequestHandler(RequestHandler):
                     # All skills failed - mark task as failed
                     task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_FAILED))
 
-                    # Send protocol-level webhook notification for failure
+                    # Delivery is gated centrally in _send_protocol_webhook (inline terminal → skip).
                     error_messages = [
                         res["error_envelope"]["errors"][0]["message"] for res in results if not res["success"]
                     ]
@@ -1039,7 +1039,7 @@ class AdCPRequestHandler(RequestHandler):
             # Mark task with appropriate status
             task.status.CopyFrom(TaskStatus(state=task_state))
 
-            # Send protocol-level webhook notification if configured
+            # Delivery is gated centrally in _send_protocol_webhook (inline terminal → skip).
             await self._send_protocol_webhook(task, status=task_status_str)
 
         except A2AError:
@@ -1085,9 +1085,9 @@ class AdCPRequestHandler(RequestHandler):
                 )
             )
 
-            # Inline terminal failed Task: no webhook (webhooks.mdx:160 MUST NOT), and
-            # the accepted push config is dropped here — a terminal task returned
-            # synchronously has no future consumer for it.
+            # No webhook: an inline terminal failed Task is gated out centrally in
+            # _send_protocol_webhook. The accepted push config is dropped — a terminal
+            # task returned synchronously has no future consumer for it.
             self._task_push_configs.pop(task_id, None)
 
         self.tasks[task_id] = task
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -221,23 +221,25 @@ DISCOVERY_SKILLS = frozenset(
 
 
 def _internal_error_for(operation: str, exc: Exception) -> InternalError:
-    """Canonical InternalError shape for non-skill A2A boundary failures.
+    """Canonical InternalError shape for non-Task A2A boundary failures.
 
     Skill handlers raise typed ``AdCPError`` (or untyped exceptions that the
     dispatcher normalizes), and ``_handle_explicit_skill`` → ``on_message_send``
-    surface those as a two-layer envelope on a failed Task's DataPart. Non-skill
-    paths (``on_message_send`` fallthrough, NL handlers) historically picked their
-    own prefixes (``"Message processing failed: "``, ``"Error in ..."``)
-    for semantically identical untyped failures — divergence on the buyer-
-    facing wire message for the same condition.
+    surface those as a two-layer envelope on a failed Task's DataPart —
+    including the top-level ``on_message_send`` fallthrough, which per AdCP
+    3.1.x transport rules (building/operating/transport-errors.mdx "Layer
+    Separation") RETURNS a failed Task carrying the envelope artifact rather
+    than raising this helper. JSON-RPC errors are reserved for genuine
+    transport faults (``A2AError``) and methods with no Task to carry the
+    envelope.
 
-    Use this helper at every non-skill ``InternalError(...)`` raise site that
-    is NOT a deliberate protocol-level convention (see push-notif handlers
-    below). The canonical prefix is ``"{operation} failed: {exc}"`` so
-    storyboard runners can parse the failure uniformly.
+    Use this helper only at ``InternalError(...)`` raise sites that have no
+    async Task to attach an envelope artifact to. The canonical prefix is
+    ``"{operation} failed: {exc}"`` so storyboard runners can parse the
+    failure uniformly.
 
     The four ``on_*_task_push_notification_config`` JSON-RPC protocol methods use
-    this helper too — they have no async Task to carry a DataPart, so the two-layer
+    this helper — they have no async Task to carry a DataPart, so the two-layer
     envelope rides in the error's ``data`` field (``error.data["errors"][0]["code"]``
     / ``error.data["adcp_error"]``). ``InternalError`` stays an ``A2AError`` so the
     SDK's ``JsonRpcDispatcher`` serializes it as a structured JSON-RPC error; raising
@@ -925,9 +927,10 @@ class AdCPRequestHandler(RequestHandler):
                 # ``_create_media_buy`` is an NL stub that always raises
                 # ``AdCPCapabilityNotSupportedError`` — the explicit-skill
                 # path is the spec contract for media buy creation. The
-                # outer error handler at on_message_send catches the raise
-                # and attaches a spec-compliant two-layer envelope to the
-                # failed Task artifact.
+                # outer error handler at on_message_send catches the raise,
+                # attaches a spec-compliant two-layer envelope to the failed
+                # Task artifact, and returns that failed Task (never a
+                # JSON-RPC error).
                 await self._create_media_buy(combined_text, identity)
             else:
                 # General help response
@@ -1022,12 +1025,14 @@ class AdCPRequestHandler(RequestHandler):
                 principal_id=err_principal_id,
             )
 
-            # Send protocol-level webhook notification for failure if configured
             task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_FAILED))
             # Attach error to task artifacts as a spec-compliant two-layer
             # envelope (same shape as failed-skill DataParts) so storyboard
             # runners can ``JSON.parse`` the artifact uniformly regardless of
-            # which failure path produced it.
+            # which failure path produced it. ``_build_error_envelope``
+            # normalizes untyped exceptions to base AdCPError (wire code
+            # SERVICE_UNAVAILABLE via ERROR_CODE_MAPPING) and passes
+            # through a typed AdCPError's own wire code.
             del task.artifacts[:]
             task.artifacts.append(
                 Artifact(
@@ -1037,10 +1042,17 @@ class AdCPRequestHandler(RequestHandler):
                 )
             )
 
+            # Send protocol-level webhook notification for failure if configured
             await self._send_protocol_webhook(task, status="failed")
 
-            # Raise A2A error instead of creating failed task
-            raise _internal_error_for("message processing", e)
+            # Per AdCP 3.1.x transport rules (spec prose:
+            # building/operating/transport-errors.mdx "Layer Separation" and the
+            # two-layer error-handling model), application/task-execution
+            # failures MUST be returned in the task response body as a failed
+            # Task carrying the envelope artifact. JSON-RPC errors are reserved
+            # for genuine transport faults (A2AError, re-raised above). So we
+            # fall through to the shared store-and-return below — never raise
+            # InternalError here.
 
         self.tasks[task_id] = task
         return task
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -220,6 +220,19 @@ DISCOVERY_SKILLS = frozenset(
 )
 
 
+# Terminal A2A task states. `_send_protocol_webhook` notifies on the INITIAL synchronous
+# response only; an inline terminal response must not emit a webhook (pinned AdCP 3.1.1
+# webhooks.mdx:160 — the buyer already has the result).
+_TERMINAL_TASK_STATES = frozenset(
+    {
+        TaskState.TASK_STATE_COMPLETED,
+        TaskState.TASK_STATE_FAILED,
+        TaskState.TASK_STATE_CANCELED,
+        TaskState.TASK_STATE_REJECTED,
+    }
+)
+
+
 def _internal_error_for(operation: str, exc: Exception) -> InternalError:
     """Canonical InternalError shape for non-Task A2A boundary failures.
 
@@ -466,6 +479,14 @@ class AdCPRequestHandler(RequestHandler):
         ``notify`` selects the payload type from the status: Task for final states,
         TaskStatusUpdateEvent for intermediate ones.
         """
+        # This sender notifies on the INITIAL synchronous response only. Per pinned AdCP
+        # 3.1.1 webhooks.mdx:160, an inline terminal response MUST NOT emit a webhook — the
+        # buyer already has the result. No async completion-delivery path exists today; if
+        # one is added it MUST use a distinct entry point, because this guard would
+        # otherwise suppress its legitimate terminal completion webhook.
+        if task.status.state in _TERMINAL_TASK_STATES:
+            return
+
         try:
             # Check if task has push notification config stored
             webhook_config = self._task_push_configs.get(task.id)
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -166,14 +166,14 @@ def _accept_a2a_push_config(url: str, scheme: str | None, credentials: str | Non
     tool path on either precondition.
 
     Why both surfaces must come through here rather than call the constructor
-    directly: ``on_message_send``'s body runs inside a ``try`` whose handlers are
-    ``except A2AError: raise`` / ``except Exception`` -> ``_internal_error_for``.
-    ``AdCPValidationError`` is not an ``A2AError``, so a raw constructor call
-    there would surface a buyer's correctable credential refusal as
-    ``INTERNAL_ERROR``. The translation below preserves the raised
-    ``AdCPValidationError`` verbatim (see
-    :func:`_invalid_params_from_ssrf_error`'s isinstance branch), which is what
-    keeps a credential refusal naming the credentials field instead of being
+    directly: a push-config refusal is a protocol-level precondition failure and
+    must surface as a JSON-RPC ``InvalidParamsError``. ``AdCPValidationError`` is
+    not an ``A2AError``, so a raw constructor call inside ``on_message_send``'s
+    ``try`` would fall to its typed-error branch and be returned as a task-level
+    ``VALIDATION_ERROR`` failed Task instead. The translation below re-raises the
+    refusal as ``InvalidParamsError`` (see
+    :func:`_invalid_params_from_ssrf_error`'s isinstance branch), keeping it on
+    the transport layer and naming the credentials field instead of being
     re-labelled as a URL problem.
     """
     try:
@@ -470,14 +470,14 @@ class AdCPRequestHandler(RequestHandler):
         result: dict[str, Any] | None = None,
         error: str | None = None,
     ):
-        """Send protocol-level push notification if configured.
+        """Send a protocol-level push notification for a non-terminal status, if configured.
 
-        Per AdCP A2A spec (https://docs.adcontextprotocol.org/docs/protocols/a2a-guide#push-notifications-a2a-specific):
-        - Final states (completed, failed, canceled): Send full Task object with artifacts
-        - Intermediate states (working, input-required, submitted): Send TaskStatusUpdateEvent
-
-        ``notify`` selects the payload type from the status: Task for final states,
-        TaskStatusUpdateEvent for intermediate ones.
+        This sender delivers only on the INITIAL synchronous response, which today is
+        always non-terminal here (``submitted``): an inline terminal response is guarded
+        out below (pinned AdCP 3.1.1 webhooks.mdx:160 — the buyer already has the result).
+        The terminal completion of a ``submitted`` operation is delivered later by an async
+        worker, which does not exist yet and would need its own entry point (see the guard).
+        For the non-terminal states that reach ``notify``, it sends a TaskStatusUpdateEvent.
         """
         # This sender notifies on the INITIAL synchronous response only. Per pinned AdCP
         # 3.1.1 webhooks.mdx:160, an inline terminal response MUST NOT emit a webhook — the
@@ -544,9 +544,9 @@ class AdCPRequestHandler(RequestHandler):
                 notification_type=None,
             )
 
-            # notify() picks the payload type: Task for final states,
-            # TaskStatusUpdateEvent for intermediate ones. protocol is "a2a"
-            # unconditionally -- this IS the A2A server.
+            # notify() picks the payload type by status; terminal states are guarded out
+            # above, so in practice this delivers a TaskStatusUpdateEvent for the
+            # non-terminal (submitted) case. protocol is "a2a" -- this IS the A2A server.
             sent = await push_notification_service.notify(
                 push_notification_config,
                 task=webhook_task,
@@ -830,15 +830,11 @@ class AdCPRequestHandler(RequestHandler):
                 successful_skills = [res["skill"] for res in results if res["success"]]
 
                 if failed_skills and not successful_skills:
-                    # All skills failed - mark task as failed
+                    # All skills failed - mark task as failed and return it inline.
+                    # No webhook: an inline terminal response is gated out centrally in
+                    # _send_protocol_webhook (webhooks.mdx:160), so a call here would be a
+                    # guaranteed no-op — the buyer already has the failed Task.
                     task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_FAILED))
-
-                    # Delivery is gated centrally in _send_protocol_webhook (inline terminal → skip).
-                    error_messages = [
-                        res["error_envelope"]["errors"][0]["message"] for res in results if not res["success"]
-                    ]
-                    await self._send_protocol_webhook(task, status="failed", error="; ".join(error_messages))
-
                     return task
                 elif successful_skills:
                     # Log successful skill invocations with rich context
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1047,9 +1047,11 @@ class AdCPRequestHandler(RequestHandler):
                 )
             )
 
-            # Send protocol-level webhook notification for failure if configured
-            await self._send_protocol_webhook(task, status="failed")
-
+            # No webhook here: this branch returns a terminal failed Task
+            # synchronously (inline response). Pinned AdCP 3.1.1
+            # webhooks.mdx:160 (MUST NOT) + a2a-guide:484 — an inline terminal
+            # response emits no push notification.
+            #
             # The exception is a typed application failure, so return its
             # failed Task. The untyped transport-crash branch returned above.
 
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -1002,7 +1002,12 @@ class AdCPRequestHandler(RequestHandler):
             await self._send_protocol_webhook(task, status=task_status_str)
 
         except A2AError:
-            # Re-raise A2AError as-is (will be caught by JSON-RPC handler)
+            # Transport fault: the request could not be routed to a skill, so the
+            # buyer receives a JSON-RPC error and no Task exists to query. Drop the
+            # Task and push registration stashed before the try so neither orphans
+            # (a2a-response-format.mdx@3.1.1 "no artifact produced" row).
+            self.tasks.pop(task_id, None)
+            self._task_push_configs.pop(task_id, None)
             raise
         except Exception as e:
             # Use identity resolved at transport boundary (if available).
@@ -1021,23 +1026,15 @@ class AdCPRequestHandler(RequestHandler):
                 principal_id=err_principal_id,
             )
 
-            # Pinned AdCP 3.1.1, docs/building/operating/transport-errors.mdx
-            # "Layer Separation": an internal crash belongs to the transport
-            # channel. Do not normalize an untyped bug to SERVICE_UNAVAILABLE /
-            # transient — that tells the buyer to retry a potentially
-            # deterministic defect. The A2A SDK serializes this InternalError as
-            # JSON-RPC -32603. Its static message and absent data deliberately
-            # keep exception text and internals off the buyer wire.
-            if not isinstance(e, AdCPError):
-                self.tasks.pop(task_id, None)
-                self._task_push_configs.pop(task_id, None)
-                raise InternalError(message="Internal server error") from e
-
+            # A Task exists and structured error data is in hand, so per pinned AdCP
+            # 3.1.1 a2a-response-format.mdx "Where the Error Lives: Decision Rule" the
+            # failure is RETURNED as a failed Task carrying the adcp_error DataPart —
+            # the same path MCP (tool_error_logging), REST, and the A2A explicit-skill
+            # loop already take. ``_build_error_envelope`` routes an untyped crash
+            # through ``normalize_to_adcp_error`` to base AdCPError (wire
+            # SERVICE_UNAVAILABLE) and passes a typed AdCPError's own wire code. JSON-RPC
+            # is reserved for failures where no Task was produced (the A2AError branch).
             task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_FAILED))
-            # Attach error to task artifacts as a spec-compliant two-layer
-            # envelope (same shape as failed-skill DataParts).
-            # ``_build_error_envelope`` preserves the typed AdCPError's wire
-            # code and recovery metadata.
             del task.artifacts[:]
             task.artifacts.append(
                 Artifact(
@@ -1047,13 +1044,10 @@ class AdCPRequestHandler(RequestHandler):
                 )
             )
 
-            # No webhook here: this branch returns a terminal failed Task
-            # synchronously (inline response). Pinned AdCP 3.1.1
-            # webhooks.mdx:160 (MUST NOT) + a2a-guide:484 — an inline terminal
-            # response emits no push notification.
-            #
-            # The exception is a typed application failure, so return its
-            # failed Task. The untyped transport-crash branch returned above.
+            # Inline terminal failed Task: no webhook (webhooks.mdx:160 MUST NOT), and
+            # the accepted push config is dropped here — a terminal task returned
+            # synchronously has no future consumer for it.
+            self._task_push_configs.pop(task_id, None)
 
         self.tasks[task_id] = task
         return task
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -223,15 +223,11 @@ DISCOVERY_SKILLS = frozenset(
 def _internal_error_for(operation: str, exc: Exception) -> InternalError:
     """Canonical InternalError shape for non-Task A2A boundary failures.
 
-    Skill handlers raise typed ``AdCPError`` (or untyped exceptions that the
-    dispatcher normalizes), and ``_handle_explicit_skill`` → ``on_message_send``
-    surface those as a two-layer envelope on a failed Task's DataPart —
-    including the top-level ``on_message_send`` fallthrough, which per AdCP
-    3.1.x transport rules (building/operating/transport-errors.mdx "Layer
-    Separation") RETURNS a failed Task carrying the envelope artifact rather
-    than raising this helper. JSON-RPC errors are reserved for genuine
-    transport faults (``A2AError``) and methods with no Task to carry the
-    envelope.
+    Skill handlers raise typed ``AdCPError`` for expected application failures,
+    and ``on_message_send`` surfaces those as a two-layer envelope on a failed
+    Task's DataPart. Untyped crashes instead become a separately sanitized
+    JSON-RPC ``InternalError`` at the outer boundary, per AdCP 3.1.1
+    ``docs/building/operating/transport-errors.mdx`` "Layer Separation".
 
     Use this helper only at ``InternalError(...)`` raise sites that have no
     async Task to attach an envelope artifact to. The canonical prefix is
@@ -269,9 +265,9 @@ class AdCPRequestHandler(RequestHandler):
         """Build a spec-compliant two-layer envelope for any exception.
 
         Single source of truth for "wrap-arbitrary-exception → wire envelope"
-        used by both the per-skill dispatcher (``_build_failed_skill_result``)
-        and the top-level ``on_message_send`` error handler. Delegates to
-        ``normalize_to_adcp_error`` for the type→AdCPError mapping
+        used by the per-skill dispatcher (``_build_failed_skill_result``) and
+        by ``on_message_send`` for typed application failures. Delegates to
+        ``normalize_to_adcp_error`` for the per-skill type→AdCPError mapping
         (``ValueError → AdCPValidationError``, ``PermissionError →
         AdCPAuthorizationError``, arbitrary ``Exception →
         AdCPError(INTERNAL_ERROR)``) so the wire output stays in
@@ -1025,14 +1021,23 @@ class AdCPRequestHandler(RequestHandler):
                 principal_id=err_principal_id,
             )
 
+            # Pinned AdCP 3.1.1, docs/building/operating/transport-errors.mdx
+            # "Layer Separation": an internal crash belongs to the transport
+            # channel. Do not normalize an untyped bug to SERVICE_UNAVAILABLE /
+            # transient — that tells the buyer to retry a potentially
+            # deterministic defect. The A2A SDK serializes this InternalError as
+            # JSON-RPC -32603. Its static message and absent data deliberately
+            # keep exception text and internals off the buyer wire.
+            if not isinstance(e, AdCPError):
+                self.tasks.pop(task_id, None)
+                self._task_push_configs.pop(task_id, None)
+                raise InternalError(message="Internal server error") from e
+
             task.status.CopyFrom(TaskStatus(state=TaskState.TASK_STATE_FAILED))
             # Attach error to task artifacts as a spec-compliant two-layer
-            # envelope (same shape as failed-skill DataParts) so storyboard
-            # runners can ``JSON.parse`` the artifact uniformly regardless of
-            # which failure path produced it. ``_build_error_envelope``
-            # normalizes untyped exceptions to base AdCPError (wire code
-            # SERVICE_UNAVAILABLE via ERROR_CODE_MAPPING) and passes
-            # through a typed AdCPError's own wire code.
+            # envelope (same shape as failed-skill DataParts).
+            # ``_build_error_envelope`` preserves the typed AdCPError's wire
+            # code and recovery metadata.
             del task.artifacts[:]
             task.artifacts.append(
                 Artifact(
@@ -1045,14 +1050,8 @@ class AdCPRequestHandler(RequestHandler):
             # Send protocol-level webhook notification for failure if configured
             await self._send_protocol_webhook(task, status="failed")
 
-            # Per AdCP 3.1.x transport rules (spec prose:
-            # building/operating/transport-errors.mdx "Layer Separation" and the
-            # two-layer error-handling model), application/task-execution
-            # failures MUST be returned in the task response body as a failed
-            # Task carrying the envelope artifact. JSON-RPC errors are reserved
-            # for genuine transport faults (A2AError, re-raised above). So we
-            # fall through to the shared store-and-return below — never raise
-            # InternalError here.
+            # The exception is a typed application failure, so return its
+            # failed Task. The untyped transport-crash branch returned above.
 
         self.tasks[task_id] = task
         return task
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -295,6 +295,24 @@ class AdCPRequestHandler(RequestHandler):
             "success": False,
         }
 
+    @staticmethod
+    def _error_artifact_parts(envelope: dict[str, Any]) -> list[Part]:
+        """Parts for a failed-Task artifact: recommended TextPart + required DataPart.
+
+        Pinned AdCP 3.1.1 a2a-response-format.mdx "Required Structure": the adcp_error
+        DataPart is required and a human/LLM-readable TextPart is recommended. Single
+        source of shape shared by the top-level ``on_message_send`` failure path and the
+        per-skill dispatcher, so both failed-Task artifacts read identically. The text is
+        the envelope's own error message (``errors[0].message``), never re-derived.
+        """
+        errors = envelope.get("errors") or []
+        text = errors[0].get("message") if errors else None
+        parts: list[Part] = []
+        if text:
+            parts.append(Part(text=text))
+        parts.append(Part(data=_dict_to_value(envelope)))
+        return parts
+
     def _get_auth_token(self, context: ServerCallContext | None = None) -> str | None:
         """Extract Bearer token from ServerCallContext.
 
@@ -766,15 +784,17 @@ class AdCPRequestHandler(RequestHandler):
                     # a reference to the dict about to go on the wire, and one of
                     # them mutated it in place (the list_creatives format_id
                     # bare-string defect). Nothing rebuilds an outbound payload.
-                    text_message = None
-                    if res["success"] and isinstance(artifact_data, dict):
-                        text_message = artifact_data.get("message")
-
-                    # Build parts list per A2A spec: optional text Part + required data Part
-                    parts = []
-                    if text_message:
-                        parts.append(Part(text=text_message))
-                    parts.append(Part(data=_dict_to_value(artifact_data)))
+                    # Build parts per A2A spec: recommended text Part + required data Part.
+                    # A failed skill goes through the shared error-artifact builder so its
+                    # DataPart+TextPart shape matches the top-level on_message_send failure.
+                    if not res["success"]:
+                        parts = self._error_artifact_parts(artifact_data)
+                    else:
+                        text_message = artifact_data.get("message") if isinstance(artifact_data, dict) else None
+                        parts = []
+                        if text_message:
+                            parts.append(Part(text=text_message))
+                        parts.append(Part(data=_dict_to_value(artifact_data)))
 
                     task.artifacts.append(
                         Artifact(
@@ -1040,7 +1060,7 @@ class AdCPRequestHandler(RequestHandler):
                 Artifact(
                     artifact_id="error_1",
                     name="processing_error",
-                    parts=[Part(data=_dict_to_value(self._build_error_envelope(e)))],
+                    parts=self._error_artifact_parts(self._build_error_envelope(e)),
                 )
             )
 
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -220,12 +220,16 @@ DISCOVERY_SKILLS = frozenset(
 )
 
 
-# Terminal A2A task states. `_send_protocol_webhook` notifies on the INITIAL synchronous
-# response only; an inline terminal response must not emit a webhook (pinned AdCP 3.1.1
-# webhooks.mdx:160 — the buyer already has the result).
-_TERMINAL_TASK_STATES = frozenset(
+# Inline terminal FAILURE states. `_send_protocol_webhook` skips these: a task that failed
+# on the initial synchronous response is returned inline, so a webhook would duplicate the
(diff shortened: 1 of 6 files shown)


KonstantinMirin (commented):
## Review — PR #2183

**Overview** — #2070 asked for one repair: `on_message_send` builds the two-layer AdCP envelope, attaches it to a failed Task, and then discards the Task, so the buyer never receives it. That repair is in this diff and it is correct: the typed branch now runs `self.tasks[task_id] = task; return task`. On top of it the PR adds a classification policy the issue did not ask for. An untyped exception becomes a bare JSON-RPC `-32603` with `data=None` and the Task is thrown away, decided by an inline `isinstance` at one entry path of one transport. That policy is the inverse of the pinned AdCP 3.1.1 A2A binding, and it makes the same `RuntimeError` produce four different buyer-visible results across MCP, REST, the A2A explicit-skill path, and the A2A fallthrough. The work below is to remove that policy, put the sanitization the PR correctly wants in the core builder where every transport inherits it, and close the two scope items #2070 named as this PR's deliverable.

### Should fix

#### 1. Untyped crashes are routed off the Task channel, against the pinned A2A binding

`src/a2a_server/adcp_a2a_server.py:1031-1034` sends any non-`AdCPError` to `raise InternalError(message="Internal server error")` and discards the Task. The pin puts this condition on the Task channel:

- `dist/docs/3.1.1/building/by-layer/L0/a2a-response-format.mdx:518` lists "System errors (database unavailable, internal service failure)" under **when to use `status: failed`**.
- `:527` places the structured error in `artifacts[0].parts[]` as a DataPart.
- `:531` and `:533` reserve `status.message` TextPart and JSON-RPC errors for failures where no task exists: parse errors, auth handshake failures, malformed requests. In `on_message_send` a Task exists and the envelope is already built.

Two further claims in the same hunk do not hold against the document they cite:

- `data=None` (`:1034`) removes the only machine-readable signal the buyer had. `transport-errors.mdx:71-89` and the detection order at `:218` both put the AdCP error in `error.data.adcp_error`. Emitting `data=None` answers `check: error_code` with `error_code_not_resolvable` (`:222`).
- The comment at `:1026-1027` says an untyped bug must not become `SERVICE_UNAVAILABLE` / `transient`. `transport-errors.mdx:622` Rule 1 and the reference implementation at `:640-657` end with exactly `{code: 'SERVICE_UNAVAILABLE', message: 'An internal error occurred', recovery: 'transient'}`. The Security Considerations list at `:663-669` forbids the internal service names, database text, and stack traces, not the `code` and `recovery` fields. `recovery: transient` is the pinned `enumMetadata` value for `SERVICE_UNAVAILABLE` in `dist/schemas/3.1.1/enums/error-code.json`, not an implementation choice.

**Fix:** delete the `if not isinstance(e, AdCPError):` branch at `src/a2a_server/adcp_a2a_server.py:1031-1034` and return the failed Task for every exception that reaches the outer handler. That is also the literal #2070 repair. Correct the comment at `:1024-1030` to cite `a2a-response-format.mdx@3.1.1:497-533`, or delete it. **Done when** `grep -n "isinstance(e, AdCPError)" src/a2a_server/adcp_a2a_server.py` returns nothing and `uv run pytest tests/unit/test_a2a_error_routing.py tests/unit/test_a2a_transport_contract.py -q` passes.

#### 2. The exception-to-channel decision and its sanitization have no owner, so a transport entry point became the place to decide both

This is one move, not a list of sites. `src/core/exceptions.py:1340 normalize_to_adcp_error()` documents itself as the single source of truth for the wrapping applied at all three transport boundaries, and `src/core/tool_error_logging.py:288` and `src/app.py:201` ship comments asserting that symmetry. This diff writes a second, contradicting policy inline at `adcp_a2a_server.py:1031` instead of changing the one that exists. One `RuntimeError` now produces:

| path | result |
| --- | --- |
| MCP `src/core/tool_error_logging.py:294` | `AdCPToolError` carrying the `SERVICE_UNAVAILABLE` envelope |
| REST `src/app.py:213`, `:272` | envelope `JSONResponse` |
| A2A explicit skill `adcp_a2a_server.py:703`, `:722` → `_build_failed_skill_result` (`:283`) | failed Task with the envelope |
| A2A fallthrough `adcp_a2a_server.py:1034` (new) | `-32603`, no envelope, no code |

The real defect the PR identified is that raw exception text reaches the buyer, and that defect is in the core mapping, not in A2A. Reproduce at head:

```
uv run python -c "
from src.core.exceptions import normalize_to_adcp_error, build_two_layer_error_envelope
print(build_two_layer_error_envelope(normalize_to_adcp_error(RuntimeError('db host secret-canary exploded')))['adcp_error'])"
# -> {'code': 'SERVICE_UNAVAILABLE', 'message': 'db host secret-canary exploded', 'recovery': 'transient'}
```

Because the scrubbing landed at one raise site, `_internal_error_for` still interpolates `{exc}` into the buyer-visible message on the four `on_*_task_push_notification_config` handlers (`:1230`, `:1312`, `:1370`, `:1415`):

```
uv run python -c "
from src.a2a_server.adcp_a2a_server import _internal_error_for
print(repr(_internal_error_for('set push notification config', RuntimeError('db host secret-canary exploded')).message))"
# -> 'set push notification config failed: db host secret-canary exploded'
```

Same crash, scrubbed on one JSON-RPC method and verbatim on four others, in one file. The diff also rewrote `_internal_error_for`'s docstring from "use this helper at **every** non-skill `InternalError(...)` raise site" to "use this helper **only** at raise sites that **have no async Task**" in the same commit that adds the one raise site the old wording covered. The new wording does not describe `:1034` either: the two lines above it are `self.tasks.pop(task_id, None)` and `self._task_push_configs.pop(task_id, None)`, so that site has a Task and discards it.

**Fix:** put the sanitized fallback in `src/core/exceptions.py` where envelopes are produced, matching `transport-errors.mdx@3.1.1:640-657`. An untyped exception normalizes to code `SERVICE_UNAVAILABLE`, message `"An internal error occurred"`, `recovery: transient`, with the real text staying on the internal log record through `record_boundary_error`, which every boundary already calls. Then make `_internal_error_for` the only way this module constructs an `InternalError` and restore its broad obligation wording. If a Layer-Separation predicate is genuinely wanted repo-wide, it is a property of the error, so it belongs beside `normalize_to_adcp_error` as `is_application_failure(exc: Exception) -> bool` (or an attribute on `AdCPError`), consumed by all six edges in one sweep: `src/core/tool_error_logging.py:294`, `src/app.py:213`, `src/app.py:272`, `adcp_a2a_server.py:283`, `adcp_a2a_server.py:248`, and the deleted `:1031`. Do not register the divergence in an allowlist, an xfail tag, or a `FIXME(#1588)` marker. **Done when** `grep -n 'InternalError(' src/a2a_server/adcp_a2a_server.py` shows construction only inside `_internal_error_for`, and the first reproduction above prints `'An internal error occurred'`.

#### 3. `on_message_send` has no terminal-exit contract, so each exit re-implements registry lifetime, webhook policy, and artifact shape

`self.tasks[task_id] = task` runs at `:601` and `self._task_push_configs[task_id] = registration` at `:656`, both before the `try`. That makes every failure path individually responsible for unregistering, and this diff added the third hand-written compensation (`:1032-1033`) instead of giving the exit one owner. Three consequences, all in the function this diff edits:

- The sibling `except A2AError: raise` at `:1026` leaves exactly the orphan the new test declares forbidden at `tests/unit/test_a2a_error_routing.py:89`, and the orphan is buyer-observable through `tasks/get` (`_get_task_or_raise`, `:1112`). Reproduced at head:

  ```
  uv run python -c "
  import asyncio
  from a2a.types import SendMessageRequest, InvalidRequestError
  from src.a2a_server.adcp_a2a_server import AdCPRequestHandler
  from tests.a2a_helpers import make_a2a_context
  from tests.utils.a2a_helpers import create_a2a_message_with_skill
  h=AdCPRequestHandler(); ctx=make_a2a_context(auth_token=None, headers={'host':'t.example.com'})
  p=SendMessageRequest(message=create_a2a_message_with_skill('create_media_buy',{'product_ids':['p1']}))
  try: asyncio.run(h.on_message_send(p, context=ctx))
  except InvalidRequestError: pass
  print('orphaned:', list(h.tasks))"
  # -> orphaned: ['task_401ad9b93e34']
  ```

  The typed branch keeps a `ValidatedWebhookRegistration`, which holds the buyer's webhook credentials, registered against a task that will never notify. `tests/unit/test_a2a_error_routing.py:155` asserts that as intended.
- The webhook rule the diff introduces at `:1050` ("an inline terminal response emits no push notification", `webhooks.mdx@3.1.1:160` plus `a2a-guide:484`, both verified against the pin) is applied at one of three inline-terminal sites. `:799` sends `status="failed"` and returns the same failed Task at `:801`. `:1002` sends `"completed"` for the Task returned inline at `:1059`. `:736` sends `"submitted"`, which is not terminal and is fine.
- The now-delivered artifact at `:1046` is `parts=[Part(data=...)]` only. The pinned A2A binding for failed tasks is the AdCP error in a DataPart plus a TextPart (`transport-errors.mdx@3.1.1:148-150`, `a2a-response-format.mdx@3.1.1:792`). This was unobservable before the PR because the Task was discarded, and it is in scope now because the PR makes it reach the buyer. The human-readable string is in hand at `envelope["errors"][0]["message"]`, and this file already pairs the two parts at `:773-777`.

**Fix:** give the function one terminal exit. Publish the lifecycle Task and the push config on commit at the single return path, or wrap the body in a registration scope that discards on any non-returning exit, and delete the two `pop()` calls. Move the inline-terminal rule into `_send_protocol_webhook` (`:435`), which is the only place that knows both the task state and the registered config, then remove the per-branch decisions at `:799` and `:1002`. Build the failed artifact through one builder that always emits TextPart plus DataPart, copying the pairing at `:773-777`. **Done when** the reproduction above prints `orphaned: []`, `grep -n "_send_protocol_webhook(task, status=" src/a2a_server/adcp_a2a_server.py` shows no terminal status at a call site, and the failed-Task artifact carries two parts.

#### 4. The behavior this PR exists to deliver is verified without crossing a transport, and the guards that do exist pass when it breaks

`TransportResult.assert_wire_error` states in its own docstring at `tests/harness/transport.py:451` that it is "the single harness-provided way to verify an error on the wire — step definitions must not hand-roll envelope parsing". No new test goes through it. `grep -rn "assert_wire_error\|wire_error_envelope" tests/unit/test_a2a_error_routing.py` returns nothing, and no `.feature` scenario or step binds this behavior on any transport. Four specific consequences:

- **The sanitization guard passes when the behavior is removed.** `tests/unit/test_a2a_transport_contract.py:253` injects the crash into `AdCPRequestHandler._resolve_a2a_identity`, before skill dispatch, then asserts `"secret-canary" not in response.text` and `"SERVICE_UNAVAILABLE" not in response.text`. Move the crash one frame later, to `_handle_get_products_skill`, and the same `get_products` POST returns `TASK_STATE_FAILED` with `{"code":"SERVICE_UNAVAILABLE","message":"adapter exploded: secret-canary","recovery":"transient"}` on the buyer wire while both assertions stay green. `tests/unit/test_error_boundary_translation.py:407` already asserts that opposite outcome for the same input, and neither test names the other.
- **The #2070 contract never leaves the process.** `tests/unit/test_a2a_error_routing.py:95`, `:125`, and `tests/unit/test_a2a_nl_auth_redundancy.py:109` call `await handler.on_message_send(...)` and read the in-process protobuf. Nothing in the repository serializes a `processing_error` artifact through JSON-RPC. Any regression between the returned `Task` and the JSON-RPC `result` leaves all three green.
- **Three assertions read state no wire can observe.** `tests/unit/test_a2a_error_routing.py:89-91` and `:155-157` assert `not handler.tasks`, `not handler._task_push_configs`, and `AsyncMock.assert_not_awaited()`. Emit the terminal push from anywhere other than that one bound attribute and the buyer gets the duplicate delivery while both assertions pass. Each has a transport-observable equivalent already built here: a follow-up `tasks/get` refusal (`adcp_a2a_server.py:1117-1127`), and a real receiver in `tests/harness/webhook_registration.py:76 A2APushRegistrationEnv` on `LocalOriginMixin`, which is how the pinned storyboard says the rule is verified (`dist/compliance/3.1.1/universal/webhook-emission.yaml:5`). Do not register the A2A variant in an e2e-internal or xfail set to absorb the difference.
- **Two assertions grade the weakest observable.** The same hunk adds `body["error"]["code"] == -32601` to `test_unknown_method_returns_error` at `:222` and then grades the unknown-skill case at `:245` only on `"nonexistent_skill" in body["error"]["message"]`. Change that raise from `MethodNotFoundError` to `InternalError` and the buyer-visible code moves from -32601 to -32603, which is the exact distinction this PR draws, and the test passes. `test_genuine_transport_fault_still_raises_json_rpc_error` (`tests/unit/test_a2a_error_routing.py:161`) has the same shape: it asserts `"authentication" in str(exc_info.value).lower()` and ignores the two-layer envelope that path carries in `data` (`adcp_a2a_server.py:632-641`).

**Fix:** grade the buyer-visible contract through `dispatch_request` → `env.call_via(transport)` → `result.assert_wire_error(code)` across REST, MCP, and A2A, letting `recovery` default to the pinned enum. The infrastructure exists: `ProductEnv` declares `MCP_TOOL`, `A2A_SKILL`, and `REST_ENDPOINT` at `tests/harness/product.py:75-88`, and `tests/harness/_base.py:857-865` already reconstructs the envelope off a failed Task's artifact. Re-express the webhook and orphan assertions on the receiver and on `tasks/get`, and delete the private-attribute reads. Add `body["error"]["code"] == -32601` to the unknown-skill test and grade the auth fault with `assert_envelope_shape(exc_info.value.data, "AUTH_REQUIRED", recovery="correctable")`. Rewrite the wire test on its own file's convention: patch `src.core.resolved_identity.resolve_identity` to return `_MOCK_IDENTITY` and put the `side_effect` on the tool seam, so it exercises a crash from tool code rather than a private method name. **Done when** `grep -rn "handler.tasks\|_task_push_configs\|assert_not_awaited" tests/unit/test_a2a_error_routing.py` returns nothing and the failed-envelope contract runs on three transports through `assert_wire_error`.

#### 5. The two harness items #2070 named as this PR's deliverable are untouched

The issue's Scope paragraph is binary on the first one: "either give the env real per-transport dispatch or delete the method."

- `tests/harness/media_buy_update.py:315` — `def call_via(self, transport: Any, **kwargs: Any) -> Any` overrides `BaseTestEnv.call_via` (`tests/harness/_base.py:577`, which dispatches through `DISPATCHERS[transport]`), never references `transport`, and returns `ImplDispatcher().dispatch(...)`. `grep -rn "\.call_via(" tests/unit/test_update_media_buy_behavioral.py` returns nothing, so it has zero callers. `tests/bdd/steps/generic/_dispatch.py:112` calls `env.call_via(transport, **kwargs)` generically, so one env selection away, a scenario reports four transports and crossed none, with `wire_error_envelope=None` and a harness-built `_synthesized_error_envelope`.
- `tests/unit/test_protocol_webhook_ssrf.py:411` and `:434` read the buyer-visible gate (`VALIDATION_ERROR`, `recovery="correctable"`, `field=...url`) off `exc_info.value.data` from a direct `await handler.on_message_send(...)`. That is the issue's item 4 in its post-refactor location, on the function this PR edits and against the registry this diff pops at `:1032`. `error.data` does round-trip through `POST /a2a`, which the diff itself demonstrates in `test_a2a_transport_contract.py`.

**Fix:** delete the `call_via` override so the env inherits per-transport dispatch, since `call_impl` already says impl-only honestly. Route the SSRF gate through a transport, either with the `client` fixture and `assert_envelope_shape(body["error"]["data"], "VALIDATION_ERROR", recovery="correctable", field="push_notification_config.url")`, or through the existing `@T-UC-002-ext-webhook-ssrf` family (`tests/bdd/features/BR-UC-002-create-media-buy.feature:432`, bound at `tests/bdd/steps/generic/given_media_buy.py:3290`) with `a2a_push_notification_config=`. `:390` (`_accept_a2a_push_config`, a helper with no transport) stays a unit assertion. **Done when** `grep -n "def call_via" tests/harness/media_buy_update.py` returns nothing and `grep -n "exc_info.value.data" tests/unit/test_protocol_webhook_ssrf.py` shows only the `:390` helper case.

#### 6. A2A test helpers have two owning modules, the new ones are untyped, and the extraction converted 2 of 3 in-diff sites

`tests/a2a_helpers.py` (5 importers) and `tests/utils/a2a_helpers.py` (7 importers) now both hold A2A message builders and artifact extraction, and `tests/unit/test_a2a_error_routing.py:28-35` imports from both. The new helpers land in the module that owns neither category:

- `tests/a2a_helpers.py:38 extract_processing_error_envelope` re-implements `json.loads(json_format.MessageToJson(part.data))`, which `tests/utils/a2a_helpers.py:39 extract_data_from_artifact` already owns and which the harness calls for this exact case at `tests/harness/_base.py:862`. The two copies have already diverged: the new one indexes `parts[0]`, the existing one scans for the first data part, so a Task with a leading text Part passes one and fails the other.
- `tests/a2a_helpers.py:67 make_nl_send_message_request` rebuilds `tests/utils/a2a_helpers.py:102 create_a2a_text_message`, whose composed form already exists at `tests/integration/test_a2a_skill_invocation.py:184` as `SendMessageRequest(message=create_a2a_text_message(...))`.
- Both consumers then rename the shared helper behind a private alias (`_make_nl_request` at `tests/unit/test_a2a_error_routing.py:41`, `_make_nl_message` at `tests/unit/test_a2a_nl_auth_redundancy.py:24`), so one function answers to three names and `grep make_nl_send_message_request` finds zero call sites.
- `make_mock_a2a_identity()` is routed at `test_a2a_error_routing.py:40` and `test_a2a_nl_auth_redundancy.py:23`. `tests/unit/test_a2a_transport_contract.py:24`, a file this same diff edits and whose new test consumes the literal through `@patch(..., return_value=_MOCK_IDENTITY)`, keeps the byte-identical copy. That is 2 of 3 in-diff sites.
- `_make_handler()` is extracted at `test_a2a_error_routing.py:53` while the in-diff sibling keeps four hand-written copies of the same three lines at `test_a2a_nl_auth_redundancy.py:37`, `:65`, `:90`, `:129`.
- Three of the four new signatures give callers no contract: `extract_processing_error_envelope(task) -> dict` (unannotated parameter, bare `dict`), `make_mock_a2a_identity()` (returns `ResolvedIdentity`), `make_nl_send_message_request(text: str)` (returns `SendMessageRequest`), and `_make_handler() -> tuple[AdCPRequestHandler, object]`, where `make_a2a_context` in the same module is already annotated `-> ServerCallContext`. Five imports sit in function bodies (`tests/a2a_helpers.py:57`, `:69`, `:71`, `tests/unit/test_a2a_nl_auth_redundancy.py:125`, `:127`) while both modules import at top and neither has an import cycle to avoid.
- `handler._get_auth_token = MagicMock(return_value="test-token")` at `test_a2a_error_routing.py:56` returns exactly what the real `_get_auth_token()` (`adcp_a2a_server.py:298`) reads out of the context `make_a2a_context` already builds. Delete the line and all four tests still pass, verified. Its only effect is to make the new suite blind to a regression in token extraction on the one path it exists to grade.

**Fix:** make `tests/utils/a2a_helpers.py` the single owner of A2A message and artifact helpers, re-exporting from `tests/a2a_helpers.py` if the import path must stay. Express the reader as the `processing_error` name assertion plus `extract_data_from_artifact(...)`, and the NL builder as `SendMessageRequest(message=create_a2a_text_message(text))`. Give the two-layer envelope a concrete type built on the SDK model per the `Library*` convention, `class ErrorEnvelope(AdCPBaseModel): adcp_error: Error; errors: list[Error]` with `from adcp import Error`, rather than a local `TypedDict` or `dict[str, Any]`. Annotate the rest (`-> ResolvedIdentity`, `-> SendMessageRequest`, `-> tuple[AdCPRequestHandler, ServerCallContext]`), hoist the five body imports, convert `test_a2a_transport_contract.py:24`, route the four `test_a2a_nl_auth_redundancy.py` setups through the shared handler fixture, drop both aliases, and delete the `_get_auth_token` stub. Much of this disappears if item 4 routes the grading through the harness and the local reader goes away. **Done when** `grep -rn "json_format.MessageToJson(part.data)" tests/` returns one file, `grep -rn "PrincipalFactory.make_identity" tests/unit/test_a2a_transport_contract.py` returns nothing, and `grep -rn "= make_nl_send_message_request" tests/` returns nothing.

#### 7. The routing rule is written in prose in four places, and three of the four are now wrong

`adcp_a2a_server.py:168-177`, `:224-243`, `:268-272`, and `:1025-1030` each restate the A2A error-routing policy. The diff updated three of them and left the fourth asserting the removed behavior:

- `adcp_a2a_server.py:168-177` (`_accept_a2a_push_config`) still says `on_message_send`'s `except Exception` reaches `_internal_error_for` and that an `AdCPValidationError` raised there would surface as `INTERNAL_ERROR`. At head neither is true: the helper has no caller in `on_message_send`, and an `AdCPValidationError` is an `AdCPError`, so it takes the typed branch and becomes a `VALIDATION_ERROR` envelope. This docstring is the stated reason both push-config ingest surfaces go through one seam, so a reader checking whether the seam is still needed gets a rationale that no longer describes the code.
- `tests/a2a_helpers.py:43` and `tests/unit/test_a2a_nl_auth_redundancy.py:118` cite "AdCP 3.1.x". These are the only two occurrences of `3.1.x` in the repository and the pin is 3.1.1, guarded by `tests/unit/test_adcp_spec_version.py`. The same diff cites the same document correctly as "Pinned AdCP 3.1.1" at `adcp_a2a_server.py:230`, `:1024`, and `tests/unit/test_a2a_error_routing.py:3`.
- `tests/unit/test_a2a_error_routing.py:6-8` claims no conformance storyboard grades this change. That holds for the Task-versus-JSON-RPC routing axis only. `dist/compliance/3.1.1/universal/webhook-emission.yaml:5` grades the inline-terminal rule the diff cites, and `dist/compliance/3.1.1/universal/error-compliance.yaml` grades `check: error_code`, which the new `-32603` with `data=None` path answers with `error_code_not_resolvable`.

**Fix:** correct `adcp_a2a_server.py:168-177` to describe the current handler, change both `3.1.x` citations to `3.1.1`, and narrow the module note to the routing axis while naming the two storyboards that do grade the adjacent behavior. Once item 2 gives the predicate one owner, the rule belongs on that function's docstring with the transport comments citing it instead of restating it. Do not fix the cross-module comments at `src/core/exceptions.py:1240`, `src/core/tool_error_logging.py:288-290`, and `src/app.py:203-206` by editing prose: they are correct statements of the design, and item 2's sweep is what makes them true again. **Done when** `grep -rn "AdCP 3\.1\.x" src/ tests/` returns nothing and `adcp_a2a_server.py:168-177` names the branch that actually runs.

- The PR description defers the other three transports to #1588. `gh issue view 1588` returns OPEN with `assignees: []`, so that deferral is not established. Ownership was lost when #2185, which was assigned to you, was closed as its duplicate. Items 1 and 2 are in scope regardless, because the diff creates the second, contradicting policy inside the file it edits. If any residue genuinely survives that sweep, self-assign #1588 and name the exact residual scope in it.
- `gh issue view 1574` returns OPEN and assigned to you. It is a valid home for the four-transport storyboard, and not a home for item 4: the serialized JSON-RPC assertion for the typed-failure path costs about 8 lines in the `TestA2AJsonRpcProtocol` fixture this diff already edits.
- `tests/unit/test_error_boundary_translation.py:407` (untyped explicit-skill failure becomes `SERVICE_UNAVAILABLE` plus the raw message) is an accurate description of today's production and is cited above only as evidence of the divergence. It becomes wrong when items 1 and 2 land, so change it in that sweep and not before.
- About 20 A2A identity literals live in test files this diff does not open, including `tests/unit/test_a2a_handler_correctness.py:90`, `tests/unit/test_a2a_auth_optional.py:28`, and `tests/unit/test_a2a_brand_manifest_parameter.py:22`. They are the same missing-owner problem as item 6 but outside this diff, so converting them is a separate sweep with `make_mock_a2a_identity` as the destination.
- `build_two_layer_error_envelope() -> dict[str, Any]` (`src/core/exceptions.py:1252`) and `AdCPRequestHandler._build_error_envelope() -> dict[str, Any]` (`adcp_a2a_server.py:264`) are the production half of the missing envelope type in item 6. Neither line is in this diff, so converting the producers is a separate boundary. Item 2's fix already edits `normalize_to_adcp_error`, so landing the model in that sweep is the cheaper end state.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1022,12 +1021,23 @@ async def on_message_send(
...
 
-            # Send protocol-level webhook notification for failure if configured
+            # Pinned AdCP 3.1.1, docs/building/operating/transport-errors.mdx
+            # "Layer Separation": an internal crash belongs to the transport
+            # channel. Do not normalize an untyped bug to SERVICE_UNAVAILABLE /
+            # transient — that tells the buyer to retry a potentially
+            # deterministic defect. The A2A SDK serializes this InternalError as
+            # JSON-RPC -32603. Its static message and absent data deliberately
+            # keep exception text and internals off the buyer wire.
+            if not isinstance(e, AdCPError):
```
KonstantinMirin: This branch routes an untyped crash off the Task channel, which is the inverse of the pinned A2A binding. `dist/docs/3.1.1/building/by-layer/L0/a2a-response-format.mdx:518` lists "System errors (database unavailable, internal service failure)" under **when to use `status: failed`**, and `:527` puts the structured error in `artifacts[0].parts[]` as a DataPart. `:531`/`:533` reserve JSON-RPC and `status.message` for failures where no task exists (parse errors, auth handshake, malformed request). Here a Task exists and the envelope is already built.

It is also a second owner for a decision `src/core/exceptions.py:1340 normalize_to_adcp_error` already makes: the same `RuntimeError` now becomes an envelope on MCP (`src/core/tool_error_logging.py:294`), REST (`src/app.py:213`, `:272`), and the A2A explicit-skill path (`:703`/`:722`), and a bare `-32603` here.

**Fix:** delete lines 1031-1034 and return the failed Task for every exception that reaches this handler, which is also the literal #2070 repair.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1022,12 +1021,23 @@ async def on_message_send(
                 principal_id=err_principal_id,
             )
 
-            # Send protocol-level webhook notification for failure if configured
+            # Pinned AdCP 3.1.1, docs/building/operating/transport-errors.mdx
+            # "Layer Separation": an internal crash belongs to the transport
+            # channel. Do not normalize an untyped bug to SERVICE_UNAVAILABLE /
+            # transient — that tells the buyer to retry a potentially
```
KonstantinMirin: This rationale contradicts the document it cites. `transport-errors.mdx@3.1.1:622` Rule 1 says "A database connection timeout becomes `SERVICE_UNAVAILABLE`", and the reference implementation at `:640-657` ends with the unclassified fallback `{code: 'SERVICE_UNAVAILABLE', message: 'An internal error occurred', recovery: 'transient'}`. `recovery: transient` is the pinned `enumMetadata` value for that code in `dist/schemas/3.1.1/enums/error-code.json`, not an implementation choice. What `:663-669` forbids is the message content (internal service names, database text, stack traces), not the `code` and `recovery` fields.

**Fix:** cite `a2a-response-format.mdx@3.1.1:497-533` instead, or delete the comment along with the branch.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1022,12 +1021,23 @@ async def on_message_send(
...
+            # deterministic defect. The A2A SDK serializes this InternalError as
+            # JSON-RPC -32603. Its static message and absent data deliberately
+            # keep exception text and internals off the buyer wire.
+            if not isinstance(e, AdCPError):
+                self.tasks.pop(task_id, None)
+                self._task_push_configs.pop(task_id, None)
+                raise InternalError(message="Internal server error") from e
```
KonstantinMirin: This bypasses `_internal_error_for`, whose docstring this same diff rewrote to declare it the canonical construction (`:232`). The four sibling sites (`:1230`, `:1312`, `:1370`, `:1415`) still route through the helper and still interpolate `{exc}` into the buyer-visible message:

```
uv run python -c "
from src.a2a_server.adcp_a2a_server import _internal_error_for
print(repr(_internal_error_for('set push notification config', RuntimeError('db host secret-canary exploded')).message))"
# -> 'set push notification config failed: db host secret-canary exploded'
```

So the same crash is scrubbed on one JSON-RPC method and verbatim on four others. `data=None` also drops the AdCP error the pin requires in `error.data.adcp_error` (`transport-errors.mdx@3.1.1:71-89`, detection order item 4 at `:218`), which answers `check: error_code` with `error_code_not_resolvable` (`:222`).

**Fix:** put the sanitized fallback in `normalize_to_adcp_error` / `build_two_layer_error_envelope` so every transport inherits it, keeping the real text on the internal log record through `record_boundary_error`.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1022,12 +1021,23 @@ async def on_message_send(
...
-            # Send protocol-level webhook notification for failure if configured
+            # Pinned AdCP 3.1.1, docs/building/operating/transport-errors.mdx
+            # "Layer Separation": an internal crash belongs to the transport
+            # channel. Do not normalize an untyped bug to SERVICE_UNAVAILABLE /
+            # transient — that tells the buyer to retry a potentially
+            # deterministic defect. The A2A SDK serializes this InternalError as
+            # JSON-RPC -32603. Its static message and absent data deliberately
+            # keep exception text and internals off the buyer wire.
+            if not isinstance(e, AdCPError):
+                self.tasks.pop(task_id, None)
```
KonstantinMirin: These two `pop()` calls compensate for `self.tasks[task_id] = task` at `:601` and `self._task_push_configs[task_id] = registration` at `:656`, both of which run before the `try`. The sibling `except A2AError: raise` at `:1026` has no such compensation, so it leaves exactly the orphan the new test forbids at `tests/unit/test_a2a_error_routing.py:89`, and the orphan is buyer-observable through `tasks/get` (`_get_task_or_raise`, `:1112`). Reproduced at head:

```
uv run python -c "
import asyncio
from a2a.types import SendMessageRequest, InvalidRequestError
from src.a2a_server.adcp_a2a_server import AdCPRequestHandler
from tests.a2a_helpers import make_a2a_context
from tests.utils.a2a_helpers import create_a2a_message_with_skill
h=AdCPRequestHandler(); ctx=make_a2a_context(auth_token=None, headers={'host':'t.example.com'})
p=SendMessageRequest(message=create_a2a_message_with_skill('create_media_buy',{'product_ids':['p1']}))
try: asyncio.run(h.on_message_send(p, context=ctx))
except InvalidRequestError: pass
print('orphaned:', list(h.tasks))"
# -> orphaned: ['task_401ad9b93e34']
```

**Fix:** publish the Task and the push config on commit at the single return path, or wrap the body in a registration scope that discards on any non-returning exit, and delete both `pop()` calls.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1022,12 +1021,23 @@ async def on_message_send(
...
             # Attach error to task artifacts as a spec-compliant two-layer
-            # envelope (same shape as failed-skill DataParts) so storyboard
-            # runners can ``JSON.parse`` the artifact uniformly regardless of
-            # which failure path produced it.
+            # envelope (same shape as failed-skill DataParts).
+            # ``_build_error_envelope`` preserves the typed AdCPError's wire
+            # code and recovery metadata.
             del task.artifacts[:]
             task.artifacts.append(
                 Artifact(
```
KonstantinMirin, line 1080: This artifact now reaches the buyer, which puts its shape in scope. The pinned A2A binding for failed tasks is the AdCP error in a DataPart **plus** a TextPart for human and LLM consumption (`transport-errors.mdx@3.1.1:148-150`, `a2a-response-format.mdx@3.1.1:792`). Line 1046 builds `parts=[Part(data=...)]` only.

The string is already in hand at `envelope["errors"][0]["message"]`, and this file pairs the two parts at `:773-777`.

**Fix:** emit `parts=[Part(text=...), Part(data=...)]`, and put the pairing in one artifact builder so the failed-skill path at `:770-777` and this path produce the same shape.

src/a2a_server/adcp_a2a_server.py
```diff
@@ -1037,10 +1047,13 @@ async def on_message_send(
                 )
             )
 
-            await self._send_protocol_webhook(task, status="failed")
-
-            # Raise A2A error instead of creating failed task
-            raise _internal_error_for("message processing", e)
+            # No webhook here: this branch returns a terminal failed Task
```
KonstantinMirin: The rule is right and it is applied at one of three inline-terminal sites in this file. `:799` sends `_send_protocol_webhook(task, status="failed")` and then returns the same terminal failed Task at `:801`. `:1002` sends `"completed"` for the Task returned inline at `:1059`. `:736` sends `"submitted"`, which is not terminal and is correct.

As written, whether a buyer gets a duplicate notification depends on whether the request arrived as an explicit skill or as NL text, and the new `AsyncMock.assert_not_awaited()` guard cannot see either sibling.

**Fix:** move the rule into `_send_protocol_webhook` (`:435`), which is the only place that knows both the task state and the registered config, then remove the per-branch decisions at `:799` and `:1002`.

tests/a2a_helpers.py
```diff
@@ -28,3 +33,46 @@ def make_a2a_context(
     """
     auth_ctx = AuthContext(auth_token=auth_token, headers=headers or {})
     return ServerCallContext(state={AUTH_CONTEXT_STATE_KEY: auth_ctx})
+
+
+def extract_processing_error_envelope(task) -> dict:
```
KonstantinMirin, line 55: This re-implements the DataPart decode that `tests/utils/a2a_helpers.py:39 extract_data_from_artifact` already owns and that the harness calls for this exact case at `tests/harness/_base.py:862`. The two copies have already diverged: this one indexes `parts[0]`, the existing one scans for the first data part, so a Task with a leading text Part passes one and fails the other.

It also bypasses `TransportResult.assert_wire_error`, whose docstring at `tests/harness/transport.py:451` says it is "the single harness-provided way to verify an error on the wire — step definitions must not hand-roll envelope parsing". Because `recovery` is hand-passed at each call site instead of defaulting to the pinned `error-code.json` classification, a drift in the pin does not make these tests fail.

The signature gives callers nothing either: `task` is unannotated and the return is a bare `dict` for the two-layer envelope, whose entries the SDK models as `adcp.Error`.

**Fix:** grade through `env.call_via(transport)` → `result.assert_wire_error(code)` and delete this helper. If a Task-level fact still needs a plain reader, put it in `tests/utils/a2a_helpers.py` as the `processing_error` name assertion plus `extract_data_from_artifact(...)`, typed `(task: Task) -> ErrorEnvelope` against a model built on `from adcp import Error`.

tests/a2a_helpers.py
```diff
@@ -28,3 +33,46 @@ def make_a2a_context(
...
+    """Read the two-layer AdCP envelope from a failed Task's processing_error artifact.
+
+    ``on_message_send``'s outer error handler attaches the envelope built by
+    ``AdCPRequestHandler._build_error_envelope`` to the failed Task as a
+    single ``processing_error`` artifact with one DataPart (AdCP 3.1.x
```
KonstantinMirin: `AdCP 3.1.x` is not a pin, so it cannot be checked against `tests/unit/test_adcp_spec_version.py` and the grounding claim is unfalsifiable. This and `tests/unit/test_a2a_nl_auth_redundancy.py:118` are the only two occurrences of `3.1.x` in the repository; the same diff cites the same document correctly as "Pinned AdCP 3.1.1" at `src/a2a_server/adcp_a2a_server.py:230`.

**Fix:** change both to `AdCP 3.1.1`.

tests/a2a_helpers.py
```diff
@@ -28,3 +33,46 @@ def make_a2a_context(
...
+    """
+    assert task.artifacts, "failed Task must carry the error envelope artifact"
+    artifact = task.artifacts[0]
+    assert artifact.name == "processing_error", f"expected processing_error artifact, got {artifact.name!r}"
+    part = artifact.parts[0]
+    assert part.HasField("data"), "envelope artifact part must be a DataPart"
+    return json.loads(json_format.MessageToJson(part.data))
+
+
+def make_mock_a2a_identity():
```
KonstantinMirin: No return annotation, and `PrincipalFactory` is imported inside the body while the module imports at top (`:10-15`) and there is no import cycle to avoid. The pre-existing helper in this same module is annotated (`make_a2a_context(...) -> ServerCallContext`), and every helper in `tests/utils/a2a_helpers.py` annotates both parameters and return.

The extraction is also 2 of 3 in-diff sites: `tests/unit/test_a2a_transport_contract.py:24` keeps the byte-identical literal, in a file this diff edits, and the new test there consumes it through `@patch(..., return_value=_MOCK_IDENTITY)`.

**Fix:** hoist the import, annotate `-> ResolvedIdentity`, and convert `test_a2a_transport_contract.py:24` to `_MOCK_IDENTITY = make_mock_a2a_identity()`.

tests/a2a_helpers.py
```diff
@@ -28,3 +33,46 @@ def make_a2a_context(
...
+
+    return PrincipalFactory.make_identity(
+        principal_id="test-principal",
+        tenant_id="test-tenant",
+        tenant={"tenant_id": "test-tenant"},
+        protocol="a2a",
+    )
+
+
+def make_nl_send_message_request(text: str):
```
KonstantinMirin: This rebuilds `tests/utils/a2a_helpers.py:102 create_a2a_text_message`, whose composed form already exists at `tests/integration/test_a2a_skill_invocation.py:184` as `SendMessageRequest(message=create_a2a_text_message(...))`. `uuid` and four `a2a.types` names are imported inside the body while the module imports at top, and there is no return annotation.

Root: two modules named `a2a_helpers` both hold A2A message builders, so a new one has two plausible homes.

**Fix:** make `tests/utils/a2a_helpers.py` the owner, implement this as `SendMessageRequest(message=create_a2a_text_message(text))`, hoist the imports, and annotate `-> SendMessageRequest`.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+    extract_processing_error_envelope,
+    make_a2a_context,
+    make_mock_a2a_identity,
+    make_nl_send_message_request,
+)
+from tests.helpers import assert_envelope_shape
+from tests.utils.a2a_helpers import create_a2a_message_with_skill
+
+_MOCK_IDENTITY = make_mock_a2a_identity()
+_make_nl_request = make_nl_send_message_request
```
KonstantinMirin: The helper was extracted so one name covers both files, and this renames it back. Same at `tests/unit/test_a2a_nl_auth_redundancy.py:24` with a different alias, so one function answers to three names and `grep -rn "make_nl_send_message_request(" tests/` finds zero call sites.

**Fix:** delete both aliases and call `make_nl_send_message_request(...)` at all six sites. `_make_nl_request_with_push` at `:44` is a real local wrapper and stays.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+def _make_nl_request_with_push(text: str, url: str) -> SendMessageRequest:
+    """NL request carrying a protocol-level push-notification config."""
+    request = _make_nl_request(text)
+    request.configuration.CopyFrom(
+        SendMessageConfiguration(task_push_notification_config=TaskPushNotificationConfig(url=url))
+    )
+    return request
+
+
+def _make_handler() -> tuple[AdCPRequestHandler, object]:
```
KonstantinMirin: The setup is extracted here and the in-diff sibling keeps four hand-written copies of the same three lines (`tests/unit/test_a2a_nl_auth_redundancy.py:37`, `:65`, `:90`, `:129`), so a change to the handler's auth seam, which is what these tests drive, needs five parallel edits.

`-> tuple[AdCPRequestHandler, object]` also gives callers no contract: `object` is the same missing-type problem as `Any`, and `make_a2a_context` in the module this calls is already annotated `-> ServerCallContext`.

**Fix:** move the fixture next to `make_a2a_context` as `make_a2a_handler() -> tuple[AdCPRequestHandler, ServerCallContext]`, expose it through `tests/conftest.py`, and route all five in-diff sites through it.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+    """Handler + authenticated call context for driving on_message_send."""
+    handler = AdCPRequestHandler()
+    handler._get_auth_token = MagicMock(return_value="test-token")
```
KonstantinMirin: This stub is inert. `make_a2a_context(auth_token="test-token", ...)` on the next line puts the token exactly where the real `_get_auth_token()` (`src/a2a_server/adcp_a2a_server.py:298`) reads it from `context.state[AUTH_CONTEXT_STATE_KEY]`. Delete the line and all four tests still pass:

```
sed 's/    handler._get_auth_token = MagicMock(return_value="test-token")//' tests/unit/test_a2a_error_routing.py > tests/unit/zz_tmp_mut_test.py
uv run pytest tests/unit/zz_tmp_mut_test.py -q -p no:randomly   # 4 passed
```

Its only effect is to make this suite blind to a regression in token extraction, on the one code path it exists to grade.

**Fix:** delete the stub and let the real extraction run against the context the fixture already builds. Keep `MagicMock` for the boundaries the unit genuinely cannot reach, identity resolution and the tool call.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+            side_effect=RuntimeError("adapter exploded: secret-canary"),
+        ),
+        pytest.raises(InternalError) as exc_info,
+    ):
+        await handler.on_message_send(params, context=ctx)
+
+    assert exc_info.value.message == "Internal server error"
+    assert exc_info.value.data is None
+    assert "adapter exploded" not in str(exc_info.value)
+    assert not handler.tasks, "a transport-rejected request must not leave a lifecycle Task"
```
KonstantinMirin: These three assertions read state no buyer and no Docker or HTTP path can observe: two private dicts on a handler instance and a mock's await count. The webhook assertion passes if the terminal push is emitted from anywhere other than that one bound attribute, for example a shared terminal-exit helper or the SDK's own notifier, and the buyer still gets the delivery.

The orphan assertion is also true only on this branch: the sibling `except A2AError: raise` at `src/a2a_server/adcp_a2a_server.py:1026` leaves the Task registered, and nothing here catches that.

Each has a transport-observable equivalent already built in this repo: a follow-up `tasks/get` refusal (`adcp_a2a_server.py:1117-1127`), and a real receiver in `tests/harness/webhook_registration.py:76 A2APushRegistrationEnv` on `LocalOriginMixin`, which is how the pinned storyboard verifies the rule (`dist/compliance/3.1.1/universal/webhook-emission.yaml:5`).

**Fix:** re-express these on the receiver and on `tasks/get`, and delete the private-attribute reads. Do not register the A2A variant in an e2e-internal or xfail set.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+            side_effect=AdCPValidationError("brief must not be empty"),
+        ):
+            result = await handler.on_message_send(params, context=ctx)
+
+    assert isinstance(result, Task), f"expected a returned Task, got {type(result).__name__}"
+    assert result.status.state == TaskState.TASK_STATE_FAILED, (
+        f"expected TASK_STATE_FAILED, got {result.status.state!r}"
+    )
+    assert_envelope_shape(
+        extract_processing_error_envelope(result),
```
KonstantinMirin, line 121: This is the #2070 contract, "the envelope the boundary builds is the envelope the buyer receives", and it never leaves the process. `await handler.on_message_send(...)` returns the in-process protobuf and this reads it directly, so any regression between that object and the JSON-RPC `result` (artifacts dropped by the task serializer, the DataPart `Value` mangled, `TASK_STATE_FAILED` remapped) leaves the test green. `grep -rn "processing_error" tests/` shows nothing in the repository ever serializes this artifact.

The wire pattern is 40 lines away in `tests/unit/test_a2a_transport_contract.py`, and `ProductEnv` declares `MCP_TOOL`, `A2A_SKILL`, and `REST_ENDPOINT` (`tests/harness/product.py:75-88`) while `tests/harness/_base.py:857-865` already reconstructs the envelope off a failed Task's artifact.

**Fix:** grade through `dispatch_request` → `env.call_via(transport)` → `result.assert_wire_error("UNSUPPORTED_FEATURE")` on REST, MCP, and A2A, letting `recovery` default to the pinned enum, and delete `extract_processing_error_envelope`.

tests/unit/test_a2a_error_routing.py
```diff
@@ -0,0 +1,180 @@
...
+        ),
+    ):
+        result = await handler.on_message_send(params, context=ctx)
+
+    assert isinstance(result, Task), f"expected a returned Task, got {type(result).__name__}"
+    assert result.status.state == TaskState.TASK_STATE_FAILED, (
+        f"expected TASK_STATE_FAILED, got {result.status.state!r}"
+    )
+    # An accepted push config WAS registered for this task ...
+    assert handler._task_push_configs, "push config should have been registered for the task"
```
KonstantinMirin: This asserts as intended that a `ValidatedWebhookRegistration`, which carries the buyer's webhook credentials, stays registered against a Task that is terminal and, per the same diff, will never notify. The untyped branch pops the same registry at `src/a2a_server/adcp_a2a_server.py:1033`, so the two terminal exits of one function disagree about retention.

**Fix:** decide retention once at the single terminal exit for both branches, then this assertion states the real contract instead of one branch's behavior.

tests/unit/test_a2a_transport_contract.py
```diff
@@ -219,13 +219,51 @@ def test_invalid_method_returns_error(self, client, auth_headers):
...
+        mocked so the request reaches skill dispatch; pre-fix this test passed
+        for the wrong reason (the unmocked DB crashed identity resolution first,
+        and the old outer handler converted that crash to a JSON-RPC
+        InternalError).
+        """
         payload = _build_jsonrpc("nonexistent_skill", {})
         response = client.post("/a2a", json=payload, headers=auth_headers)
         body = response.json()
         assert "error" in body, "Unknown skill should return JSON-RPC error"
+        assert "nonexistent_skill" in body["error"].get("message", ""), (
```
KonstantinMirin, line 246: The same hunk adds `body["error"]["code"] == -32601` to `test_unknown_method_returns_error` three lines up, and grades the unknown-skill case only on the message. The code is what a buyer branches on and is exactly what this PR reclassifies: change the raise from `MethodNotFoundError` to `InternalError` while keeping the message, the buyer-visible code moves from -32601 to -32603, and this test passes.

**Fix:** add `assert body["error"]["code"] == -32601` here, matching the assertion added at `:222`.

tests/unit/test_a2a_transport_contract.py
```diff
@@ -219,13 +219,51 @@ def test_invalid_method_returns_error(self, client, auth_headers):
...
+            f"MethodNotFound error should name the unknown skill: {body['error']}"
+        )
+
+    @patch(
+        "src.a2a_server.adcp_a2a_server.AdCPRequestHandler._resolve_a2a_identity",
```
KonstantinMirin, line 251: This is the only test in the diff that crosses a real transport, and it injects the crash into a private method of the class under test, before skill routing and tool invocation. The file's own convention at `:307`, `:326`, `:371`, `:389`, `:411`, `:433` patches `src.core.resolved_identity.resolve_identity` and pairs it with a patch of the tool under test, which is what the new unit test does at `test_a2a_error_routing.py:76-79`.

The injection point is also why the guard cannot fail. Move the crash one frame later, to `_handle_get_products_skill`, and the same `get_products` POST returns `TASK_STATE_FAILED` with `{"code":"SERVICE_UNAVAILABLE","message":"adapter exploded: secret-canary","recovery":"transient"}` on the buyer wire while `"secret-canary" not in response.text` and `"SERVICE_UNAVAILABLE" not in response.text` both stay green. `tests/unit/test_error_boundary_translation.py:407` already asserts that opposite outcome for the same input.

**Fix:** patch `src.core.resolved_identity.resolve_identity` to return `_MOCK_IDENTITY` and put the `side_effect=RuntimeError("... secret-canary ...")` on the tool seam, so this exercises a crash from tool code, which is the condition #2070 is about.

tests/unit/test_a2a_nl_auth_redundancy.py
```diff
@@ -119,40 +106,46 @@ async def test_nl_targeting_query_calls_resolve_identity_once():
...
     """NL media buy is not supported — invocation must surface a typed AdCPError.
 
     Pre-fix the NL stub returned ``{"success": False, "message": "use explicit"}``
     which bypassed the two-layer-envelope contract — storyboard runners
     parsed the artifact as ``MCP_ERROR``. The handler now raises
     ``AdCPCapabilityNotSupportedError`` which propagates to the outer
     ``on_message_send`` error handler that attaches a spec-compliant
-    envelope to the failed Task artifact.
+    envelope to the failed Task artifact and RETURNS the failed Task
+    (per AdCP 3.1.x transport-errors.mdx "Layer Separation" — JSON-RPC
```
KonstantinMirin: `AdCP 3.1.x` is not the pin. Change it to `AdCP 3.1.1`, matching `tests/unit/test_a2a_error_routing.py:3` and `src/a2a_server/adcp_a2a_server.py:1024` in this same diff; `tests/a2a_helpers.py:43` has the other copy.

The claim is also inverted for untyped exceptions after this PR: at `src/a2a_server/adcp_a2a_server.py:1034` an application-layer `RuntimeError` from NL code now becomes a JSON-RPC error with no envelope.

While here, `TaskState` and `assert_envelope_shape` are imported inside the test body at `:125` and `:127`; `tests/unit/test_a2a_error_routing.py:16-29` imports both at module scope without an import cycle.

tests/unit/test_a2a_nl_auth_redundancy.py
```diff
@@ -119,40 +106,46 @@ async def test_nl_targeting_query_calls_resolve_identity_once():
...
 
     # Identity is resolved once during route dispatch before the raise.
     assert mock_resolve.call_count == 1, (
         f"resolve_identity called {mock_resolve.call_count} times for media buy NL request. Expected 1."
     )
-    # InternalError message names the original capability failure (sanity check).
-    assert "create_media_buy" in str(exc_info.value).lower() or "explicit" in str(exc_info.value).lower(), (
-        f"InternalError should reference the unsupported capability; got: {exc_info.value}"
+    assert task.status.state == TaskState.TASK_STATE_FAILED, f"Expected failed Task, got state {task.status.state!r}"
+    envelope = extract_processing_error_envelope(task)
```
KonstantinMirin, line 134: Same problem as `tests/unit/test_a2a_error_routing.py:117`: this reads the envelope off the in-process protobuf returned by `handler.on_message_send(...)`, so nothing verifies the envelope survives serialization, and `recovery="correctable"` is hand-passed instead of defaulting to the pinned `error-code.json` classification.

**Fix:** obtain the result through `env.call_via(transport)` and assert with `result.assert_wire_error("UNSUPPORTED_FEATURE")`, letting `recovery` come from the pin.

PR #2189, opened 2026-09-03
perf(test): read what the run published instead of re-running it

### LINKED ISSUE #2168: test setup: tests that observe a run spawn their own pytest — ten redundant collections of tests/bdd per suite run
Several tests answer questions about **what the suite collects** or **what it
actually ran**, and each answers it by spawning its own `pytest` subprocess. They
pay a full import + collection every time, they duplicate each other's work, and
because their cost is wall-clock they are the first tests to fail when the
machine is busy.

## The duplication, measured

`--collect-only` invocations of `tests/bdd` in one suite run:

| file | invocations |
|------|-------------|
| `tests/unit/test_architecture_env_route_agreement.py` | 4 |
| `tests/unit/test_architecture_ci_bdd_shard_manifest.py` | 3 |
| `tests/unit/test_architecture_bdd_xfail_reason_tokens.py` | 1 |
| `tests/unit/test_e2e_rest_ledger_fitness.py` | 1 |
| `tests/integration/test_storyboard_ledger_fitness_real_session.py` | 1 |

**Ten collections of the same 7223-item tree**, and three of them run the
byte-identical command `python -m pytest tests/bdd --collect-only -q`.

Cost in one full run, for three of them alone:

    51.3 s  test_architecture_bdd_xfail_reason_tokens
    47.8 s  test_architecture_ci_bdd_shard_manifest
    43.4 s  test_e2e_rest_ledger_fitness
    ------
   142.5 s

Collection breadth is not the driver — each process pays a fixed ~58 s import
chain (`src.app` + `tests.harness` + `tests.factories`) before it collects
anything, so narrowing the target files does not help.

## Why it is fragile, not just slow

These are the tests that go red when the box is loaded:
`test_every_ledger_entry_resolves_to_a_collected_item` (44.8 s) and
`test_bdd_shards_partition_suite` (41.7 s) are the two slowest tests in the
suite, and both failed on a busy machine while passing on an idle one. The
liveness tests carry `@pytest.mark.timeout(300)` with a comment recording
"Measured 65.17 s on an idle 40-core box against CI's `--timeout=60`" — a test
whose pass margin is a wall-clock measurement has no headroom by construction.

## Proposal: publish once, consume many

Two questions, two artifacts, both already half-built.

### 1. "What is collected?" — a collection manifest

The suite already collects everything, once, for real. Emit the collected node
ids (per suite) as an artifact of that run, and let the four `--collect-only`
observers read it instead of re-collecting. Ten collections become zero.

This is exactly the move #2006 made for `entity_markers_for_path`: the rule moved
in-process, the guard read it, and the cost went 21 s -> 0.03 s.

### 2. "What actually ran, and how?" — the liveness artifact

`tests/bdd/scenario_liveness.py` is already registered as a plugin
(`tests/bdd/conftest.py:56`) and already writes `bdd_scenario_liveness.json` at
`pytest_sessionfinish` from **every real BDD run**. It also already has the merge
this needs — `_merge_record` / `_merge_shard`, with defined semantics
(`steps_bound` AND-ed, `ledgered` OR-ed, `harness_wired` three-valued,
observations concatenated) — used today to fold xdist workers into the
controller.

`tests/integration/test_bdd_scenario_liveness_real_run.py` spawns **five** nested
pytest runs to reconstruct, in miniature, an artifact the real run already
produced. Three of its four tests can read the real artifact instead:

- `..._records_uc006_..._as_ledgered_or_live`
- `..._records_uc005_format_id_roundtrip_..._as_live`
- `..._provenance_tag_is_a_recorded_field_not_a_collection_filter`

and they get **stronger** doing so: they would grade the whole suite that really
ran rather than one hand-picked file.

`test_storyboard_ledger_fitness_real_session.py` wants the same data and should
read the same artifact.

The one test that cannot move is
`..._identical_under_xdist_and_serial`: it compares a sharded run against a
serial one by set equality, and one real run yields only one shape. It can
compare the **real sharded artifact** against a single serial slice — one
subprocess instead of five.

## Wiring it, so local / cassini / CI grade the same thing

- **In-network (`run_all_tests.sh`, and therefore `cassini run`)**: one
  `bdd_inprocess` env under xdist, so the controller already merges its workers
  and writes one artifact into `$RESULTS_DIR`. A `tox -e liveness` with
  `depends = bdd_inprocess` reads it — the same ordering mechanism `bdd_e2e`
  already uses.
- **CI**: `bdd-tests-shard` is a 2-way matrix that already uploads artifacts, so
  it publishes its liveness artifact alongside its coverage files (one added
  path, no new step). The existing `bdd-tests` job — already
  `needs: [bdd-tests-shard]`, today only a result check — downloads both and
  applies `_merge_record` one level up, across shards instead of across workers.

The merge function is the same in both paths, which is the property worth
having: one implementation grading local, remote and CI identically.

## Acceptance

- No test spawns `pytest --collect-only` against `tests/bdd`.
- Exactly one nested pytest run remains, in the xdist-vs-serial comparison, and
  its purpose is documented at the call site.
- The liveness assertions grade the artifact from the suite that actually ran.
- No `@pytest.mark.timeout` is load-bearing for any of these tests.
- Local, `cassini run` and CI use the same merge code.

## Notes

`tests/unit/test_xdist_report_serialization.py` also spawns pytest, and should
**not** be migrated: it runs a deliberately isolated 4-test suite to prove a
session is silently truncated without the report hook, which requires a real
child session crashing at the xdist boundary. It costs 8-9 s and its subprocess
is the point.

KonstantinMirin:
Closes #2168.

Tests that observed a run were starting their own. Six of them answered "what did the suite collect?" and "what actually ran?" by spawning a fresh pytest, so a full run collected `tests/bdd` ten times and executed the UC-006 slice five times. Each nested process paid ~58s of import before doing anything, which is why narrowing their target files never helped.

They now read what the run publishes.

## Result

| Measure | Before | After |
|---|---|---|
| `--collect-only` argv sites under `tests/` | 4 (10 invocations per run) | 0 |
| Nested pytest runs in this area | 5 | 0 |
| Cost of the four collection observers | 934.3s / 46 tests | 10.76s / 47 tests |
| Liveness artifact coverage | 9 of 21 modules | all 21 |
| CI jobs grading these observers | 0 | 1 |

Verified on a full in-network run (`exit=0`, eight suites). Against the previous verified run every suite is identical except `collection` 52 → 51, the deleted test, and the 53-node xpassed set is identity-equal.

## What each commit does

**Publish the collection manifest.** `tests/_collection_manifest.py` records what each session collected. A record stamps its own scope — `target` (the paths pytest was invoked with) and `filter` (the `-k`/`-m` narrowing) — so a reader names the collection it needs instead of filtering nodeids at the call site. `filter` is what makes each row's `selected` flag interpretable: `bdd_e2e` runs `-k e2e_rest`, where mcp/rest rows are deselected by the filter rather than by the conftest exemption a caller might be grading.

Two hooks, because one cannot do it. `pytest_itemcollected` sees the pre-deselection population; `pytest_collection_finish` runs after every `modifyitems`, so marker sets include the auto-applied entity markers and the transport xfail markers.

**Grade the liveness artifact the real run writes.** The published artifact described one shard — 9 features, `run.collected` 3898, against a 7243-item suite — because both shard sessions wrote one fixed path and the last one won. Nothing said so: `_run_scope` recorded `selection` and `markers`, and a shard narrows by neither. Sessions now write per-session files carrying `target`, and `load_run` folds them with the `_merge_record` the controller already uses for its workers.

**Make a lost shard unrepresentable.** `pytest_testnodedown` skipped a worker that shipped nothing and ignored its `error` argument, so a dead worker took its scenarios with it while the controller published a file that looked whole. The controller now records which workers shut down and which shipped; `workers_missing` carries the difference and `load_run` refuses a file that names one. This replaces a deleted test that ran the suite twice and compared the two files.

**Grade both artifacts in CI.** The observers moved into a `collection` tox env that no CI job ran, so on CI they graded nothing. `bdd-tests-shard` publishes unfiltered path-scoped records; `bdd-in-network` publishes the full-tree record three observers name; `bdd-tests` downloads both and runs `tox -e collection`. The merge is not reimplemented in YAML — the tests call `load()`/`load_run()`, so local runs, `cassini run` and CI grade through one code path.

## Deliberately not done

`test_liveness_artifact_is_identical_under_xdist_and_serial` is deleted rather than migrated. It started two pytest subprocesses and compared their output, which grades the recorder's plumbing rather than the suite, and its substitute is only genuinely parallel on one of four supported setups. The failure it guarded is now prevented at the shipping boundary instead.

The epic's net line count under `tests/` is **+273**, where the structural-change formula asks for a negative number. Two artifacts and their readers cost more than the ten subprocess sites they replace; `tests/_collection_manifest.py` alone is 261 lines. Recorded rather than argued away.

## Carried, not planned

- A merge of `origin/main` (#1802's outbound SSRF seam). Main independently restructured the same six functions this branch had refactored for the complexity ratchet, reaching the same counts and adding the seam, so those files take main's version wholesale. Full resolution ledger in the merge commit.
- `repo-invariants` runs at the commit stage only. It declared no `stages:` and so ran at pre-push too, where the file list is the whole pushed range: 272 violations across 110 files, none authored by this branch. No CI job runs that hook, which is how the debt grew unseen.

## Merge order: this PR is stacked on #2158

**Land #2158 first. This branch resolves against it afterwards.** GitHub cannot
express the stack because both PRs come from a fork, so the ordering lives here
instead of in a base-branch setting.

The two branches are independent lineages — #2158's head is not an ancestor of
this one — and `git merge-tree` conflicts in four files: `run_all_tests.sh`,
`tests/_worker_profile.py` (add/add), `tests/conftest.py` and `tox.ini`.

This matters beyond the conflict count. This branch carries its own copy of
`tests/_worker_profile.py` in its PRE-REVIEW form: `record_test_duration(seconds)`
with no `worker_id` discriminator, the unguarded `mkdir`/`write_text`, the
`max(0.0, ...)` clamp, `controller_collect_s`, and no reader. Merging this after
#2158 without resolving would silently reinstate every defect that review fixed,
with the profile default-on for every in-network run. So on resolution
`tests/_worker_profile.py` takes **#2158's version wholesale**, and the other
three files take the union.

Thanks to @ChrisHuie for catching this — the overlap was reported here earlier as
a single shared commit, which understated it by three files and a whole module.

## Carried from #2158's review

`d1f3700a3` guards the liveness artifact write. `pytest_sessionfinish` raising
is an unhandled error, so the unguarded `mkdir` + `write_text` pair turned a
session whose scenarios all passed into a nonzero exit with no summary line —
the same defect a reviewer found in `tests/_worker_profile.py` on #2158, and
this plugin is registered for **every** BDD session.

Note the scope limit, per B1: that guard reaches one of the three plugins that
write a per-session artifact at `pytest_sessionfinish`. The manifest writer and
the profile writer still crash a passing session on an unwritable path. The fix
is one shared writer, not a third copy of the guard.

## Follow-ups filed

- `salesagent-wkvy2` — `playwright install` stops responding while holding its own lock and wedges an in-network run. Found here, not caused here.
- `salesagent-jmrui` — three UC-004 `e2e_rest` scenarios were passing while reading in-process state; #1802's guards made that visible. They xfail, so nothing is red. Needs header storage in the webhook-capture service, or a ledger reason recording the topology limit.


agentzyx (commented):
.

tests/_worker_profile.py
```diff
@@ -0,0 +1,214 @@
...
+        global _test_seconds
+        _test_seconds += seconds or 0.0
+
+
+def on_session_finish() -> None:
+    if not enabled:
+        return
+    _mark("finish")
+    out = Path(_OUT_DIR)
+    out.mkdir(parents=True, exist_ok=True)
```
ChrisHuie, line 113: **[BLOCKER] A1 — this branch carries the pre-review `tests/_worker_profile.py`; #2158's fixes are absent, and the two PRs conflict in four files.** (`tests/_worker_profile.py:101,113,133,136,212`; `tests/conftest.py:840-845`; `run_all_tests.sh:244`)
`refs/pr/2158` (4af6e0a33) is not an ancestor of this head, and the five commits that answered that review are not here. This copy still has `record_test_duration(seconds)` with no `worker_id` discriminator, the unguarded `mkdir`/`write_text`, the `max(0.0, …)` clamp, `controller_collect_s`, the "reported separately as the gap against the process's own start time" claim, `pytest_runtestloop` without `tryfirst`, and no reader — with the profile default-on for every in-network run. On this head at `-n 2` the controller record's `tests_s` (2.622) equals the sum of both workers (2.621), and `PYTEST_WORKER_PROFILE=<regular file>/sub` turns a passing session into exit 1 with `NotADirectoryError` from `:113` and no summary line. `git merge-tree` of either head into the other conflicts in `run_all_tests.sh`, `tests/_worker_profile.py` (add/add), `tests/conftest.py` and `tox.ini`. Decide the order before merge — rebase this branch onto #2158's head, or land #2158 first and resolve the four files here — and say which in the "Carried from #2158's review" paragraph.

tests/_collection_manifest.py
```diff
@@ -0,0 +1,267 @@
...
+        ],
+    }
+
+
+def on_session_finish() -> None:
+    """Write this process's record."""
+    if not enabled or _record is None:
+        return
+    out = Path(str(_OUT_DIR))
+    out.mkdir(parents=True, exist_ok=True)
```
ChrisHuie, line 156: **[SHOULD-FIX] B1 — writing a per-session artifact at `pytest_sessionfinish` has no shared home; the guard added here reaches one writer of three.** (`tests/_collection_manifest.py:156,170`; `tests/_worker_profile.py:113,136`; `tests/bdd/scenario_liveness.py:550-577`; registered at `tests/conftest.py:848-850`; default-on at `run_all_tests.sh:251`; `ci.yml:638`)
`PYTEST_COLLECTION_MANIFEST=<regular file>/sub` makes a 6-test session that passed exit 1 with `NotADirectoryError` from `_collection_manifest.py:156` and no summary line; the manifest writer is registered by the root conftest, so this holds for every session of every suite. "Both now go through one `_write`, so the guard cannot be applied to one and forgotten on the other" is true inside `scenario_liveness.py` and false across the three plugins: the write is implemented three times, and the import-time `_OUT_DIR`/`enabled` pair, the `PYTEST_XDIST_WORKER` read, the `TOX_ENV_NAME` stamp (defaults `""` here, `"adhoc"` in the profile), the `sha256(...)[:12]` fingerprint, the refuse-to-default reader and the `glob("*.json")` loop are each written two or three times. After A1 the profile brings its own separate guard, which leaves two guard implementations and one unguarded writer — still no seam. One home for `write_json`, `read_json_records`, `require_dir`, `xdist_worker_id` and `relative_target` (e.g. `tests/_collection_manifest.py`, already imported by the root conftest and stdlib-only), adopted by all three. Pair the profile's guard with a fail-closed reader: `summarise` returns `{"suites": {}}` on an empty directory (`_worker_profile.py:149-150`), so a guarded write alone turns a loud crash into a silent loss.

tests/bdd/scenario_liveness.py
```diff
@@ -434,7 +536,144 @@ def pytest_sessionfinish(session: pytest.Session) -> None:
...
-    path.parent.mkdir(parents=True, exist_ok=True)
     artifact = {"run": _run_scope(session), "scenarios": [_SHARDS[k] for k in sorted(_SHARDS)]}
-    path.write_text(json.dumps(artifact, indent=2, sort_keys=False) + "\n", encoding="utf-8")
+    body = json.dumps(artifact, indent=2, sort_keys=False) + "\n"
+
+    # ALSO write a per-session copy, so a second session cannot silently replace
+    # this measurement with its own. The merged path is kept as-is because other
+    # consumers read it by name; the directory is what `load_run` folds.
+    sessions = _write_sessions_dir()
+    _write(artifact_path(), body)
```
ChrisHuie, line 546: **[SHOULD-FIX] B2 — the merged liveness artifact is still last-writer-wins between concurrent shards, and its reader rejects only `-k`/`-m` narrowing.** (`tests/bdd/scenario_liveness.py:546`; `scripts/audit/scenario_liveness_join.py:146-168`; `tox.ini:260-299`)
Every session still writes `artifact_path()`, and `bdd-inprocess-s{1,2}` and `bdd_e2e` have no `depends` on each other, so under `tox -p` the last to finish owns the merged file. `_reject_narrowed_run` checks `selection` and `markers` and never `target`: given the shard-1 session this CI run published (target = 12 files, no `-k`), `load_artifact` returns its 266 scenarios as the suite, while the `bdd_e2e` session is rejected. That is the "published artifact described one shard" defect from the description, closed for `load_run` and open for the reader behind `storyboard_check_index.build_index`. Refuse a `run.target` that is not the whole tree in `_reject_narrowed_run` (reuse `_covers`), or stop writing the merged path from a target-scoped session; one test with a shard-shaped `run`.

tests/unit/test_scenario_liveness_write_guard.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from tests.bdd.scenario_liveness import _write
+
+
+def test_a_write_failure_does_not_raise(tmp_path: Path, capsys: pytest.CaptureFixture[str]) -> None:
```
ChrisHuie, line 22: **[SHOULD-FIX] B3 — the write guard is tested at the helper, not at either call site.** (`tests/bdd/scenario_liveness.py:546-547`; `tests/unit/test_scenario_liveness_write_guard.py:22-37`)
Removing the `try/except` inside `_write` reddens the guard test; re-inlining an unguarded `write_text` at `:546` or at `:547` leaves everything green. Either an AST pin (no `.write_text(`/`.mkdir(` outside the shared writer — across all three plugins once B1 lands) or one session-driven test that points `BDD_LIVENESS_SESSIONS` at a regular file and asserts no exception plus the stderr line.

ChrisHuie (commented):
Reviewed at 8a1f83e6e (CI run 33792387825, 33/33 green; a trial merge into main 998464819 is conflict-free and lands the 37 files). 1 BLOCKER / 9 SHOULD-FIX. Evidence below comes from this head and from the collection and liveness artifacts that CI run published; `tests/collection/` run against those artifacts gives the same 51 passed as the job.

**Confirmed first.** No `--collect-only` argv site remains under `tests/`. The manifest record is retained before deselection and serialized after every `modifyitems` hook: in the `bdd_e2e` record (`-k e2e_rest`) all 7,647 node ids it shares with the shard records carry identical marker sets, so the env-route grader reads the superset its docstring promises. `load_run` refuses a missing shard session and a session with `workers_missing` set; `load()` enforces target scoping and the `filter=""` demand. The tox substitution of `BDD_SHARD_PATHS_*` splits into separate argv entries, and an empty value is dropped exactly as the tox.ini comment predicts. Guard suite on this head: 1093 passed, 0 failed. pylint R0801 on `tests/` is 69 at this head and 69 at the merge base, so `.duplication-baseline` 71→69 is the ratchet correcting a stale-high value, not a reduction here. ruff check and format are clean on all 26 changed Python files.

Findings are posted as inline threads: A1, B1–B5, C1–C4.

**Outside this PR.** The plugin creates a liveness record only when a scenario reaches `pytest_bdd_before_scenario`, so setup-phase xfails never enter the artifact: on this run's shards, 504 and 518 identity-tagged in-process scenarios were collected and 266 and 260 observed, and the `harness_not_wired` category has zero observations. Pre-existing semantics, unchanged here; the observers grade uc005/uc006 only.

**Not verified:** a full in-network run on this head (B2's concurrent-writer race and the `tox -p` ordering are read from `tox.ini` and tox's scheduler, and match the description's own account); `make quality` as a whole; the remote runner's consumer of `.suites`; the description's timing figures and the 53-node xpassed set.

Actions: A1 → merge-order decision first; B1–B5, C1–C4 → fix in this PR; C3's CI-grading half → a filed follow-up linked from the description.

tests/collection/test_architecture_bdd_xfail_reason_tokens.py
```diff
@@ -369,24 +371,25 @@ def test_a_per_transport_reason_keeps_its_mcp_and_rest_rows(self):
...
         """
-        import subprocess
-
-        completed = subprocess.run(
-            ["uv", "run", "pytest", str(_UC003_MODULE), "--collect-only", "-q", "-p", "no:randomly"],
-            capture_output=True,
-            text=True,
-            cwd=_REPO_ROOT,
-            timeout=300,
+        rows = load(manifest_dir(), target=None, filter="")
```
ChrisHuie, line 382: **[SHOULD-FIX] B4 — fifteen tests that read neither artifact left `make quality`, `tox -e unit`, CI `unit-tests` and `make test-entity`, and the three relocated `test_architecture_*` guards left the meta-guards' scan set.** (pure tests: `tests/collection/test_architecture_ci_bdd_shard_manifest.py:54-92`, `test_architecture_bdd_xfail_reason_tokens.py:123-358`, `test_architecture_env_route_agreement.py:330-349`; artifact readers: `test_architecture_ci_bdd_shard_manifest.py:37`, `test_architecture_bdd_xfail_reason_tokens.py:382`, `test_architecture_env_route_agreement.py:121`; `tests/unit/_architecture_helpers.py:478`)
With both environment variables unset, `tests/collection` is 36 failed / 15 passed: five shard-config and split tests, nine xfail-reason parse and AST tests, and `test_scenario_never_measured_this_run_resolves_not_wired` need no artifact, yet now run only in the `bdd-tests` job after the shards, so a `parse_xfail_reason` or `SHARD_COUNTS` regression surfaces half an hour later and never locally. `iter_architecture_guard_trees` globs `tests/unit/test_architecture_*.py`, so `test_architecture_no_handrolled_call_walk` and `..._allowlist_diff` no longer see the three moved guards (widening the glob to `tests/*/test_architecture_*.py` takes the scan from 134 to 137 modules and both still pass). Split each module: artifact readers stay in `tests/collection/`, the fifteen return to `tests/unit/`; widen the glob.

scripts/ci/merge_shard_reports.py
```diff
@@ -0,0 +1,82 @@
...
+
+    reports = []
+    for path in shard_paths:
+        try:
+            reports.append(json.loads(Path(path).read_text(encoding="utf-8")))
+        except (OSError, json.JSONDecodeError) as exc:
+            print(f"merge_shard_reports: cannot read {path}: {exc}", file=sys.stderr)
+            return 1
+
+    merged = dict(reports[0])
```
ChrisHuie, line 57: **[SHOULD-FIX] B5 — the shard, merge and rename path has no test anywhere, and the merged report carries shard 1's `exitcode`.** (`scripts/ci/merge_shard_reports.py:25-75`, `:57`; `scripts/ci/assert_full_tree_record.py:24-45`; `run_all_tests.sh:183-220`, `:616-650`; `tests/unit/test_run_all_tests_parallel_tox.py:33-60`)
No test names either script. The contract test's docker stub answers nothing for `docker info`, so `_docker_mem_gb=0` selects the laptop tier and `E2E_WORKERS=0`: the `bdd` → shards swap and the merge never run under test, and the tox stub pre-writes `bdd_inprocess.json`, the merge's output name. CI's in-network job runs `./run_all_tests.sh bdd_e2e` with `e2e_workers=0`, so it takes neither path; the only executions are manual in-network runs. `merged = dict(reports[0])` keeps shard 1's `exitcode`, `collectors` and `warnings` while `summary` counts are summed. Tests for `merge()` — summed summary, concatenated `tests`, `collectors` and `warnings`, one shard of two returns 1 and writes nothing, an unreadable file returns 1, `exitcode = max(...)` — and for `assert_full_tree_record.main()` (present → 0, shard-only → 1, empty → 1); one contract-test variant forcing `E2E_WORKERS=2` with `bdd-inprocess-s1/s2.json` stubs that asserts `.tox/bdd_inprocess.json` and `.suites`, and a missing-shard variant that asserts RC=1.

.github/workflows/ci.yml
```diff
@@ -634,29 +656,87 @@ jobs:
...
     permissions:
       contents: read
-    timeout-minutes: 5
+    # Was 5 while this job only read a result string. It now runs a real pytest
+    # env, so it needs setup time; the suite itself is set operations over JSON.
+    timeout-minutes: 15
     steps:
       - uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
```
ChrisHuie, line 680: **[SHOULD-FIX] C1 — `bdd-tests` dropped `disable-sudo-and-containers: true` from harden-runner, and no step in the job needs it.** (`.github/workflows/ci.yml:678-680`; the merge base had it at `:652`)
The job runs checkout, `_setup-env`, `_adcp-bundle`, two `download-artifact` steps, a shell gather and `_pytest`. Every one of those step types runs under the flag elsewhere on this head (`unit-tests` `:170-183`, `coverage` `:785-800`, `smoke-tests` `:143-155`), and none of the composite actions uses sudo, docker or apt. Of the nine jobs without the flag, eight have `services:`, docker or sudo; this is the only one with none, and neither commit 90cd12e1e nor the description mentions the removal. With `egress-policy: audit` on every job, the flag is this job's only preventive runtime control. Restore the line.

tests/bdd/scenario_liveness.py
```diff
@@ -357,10 +365,94 @@ def pytest_testnodedown(node: Any, error: Any) -> None:
...
+    worker_id = getattr(getattr(node, "gateway", None), "id", None) or repr(node)
+    _WORKERS_SEEN.add(worker_id)
+
     workeroutput = getattr(node, "workeroutput", None)
     if workeroutput and _WORKEROUTPUT_KEY in workeroutput:
         _merge_shard(workeroutput[_WORKEROUTPUT_KEY])
+        _WORKERS_REPORTED.add(worker_id)
+
+
+IN_PROCESS_TRANSPORTS = frozenset({"mcp", "a2a", "rest"})
```
ChrisHuie, line 389: **[SHOULD-FIX] C2 — `_relative_target` is implemented twice, and the in-process transport set now has three literal homes plus a same-named constant with a different value.** (`tests/_collection_manifest.py:84-106` ⟷ `tests/bdd/scenario_liveness.py:407-417`; `tests/bdd/scenario_liveness.py:389` ⟷ `tests/bdd/conftest.py:3244-3245` ⟷ `tests/harness/transport.py:143-152`; `scripts/detect_misrouted_transport.py:28`)
The two `_relative_target` bodies produce identical output on 13 argument shapes and share the dead `rstrip`. `IN_PROCESS_TRANSPORTS = frozenset({"mcp", "a2a", "rest"})` restates the parametrize list as a hand-copied subset of the `Transport` enum, while `scripts/detect_misrouted_transport.py:28` already defines `IN_PROCESS_TRANSPORTS = {"impl", "a2a", "mcp", "rest"}`; a transport added to the enum is silently dropped by `_project_in_process`. One `_relative_target` in `tests/_collection_manifest.py`, imported by the plugin; one `BDD_IN_PROCESS_TRANSPORTS = (Transport.A2A, Transport.MCP, Transport.REST)` beside the enum, used by the conftest and the plugin (`transport.py` imports pydantic and `tests.helpers.pinned_schema`, so check that weight is acceptable in the plugin first); rename one of the two colliding constants.

.pre-commit-hooks/check_repo_invariants.py
```diff
@@ -90,7 +91,35 @@ def check_no_unresolvable_citations(files: list[Path]) -> list[str]:
...
+
+    The guard is not weakened for authored work -- the same files are graded on
+    the next ordinary commit that touches them, by whoever touches them. What is
+    given up is catching a violation at the moment it ARRIVES via a merge rather
+    than when it is next edited.
+    """
+    git_dir = subprocess.run(
+        ["git", "rev-parse", "--git-dir"], capture_output=True, text=True, check=False
+    ).stdout.strip()
+    return bool(git_dir) and (Path(git_dir) / "MERGE_HEAD").exists()
```
ChrisHuie, line 115: **[SHOULD-FIX] C3 — `check_repo_invariants` gains an untested whole-hook bypass during merges, justified by a population the hook's own regex creates.** (`.pre-commit-hooks/check_repo_invariants.py:61-66`, `:94-122`)
`_merge_in_progress()` returns 0 for all three checks whenever `MERGE_HEAD` exists, though the docstring justifies only the citation check; forcing it true makes `main([bad_file])` return 0 and no test reddens (`test_architecture_repo_invariants.py` binds only the other two predicates). The comment at `:61-63` says beads ids are "deliberately NOT covered" while the alternation at `:66` covers them: over `src/` and `tests/` at this head the citation check reports 272 hits in 110 files (269 beads ids, three `.claude/notes/` references), none in files this PR touched — the 272/110 the description reports for the pre-push run — so the wall is the pre-existing population the regex matches, not anything the merge authored. pre-commit already narrows a conflict merge to the conflicted files, so the exemption removes grading from exactly the files the author resolved. The hook is local-only (no CI job runs it). Add a self-test with a temporary `git init` (graded without `MERGE_HEAD`, skipped with it), then either narrow the skip to files whose staged blob equals `MERGE_HEAD`'s, or reconcile the regex with its comment and delete the exemption. The description's "No CI job runs that hook" needs an owner; file the follow-up with the fix.

src/adapters/triton_digital.py
```diff
@@ -410,6 +410,45 @@ def check_media_buy_status(self, media_buy_id: str, today: datetime) -> CheckMed
                 self.log(f"Error checking Triton Campaign status: {e}")
                 return CheckMediaBuyStatusResponse(media_buy_id=media_buy_id, status="unknown")
 
+    def _dry_run_delivery(
```
ChrisHuie, line 413: **[SHOULD-FIX] C4 — five statements in the description and docs do not match the tree.** (`src/adapters/triton_digital.py:413-450` and four comment-only `src/` files; `tests/integration/test_storyboard_ledger_fitness_real_session.py:170-186`; `docs/development/e2e-testing.md:29,33,178,324`; `tox.ini:238`)
(1) "those files take main's version wholesale": `git diff origin/main -- src/` is 47+/36− — the `_dry_run_delivery` extraction (behavior-preserving, 412901b04) and seven citation rewrites to `#1802`; main has not touched these files since the merge base. (2) "all 21" modules: `tests/bdd/test_*.py` is 22 at base and head, and `_covers` demands 22. (3) "+273 net under tests/": `git diff --shortstat 6c378a676 HEAD -- tests/` is +989/−377 = +612; the scoping behind 273 is not stated. (4) Issue #2168's "Exactly one nested pytest run remains": two do — `test_xdist_report_serialization.py` by design, and `test_storyboard_ledger_fitness_real_session.py:170-186`, which the issue listed for migration, still spawns a full nested session and is touched here only in its docstring. (5) The docs and the `tox.ini` comment still name `tox -e bdd_inprocess`, an env this PR renamed away (`tox config -e bdd_inprocess` → "did you mean bdd-inprocess-s2?"). Correct each, or state the intended scope where the number is deliberate.

tests/_worker_profile.py
KonstantinMirin: Confirmed, and the decision is: **#2158 lands first; this branch resolves against it.** Recorded in the description under "Merge order: this PR is stacked on #2158" — GitHub cannot express the stack because both PRs come from a fork, so the ordering is stated there instead.

I verified every part of this rather than taking it on trust. `fork/fix/2104-worker-profile-and-ui-output` is not an ancestor of this head; `tests/_worker_profile.py` here still has `record_test_duration(seconds)`, one `max(0.0` clamp, `controller_collect_s`, and zero `except OSError`; and `git merge-tree` conflicts in exactly the four files you name, with `tests/_worker_profile.py` as add/add.

On resolution, `tests/_worker_profile.py` takes #2158's version wholesale — it is the reviewed one, and there is nothing in this branch's copy that the other lacks. The other three take the union.

Worth saying plainly: I reported this overlap on this PR a week ago as a single shared commit (`a584c6c09`, the `repo-invariants` stage fix). That understated it by three files and an entire module, and the consequence — merging this after #2158 would have silently reinstated the defects that review had just fixed, with the profile default-on for every in-network run — is worse than the bookkeeping. Thanks for catching it.

Issue #1119, opened 2026-03-12
refactor: migrate remaining raw select() calls to repository pattern (292 violations)


KonstantinMirin:
## Summary

PR #1097 introduced the repository/Unit of Work pattern and a structural guard (`test_architecture_no_raw_select.py`) that prevents new raw `select()` calls outside repositories. 292 pre-existing violations are allowlisted.

This issue tracks the remediation of those violations — migrating all raw `select()` calls to go through repository classes.

## Violations by layer

| Layer | Count | Files |
|-------|-------|-------|
| Admin blueprints | 133 | creatives, workflows, products, principals, settings, etc. |
| Services | 65 | background services, GAM inventory, order approval, etc. |
| Core | 34 | config_loader, context_manager, auth, strategy, etc. |
| Adapters | 24 | GAM managers, reporting, mock adapter |
| Admin utilities | 26 | domain_access, sync_api, tenant_management, helpers |
| Routes | 2 | health debug endpoints |

## Approach

- Work layer by layer, starting with the smallest (routes, admin utilities)
- For each file: create/extend a repository class, move queries into typed methods, update callers
- Remove allowlist entries as violations are fixed (the stale-entry test enforces this)
- The allowlist can only shrink — new violations fail the build immediately

## Acceptance criteria

- [ ] `ALLOWLIST` in `test_architecture_no_raw_select.py` is empty
- [ ] `make quality` passes
- [ ] `./run_all_tests.sh` passes — full suite, 0 failures

## Context

- Guard: `tests/unit/test_architecture_no_raw_select.py`
- Architecture pattern: CLAUDE.md Pattern #3 (Repository Pattern)
- Origin: PR #1097 review by @chrishuie

PR #2191, opened 2026-09-03
fix(a2a): make A2A and MCP conformance gradable — 0 graded checks to 49 passing on both protocols

### LINKED ISSUE #1440: fix(a2a): agent-card discovery serves only one non-canonical path and never dual-emits for v0.3 clients
## Summary

A2A agent-card discovery on `main` has two independent, verified defects that together block any v0.3 / non-Python client (and any spec-canonical-path client) from discovering this agent:

1. **Only `/.well-known/agent-card.json` is actually served.** The spec-canonical `/.well-known/agent.json` and `/agent.json` return **404** — even though the app intends to serve all three and logs a warning that it couldn't.
2. **The served card is A2A-1.0-only.** It exposes `supportedInterfaces[].url` but **no top-level `url`**, so a v0.3 reader that looks for a top-level `url` fails even at the one path that works.

The decision to support v0.3 was already made (non-Python clients need it); this issue is the implementation of that decision, plus the canonical-path fix discovered alongside it.

Supersedes #1289 (its premises were stale: `enable_v0_3_compat=True` was already set, and the "local patch to revert" was never in `main`).

## Verified on

- HEAD `19d3efd54` (2026-06-12), clean tree
- `adcp==4.3.0` (spec 3.0.1), `a2a-sdk==1.0.1`

## Current behavior (observed, via Starlette TestClient — no Docker)

```
/.well-known/agent.json       -> 404
/.well-known/agent-card.json  -> 200   top-level url=None   supportedInterfaces=1
/agent.json                   -> 404
```

App log on startup:
```
WARNING:src.app:_replace_routes: expected SDK routes not found for paths: ['/.well-known/agent.json', '/agent.json']
```

nginx (`config/nginx/*.conf`) has no rewrite from `agent.json` → `agent-card.json`, so the 404 reaches clients.

## Root cause

**Defect 1 — missing paths.** `_replace_routes()` in `src/app.py` *replaces by path match*: it only swaps in a dynamic handler for routes the SDK already created. `create_agent_card_routes()` (a2a-sdk 1.0.1) mounts exactly one route (`AGENT_CARD_WELL_KNOWN_PATH = '/.well-known/agent-card.json'`) and has no legacy-alias option, so the other two entries in `_AGENT_CARD_PATHS` are never mounted — only warned about.

**Defect 2 — no dual-emit.** The card is built with a single interface at `protocol_version="1.0"` (`src/a2a_server/adcp_a2a_server.py`). The SDK's dual-emit (`agent_card_to_dict` → `to_compat_agent_card`) only injects the v0.3 top-level `url` when the card advertises a *legacy* interface; with only a 1.0 interface it raises `VersionNotSupportedError` ("AgentCard must have at least one interface with compatible protocol version"), which `agent_card_to_dict` swallows → no top-level `url`. The `enable_v0_3_compat=True` flag already set in `src/app.py` is on the JSON-RPC *message* transport, not the card.

Confirmed fix mechanism (TestClient, after adding a legacy interface):
```
top-level url   = http://.../a2a
protocolVersion = 0.3
supportedInterfaces = 2   # 1.0 readers still find their entry
```

## Decisions needed

1. **Is v0.3 / legacy discovery still required, and at what priority?** The #1289 thread says yes (non-Python clients). Confirm whether a live consumer is actually v0.3-only today, or whether all current readers parse `supportedInterfaces[]`. This sets priority: real interop blocker vs. forward-compat hygiene. (Note: A2A defines no sunset for the v0.3 compatibility period.)
2. **Always-on or config-gated legacy interface?** The RPC-side compat flag is currently hardcoded on. Match that (hardcode the legacy interface) or gate both behind one env switch so the compat period can be ended cleanly later?
3. **Path coverage** (recommendation, confirm): serve **both** `/.well-known/agent-card.json` (A2A-1.0 default) and `/.well-known/agent.json` (AdCP-guide canonical + v0.3 legacy). `/agent.json` is currently intended too — keep or drop?

## Implementation guidance

- **`src/a2a_server/adcp_a2a_server.py` (`create_agent_card`)** — append a second interface so the card dual-emits:
  ```python
  from a2a.utils.constants import PROTOCOL_VERSION_0_3
  # supported_interfaces=[
  #     AgentInterface(url=server_url, protocol_version="1.0"),
  #     AgentInterface(url=server_url, protocol_version=PROTOCOL_VERSION_0_3),  # "0.3"
  # ]
  ```
- **`src/app.py` (`_create_dynamic_agent_card`)** — the per-request URL rewrite currently sets `supported_interfaces[0].url` only. Set it on **every** interface, or the new v0.3 interface keeps the static fallback host and tenant routing breaks for v0.3 readers.
- **`src/app.py` (`_replace_routes`)** — mount dynamic routes for **all** `_AGENT_CARD_PATHS`, not just those the SDK created: for the `missing` set, append a new `Route(path, dynamic_agent_card, ...)` instead of only warning. Ensure they sit before the admin catch-all mount.
- **`tests/unit/test_a2a_transport_contract.py`** — extend `test_agent_card_has_required_fields` to assert the dual shape (top-level `url` **and** non-empty `supportedInterfaces`), and assert `/.well-known/agent.json` returns 200. Update the "URL is inside supportedInterfaces, not top-level" comment.
- Sanity-check the e2e card tests (`tests/e2e/test_a2a_endpoints_working.py`, `tests/e2e/test_a2a_regression_prevention.py`) — they assert on `supported_interfaces[0]`, so they survive an additive second interface, but confirm.

## Verification

- TestClient: GET all three paths → 200; top-level `url` present; `len(supportedInterfaces) >= 2`.
- `tox -e unit -- -k a2a` then `./run_all_tests.sh ci` (e2e/admin exercise the live routes).
- Running stack: `curl -s http://localhost:8000/.well-known/agent.json | jq '{url, n:(.supportedInterfaces|length)}'`.

## Out of scope

- The compliance runner probing `/a2a/.well-known/agent-card.json` (the `/a2a/` prefix) is an upstream `@adcp/sdk` / a2a-js matter (v0.3 back-compat tracked upstream). Do **not** add a bridge route for it — it is not spec-canonical and would mask the upstream defect.
- The AdCP spec-version bump (3.0.1 → 3.1) is tracked separately in #1388 and is independent of this change.

### LINKED ISSUE #1845: Product.publisher_properties fabricates <subdomain>.example.com when tenant has no virtual_host
Split out of a fix for get_adcp_capabilities' publisher_domains fabrication (see the sibling fix for that tool) after review found the fix shape does not transfer to this call site.

`src/core/database/models.py` (`Product.publisher_properties`, AdCP 2.0.0 by_id/by_tag/all discriminated union): same `f'{tenant.subdomain}.example.com'` fabrication pattern as `get_adcp_capabilities`, but the fix cannot be "return None"/omit the way that tool's optional `Portfolio` object allows:

1. The pinned AdCP v3.1.1 schema (`core/publisher-property-selector.json`) REQUIRES `publisher_domain` (or `publisher_domains`) on EVERY `publisher_properties` entry — for `by_id`: `required=[publisher_domain, selection_type, property_ids]`; for `all`/`by_tag`: an XOR of `publisher_domain`/`publisher_domains` is required. There is no "omit the field" escape hatch.
2. `src/core/schemas/product.py` (`validate_publisher_properties`) independently requires the `publisher_properties` list be non-empty — this is why the current fallback exists at all (it's a safety net for that validator, not incidental).
3. Consequence: naively fixing this the way `get_adcp_capabilities` was fixed (return None / omit) would make `get_products` raise `ValidationError` for every product on every tenant without a configured `virtual_host` — likely the common case, since subdomain routing is the default and `virtual_host` is opt-in. That would be a worse regression than the current fabrication bug.

Needs its own research/design cycle to decide:
- (a) read the real `publisher_domain` from `PublisherPartner` rows (the same source `get_adcp_capabilities` now correctly reads) as a fallback chain before `virtual_host`;
- (b) what the spec-legal, honest behavior is when NEITHER a `PublisherPartner` row nor `virtual_host` exists for a tenant that has products — exclude/degrade the product from `get_products` responses? raise a typed provisioning error? something else? There is no "just omit it" option like `Portfolio` has.

`example.com` is RFC 2606 reserved, so the fabricated value is guaranteed unreachable rather than merely unverified — same class of dishonesty as the sibling capabilities.py bug.

### LINKED ISSUE #2252: Storyboard Conformance seeds no tenant: 86 of 121 CI failures are 401 artifacts, not conformance gaps
## What happens

The `Storyboard Conformance` CI job runs `./run_all_tests.sh storyboard`, a suite list of
exactly one suite. Nothing seeds the `ci-test` tenant or its principal into that stack's
database, so the agent rejects the runner's credential and the run grades nothing:

```
121 failed, 22 skipped in 38.03s
mcp  … 0 passed
```

Of the 121, 86 are artifacts of the missing seed rather than conformance gaps:

| Count | Symptom | Cause |
|---|---|---|
| 26 | `The agent rejected the configured credential with HTTP 401` (mcp) | no seeded principal holds `ci-test-token` |
| 60 | `security_transport/signed_requests/*` expecting 401 | the `signed_requests` storyboard self-gates on the seller's advertisement, and the runner cannot read `get_adcp_capabilities` through a 401, so it grades the track instead of skipping it |
| 26 | `Refusing to fetch: storyboard.adcp.test resolves to a private/loopback address` (a2a) | separate: the 14-rc A2A probe path does not honour `--allow-http` |

We advertise `request_signing: {supported: false}` (`src/core/tools/capabilities.py:328`),
so those 60 checks skip once the runner can read the advertisement.

## Why CI and the box disagree

`tox.ini`'s storyboard env declares `depends = e2e, ui, bdd, bdd_e2e`, and the `e2e` suite
is what seeds `ci-test` (`tests/e2e/conftest.py:491`, `scripts/setup/init_database_ci.py`).
A full-suite run therefore seeds before the storyboard env starts, and a full-suite run on
the CI box reports 47 storyboard failures whose causes are real (`ACCOUNT_NOT_FOUND`,
schema divergences) rather than 401s.

`depends` orders envs that are in the run. With a one-suite list there is nothing to order,
so `e2e` never runs and the seed never happens. The dependency is therefore unenforced in
the one configuration that CI uses.

## Why it is newly visible

`adcp-server-storyboard` does not exist on `main`. The job pointed at a host that resolved
to nothing, the runner reported the agent unreachable, and the suite graded zero checks in
about 6 seconds while reporting success (run 34845326360, 2026-09-14:
`storyboard: OK (27.81=setup[16.25]+cmd[5.51,6.06] seconds)`). Adding the dedicated agent
turned the job into a real measurement, which is what exposed the missing seed.

The job is report-only by design — `storyboard-conformance` is absent from the `summary`
job's `needs` — so this does not block merges.

## Fix

Seed the tenant from the storyboard suite itself rather than relying on `depends`: a
session-scoped fixture in `tests/storyboard/conftest.py` that runs
`scripts/setup/init_database_ci.py` against the in-network database, the same way
`tests/e2e/conftest.py` does. Then the failure count measures conformance.

Verifying the fix needs an in-network storyboard-only run, because a full-suite run seeds
through `e2e` and cannot observe the gap.

## Separately

The 26 A2A refusals are a second defect. `--allow-http` sets `allow_http: true`, which
`runner.mjs` reads as `allowPrivateIp` at its probe sites, and a local
`adcp storyboard run https://localhost:8443/mcp/ … --allow-http` does reach connection
attempts instead of refusing. The A2A axis still refuses, so the flag is not reaching that
path.

### LINKED ISSUE #2253: test_template_url_validation grades the ambient ADCP_AUTH_TEST_MODE, not the templates
## What happens

Two tests in `tests/integration/test_template_url_validation.py` pass or fail according to
the ambient value of `ADCP_AUTH_TEST_MODE`, not according to anything they set up:

```
FAILED test_all_template_url_for_calls_resolve
FAILED test_form_actions_point_to_valid_endpoints
E   Failed: Found unresolvable url_for calls in templates:
E   Template: login.html
E     url_for('test_auth.test_auth')
E     Error: Could not build url for endpoint 'test_auth.test_auth'.
```

They fail in CI and on a bare agent-db box, and they pass on the full-suite CI box, on the
same commit (`84c788730`). That pair is the entire difference between the two runs'
integration results: 71 of 73 failures are identical.

## Why

`src/admin/app.py:351` registers `test_auth_bp` conditionally:

```python
if settings.testing.adcp_auth_test_mode and not is_production:
    app.register_blueprint(test_auth_bp)
```

`templates/login.html:77` and `:87` call `url_for('test_auth.test_auth')`, and both sit
inside `{% if test_mode %}`, which is driven by the same setting. **The templates are
correct** — at render time the guard and the registration agree, so the block never renders
without the blueprint.

The tests walk template SOURCE and call `url_for` on every call site they find, ignoring the
enclosing `{% if %}`. So they resolve the endpoint only when the app under test happens to
have been built with test-auth registered. What they grade is the ambient flag, not whether
the templates are right.

## Fix

Either build the app under test with `adcp_auth_test_mode` on, so the scan sees the routes
the guarded blocks depend on, or have the scanner skip call sites inside a `{% if %}` whose
condition it cannot evaluate. The first is smaller and keeps the scan total.

Do not set the flag through the environment: production reads a settings FIELD built once,
so inject the typed value.

## Why it matters beyond these two tests

This is the shape `docs/development/fixture-layer-audit.md` describes — a test whose verdict
comes from ambient environment rather than from state it establishes. It cannot be read as
coverage in either direction: green does not mean the templates resolve, and red does not
mean they do not.

### LINKED ISSUE #2259: Tenant resolution falls back to the 'default' tenant for a loopback Host; every deployment has one
## The rule

Tenant identification must come from something the request **names**. A caller that names
no tenant has not addressed a seller, and the answer is a refusal — not a guess. A default
here is the same class of defect as a production path branching on `ADCP_TESTING` (#2255):
an allowance that exists for the convenience of the machine running the code, which then
survives into the path that serves buyers.

## The fallback

`src/core/resolved_identity.py:214-215`, the fourth and last strategy of `_detect_tenant`:

```python
if not tenant_id and host.split(":")[0] in ["localhost", "127.0.0.1", "localhost.localdomain"]:
    tenant_id = tenant_id_for(subdomain="default")
```

Three strategies before it read something the caller supplied — `Host` as a virtual host,
`Host` as a subdomain, `x-adcp-tenant`, `Apx-Incoming-Host`. This one reads nothing. It
fires precisely when the request named no tenant, and answers with one anyway.

Note the port is stripped, so `localhost:8080`, `127.0.0.1:8000` and
`localhost.localdomain` all match.

## Every deployment has a tenant for it to land on

This is the part that makes it more than a development convenience.
`src/core/database/database.py:45-72` creates `tenant_id="default"`,
`subdomain="default"` **unconditionally**. `CREATE_DEMO_TENANT` does not gate creation —
it only chooses between two shapes:

| `create_demo_tenant` | what `default` becomes |
|---|---|
| `true` | fully configured: mock adapter, currency limits, sample data, **and a principal holding the hardcoded token `ci-test-token`** (`database.py:96-103`) |
| `false` (production default) | "My Sales Agent", blank, awaiting setup — still `subdomain="default"`, still resolvable |

So the fallback is never inert. On a production deployment it resolves an unconfigured
seller; on anything started with `CREATE_DEMO_TENANT=true` it resolves a seller whose
credential is a constant published in this repository and copied into
`scripts/setup/init_database_ci.py`.

## What reaches it

nginx forwards `Host` verbatim (tenant routing is an exact `Host` match), so an external
caller's own `Host` normally reaches the resolver unchanged — an internet caller cannot
simply assert `localhost`. The exposure is everything that is NOT that path:

- a request reaching the app container directly on `:8080`, bypassing the proxy — in-cluster
  callers, a sidecar, a port-forward, a debug shell;
- a health check or internal poller dialling loopback;
- any proxy or ingress in front that normalizes or rewrites `Host` (the app cannot tell).

On a public tool — `get_products`, `list_creative_formats`, `get_adcp_capabilities`, all
declared `PublicIdentity` — such a request is served anonymously **against the default
tenant's data**, having named no seller. With `CREATE_DEMO_TENANT=true` and the published
token, it authenticates as a real principal of that tenant.

## The fix

Delete the branch. `_detect_tenant` returns `None`, and the layers above already handle
that honestly: a protected tool answers `AUTH_MISSING` because no tenant means no principal
lookup, and a public tool proceeds with `PublicIdentity(tenant=None)`, which its
implementation already has a branch for (`capabilities.py` returns the minimal description;
the agent card, after #1440, answers 404).

What that costs, and it is the reason the branch exists: `docker compose up` →
`http://localhost:8000` stops resolving a tenant. Local development would address the
demo tenant the same way every test already does — `x-adcp-tenant: default`, or a
subdomain under the configured `SALES_AGENT_DOMAIN`. That is a documentation and
quickstart change, not a code one, and it makes the local stack exercise the same
resolution path a deployment uses instead of a fourth one nothing else runs.

Worth doing in the same pass: `CREATE_DEMO_TENANT=true` seeding a principal with a
repository-constant token is its own sharp edge, independent of this fallback.

## Why it will not be caught by tests

Nothing grades the absence of a fallback. The e2e and storyboard suites address tenants
by `x-adcp-tenant` (`tests/storyboard/test_storyboard_conformance.py:145`,
`scripts/setup/init_database_ci.py`), so strategy 4 is dead in CI and alive only where
a human points a browser at loopback. Deleting it therefore breaks no test, which is
also why it has survived: nothing depends on it except the quickstart.

KonstantinMirin:
Closes #1440 (the canonical-path half; the v0.3 dual-emit stays out — see below).
Closes #2252 — the storyboard job seeds its own tenant now, so its runs grade conformance
instead of 401 artifacts.
Closes #2253 — `url_for('test_auth.test_auth')` no longer exists, so the two template tests
stop grading the ambient `ADCP_AUTH_TEST_MODE` and pass in every environment.
Closes #2259 — the loopback→`default` fallback is gone; a request naming no tenant is
refused rather than guessed at.

**What #2259 asked for, and what it did not.** Its stated fix is "delete the branch", and
that branch is deleted: `_detect_tenant` has no `localhost` / `127.0.0.1` case, and
`grep -n localhost src/core/resolved_identity.py` is empty. Two things it raises are NOT in
this PR, and neither is part of its fix:

- **The `default` tenant still exists as a concept**, in roughly 17 sites — `database.py`
  creates and looks it up in 10 places, `admin/blueprints/auth.py:251,257,273` falls back to
  it in single-tenant mode, `core.py:213,302` redirects to it, `landing_page.py:90` defaults
  to it. What died is the RESOLVER's guess, not the row. Removing the concept is a separate
  change with real deployment consequences.
- **`CREATE_DEMO_TENANT=true` seeding a principal whose token is a repository constant** is
  untouched. #2259 itself calls this "its own sharp edge, independent of this fallback".

This PR does pay the cost #2259 names: it predicted that deleting the branch stops
`docker compose up` → `http://localhost:8000` from resolving a tenant, and called that "a
documentation and quickstart change, not a code one". `docs/quickstart.md` now names the
tenant explicitly with `x-adcp-tenant` in the MCP example and says which endpoints need it,
so the local stack exercises the same resolution path a deployment uses. Without that, the
quickstart's own MCP snippet would have failed with `CONFIGURATION_ERROR` and nothing in CI
would have caught it — as #2259 wrote, "nothing grades the absence of a fallback".

Closes #1845 — `primary_domain` no longer builds `<subdomain>.example.com`; it returns None,
and the `publisher_properties` call sites keep their existing `"unknown"` placeholder.

## What this is

#1440 was filed as "two agent-card paths 404". Fixing it uncovered that the thing which
should have caught it — the AdCP storyboard-conformance job — **had never graded a single
check**, and exited green while doing so. Everything since has been making that instrument
real and then fixing what it found.

| | mcp | a2a |
|---|---|---|
| before this PR | never graded (`overall_status=unreachable`) | never graded |
| now | **49 passed / 16 failed** | **49 passed / 16 failed** |

Both protocols identical, 72 storyboards executed, and the 16 that remain are genuine
conformance gaps in creative, notification-config and validation behaviour — not
infrastructure, and no longer misattributed.

## How it got there

**1. The conformance job never collected anything.** `_missing_env()` asked whether
`STORYBOARD_COMPLIANCE_DIR` / `STORYBOARD_SCHEMA_ROOT` were *set*. Nothing sets them —
`tox.ini` defaults both to empty, the CI job sets neither, and `git log -S` shows they were
never in the workflow. The gate short-circuited before the derivation that builds the
runner's own arguments could run. It now asks that resolver, so the gate cannot disagree
with what the runner is handed.

**2. Serve every declared agent-card path** — the original #1440. `_AGENT_CARD_PATHS` is the
one declaration; one handler serves all three, so their bodies are byte-identical by
construction.

**3. The card published whatever host the caller asked for.** A ladder over
`Apx-Incoming-Host` / `Host` / `X-Forwarded-Proto` meant `Host: evil.example.com` came back
as `supportedInterfaces[0].url`. The card now publishes the tenant's **stored** origin, so
it and `brand.json` agree byte-for-byte — `security.mdx` step 5 compares them, and two
derivations are two chances to disagree. Both holes close by construction: an unregistered
host never reaches the URL.

**4. A tenant is named by its host or by a header, and by nothing else.** Subdomain
resolution is gone — the `_detect_tenant` branch, the `x-adcp-tenant`-as-subdomain lookup,
five `domain_config` helpers, nginx's `map $host $tenant` and the headers it injected, and
the subdomain URL derivations in four templates and four JS sites (now one `agentOrigin()`).
`SALES_AGENT_DOMAIN` survives only as this deployment's own domain. Two derivations outlived
the strategy and both invented a name: `_agent_host` fell back to
`f"{subdomain}.{SALES_AGENT_DOMAIN}"` — which is what published
`ci-test.sales-agent.example.com` on the CI tenant's card, a name nothing served — and
`primary_domain` built `f"{subdomain}.example.com"`, a domain nobody owns on a reserved TLD,
handed to buyers as the publisher's own (#1845). Both are gone. A request naming neither host
nor tenant is **refused** with `CONFIGURATION_ERROR` / `terminal`, carrying the host it could
not serve.

Making `virtual_host` *mandatory* is deliberately **not** done here: a publisher without a
domain is a real seller — a print title sells inventory that has no host — so the
`publisher_properties` sites keep an explicit placeholder rather than refusing, and the wider
removal of the `subdomain` column (676 references, 33 src files, a migration) is the next
change, not this one.
A deployment that cannot tell which seller a request addresses has no rule of that seller's
to apply. Graded by four `@tenantid` scenarios on a2a/mcp/rest.

**5. `virtual_host` is an origin, not a hostname.** Three readers want different parts of one
stored fact. Dropping the port fixed `publisher_domain` (AdCP's pattern admits no colon, and
a port there failed every product of the tenant) and broke the card, which published
`https://storyboard.adcp.test/a2a` for an agent listening on `:8443` — sending every A2A
client to a closed port and taking the axis **27 → 0**. The column stores the origin; the
hostname is projected at the two readers that need one.
`tests/integration/test_tenant_origin_with_port.py` pins all three and goes red if the strip
returns.

**6. The seeded accounts come from the bundle.** The storyboard tenant's account was
hand-typed and wrong the day it was written (`acme-outdoor.example` for a kit whose
`brand.house.domain` is `acmeoutdoor.example`; that same string again where the operator
belongs). The natural key is `(brand_domain, operator, sandbox)`, so nothing matched and 251
steps answered `ACCOUNT_NOT_FOUND` about products they never reached. Accounts are now READ
from the bundle, as the inventory already was: **30 → 49 on both protocols**, and
`ACCOUNT_NOT_FOUND` from 14 to 1.

That also exposed a runner-side defect — a step named for one tool graded as that tool's
failure when an unrelated prerequisite cannot be met. Filed upstream as
[adcp#7588](https://github.com/adcontextprotocol/adcp/issues/7588) with
[PR #7589](https://github.com/adcontextprotocol/adcp/pull/7589).

**7. The test-credential login path is deleted.** `ADCP_AUTH_TEST_MODE` decided whether
`create_app` **composed** a blueprint serving `/test/login` and `/test/auth` — a password
form with three default credentials. A flag that changes which blueprints exist makes the app
under test a different app from the deployed one, and it cost real signal: two template tests
passed on the box (which set the flag) and failed on a laptop (which did not), sitting in the
integration ledger as a "gap" that was neither. Both now pass everywhere and are graduated.
A test that needs an admin session signs one, including Playwright, verified in a browser
against a live stack. A deployment reaches its first admin through its own OAuth plus
`SUPER_ADMIN_EMAILS`, with per-tenant Setup Mode — the successor the docs already named —
covering the interval before a tenant's SSO is on. `TestingSettings` drops 8 fields to 1.

**8. Three broken admin pages**, found by inventorying all 56 admin page routes against their
templates, inbound links and tests: two rendered a template that was never written (one
swallowed by the route's own `except Exception`, one linked from nowhere); the
authorized-properties family had no nav entry at all; and **enabling a tenant's SSO silently
did nothing** — `enable_oidc` nested a scoped session inside its own, so the `commit()` wrote
nothing while the service logged "Enabled OIDC". `test_architecture_admin_templates_exist`
now fails the build for any `render_template()` naming a file that is not in the tree, and
`tests/ui` gains three Playwright tests that grade reachability rather than rendering.

## Deliberately out of scope

- **The v0.3 dual-emit.** #1440's body prescribes a second `AgentInterface` at
  `protocol_version=0.3`; decided against on 2026-08-31, and the decision is now *more*
  correct because the only identified consumer of the top-level `url` fixed itself upstream.
  A prior implementation exists at `c5379a515` — recorded so nobody rediscovers it and
  assumes it was an oversight. Do not cherry-pick it.
- **`tests/admin/test_sell_readiness_browser.py`** is marked `slow`, no tox env selects
  `slow`, and no retained run has ever collected it. Its login helper is repaired here; its
  five flows still fail on seeding assumptions that were never validated. Pre-existing and
  reported rather than hidden.

## Verification

Every claim above is a measurement on the CI box, not an argument. Two regressions this
branch introduced were caught there and fixed: the `27 → 0` card-URL one above, and a
database read inside `require_tenant_access` whose nested scoped session detached the
caller's and discarded uncommitted rows — the same defect as the `enable_oidc` bug, three
files away. Removing it took a `get_db_session` out of every admin request and tightened two
complexity ratchets.

Local on the final tree: unit 5342, integration 2765, admin 143, ui 8 — the one integration
failure is `Errno 49` (macOS cannot bind the second hostname) and passes on the box.

---

## One way to name a tenant, one way to look one up

The card fix above removed a host ladder from ONE endpoint. Reviewing that surfaced the
same shape everywhere else, so this branch finishes the job: after it, the application
answers "which tenant is this request for" from exactly two inputs, and asks one component
for the row.

**Two inputs, not three.** `Host` → `tenants.virtual_host`, and an `x-adcp-tenant` the
client sent. `Apx-Incoming-Host` is gone from the application entirely — code, comments and
tests. It existed because the optional Approximated proxy forwards to
`APPROXIMATED_BACKEND_URL`, so the `Host` it sends names the backend identically for every
publisher and the publisher's own domain arrives in a vendor header. That is edge
configuration, not application logic: `config/nginx/nginx-multi-tenant.conf` now folds the
header into `Host` and drops it at all 14 proxy sites, verified against a real nginx with an
echo upstream — `Host: adcp.fly.dev` + `Apx-Incoming-Host: sales.publisher.com` reaches the
upstream as `Host: sales.publisher.com` with no vendor header, and a request without one
passes through untouched. The feature keeps working; the app simply never learns the vendor
exists.

**It was not a redundancy, it was a disagreement.** `domain_routing` read
`apx_host or host_header` — vendor wins. `_detect_tenant` tried `Host` first and fell back to
the vendor header — `Host` wins. The same request could resolve to two different tenants
depending on which code asked, which is the concrete form of the rule
`ruff-boundary.toml` already states: "a caller that detects its own tenant is a second tenant
resolver, and the two WILL disagree". Eleven copies of that ladder existed across `src/`.

**And the test that covered it could not fail.** `test_multiple_routing_headers_priority`
asserted the vendor header took priority; it built a `MockContext`, reimplemented the
handler's ladder in the test body, and asserted a dict returned what it had put there. It
would have passed identically under the opposite priority, and was green throughout the
period the two real ladders disagreed. Eleven tests across four files were that shape —
self-simulations calling no production code — and are deleted rather than migrated. Fifteen
tests deleted in total; every deletion checked for surviving coverage first.

**One lookup.** `TenantLookupRepository` gains the four active-filtered routing lookups, and
`config_loader`'s functions plus the three admin blueprint sites delegate to it. The admin
plane's copy matched `virtual_host` as an exact string and ignored `is_active`, so a
deactivated tenant's host still resolved there and a host naming the same origin with a port
did not — the two planes answered the same question differently. `no_raw_select` allowlist
**250 → 244**, with no stale entries.

**Deleted, not replaced.** A failed tenant lookup now raises instead of reporting "no
tenant": four lookups caught every exception, matched the driver's message for `"no such
table"` (SQLite's wording, from before this project was PostgreSQL everywhere) or
`"does not exist"` (a substring of many unrelated PostgreSQL errors), and returned `None` —
which every caller reads as "no such tenant". `get_single_tenant` had zero callers and
`get_default_tenant`'s only caller was `get_single_tenant`; both deleted.

**What is deliberately NOT here.** The `default` tenant still exists as a row and a concept
(~17 sites); `ensure_default_tenant_exists` still runs from `scripts/ops/migrate.py`. And 89
`select(Tenant)` entries remain — all hydration with an already-known id arriving from a URL
segment or an argument, none of them tenant resolution. Those are #1119, and #2263 proposes
making `Tenant` repository-private the way `Principal` already is, so a 90th cannot be
written. Both are separate changes; neither belongs in a branch this size.


ChrisHuie:
## Review — `d3409d42e`

Merge base is current `main` (`e7c6d57e6`). Everything below I reproduced myself against both revisions; anything that behaves identically at the merge base is in the follow-ups section rather than here.

---

### A tenant whose `virtual_host` has any uppercase letter becomes unroutable

`tenant_lookup.py:117` compares `func.split_part(Tenant.virtual_host, ":", 1)` — case-preserving on the SQL side — against `hostname_of(requested)`, which is `urlsplit(f"//{host}").hostname` and lowercases. Two rows seeded, calling the real repository method:

```
stored "Probe-Case.AdCP.test", client sends the exact spelling -> None
stored "Probe-Case.AdCP.test", client sends lowercase          -> None
stored "all-lower.adcp.test"  (control)                        -> resolves
base-form exact match, exact spelling                          -> resolves
```

Base's `filter_by(virtual_host=..., is_active=True)` was exact, so it resolved for the spelling the client sent; head resolves for none. `settings.py:202` accepts uppercase, `find_by_virtual_host` still matches exactly, so the value saves cleanly and the tenant goes dark across resolver, card and landing page at once. Nothing lowercases on write and no test uses a mixed-case host, so this is latent rather than observed in the wild — but silent, and total when hit.

The write boundary looks like the better place to fix it than `_same_host`: the column has no canonical form, and the readers now disagree about case as well as port.

### The card contradicts its own documented behaviour in two directions

`SellerCapabilities`' docstring (`seller_capabilities.py:318-321`) states the intent plainly: "`agent_url` is `None` exactly when no tenant resolved ... the agent card answers 404, because a host this deployment does not serve has no card." Neither half holds. Probing the real app:

```
tenant with virtual_host, Host with port     -> 200  https://probe.adcp.test:8443/a2a
tenant with virtual_host, Host without port  -> 200  https://probe.adcp.test:8443/a2a
tenant with NO virtual_host (x-adcp-tenant)  -> 200  http://localhost:8080/a2a
host this deployment does not serve          -> 500  CONFIGURATION_ERROR
```

The first two are the port fix working as intended. The third publishes a localhost URL as a tenant's public agent endpoint — and since `core.py:464-467` never sets `virtual_host` at creation, that is the default state of a tenant made through the UI. `agent_url` is never `None`: `canonical_agent_url` falls through to `local_base_url`, which always returns a string, so the guard at `app.py:416` cannot fire and `app.py:402-405` ("a tenant that declares no host of its own is refused here") is false. `agent_identity.py:52-53` says "inventing one produces a name nobody can reach" twenty lines above the code that invents one.

The fourth is the error-code question. `CONFIGURATION_ERROR`'s pinned description scopes it to a seller deployment fault — "account declared with `mode: 'mock'` but no `mock_upstream_url` populated ... required environment variable unset on the seller process" — and distinguishes it from buyer-fixable conditions. A caller that simply addressed the wrong host has a lever, so `terminal` misdirects, the emitted suggestion ("surface to a human at the seller") sends your operator a scanner's bogus Host, and because `codes.py:421` bands this code 5xx, every uptime check and load balancer reads the deployment as broken. I'd stop short of prescribing the replacement — `REFERENCE_NOT_FOUND` maps to 404 but its own text scopes it to referenced identifiers and resources and carries a uniform-response MUST that may not fit an addressing failure. Your docstring already says 404; the code you picked is what makes it 500.

Whatever you choose has to land with the BDD change: mutating the exception reddens exactly the three `@T-TENANTID-unserved-host` tests, and the code/recovery pair is pinned at `local-tenant-identification-routes.feature:66-67`, `tenant_identification.py:111` and `BR-UC-010-discover-seller-capabilities.feature:360-361`.

One related consequence worth making deliberately. `_resolve_identity` step 3 went from `tenant = TenantContext.load(...) if tenant_id else None` to `_addressed_tenant(headers)`, which raises, and `_boundary.py:270` calls it for every tool on every transport. The refusal itself you document and grade. What moves quietly is the credentialed case: base fell through to step 4's `if auth_token and principal is None: raise AdCPAuthenticationError()` and answered 401 with a challenge, so a buyer with a stale token and a stale host could tell "rotate my credential" from "the seller is down". Head answers 5xx with no challenge. Refusing before the credential check is defensible — it removes an auth oracle — and at 404 it largely stops mattering.

### The landing page and the card disagree about the same tenant's origin

`generate_tenant_landing_page` derives `mcp_url`, `a2a_url` and `agent_card_url` from one `base_url = _determine_base_url(virtual_host)`, where `virtual_host` is the request's effective host. The card uses the stored origin. For a tenant stored at `probe.adcp.test:8443`, one request with `Host: probe.adcp.test` gets `https://probe.adcp.test:8443/a2a` from the card and `https://probe.adcp.test/...` from the page — I observed the port loss on the admin link the page rendered, and the endpoint URLs come off the same `base_url`. At base the custom-domain lookup was exact, so the two were guaranteed equal; `_same_host` removes that guarantee.

### `6e1bea914` publishes a session signing key and widens the gate its deleted tests covered

`FLASK_SECRET_KEY: "${FLASK_SECRET_KEY:-adcp-test-stack-session-signing-key}"` lands in `docker-compose.yml:88`, `docker-compose.multi-tenant.yml:102` and `docker-compose.e2e.yml:92,575`. The literal doesn't exist at the merge base and no compose file set the variable, so the key was `secrets.token_hex(32)`. Signing a cookie with it: a same-key server decodes `is_super_admin: True`, a random-key server raises `BadTimeSignature`. `app.secret_key = settings.runtime.flask_secret_key` is unconditional and `is_super_admin` short-circuits on session content with no DB read (`helpers.py:154`).

The line it replaced was a dev credential too — the difference is that `ADCP_AUTH_TEST_MODE` was consumed as `if adcp_auth_test_mode and not is_production:`, fail-closed in production by code, and `app.secret_key` has no equivalent. Dropping the default from the two non-e2e files costs nothing: `admin_session.py:70` already falls back to the constant, and `tests/ui`/e2e run against the e2e stack.

Same commit, `_setup_session_reaches` (`helpers.py:316-318`) now reads no tenant state where base required the target tenant's `auth_setup_mode`; the only gate left is `not is_production`, which is `production or environment=="production" or bool(fly_app_name)` and set by neither non-e2e compose file. Deleting that gate entirely leaves `tests/admin` at 143 passed / 1 skipped — identical to clean. Bounded, and worth saying: nothing in `src/` writes `session["test_user"]` any more, so the branch has no minter in a deployment, and anyone who could forge the cookie would take the ordinary `user` branch anyway. It's a weakened condition shipped without its tests rather than a live path.

That commit also removed the only coverage for the open-redirect defence. `_safe_redirect` has 8 live call sites plus an unguarded `session.pop("login_next_url", None)` at `oidc.py:396`; `git grep -ln "_safe_redirect\|login_next_url" -- tests/` returned one file at base and nothing at head. Two of its three tests drove `/test/auth`, but `test_external_next_url_not_stored_at_login` drove `GET /login?next=...`, which still exists.

### The storyboard runner's tenant header stopped resolving

It sends `x-adcp-tenant: storyboard` while `seed_storyboard_tenant.py:240` mints a uuid4 `tenant_id`. At base that worked — strategy 2 was "subdomain, then the literal id" (`tenant_id_for(subdomain=hint) or hint`). This branch deletes the subdomain lookup, so the header can never match and the runs pass on `Host` alone. The comment above it still describes the old two-step behaviour and calls the header the credential's safety net, which no longer exists. You made exactly this fix in `tenant_settings.html` when the snippet's `x-adcp-tenant` moved from `tenant.subdomain` to `tenant.tenant_id` — the storyboard test is the site that missed it.

### An allowlist entry was removed without the violation being fixed

`0487e4fca` renames `Tenant` → `ModelTenant` in `health.py` and deletes `("src/routes/health.py","debug_db_state")` from the `no_raw_select` allowlist in the same commit. `debug_db_state` still makes four raw `select()` calls — `:71 select(ModelProduct)`, `:83 select(ModelTenant.tenant_id)`, `:99 select(ModelTenant)`, `:108 select(ModelProduct)`. The matcher resolves no import aliases, so the rename moved the violation out of view and the stale-entry check then instructed you to delete the entry. Reverting only the alias and re-running the guard gives "Expected 244 allowlisted violations, found 245". The ratchet self-lowers on a rename, and eleven alias-shadowed sites are invisible tree-wide. Mitigating: the route is behind `debug_routes_enabled`, so it's opt-in.

### The conformance job is red here and green on main

At the merge base it scores `30/21/249/72` on both protocols with `storyboard: OK`. `_missing_env`, the gate the body blames, is absent from `main` — removed by `a690ccfee`, which landed inside `2a9613498` (#1721). So the delta is 30 → 49 passed and green → red, not 0 → 49. Two of the three causes are closable here:

- 26 ledger entries resolve to no collected check. `known_failures.txt` is byte-identical at base and head and is fit on main — your account fix graduated them, and the ledger only shrinks, so removing them *is* the shrink.
- `.storyboard-pass-floor` ships at 30 against a measured 49. The floor is `ec2b90815` (09-17), the 30→49 fix is `4a6da9f24` (09-18), and `git log` on the file lists only the first. `_surplus_note` prints "(+19, raise the floor)" on every run. As it stands the +19 has no failing oracle, in the direction the module docstring calls the dangerous one.

### Three assertions with no mechanism behind them

`app.py:399-400` says "Neither the URL nor any field is derived here". I put the header echo back — `seller.model_copy(update={"agent_url": f"https://{request.headers.get('host')}/a2a"})` after `describe_seller` — and the four card test files went **53 passed clean, 53 passed mutated**. The new test can't discriminate because `test_tenant_origin_with_port.py:22,48` sends `Host: ORIGIN` where `ORIGIN` is also the stored `virtual_host`. One test where they differ closes it: `x-adcp-tenant: <id>` with `Host: attacker.invalid`.

`protocol_binding` has no grader at all after `test_agent_card_protocol_binding.py` was deleted — setting it to `"MUTANT"` at `adcp_a2a_server.py:553` leaves those 53 green plus 282 unit tests matching `a2a or card or agent`. The comment two lines above calls it "REQUIRED in practice, not decorative". One assertion in the test that already parses the card covers it.

And the nginx fold — the premise the whole header deletion rests on — is exact today at 14/14/14, with no test parsing `config/nginx/nginx-multi-tenant.conf`. A fifteenth `location` without the pair re-opens it silently.

Smaller: `_detect_tenant`'s new docstring says "`None` is a real answer ... a public tool proceeds with no tenant", but its only caller refuses `None`, so `tenant is not None` at `:373`, `:383`, `:389` and the `assert` at `:410` are now permanently true. And `tenant_settings.html:2639` lost its `{% elif is_production %}` arm while `:2622` kept it, so a production tenant with no `virtual_host` gets `http://localhost:<port>/mcp/` in a copy-paste block two lines under one saying the domain isn't configured.

---

### Pre-existing — follow-ups, not asks here

Each reproduces identically at `e7c6d57e6`.

- Eleven `url_for()` calls name endpoints absent from the live `url_map`. `POST /settings/update` raises `BuildError` on `superadmin_settings.superadmin_settings`, and `tenants.py:466` sits after the `except`, so `update_tenant` commits and then 500s. `def admin_dashboard` exists nowhere while `core.admin_dashboard` is referenced 5 times at both revisions.
- Four nested scoped sessions remain in the auth/OIDC plane (`auth_config_service.py:329,:395`, `auth.py:306`, `oidc.py:191`) — the same shape as the `enable_oidc` bug you fixed. `:395` already holds both rows, so `_config_is_verified_for` is a one-line drop-in.
- CodeQL `py/log-injection` on `auth_config_service.py:286` is relocated content (`:302` at base); `main` carries 31 open alerts of that rule. Worth answering the thread rather than leaving it open.
- `/debug/headers` is unauthenticated with no production gate and returns the resolved tenant's metadata; and the GAM OAuth redirect builds its target from a session-stored `Host`, whose trust premise changed with the fold even though the code didn't.

PR #2254, opened 2026-09-16
fix(admin): AI banners and setup checklist follow the tenant's AI configuration, not the legacy Gemini column

### LINKED ISSUE #2224: Setup checklist reports "AI not configured" for tenants configured through ai_config
`src/services/setup_checklist_service.py:730` and `:1116` both compute:

```python
gemini_configured = bool(tenant.gemini_api_key)
```

backing a checklist task keyed `gemini_api_key` / "Gemini AI Features". A tenant that configured AI through `tenants.ai_config` — the path the Admin UI writes — is told AI is not configured, while the features themselves work.

This is the same divergence shape as the one fixed in `src/admin/blueprints/creatives.py` in #2167, where the page flag and the impl beside it disagreed about whether AI was available. Here the two sites are in a different service.

**Why it was left out of #2167**

The consumer is different in kind: a checklist item, not a gate on running a call, and the task is literally named after the Gemini key. Converting it changes what the checklist *means*, which is a product decision rather than a mechanical substitution — for instance whether an Anthropic-configured tenant should see a task called "Gemini AI Features" marked done.

**Acceptance**

- The checklist reflects whether the tenant has a usable AI configuration, resolved through `TenantAIConfig.from_tenant`, not whether one specific legacy column is populated.
- The task's key and label are reconciled with what it now measures.
- A test that fails when the resolution is reverted, for a tenant carrying only `ai_config`.

Related: #2223. Context: #2167 introduced `TenantAIConfig.from_tenant` as the single owner and named this site as deliberately out of scope.

ahmed-karim:
Closes #2224. Depends on #2167 and is stacked on its head; the four commits below this one are that PR's. Draft until it merges, then rebased onto `main`.

#2224 asks that the setup checklist reflect whether the tenant has a usable AI configuration, resolved through `TenantAIConfig.from_tenant`, with the task's key and label reconciled to what it measures, and a test that fails when the resolution is reverted for a tenant carrying only `ai_config`. This PR does that, and takes the product decision the issue names: the task is called **"AI Features"** (key `ai_provider_key`), so an Anthropic-configured tenant sees a task marked done that is named for what it measures, not for one vendor. The Settings → Business Rules banners are the same divergence on the adjacent page (same column, same owner, same class of seller confusion) and are converted in the same change rather than left as a fourth site; no issue exists for them, so they are described here in full.

@RATO-tdunkel filed #2224 as a deliberate follow-up to #2167. If you already had it in progress, say so and I will close this in favour of yours.

## The defect

Settings → Business Rules carries three banners (creative review, advertising policy, product ranking) that tell the seller whether AI is available, the naming-template help says `{auto_name}` "requires Gemini API key", and the dashboard's setup checklist carries an optional "Gemini AI Features" task. All of them read `tenant.gemini_api_key`.

The Admin UI's own AI Services form (`src/admin/blueprints/settings.py:update_ai`) writes `tenant.ai_config`, and nothing in `src/` writes the legacy column any more. So a seller who configures a key exactly the way the UI asks — any provider, Gemini included — keeps seeing "⚠️ Gemini API Key Required" on every Business Rules card and an incomplete "Gemini AI Features" item on the dashboard. The Integrations section one click away shows the same tenant as "AI Services Configured" with fourteen providers to choose from. The two halves of the page contradict each other.

#2167 fixes the feature side: `TenantAIConfig.from_tenant` becomes the one owner of "given a tenant, what is its AI configuration?" (`ai_config` carrying a usable key first, the legacy column second, `None` otherwise), and the policy check, AI review and ranking resolve through it. It does not touch the banners or the checklist, so once it merges the banners are wrong in the other direction: a tenant configured for Anthropic runs policy checks while the page still demands a Gemini key.

## The fix

- `src/admin/blueprints/tenants.py`: `has_gemini_key = bool(tenant.gemini_api_key)` becomes `has_ai_key = TenantAIConfig.from_tenant(tenant) is not None`. Tenant-only on purpose, as the previous comment already said: the banners tell the seller to set *their* key, and the policy gate in `products.py` is tenant-only for the same reason #2167 gives.
- `templates/tenant_settings.html`: the three banners and the `{auto_name}` help name what is required (an API key for an AI provider) instead of one vendor; the advertising-policy card stops reading `tenant.gemini_api_key` directly and uses the same flag as the other two.
- `src/services/setup_checklist_service.py`: the optional task resolves through `from_tenant` in both the single-tenant and bulk paths; key `ai_provider_key`, name "AI Features", vendor-neutral details.
- `docs/deployment/environment-variables.md`: the sentence "Each tenant sets their own Gemini API key" now describes the multi-provider page.

Wording only where wording was the defect; no feature behaviour changes. The keyless `{"provider", "model"}` row the form writes when the key field is left blank still reads as "not configured", because `from_tenant` already treats usability as the contract.

## Tests

- `tests/integration/test_admin_settings_ai_key_wording.py` (new): renders the settings page for four tenants built through `IntegrationEnv` + `TenantFactory` — nothing configured, `ai_config` with an Anthropic key (the regression), the keyless row, the legacy encrypted column — and asserts on the banner strings. All four fail against the pre-fix code: the Anthropic tenant on the flag itself, the other three on the wording alone.
- `tests/integration/test_setup_checklist_service.py`: `test_gemini_api_key_detection` becomes `test_ai_provider_key_detection` covering the same four cases; rewritten on the harness pattern, so its `GET_DB_SESSION_IN_TESTS_ALLOWLIST` entry is removed rather than renamed.
- `tests/unit/test_inventory_sync_status.py` and `tests/unit/test_setup_checklist_mock_adapter.py`: their `MagicMock` tenants now declare `ai_config = None` and `gemini_api_key = None`. `from_tenant` reads both and, by design, refuses a non-dict `ai_config` (a bare mock attribute) rather than degrading it; the mocks previously passed only because `bool(MagicMock())` is true.
- Run locally: `make quality` equivalent (ruff format/check, egress ruleset, mypy over `src/`, duplication baseline unchanged, `tests/unit` + `tests/harness`: 6638 passed), the two integration files against PostgreSQL (23 passed), and the pre-commit hooks on the changed files. `git diff` on the allowlists shows one removal and no additions.

## Not in this PR

- The "Legacy Gemini Key Detected" notice in the Integrations section is accurate and stays.
- `docs/encryption.md` and `src/core/config.py` describe the legacy column and the platform `GEMINI_API_KEY` variable respectively; both are still real.
- Generative creative building (`creative_agent_registry.build_creative`) genuinely requires a Gemini key, because the creative agent's tool takes `gemini_api_key`. That is a separate seam and is left alone.

### DIFF
--- a/.duplication-baseline
+++ b/.duplication-baseline
@@ -1,5 +1,5 @@
 {
   "src": 31,
-  "tests": 71,
+  "tests": 69,
   "scripts": 0
 }
--- a/src/a2a_server/adcp_a2a_server.py
+++ b/src/a2a_server/adcp_a2a_server.py
@@ -251,6 +251,44 @@ def _internal_error_for(operation: str, exc: Exception) -> InternalError:
     )
 
 
+def _response_failed(response_data: dict[str, Any]) -> bool:
+    """Did this task FAIL, as opposed to completing with something to say about it?
+
+    The A2A ``success`` marker (see ``_stamp_a2a_protocol_fields``) answers this
+    question, and until now it answered a different one: "is ``errors[]`` non-empty".
+    Pinned ``core/protocol-envelope.json`` is explicit that those are not the same
+    question — on ``adcp_error``: "a fatal task failure SHOULD populate both this
+    envelope-level field AND the payload's ``errors[]`` array ... Non-fatal warnings
+    populate ONLY ``payload.errors[]`` with ``severity: warning`` — the envelope MUST
+    NOT carry ``adcp_error`` for non-failures." The envelope's own two examples carry
+    exactly that contrast: ``severity: "warning"`` on a completed response, and
+    ``severity: "error"`` on ``status: "failed"``.
+
+    So an entry marked ``severity: "warning"`` is, by the schema's own definition, not a
+    failure and must not flip the marker. Anything else is: an entry with no severity
+    (every fatal ``errors[]`` this codebase emits — ``UpdateMediaBuyError``,
+    ``get_media_buys``' AUTH_REQUIRED degradation) keeps reporting ``success=False``, so
+    the marker only moves for entries that have explicitly declared themselves advisory.
+
+    Deliberately not keyed on the tool or the code. ``get_products``' unranked-ranking
+    advisory is the entry that surfaced this, and a per-tool branch here would leave
+    every sibling to be found separately: ``get_media_buys``'
+    ``_omitted_row_advisory``, and the ``property_list_unsupported_advisories`` that
+    ``create_media_buy`` / ``update_media_buy`` attach, are all non-fatal notices riding
+    a completed response, and all of them still report ``success=False`` today because
+    none of them carries the ``severity`` marker. That is their defect to fix, in one
+    keyword each, and this predicate is what makes fixing it work — the alternative
+    (inferring "advisory" from the code, or from which tool answered) would have to be
+    re-litigated for every new advisory.
+    """
+    if response_data.get("adcp_error"):
+        return True
+    errors = response_data.get("errors") or []
+    # A non-dict entry cannot be shown to be advisory, so it counts as a failure —
+    # unrecognised shapes must not silently upgrade a response to success.
+    return any(not isinstance(entry, dict) or entry.get("severity") != "warning" for entry in errors)
+
+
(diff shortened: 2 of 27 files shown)


ahmed-karim:
I have read the IPR Policy

